PR #4, opened 2024-11-14
Showcases the PoC implementation from Bloomberg and Tetrate

mathetake:
This is not intended to land the main branch but to showcase what the PoC implementation Bloomberg and Tetrate folks worked on looks like!


missBerg:
Video demonstrating some of the features: https://share.descript.com/view/Xp87gpK3PGy

Issue #2, opened 2024-10-23
Create GOALS.md


missBerg:
The GOALS.md should include the following sections.

- Objectives
- Non-Objectives
- Personas

missBerg:
Created a draft in a Google doc for now. Please review 
https://docs.google.com/document/d/19UGS1NJcfyTqkimvvPFfdPT4kemKKzAYcu4PtHqa4eo/edit?usp=sharing

FYI @mathetake @wengyao04 @aabchoo

PR #5, opened 2024-11-25
Adding GOALS.md

missBerg:
To close issue Create GOALS.md #2
https://github.com/envoyproxy/ai-gateway/issues/2

This initial set of goals are intended to be lean.
We can always expand the goals later, but it is harder to revise or remove goals.

For ideas for additional goals, please raise an issue so we can track.

For thoughts on removing, rewording or improving the goals stated in the PR, please leave comments on this PR.

### DIFF
--- a/GOALS.md
+++ b/GOALS.md
@@ -1,0 +1,39 @@
+
+# Envoy AI Gateway GOALS.md
+
+## Envoy AI Gateway Goals
+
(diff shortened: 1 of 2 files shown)


mathetake (approved):
I've already reviewed in the doc, so I am only giving a stamp here

aabchoo (approved):

PR #9, opened 2024-11-27
Remove duplicate linters and use gci for imports

anuraaga:
Randomly was reading through this repo and found this potential cleanup, feel free to reject if not needed

- `gofmt` and `gofumpt` fully overlap, so remove `gofmt` linter since `gofumpt` is more advanced
- `gci` and `goimports` fully overlap, so remove `goimports` linter since `gci` is more avanced.
  - If wanting to use `goimports`, can do the reverse, but I don't recommend it since it formats much worse IMO
- Replace `goimports` with `gci` in Makefile to correspond to above change
  - Removes `awk` hack since not needed for `gci`
  - Should be reverted if picking `goimports` instead of `gci`, basically both Makefile and golangci.yml should define the same thing

A change that came to mind but was larger than the above was replacing manual invocation of `gofumpt` and `gci` in Makefile with `golangci-lint run --fix`, which works very similarly because it is configured for the two. I personally like it since it means no need to reproduce configuration from yaml file into Makefile (notably the local import prefix). The downside is it would be a bit slower, so also understand why not everyone would want to do that.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -9,9 +9,7 @@ linters:
     - exportloopref
     - importas
     - gci
-    - gofmt
     - gofumpt
-    - goimports
     - gocritic
     - gosec
     - govet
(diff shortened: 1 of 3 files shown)


mathetake (approved):
Thanks!

PR #11, opened 2024-11-27
build: setups for API generation

mathetake:
This adds setups for API generation in preparation for
API discussion. Just migrated from the PoC, and 
nothing special happening in this commit.


Makefile
```diff
@@ -32,7 +43,8 @@ check:
 		git diff --exit-code; \
 	fi
 
-.PHONY: test
+# This runs the unit tests for the codebase.
+.PHONY: testf
```
aabchoo, line 47: Is this a typo `testf`?

aabchoo (approved):

Makefile
mathetake: good catch!

PR #18, opened 2024-12-03
build: setups for container images

mathetake:
This commit set ups the container images build in Makefile
as well as in the CI. A new GHA job template is added also,
which will be used in the release workflow in the future.

All makefile recipes come with the detail comments in order to
make it easier.


.github/workflows/docker_builds_template.yaml
```diff
@@ -0,0 +1,63 @@
+name: Docker Builds Template
```
mathetake, line 1: note: make this as a template since it will be used "release" workflow in the future

wengyao04 (approved):

Issue #13, opened 2024-11-27
Define Control Plane API v0.1.0


missBerg:
We need to define the initial control plane API and the custom resources.

Propose we keep the v0.1.0 API definitions lean, so that we can add more fields as we evolve future versions.
Propose we finalize API by 2024-12-19

- [ ] Propose v0.1.0 custom resources
- [ ] Finalize resource definitions

mathetake:
ref https://github.com/envoyproxy/ai-gateway/pull/35

mathetake:
the only thing left is 
* The configuration for the "filter component" (which is currently extproc), relevant to #82 
* RateLimit related API which should be extremely simple given that we can rely on https://github.com/envoyproxy/gateway/pull/4957 

I will take care of these two

mathetake:
I think the initial version has landed the main branch - now in the polishing phase if necessary

Issue #10, opened 2024-11-27
[Discussion] Semantic Caching for LLM/AI Gateways


sanjeewa-malalgoda:
Created this issue to discuss Semantic Caching requirement in high level. Would like to hear what community think of this. Supporting caching at gateway is not something everyone agrees. Hence lets have a discussions here and see what is the best approach to follow.

**Why Implement Semantic Caching for LLM/AI Gateways?**
LLM calls are expensive and time-consuming, making a caching layer highly valuable for consumers of API gateways. Users expect caching capabilities at the gateway level to reduce backend load and improve response times. While most API gateways support request-based caching (based on request attributes), semantic caching extends this concept to the AI/LLM domain by focusing on the meaning of requests rather than their syntax.

Some LLMs provide built-in caching, but these implementations often consume additional resources on the LLM backend. By introducing semantic caching at the gateway level, we can provide vendor-agnostic caching that works independently of the LLM provider. This approach guarantees better resource utilization but also supports multi-LLM backends where a single API may route requests to different LLMs. Gateway-level caching enhances performance, offering faster responses and reducing backend processing costs.

Technical Considerations for Semantic Caching
In semantic caching, cache keys are determined by the meaning of a request rather than its exact syntax. This involves using embeddings, metadata, and context to capture differences in AI behavior and ensure accurate matches for similar requests.

To implement semantic caching(storages):
- Vector Databases: Store and manage embeddings for semantic similarity searches.
- In-Memory Stores (e.g., Redis): Handle fast lookups for associated metadata and cache entries.
- Scalable Object Storage: Manage large payloads, such as LLM responses, efficiently.

mathetake:
let's discuss this API after the basic API definitions are in (like the one transformation, token rate limit and upstream auth stuff) - hopefully start discussing by the end of the year

Krishanx92:
Design : https://github.com/envoyproxy/ai-gateway/issues/30

PR #20, opened 2024-12-03
api: initial skeleton of LLMRoute and LLMBackend

mathetake:
This adds the skeleton API of LLMRoute and LLMBackend.
These two resources would be the foundation for the future
iterations, such as authn/z, token-based rate limiting, 
schema transformation and more advanced thingy like #10 

Note: we might / will break APIs if necessity comes up until
the initial release.

part of #13 

cc @yuzisun


aabchoo (approved):

mathetake:
cc @arkodg

mathetake:
cc @sanjeewa-malalgoda (i would appreciate it if you can ping other folks from your org)

mathetake:
will add more concrete stuff so that this will be more interesting enough for reviews 😉

sanjeewa-malalgoda:
@mathetake We reviewed this and it looks good for us.

sanjeewa-malalgoda (approved):

Krishanx92 (approved):

api/v1alpha1/api.go
```diff
@@ -1,3 +1,118 @@
...
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []LLMRoute `json:"items"`
+}
+
+// LLMRouteSpec details the LLMRoute configuration.
+type LLMRouteSpec struct {
+	// APISchema specifies the API schema of the input that the target Gateway(s) will receive.
+	// Based on this schema, the ai-gateway will perform the necessary transformation to the
+	// output schema specified in the selected LLMBackend during the routing process.
+	APISchema LLMAPISchema `json:"inputSchema"`
```
mathetake, line 39:
```suggestion
	APISchema LLMAPISchema `json:"apiSchema"`
```

api/v1alpha1/api.go
```diff
@@ -1,3 +1,118 @@
...
+}
+
+// LLMBackendSpec details the LLMBackend configuration.
+type LLMBackendSpec struct {
+	// APISchema specifies the API schema of the output that this LLMBackend can accept.
+	// Based on this schema, the ai-gateway will perform the necessary transformation for
+	// the pair of LLMRoute.APISchema and LLMBackend.APISchema.
+	//
+	// This is required to be set.
+	APISchema LLMAPISchema `json:"outputSchema"`
```
mathetake, line 86:
```suggestion
	APISchema LLMAPISchema `json:"apiSchema"`
```

mathetake (commented):
Reviewed myself🏃

api/v1alpha1/api.go
yuzisun: Are we going to allow different APISchema for different LLM route if we define at the route level?

api/v1alpha1/api.go
```diff
@@ -1,3 +1,118 @@
...
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []LLMBackend `json:"items"`
+}
+
+// LLMBackendSpec details the LLMBackend configuration.
+type LLMBackendSpec struct {
+	// APISchema specifies the API schema of the output that this LLMBackend can accept.
+	// Based on this schema, the ai-gateway will perform the necessary transformation for
+	// the pair of LLMRoute.APISchema and LLMBackend.APISchema.
```
yuzisun, line 83: I am not sure if we want to support arbitrary transformation between any two of the APISchemas

api/v1alpha1/api.go
Krishanx92: If my understanding is correct, this indicates which vendor and model the route belongs to. Therefore, a single LLMRoute can correspond to only one vendor and model

mathetake: > Are we going to allow different APISchema for different LLM route if we define at the route level?

I think the answer would be no at the moment since i am not sure why a user wants to access API like that - clients need to use different set of API clients and switching it from their end depending on a path? idk.

api/v1alpha1/api.go
mathetake: https://github.com/envoyproxy/ai-gateway/pull/20/commits/865ec35246232ff7ef24bffa0411798bba50fd91 done, thanks for the comment!

mathetake:
on second thought i kept the field name `inputSchema` and `outputSchema` for LLMRoute and LLMBackend respectively, instead of the same `apiSchema` as I left comment by myself yesterday. I think it's almost ready to go

zirain (approved):

api/v1alpha1/api.go
```diff
@@ -1,3 +1,122 @@
...
+	Version string `json:"version,omitempty"`
+}
+
+// APISchema defines the API schema.
+type APISchema string
+
+const (
+	// APISchemaOpenAI is the OpenAI schema.
+	//
+	// https://platform.openai.com/docs/overview
```
codefromthecrypt, line 116: In the case of OpenAI, they publish an OpenA*P*I spec, so maybe that's the more helpful link?

```suggestion
	// https://github.com/openai/openai-openapi
```

manifests/charts/ai-gateway-helm/crds/aigateway.envoyproxy.io_llmroutes.yaml
```diff
@@ -0,0 +1,149 @@
...
+                  type: object
+                maxItems: 128
+                type: array
+              inputSchema:
+                description: |-
+                  APISchema specifies the API schema of the input that the target Gateway(s) will receive.
+                  Based on this schema, the ai-gateway will perform the necessary transformation to the
+                  output schema specified in the selected LLMBackend during the routing process.
+
+                  Currently, the only supported schema is OpenAI as the input schema.
```
codefromthecrypt, line 74: would it be less problematic to remove this text constraint and introduce bedrock later once supported?

codefromthecrypt (commented):
drive-by notes

manifests/charts/ai-gateway-helm/crds/aigateway.envoyproxy.io_llmroutes.yaml
mathetake: the constraint will be enforced in the cel validation rule that happens at k8s API server - will do the follow up soon

mathetake: good comment anyways adrian as usual

PR #27, opened 2024-12-05
Adds basic translator and api schemas

mathetake:
This migrates the translation logic from the PoC code in poc-showcasing
branch. Nothing special in here - just the manual translation plus the 
barebone API definitions for each OpenAI and AWS Bedrock.

Just in case anyway curious about why not using OpenAPI generator for
the OpenAI part, here's the comment in the apischema/openai package:
```
// Package openai contains the followings are the OpenAI API schema definitions.
// Note that we intentionally do not use the code generation tools like OpenAPI Generator not only to keep the code simple
// but also because the OpenAI's OpenAPI definition is not compliant with the spec and the existing tools do not work well.
```

Especially, the OpenAPI schema is a bit tricky especially on the streaming=true/false
conditionals. Given that only a few endpoints will be supported here anyways,
it's not that problematic to have the manually written data structure.


internal/extproc/translator/openai_openai.go
```diff
@@ -0,0 +1,102 @@
...
+	if req.Stream {
+		o.stream = true
+		override = &extprocv3http.ProcessingMode{
+			ResponseHeaderMode: extprocv3http.ProcessingMode_SEND,
+			ResponseBodyMode:   extprocv3http.ProcessingMode_STREAMED,
+		}
+	}
+	headerMutation = &extprocv3.HeaderMutation{
+		SetHeaders: []*corev3.HeaderValueOption{{Header: &corev3.HeaderValue{Key: "host", Value: "api.openai.com"}}},
+	}
```
wengyao04, line 52: could we allow user to rewrite the host instead of force to api.openai.com ?

mathetake: good catch - actually this needs to be handled in a different way

mathetake: https://github.com/envoyproxy/ai-gateway/pull/27/commits/33c5c4183f7d8476e7cb688623da2e434a02f4c3

mathetake:
Will merge next Monday unless someone screams!

Krishanx92:
LGTM. Create https://github.com/envoyproxy/ai-gateway/issues/37 to discussed concerns separately.

PR #35, opened 2024-12-06
add LLMBackendTrafficPolicy

wengyao04:
- add LLMBackendTrafficPolicy, which controls the flow of traffic to the backend.
- add ratelimit in LLMBackendTrafficPolicy


api/v1alpha1/api.go
```diff
@@ -123,3 +123,204 @@ const (
...
+	//
+	// +optional
+	// +kubebuilder:validation:MaxLength=1024
+	Value *string `json:"value"`
+}
+
+// LLMPolicyRateLimitHeaderMatch defines the match attributes within the HTTP Headers of the request.
+type LLMPolicyRateLimitHeaderMatch struct {
+	// Type specifies how to match against the value of the header.
+	Type LLMPolicyRateLimitStringMatchType `json:"type"`
```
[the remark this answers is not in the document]
wengyao04: enum is defined in
```
// +kubebuilder:validation:Enum=Exact;RegularExpression;Distinct
type LLMPolicyRateLimitStringMatchType string
```

api/v1alpha1/api.go
```diff
@@ -123,3 +123,138 @@ const (
...
+	// Headers is a list of request headers to match. Multiple header values are ANDed together,
+	// meaning, a request MUST match all the specified headers.
+	// At least one of headers or sourceCIDR condition must be specified.
+	Headers []LLMPolicyRateLimitHeaderMatch `json:"headers,omitempty"`
+	// +kubebuilder:validation:MinItems=1
+	Limits []LLMPolicyRateLimitValue `json:"limits"`
+}
+
+// LLMPolicyRateLimitHeaderMatch defines the match attributes within the HTTP Headers of the request.
+type LLMPolicyRateLimitHeaderMatch struct {
```
yuzisun, line 173: Can we reuse the generic envoy gateway headerMatch type?

https://github.com/envoyproxy/gateway/blob/14fb56e9b01da0aa333945be4caafb35cf2b7fbb/api/v1alpha1/ratelimit_types.go#L142

mathetake:
sorry the CI has a bug https://github.com/envoyproxy/ai-gateway/pull/36 this will fix it 🙏

api/v1alpha1/api.go
```diff
@@ -123,3 +123,138 @@ const (
...
+type LLMTrafficPolicyRateLimit struct {
+	// Rules defines the rate limit rules.
+	Rules []LLMTrafficPolicyRateLimitRule `json:"rules,omitempty"`
+}
+
+// LLMTrafficPolicyRateLimitRule defines the details of the rate limit policy.
+type LLMTrafficPolicyRateLimitRule struct {
+	// Headers is a list of request headers to match. Multiple header values are ANDed together,
+	// meaning, a request MUST match all the specified headers.
+	// At least one of headers or sourceCIDR condition must be specified.
```
yuzisun, line 166: it is not matching by sourceCIDR here, we can also document the canonical header such as `x-ai-gateway-llm-model-name` used to apply the rate limiting.

api/v1alpha1/api.go
```diff
@@ -123,3 +123,138 @@ const (
...
+	Items           []LLMBackendTrafficPolicy `json:"items"`
+}
+
+// LLMBackendTrafficPolicySpec defines the details of llm backend traffic policy
+// like rateLimit, timeout etc.
+type LLMBackendTrafficPolicySpec struct {
+	// BackendRefs lists the LLMBackends that this traffic policy will apply
+	// The namespace is "local", i.e. the same namespace as the LLMRoute.
+	//
+	BackendRef LLMBackendLocalRef `json:"backendRef,omitempty"`
```
aabchoo, line 152: Do we want to have the option to have the traffic policy apply to the multiple backends? Ie two very similar models that require the same rate limit.

api/v1alpha1/api.go
mathetake: yeah, I would like to reuse EG native type as much as possible

mathetake:
marked this as a draft since I landed the upstream generic rate limit feature based on the response content in Envoy upstream: https://github.com/envoyproxy/envoy/pull/37548 - and the corresponding feature in EG is being worked on https://github.com/envoyproxy/gateway/pull/4957. Once they are all done, then the token rate limit API is not needed at ai-gateway at all

PR #39, opened 2024-12-09
api: make LLMRoute reference HTTPRoute

mathetake:
This commit is a follow up on #20. Basically, this makes LLMRoute 
a pure "addition" to the existing standardized HTTPRoute. 
This makes it possible to configure something like 
```
kind: LLMRoute
metadata:
  name: llm-route
spec:
  inputSchema: OpenAI
  httpRouteRef:
    name: my-llm-route
---
kind: HTTPRoute
metadata:
  name: my-llm-route
spec:
  matches:
     - headers:
         key: x-envoy-ai-gateway-llm-model
         value: llama3-70b 
       backendRefs: 
       - kserve:
         weight: 20
       - aws-bedrock:
         weight: 80
```

where LLMRoute is purely referencing HTTPRoute and 
users can configure whatever routing condition in a standardized way
via HTTPRoute while leveraging the LLM specific information, in this case
x-envoy-ai-gateway-llm-model header.

In the implementation, though it's not merged yet, we have to do the 
routing calculation in the extproc by actually analyzing the referenced 
HTTPRoute, and emulate the behavior in order to do the transformation.
The reason is that the routing decision is made at the very end of filter chain 
in general, and by the time we invoke extproc, we don't have that info.
Furthermore, `x-envoy-ai-gateway-llm-model` is not available before extproc.

As a bonus of this, we no longer need TargetRef at LLMRoute level since that's within 
the HTTPRoute resources. This will really simplify the PoC implementation.


api/v1alpha1/api.go
```diff
@@ -91,6 +101,11 @@ type LLMBackendSpec struct {
 	//
 	// This is required to be set.
 	APISchema LLMAPISchema `json:"outputSchema"`
+	// BackendRef is the reference to the Backend resource that this LLMBackend corresponds to.
+	//
+	// The backend can be of either k8s Service or [egv1a1.Backend] kind plus the resource must be in the same namespace as the LLMRoute.
+	// For the sake of simplicity, currently the same backend cannot be referenced by multiple LLMBackend(s).
+	BackendRef gwapiv1.BackendObjectReference `json:"backendRef"`
```
mathetake, line 108: oh man this is unrelated change - will delete

api/v1alpha1/api.go
```diff
@@ -43,22 +46,21 @@ type LLMRouteSpec struct {
...
-	// +optional
-	// +kubebuilder:validation:MaxItems=128
-	TargetRefs []gwapiv1a2.LocalPolicyTargetReferenceWithSectionName `json:"targetRefs"`
-	// BackendRefs lists the LLMBackends that this LLMRoute will route traffic to.
-	// The namespace is "local", i.e. the same namespace as the LLMRoute.
+	// In the matching configuration of the referenced HTTPRoute, `x-envoy-ai-gateway-llm-model` header
+	// can be used to describe the routing behavior.
 	//
-	// +kubebuilder:validation:MaxItems=128
-	BackendRefs []LLMBackendLocalRef `json:"backendRefs,omitempty"`
```
Krishanx92, line 56: If we follow this approach do we need LLMBackend Route? we can utilize the Backend CR coming from gateway itself. and missing backend apischema can handle by checking the LLMModelHeaderKey from Httproute.

mathetake: not only the schema but another backend-specific configs are needed anyways - rate limit, upstream authz, etc.

api/v1alpha1/api.go
```diff
@@ -43,22 +46,21 @@ type LLMRouteSpec struct {
 	// +kubebuilder:validation:Required
 	// +kubebuilder:validation:XValidation:rule="self.schema == 'OpenAI'"
 	APISchema LLMAPISchema `json:"inputSchema"`
-	// TargetRefs are the names of the Gateway resources this policy is being attached to.
+	// HTTPRouteRef is the name of the HTTPRoute resource that the Gateway will use to route the traffic.
```
yuzisun, line 49: One downside of `HTTPRouteRef` is that we lose the owner reference between LLMRoute and HTTPRoute

mathetake: https://github.com/envoyproxy/ai-gateway/pull/39/commits/eecdbb873b8a20bda944224fb979addb8f120594 agree. as discussed offline, I went ahead and made a change to use HTTPRouteSpec

mathetake:
thanks @Krishanx92 @yuzisun for the reviews!

PR #40, opened 2024-12-11
api: adds BackendRef into LLMBackendSpec

mathetake:
This adds `backendRef` to LLMBackendSpec which specifies
the "backend" either Service or Backend resource of Envoy Gateway.

The choice of not embedding is intentional - A backend can be a target
of routing in HTTPRoute and can be either of these two types. Hence
this is not suitable for embedding. In addition, in the implementation, 
we won't directly use or reference them, but just simply attach the 
necessary logic by the names, so basically no benefit by doing so.


Krishanx92:
LGTM

Krishanx92 (approved):

yuzisun (approved):

Issue #29, opened 2024-12-05
Set up Docs Site


missBerg:
- [ ] Set up Site to build and deploy
- [ ] Set up URL to Site
- [ ] Set up Docs Site Section and templates

PR #41, opened 2024-12-12
Initial creation of the Envoy AI Gateway Site

missBerg:
- Simple landing page
- Links to get involved in community
- Basis for future Docs
- Blog setup, and a copy of the original announcement post from CNCF blog

We will need to set up the build and publish to github pages, and I've got the domain envoyaigateway.io for us to use.

Related to https://github.com/envoyproxy/ai-gateway/issues/29


mathetake (changes requested):
Remove `node_modules`? It's going to explode the repository.
Also Could you remove .DS_Store?

mathetake:
after that let's setup a GitHub Actions workflow that ensures it's buildable

.gitignore
```diff
@@ -2,4 +2,3 @@
 .bin
 out/
 /site/node_modules/.cache
-/site/node_modules
```
mathetake: the entire `node_modules` should be ignored, there's no reason to retain the entire copies of dependencies checked in the repo here

missBerg:
Got myself into a mess with unsigned commits, so I raised a new PR https://github.com/envoyproxy/ai-gateway/pull/46

PR #43, opened 2024-12-12
add BackendSecurityPolicy for traffic (authN/authZ) from gateway to provider

aabchoo:
- add backend security policy for traffic from gateway to backend provider
- introduces api key, oidc, and static key auth
- add reference to backendSecurityPolicy on llmbackend
- add provider type on llmbackend


api/v1alpha1/api.go
```diff
@@ -144,3 +144,71 @@ const (
...
+const (
+	LLMProviderTypeAPIKey LLMProviderType = "APIKey"
+)
+
+// +kubebuilder:object:root=true
+
+// LLMProviderPolicy specifies the provider specific configuration.
+//
+// This is a provider specific-configuration, e.g.AWS Bedrock, Azure etc.
+type LLMProviderPolicy struct {
```
mathetake, line 160: i wonder at this poinst if `LLMProviderPolicy` is a write naming... maybe SecurityPolicy? Do we have anything other than security related stuff to this CRD? cc @wengyao04 @envoyproxy/ai-gateway-assignable

aabchoo: At this moment, we are only setting auth. Not too familiar with what other provider specific config we would need. Alternatively to Security policy, we could revamp this to `LLMAuthPolicy` if nothing else is added? But it's very generic and not ai-gateway specific

Krishanx92: I also think SecurityPolicy is a better choice than ProviderPolicy because ProviderPolicy feels too generic atm. SecurityPolicy can handle not only authentication methods like auth, JWT, and API key, but it can also be extended to support future requirements like data masking or data protection algorithms etc. we may can keep LLM part as it is.

api/v1alpha1/api.go
```diff
@@ -144,3 +144,74 @@ const (
...
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+	Spec              LLMProviderPolicySpec `json:"spec,omitempty"`
+}
+
+// LLMProviderPolicySpec specifies a provider (e.g.AWS Bedrock, Azure etc.) specific-configuration like auth
+type LLMProviderPolicySpec struct {
+	// BackendRefs lists the LLMBackends that this provider policy will apply
+	// The namespace is "local", i.e. the same namespace as the LLMRoute.
+	//
+	BackendRefs []egv1a1.BackendRef `json:"backendRefs,omitempty"`
```
Krishanx92, line 169: Shouldn't this ref to LLMBackend? so we can remove providerType field as well.

api/v1alpha1/api.go
```diff
@@ -144,3 +144,74 @@ const (
...
+	SecretRef *gwapiv1.SecretObjectReference `json:"secretRef"`
+
+	// Inline specifies the inline API key.
+	//
+	// +optional
+	Inline *string `json:"inline,omitempty"`
+
+	// BackendRefs lists the LLMBackends that this API Key will apply
+	//
+	BackendRefs []egv1a1.BackendRef `json:"backendRefs"`
```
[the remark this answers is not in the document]
aabchoo: Good catch

api/v1alpha1/api.go
aabchoo: @mathetake do you have advice here? I saw that you removed the [LLMBackendRef in this commit](https://github.com/envoyproxy/ai-gateway/commit/8299827281dd4b92f26a3881744ca7687cca3fa1#diff-788b9806fbf2a26408470898083b955fff036495d42d54707301ed2d38137ad0L60-L63)

api/v1alpha1/api.go
aabchoo: I'll update this to be LLMSecurityPolicy if that works for everyone

api/v1alpha1/api.go
aabchoo: Going to update this so LLMBackend points to SecurityPolicy

aabchoo: Could you clarify why we should remove the providerType?

mathetake: yeah to be honest the existence of "Type" is just a convention following the EG APIs (e.g. `type` field of [HealthCheck](https://gateway.envoyproxy.io/latest/api/extension_types/#activehealthcheck).) For me personally I would rather remove it (making everything optional is enough).

api/v1alpha1/api.go
```diff
@@ -106,6 +106,10 @@ type LLMBackendSpec struct {
 	//
 	// +kubebuilder:validation:Required
 	BackendRef egv1a1.BackendRef `json:"backendRef"`
+
+	// LLMSecurityPolicyName list the LLMSecurityPolicy that this backend will depend on
+	//
+	LLMSecurityPolicyName *string `json:"llmSecurityPolicyName,omitempty"`
```
mathetake, line 112:
```suggestion
	securityPolicyName *string `json:"securityPolicyName,omitempty"`
```

api/v1alpha1/api.go
```diff
@@ -106,6 +106,10 @@ type LLMBackendSpec struct {
 	//
 	// +kubebuilder:validation:Required
 	BackendRef egv1a1.BackendRef `json:"backendRef"`
+
+	// LLMSecurityPolicyName list the LLMSecurityPolicy that this backend will depend on
+	//
```
mathetake, line 111: finish the sentence & cel validation?

api/v1alpha1/api.go
```diff
@@ -144,3 +148,65 @@ const (
...
+// LLMProviderType specifies the type of the LLMSecurityPolicy.
+type LLMProviderType string
+
+const (
+	LLMProviderTypeAPIKey LLMProviderType = "APIKey"
+)
+
+// +kubebuilder:object:root=true
+
+// LLMSecurityPolicy specifies the provider specific configuration.
```
mathetake, line 161: could you add more description?

api/v1alpha1/api.go
mathetake: btw I like the current form where LLMBackend referencing LLMSecurityPolicy rather than the other way around

api/v1alpha1/api.go
aabchoo: I've updated this, but I'm worried this will get confused with the envoy gateway security policy

api/v1alpha1/api.go
```diff
@@ -144,3 +150,65 @@ const (
...
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+	Spec              LLMSecurityPolicySpec `json:"spec,omitempty"`
+}
+
+// LLMSecurityPolicySpec specifies a provider (e.g.AWS Bedrock, Azure etc.) specific-configuration like auth
+type LLMSecurityPolicySpec struct {
+	// Type specifies the type of the provider. Currently, only "APIKey" and "AWSBedrock" are supported.
+	//
+	// +kubebuilder:validation:Enum=APIKey;AWSBedrock
+	Type LLMProviderType `json:"type"`
```
yuzisun, line 175: I think this is not provider type, it is the authentication mechanism with APIKey or AWS Auth.
"API Key" is a simple authentication method used to grant access to an API, while "AWS Auth" refers to the broader system of authentication within Amazon Web Services (AWS)

api/v1alpha1/api.go
```diff
@@ -144,3 +150,65 @@ const (
...
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+	Spec              LLMSecurityPolicySpec `json:"spec,omitempty"`
+}
+
+// LLMSecurityPolicySpec specifies a provider (e.g.AWS Bedrock, Azure etc.) specific-configuration like auth
+type LLMSecurityPolicySpec struct {
+	// Type specifies the type of the provider. Currently, only "APIKey" and "AWSBedrock" are supported.
+	//
+	// +kubebuilder:validation:Enum=APIKey;AWSBedrock
```
yuzisun, line 174: suggest `APIKey` and `AWS_IAM`, AWSBedrock is a AWS service not authentication mechanism

api/v1alpha1/api.go
yuzisun: maybe the envoy gateway SecurityPolicy is already good enough, they just do not support AWS Auth
https://gateway.envoyproxy.io/contributions/design/security-policy/#features--api-fields

api/v1alpha1/api.go
```diff
@@ -144,3 +166,87 @@ const (
...
+	Inline *string `json:"inline,omitempty"`
+}
+
+// LLMProviderAPIKeyType specifies the type of LLMProviderAPIKey.
+type LLMProviderAPIKeyType string
+
+// LLMProviderStaticKey specifies the static access key and secret key.
+type LLMProviderStaticKey struct {
+	// AccessKey is the static access key.
+	AccessKey *string `json:"accessKey,omitempty"`
```
yuzisun, line 246: AccessKey is AWS specific, user can probably include it in the secret

api/v1alpha1/api.go
```diff
@@ -144,3 +166,87 @@ const (
...
+	// ai-gateway must be given the permission to read this secret.
+	// The key of the secret should be "apiKey".
+	//
+	// +optional
+	SecretRef *gwapiv1.SecretObjectReference `json:"secretRef"`
+
+	// Inline specifies the inline API key.
+	//
+	// +optional
+	Inline *string `json:"inline,omitempty"`
```
yuzisun, line 237: We probably do not want to support inline API key as it is a security leak.

api/v1alpha1/api.go
```diff
@@ -144,3 +166,87 @@ const (
...
+
+// BackendSecurityPolicy specifies configuration for authentication and authorization rules on the traffic
+// exiting the gateway to the backend.
+type BackendSecurityPolicy struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+	Spec              BackendSecurityPolicySpec `json:"spec,omitempty"`
+}
+
+// BackendSecurityPolicySpec specifies authentication and authorization rules on access the provider from the Gateway.
```
yuzisun, line 189: The authorization rules are checked from client to gateway, gateway to LLM provider is only authentication.

api/v1alpha1/api.go
```diff
@@ -106,8 +106,30 @@ type LLMBackendSpec struct {
...
+	// exiting the gateway to the backend.
+	//
+	// +optional
+	BackendSecurityPolicyName *string `json:"backendSecurityPolicyName,omitempty"`
+
+	// LLMProviderType is the provider type specifies the provider associated with the backend.
+	//
+	// +kubebuilder:validation:Enum=AWS;Unspecified
+	// +kubebuilder:default=Unspecified
+	ProviderType *LLMProviderType `json:"providerType"`
```
Krishanx92, line 122: Do we need this again? we can derive provider type from outputSchema(LLMAPISchema--> APISchema)

api/v1alpha1/api.go
```diff
@@ -144,3 +166,87 @@ const (
...
+	// +kubebuilder:validation:Enum=APIKey;StaticKey;OIDC
+	Type LLMProviderAuthenticationType `json:"type"`
+
+	// APIKey specific configuration. The API key will be injected into the Authorization header.
+	// +optional
+	APIKey *LLMProviderAPIKey `json:"apiKey,omitempty"`
+
+	// OIDC tokens are retrieved from the following configuration. The token and backend type will determine how Authorization is configured.
+	// +optional
+	OIDC *egv1a1.OIDC `json:"oidc,omitempty"`
```
zhaohuabing, line 202: API key and OIDC are not AI traffic specifc, is it possible to reuse the EG SecurityPolicy?

zhaohuabing: Oh, I see, this is for LLM Backend authentication, and EG SecurityPolicy is used for client side authentication, right?

mathetake: yea

api/v1alpha1/api.go
```diff
@@ -106,8 +106,30 @@ type LLMBackendSpec struct {
...
 	// +kubebuilder:validation:Required
 	BackendRef egv1a1.BackendRef `json:"backendRef"`
+
+	// BackendSecurityPolicyName list the BackendSecurityPolicy that this backend will depend on.
+	//
+	// A BackendSecurityPolicy is for configuring authentication and authorization rules on the traffic
+	// exiting the gateway to the backend.
+	//
+	// +optional
+	BackendSecurityPolicyName *string `json:"backendSecurityPolicyName,omitempty"`
```
zhaohuabing, line 116: Nit:
If Backend and BSP are 1:1, then BSP can be embedded here,
IF Backend and BSP are n:1, should we adopt the Gateway API way, with a `TargetRefs` list in the BTP referecing th Backend?

api/v1alpha1/api.go
zhaohuabing: I think this is for LLM Backend authentication with the provider, and EG SecurityPolicy is used to authenticate the clients/users.

api/v1alpha1/api.go
aabchoo: I wanted to keep them distinct because I didn't think the schemas are uniquely supported by each provider. Open to changing it if it's guaranteed schema == provider

api/v1alpha1/api.go
aabchoo: Won't the API key be used to check authorization in addition to authentication ie using AWS credentials for the gateway to access the LLM?

api/v1alpha1/api.go
Krishanx92: I think schema is provider specific atm, not sure will it change in future. @mathetake 
Also currently APISchema is string type which allow constant such as "OpenAI, AwsBedrock" It may redundant to keep both field.

yuzisun: There are cases that LLM provider adopts the OpenAI schema, so the provider name and APISchema name may not be necessarily the same. The provider determines the authentication mechanism and APISchema.

api/v1alpha1/api.go
zhaohuabing: @aabchoo Here is an example using EG SecurityPolicy:

```
kind: SecurityPolicy
metadata:
  name: oidc-example
spec:
  targetRefs:
      kind: HTTPRoute
      name: myapp
  oidc:
    provider:
      issuer: "https://accounts.google.com"
    clientID: "${CLIENT_ID}"
    clientSecret:
      name: "my-app-client-secret"
    redirectURL: "https://www.example.com:8443/myapp/oauth2/callback"
    logoutPath: "/myapp/logout"
```

API: https://github.com/envoyproxy/gateway/blob/main/api/v1alpha1/securitypolicy_types.go

api/v1alpha1/api.go
```diff
@@ -144,3 +166,74 @@ const (
...
+	// +optional
+	BackendRefs []egv1a1.BackendRef `json:"backendSecurityPolicyRefs,omitempty"`
+
+	// APIKey specific configuration. The API key will be injected into the Authorization header.
+	// +optional
+	APIKey *LLMProviderAPIKey `json:"apiKey,omitempty"`
+
+	// OIDC tokens are retrieved from the following configuration. The token and backend type will determine how Authorization is configured.
+	// +optional
+	OIDC *egv1a1.OIDC `json:"oidc,omitempty"`
```
zhaohuabing, line 207: I wonder can we use EG OIDC here?
The EG OIDC need users to provide their identity in the authentication flow, which is not applicable for the authentication at the backend service side.

yuzisun: @zhaohuabing you get the JWT authentication token from OIDC provider for exchanging the temporary AWS credential. User needs to provider the `client_id` and `client_secret` to get the OIDC auth token.
https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_providers_oidc.html

yuzisun: I think the authentication type here should be AWS IAM, the OIDC token is an intermediate step to get the JWT auth token for exchanging the AWS creds.

zhaohuabing: @yuzisun Thanks for the explaination! 

Are there a native envoy filter that supports the token exchange, or  we're handling this in the ext proc?

aabchoo:
Just some quick updates on the PR:
- Still using the `BackendSecurityPolicy` type
- Explicitly defining what is supported vs not (ie dedicated AWS security policy for oidc/credential file instead of a generic OIDC field)
- Introduced Credential file and OIDC auth for aws
- Authentication Types will either be `APIKey` or `AWS_IAM`
- Introduced targetRef in `LLMBackendSpec` instead of creating an array of backends in `BackendSecurityPolicy`

api/v1alpha1/api.go
```diff
@@ -106,8 +106,28 @@ type LLMBackendSpec struct {
 	//
 	// +kubebuilder:validation:Required
 	BackendRef egv1a1.BackendRef `json:"backendRef"`
+
+	// LLMProviderType is the provider type specifies the provider associated with the backend.
```
mathetake, line 110: how this will be used?

mathetake: I believe we don't need this at all as we have LLMProviderAuthenticationType, am i missing anything

api/v1alpha1/api.go
```diff
@@ -144,3 +164,130 @@ const (
...
+	// +kubebuilder:validation:Enum=APIKey;AWS_IAM
+	Type LLMProviderAuthenticationType `json:"type"`
+
+	// APIKey specific configuration. The API key will be injected into the Authorization header.
+	// +optional
+	APIKey *LLMProviderAPIKey `json:"apiKey,omitempty"`
+
+	// AwsSecurityPolicy specifies configuration to access aws via credential file and OIDC.
+	AwsSecurityPolicy *AWSSecurityPolicy `json:"awsSecurityPolicy,omitempty"`
+}
```
mathetake, line 199: this overall looks weird to me in terms of wording... on one hand, the top level crd lacks `LLM` prefixes so it has much broader namespace vs inside the spec, we have the top level `Type LLMProviderAuthenticationType `json:"type"` which makes this bound to only LLM context.

api/v1alpha1/api.go
aabchoo: This should be removed - thanks for catching that

api/v1alpha1/api.go
aabchoo: I've removed the LLM prefix since authentication isn't specific to LLMs. Hopefully that makes more sense.

api/v1alpha1/api.go
```diff
@@ -144,3 +150,128 @@ const (
...
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+	Spec              BackendSecurityPolicySpec `json:"spec,omitempty"`
+}
+
+// BackendSecurityPolicySpec specifies authentication rules on access the provider from the Gateway.
+type BackendSecurityPolicySpec struct {
+	// Type specifies the auth mechanism used to access the provider. Currently, only "APIKey", AND "AWS_IAM" are supported.
+	//
+	// +kubebuilder:validation:Enum=APIKey;AWS_IAM
+	Type ProviderAuthenticationType `json:"type"`
```
mathetake, line 177: can this be simply `BackendSecurityPolicyType` ? you use three words in here Provider Authentication and Security and should stick to consistent wording

mathetake: if `BackendSecurityPolicySpec` won't be only about authnz then this shouldn't be the top level - anyways the word "provider" here doesn't help anything i guess

api/v1alpha1/api.go
```diff
@@ -144,3 +150,128 @@ const (
...
+// ProviderAWSOIDCCredential specifies credentials to obtain oidc token from a sso server.
+// For AWS, the controller will query STS to obtain AWS AccessKeyId, SecretAccessKey, and SessionToken,
+// and store them in a temporary credentials file
+type ProviderAWSOIDCCredential struct {
+	// OIDC is used to obtain oidc tokens via an SSO server which will be used to exchange for temporary AWS credentials
+	OIDC egv1a1.OIDC `json:"oidc"`
+
+	// CredentialFileName is the name of the configmap that will store the temporary credentials
+	// file for the actual credentials will be `credentials`
+	CredentialFileName string `json:"credentialFileName"`
```
yuzisun, line 253: There is already credentialFile defined outside oidc type. I think it needs to be used in conjunction of credentialFile. Oidc federation is a way to exchange the aws creds, need to think a better name, maybe OIDCTokenExchange.

api/v1alpha1/api.go
```diff
@@ -144,3 +150,128 @@ const (
...
+	//
+	// +kubebuilder:validation:Enum=APIKey;AWS_IAM
+	Type ProviderAuthenticationType `json:"type"`
+
+	// APIKey specific configuration. The API key will be injected into the Authorization header.
+	// +optional
+	APIKey *ProviderAPIKey `json:"apiKey,omitempty"`
+
+	// AwsSecurityPolicy specifies configuration to access aws via credential file and OIDC.
+	AwsSecurityPolicy *AWSSecurityPolicy `json:"awsSecurityPolicy,omitempty"`
```
mathetake, line 184: could you use the consistent prefix for the things at the same abstraction - `ProviderAPIKey` and `AWSSecurityPolicy` are the same concrete implementation of `BackendSecurityPolicySpec`, so they should share the same prefix to indicate they are siblings. Currently when a user look at the these two types, they will have no idea how they are related and they are mutually exclusive abstract things

api/v1alpha1/api.go
```diff
@@ -144,3 +150,128 @@ const (
...
+	// +optional
+	Aud string `json:"aud,omitempty"`
+
+	// AwsRoleArn is the identifier for which role the user can access
+	//
+	// +optional
+	AwsRoleArn string `json:"awsRoleArn,omitempty"`
+
+	// AuthBearToken is an optional authorization token that can be passed with OIDC request
+	AuthBearToken string `json:"authBearToken,omitempty"`
```
yuzisun, line 271: Why do we need the bear token ?

api/v1alpha1/api.go
```diff
@@ -144,3 +150,128 @@ const (
...
+	// Profile is the profile to use in the credentials file.
+	//
+	// +kubebuilder:default=default
+	Profile string `json:"profile,omitempty"`
+}
+
+// ProviderAWSOIDCCredential specifies credentials to obtain oidc token from a sso server.
+// For AWS, the controller will query STS to obtain AWS AccessKeyId, SecretAccessKey, and SessionToken,
+// and store them in a temporary credentials file
+type ProviderAWSOIDCCredential struct {
```
yuzisun, line 247:
```suggestion
type AWSOIDCFederation struct {
```

yuzisun: Reference https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_providers_oidc.html

api/v1alpha1/api.go
```diff
@@ -144,3 +150,128 @@ const (
...
+	//
+	// +optional
+	GrantType string `json:"grantType,omitempty"`
+
+	// Aud defines the resource the application can access
+	//
+	// +optional
+	Aud string `json:"aud,omitempty"`
+
+	// AwsRoleArn is the identifier for which role the user can access
```
yuzisun, line 265:
```suggestion
	// AwsRoleArn is the AWS IAM Role with the permission to use specific resources in AWS account which maps to the temporary AWS security credentials exchanged using the authentication token issued by OIDC provider.
```

api/v1alpha1/api.go
```diff
@@ -144,3 +150,128 @@ const (
...
+	//
+	// +optional
+	AwsRoleArn string `json:"awsRoleArn,omitempty"`
+
+	// AuthBearToken is an optional authorization token that can be passed with OIDC request
+	AuthBearToken string `json:"authBearToken,omitempty"`
+
+	// SSORequestMethod specifies the expected http method for an SSO server
```
yuzisun, line 273: SSO is related to OIDC but they are different, OIDC is more specific and enables SSO functionality by providing  away to verify a user's identity and share user information. Essentially OIDC is an implementation of SSO.

api/v1alpha1/api.go
```diff
@@ -144,3 +150,128 @@ const (
...
+	//
+	// +kubebuilder:default=POST
+	SSORequestMethod string `json:"ssoRequestMethod,omitempty"`
```
yuzisun, line 276: according to https://openid.net/specs/openid-connect-core-1_0.html#TokenEndpoint, it should be always POST

api/v1alpha1/api.go
yuzisun: How about `APIKey` and `AWSCredentials` ?  We are going to add `AzureCredentials` and `GoogleCredentials` later.

api/v1alpha1/api.go
yuzisun: We probably do not need this field at all, we can validate with one of the provider authentication methods.

api/v1alpha1/api.go
```diff
@@ -144,3 +150,128 @@ const (
...
+}
+
+// ProviderAPIKey specifies the API key.
+type ProviderAPIKey struct {
+	// Type specifies the type of the API key. Currently, "SecretRef" is supported.
+	// This defaults to "SecretRef".
+	//
+	// +kubebuilder:validation:Enum=SecretRef
+	// +kubebuilder:default=SecretRef
+	Type ProviderAPIKeyType `json:"type"`
```
yuzisun, line 203: I think this is unnecessary, we only need the secret ref for API key

api/v1alpha1/api.go
mathetake: sorry, what I meant here is about the Go type `ProviderAPIKey` and `AWSSecurityPolicy`, not about the field name inside `json:...`. I am fine with the embedded field name not having prefix

api/v1alpha1/api.go
yuzisun: I agree that we can remove the word "provider" across board.

mathetake: @arkodg have some opinion about this - this is more generally about Gateway API's convention (hence EG) - I am completely fine with removing this "type" thingy but I am more concerned about the parity with EG

yuzisun: Do you have an example? I do not see the "type" field on EG SecurityPolicy
https://gateway.envoyproxy.io/docs/api/extension_types/#securitypolicyspec

mathetake: this is not specific about security policy but more about enum type switching - e.g. https://gateway.envoyproxy.io/docs/api/extension_types/#activehealthcheck having the "type" field to indicate a enum fields there - I personally think it's redundant but it's coming from Gateway API working group convention https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/1027-api-unions/README.md

mathetake: personally I want to remove 😉 but deferring to @arkodg

api/v1alpha1/api.go
mathetake: anyways I like "*Credentials" variants for cloud providers

api/v1alpha1/api.go
aabchoo: I used the bear tokens to access Github OIDC token when running GH actions to using [ACTIONS_ID_TOKEN_REQUEST_URL and ACTIONS_ID_TOKEN_REQUEST_TOKEN](https://docs.github.com/en/actions/security-for-github-actions/security-hardening-your-deployments/configuring-openid-connect-in-amazon-web-services#adding-permissions-settings). Removing it for now since this is not the expected way to access.

api/v1alpha1/api.go
aabchoo: Noted. I needed to use a GET method to access the action id token request but we can remove this to follow standards.

api/v1alpha1/api.go
aabchoo: I'll add the prefix `Authentication` to `CloudProviderCredentials` and `APIKey` to make it clear they're both authentication methods to access a backend

api/v1alpha1/api.go
aabchoo: I've updated the documentation to note that OIDCFederation required CredentialsFile (outside OIDC type) to be defined as well. What do you think of that? ([commit](https://github.com/envoyproxy/ai-gateway/pull/43/commits/5ff45c25b8489f354430080c7c3e2f45b0ab3fae#diff-788b9806fbf2a26408470898083b955fff036495d42d54707301ed2d38137ad0R227))

api/v1alpha1/api.go
```diff
@@ -144,3 +150,120 @@ const (
...
+	// Aud defines the resource the application can access.
+	//
+	// +optional
+	Aud string `json:"aud,omitempty"`
+
+	// AwsRoleArn is the AWS IAM Role with the permission to use specific resources in AWS account
+	// which maps to the temporary AWS security credentials exchanged using the authentication token issued by OIDC provider.
+	//
+	// +optional
+	AwsRoleArn string `json:"awsRoleArn,omitempty"`
```
yuzisun, line 268: This should be required field

yuzisun:
@mathetake @arkodg @zhaohuabing can you take another look?

api/v1alpha1/api.go
```diff
@@ -144,3 +150,120 @@ const (
...
+
+// BackendSecurityPolicyList contains a list of BackendSecurityPolicy
+type BackendSecurityPolicyList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []BackendSecurityPolicy `json:"items"`
+}
+
+// AuthenticationAPIKey specifies the API key.
+type AuthenticationAPIKey struct {
```
mathetake, line 200:
```suggestion
type BackendSecurityPolicyAPIKey struct {
```

api/v1alpha1/api.go
```diff
@@ -144,3 +150,120 @@ const (
...
+	// SecretRef is the reference to the secret containing the API key.
+	// ai-gateway must be given the permission to read this secret.
+	// The key of the secret should be "apiKey".
+	//
+	// +optional
+	SecretRef *gwapiv1.SecretObjectReference `json:"secretRef"`
+}
+
+// AuthenticationCloudProviderCredentials specifies supported cloud provider authentication methods
+type AuthenticationCloudProviderCredentials struct {
```
mathetake, line 210:
```suggestion
type BackendSecurityPolicyCloudProviderCredentials struct {
```

api/v1alpha1/api.go
```diff
@@ -144,3 +150,120 @@ const (
...
+	AWSCredentials AWSCredentials `json:"awsCredentials"`
```
mathetake, line 211: shouldn't this be an optional given there will be another?
```suggestion
	AWSCredentials *AWSCredentials `json:"awsCredentials,omitempty"`
```

api/v1alpha1/api.go
```diff
@@ -144,3 +150,120 @@ const (
 	// This can be used to describe the routing behavior in HTTPRoute referenced by LLMRoute.
 	LLMModelHeaderKey = "x-envoy-ai-gateway-llm-model"
 )
+
+// BackendSecurityPolicyType specifies the type of auth mechanism used to access a backend.
+type BackendSecurityPolicyType string
+
+const (
+	BackendSecurityPolicyTypeAPIKey BackendSecurityPolicyType = "APIKey"
+	BackendSecurityPolicyTypeAWSIAM BackendSecurityPolicyType = "AWS_IAM"
```
mathetake, line 159: can we have a consistent style?
```suggestion
	BackendSecurityPolicyTypeAPIKey BackendSecurityPolicyType = "APIKey"
	BackendSecurityPolicyTypeAWSIAM BackendSecurityPolicyType = "AWSIAM"
```

ack AWSIAM looks confusing though ... @arkodg @zhaohuabing how do you guys usually name types like this

arkodg: ha they are both acryonyms, so accoding to https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#constants it should be `AWSIAM`, but I agree this doesnt read well

mathetake: yeah let's leave it as-is!

mathetake (commented):
LGTM per nits above!

api/v1alpha1/api.go
arkodg: yeah for unions, EG uses `Type` which is references from Gateway API style which references from K8s :)  , would be nice if AI Gateway follows the same style

api/v1alpha1/api.go
```diff
@@ -144,3 +150,120 @@ const (
...
+}
+
+// AuthenticationAPIKey specifies the API key.
+type AuthenticationAPIKey struct {
+	// SecretRef is the reference to the secret containing the API key.
+	// ai-gateway must be given the permission to read this secret.
+	// The key of the secret should be "apiKey".
+	//
+	// +optional
+	SecretRef *gwapiv1.SecretObjectReference `json:"secretRef"`
```
mathetake, line 206:
```suggestion
	SecretRef *gwapiv1.SecretObjectReference `json:"secretRef,omitempty"`
```

api/v1alpha1/api.go
```diff
@@ -106,6 +106,12 @@ type LLMBackendSpec struct {
 	//
 	// +kubebuilder:validation:Required
 	BackendRef egv1a1.BackendRef `json:"backendRef"`
+
+	// BackendSecurityPolicyRef is the name of the BackendSecurityPolicy resources this backend
+	// is being attached to.
+	//
+	// +optional
+	BackendSecurityPolicyRef *gwapiv1.LocalObjectReference `json:"backendSecurityPolicyRef,omitempty"`
```
mathetake, line 114: can you add a CEL validation to ensure that both `apiKey` and `cloudProviderCredentials` won't be defined at the same time

api/v1alpha1/api.go
```diff
@@ -144,3 +150,120 @@ const (
...
+
+	// APIKey is a mechanism to access a backend(s). The API key will be injected into the Authorization header.
+	//
+	// +optional
+	APIKey *AuthenticationAPIKey `json:"apiKey,omitempty"`
+
+	// CloudProviderCredentials is a mechanism to access a backend(s). Cloud provider specific logic will be applied.
+	//
+	// +optional
+	CloudProviderCredentials *AuthenticationCloudProviderCredentials `json:"cloudProviderCredentials,omitempty"`
```
mathetake, line 187: so i wonder what kind of logic will make cloud providers special compared to others (currently only apiKey though). how do you envision this additional indirect layer will be used? in other words, what kind of logic will be shared among cloud providers

yuzisun: - ApiKey is simple authentication mechanism distributed to user and user is responsible to rotate the keys. 
- Cloud authentication is more sophisticated and support issuing temporary tokens like AWS STS, Google STS to access cloud resources. You do not have to distribute the token and major clouds also support OIDC federation to allow managing your user identity in external system.
see https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp.html#:~:text=You%20can%20use%20the%20AWS,has%20permissions%20to%20do%20so.

mathetake: yeah i get that it's complex, but my question is why this another layer is needed and what kind of logic will be shared among cloud providers? what becomes difficult to achieve when you have `*AWSCredentials` directly here without another layer of `AuthenticationCloudProviderCredentials`. How does having `AuthenticationCloudProviderCredentials` helps implement various cloud providers vs having them in here

api/v1alpha1/api.go
```diff
@@ -144,3 +150,120 @@ const (
...
+type AWSCredentials struct {
+	// Region specifies the AWS region associated with the policy.
+	//
+	// +kubebuilder:validation:MinLength=1
+	Region string `json:"region"`
+
+	// CredentialsFile specifies the credentials file to use for the AWS provider.
+	//
+	// +optional
+	CredentialsFile *AWSCredentialsFile `json:"credentialsFile,omitempty"`
```
zhaohuabing, line 224: Should a SecretRef be used instead of a loacal file here?

The SecretRef appraoch is easier to deploy and more secure compared with a file.

api/v1alpha1/api.go
```diff
@@ -144,3 +150,120 @@ const (
...
+	OIDCExchangeToken *AWSOIDCExchangeToken `json:"oidcExchangeToken,omitempty"`
+}
+
+// AWSCredentialsFile specifies the credentials file to use for the AWS provider.
+// Envoy reads the credentials from the file pointed by the Path field, and the profile to use is specified by the Profile field.
+type AWSCredentialsFile struct {
+	// Path is the path to the credentials file.
+	//
+	// +kubebuilder:default=~/.aws/credentials
+	Path string `json:"path,omitempty"`
```
zhaohuabing, line 239: Should a SecretRef be used instead of a loacal file here?

The SecretRef appraoch is easier to deploy and more secure compared with a file.

api/v1alpha1/api.go
aabchoo: I set originally this as optional because users can technically set it as an env. Updated now

api/v1alpha1/api.go
aabchoo: It's the only option at the moment, so I wanted to make it required. Marking it optional

api/v1alpha1/api.go
aabchoo: Behind the scenes the file should be a mounted secret. The path should equal the mount path of the secret plus `/credentials`. What do you think of that?

aabchoo: For context, the path was needed because of envoy aws signing filters. I don't think that's the way we will handle signing now, so I will update this to secretRefs. That will make it easier like you said for the external proc to interact with the resource.

api/v1alpha1/api.go
yuzisun: This is mainly a yaml UX, the additional layer for grouping is there for readability, if we list the individual types then it is not intuitive for user to see the classification. The cloud provider security implementations are different but they share the same high level concepts.

api/v1alpha1/api.go
```diff
@@ -205,3 +211,117 @@ const (
...
+	Type BackendSecurityPolicyType `json:"type"`
+
+	// APIKey is a mechanism to access a backend(s). The API key will be injected into the Authorization header.
+	//
+	// +optional
+	APIKey *BackendSecurityPolicyAPIKey `json:"apiKey,omitempty"`
+
+	// CloudProviderCredentials is a mechanism to access a backend(s). Cloud provider specific logic will be applied.
+	//
+	// +optional
```
mathetake, line 250: could you add comment about AWS_IAM will be interpreted to this field?

mathetake (approved):

api/v1alpha1/api.go
mathetake: well, shouldn't the type be "CloudProviders" instead of AWS_IAM? The "type" field is for union type assertion and this feels a weird because AWS_IAM is not for the union type assertion but for the higher layer. @yuzisun @arkodg

api/v1alpha1/api.go
```diff
@@ -205,3 +211,117 @@ const (
...
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+	Spec              BackendSecurityPolicySpec `json:"spec,omitempty"`
+}
+
+// BackendSecurityPolicySpec specifies authentication rules on access the provider from the Gateway.
+//
+// Only one type of BackendSecurityPolicy can be defined.
+// +kubebuilder:validation:MaxProperties=2
+type BackendSecurityPolicySpec struct {
+	// Type specifies the auth mechanism used to access the provider. Currently, only "APIKey", AND "AWS_IAM" are supported.
```
mathetake, line 238: well, shouldn't the type be "CloudProviders" instead of AWS_IAM? The "type" field is for union type assertion and this feels a weird because AWS_IAM is not for the union type assertion but for the higher layer. @yuzisun @arkodg

mathetake: This is relevant to the thread to https://github.com/envoyproxy/ai-gateway/pull/43#discussion_r1902265996 I am still not convinced by the "Yaml UX" stuff there. I am fine with that but at least to be consistent could you make this CloudProviders and have a type inside it for a specific cloud provider? (which i think is a worse UX vs the flat one)

aabchoo: I've flattened it out per req

mathetake (approved):
LGTM. Thank you for your hard word and multiple iterations here! @aabchoo 💯

PR #47, opened 2024-12-18
api: do not embed HTTPRoute as-is

mathetake:
This is a follow up on #39, and stops embedding HTTPRoute as-is.
The rationale here is somewhat complicated; We have to know which
backend we route traffics to **before** Envoy decides the routing as
we need to do perform the transformation etc. Hence, we need to 
implement the routing logics (e.g. header matching, weights, etc.) 
by ourselves, and cannot rely on the native router code path in Envoy.
That results in us only being able to support only a small subset of 
HTTPRoute functionality as well as we cannot simply use the embedded
HTTPRoute as a "base" in the translation. If we embed HTTPRoute here, 
it would create an impression that  all HTTPRoute functionality can be 
supported  here no matter how much we document otherwise.

We discussed about this issue, and reached the consensus that we should
have our own LLMRoute rule definition and only have the fields that 
actually can be supported.

The example API would look like;
```
kind: LLMRoute
metadata:
  name: 
  namespace: default
spec:
  inputSchema:
    schema: OpenAI
  rules:
    - matches:
      - headers:
        - type: Exact
          name: x-envoy-ai-gateway-llm-model
          value: llama3-70b
      backendRefs:
        - name: kserve
          weight: 20
        - name: aws-bedrock
          weight: 80
```


api/v1alpha1/api.go
```diff
@@ -47,18 +47,68 @@ type LLMRouteSpec struct {
...
+	// Weight is the weight of the LLMBackend. This is exactly the same as the weight in
+	// the BackendRef in the Gateway API. See for the details:
+	// https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io%2fv1.BackendRef
+	//
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Minimum=0
+	Weight int `json:"weight"`
+}
+
+type LLMRouteRuleMatch struct {
```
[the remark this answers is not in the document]
mathetake: in the future, yes - but not sure whether it would help simplify the transformer or not

mathetake:
after this i finally will work on implementing the initial controller/extproc code

Krishanx92 (approved):

api/v1alpha1/api.go
```diff
@@ -47,18 +47,68 @@ type LLMRouteSpec struct {
...
+	// Rules is the list of LLMRouteRule that this LLMRoute will match the traffic to.
+	// Each rule is a subset of the HTTPRoute in the Gateway API (https://gateway-api.sigs.k8s.io/api-types/httproute/).
 	//
-	// In the matching rules in the HTTPRoute here, `x-envoy-ai-gateway-llm-model` header
-	// can be used to describe the routing behavior.
+	// AI Gateway controller will generate a HTTPRoute based on the configuration given here with the additional
+	// modifications to achieve the necessary jobs, notably inserting the AI Gateway external processor filter.
 	//
-	// Currently, only the exact header matching is supported, otherwise the configuration will be rejected.
+	// In the matching conditions in the LLMRouteRule, `x-envoy-ai-gateway-llm-model` header
```
yuzisun, line 56: we also match by `x-envoy-ai-gateway-llm-backend` and `x-envoy-ai-gateway-llm-model` is used for a more specific model level match. When user add these rules, they need to put more specific rules on the top right?

mathetake: well who sets `x-envoy-ai-gateway-llm-backend` in that case? Users are responsible setting the header and then they can do whatever they want not limited to the specific key `x-envoy-ai-gateway-llm-backend`?

yuzisun: ye the key name does not matter, just to clarify the priority when there are multiple matches as we need to implement this in the extproc.

api/v1alpha1/api.go
```diff
@@ -47,18 +47,68 @@ type LLMRouteSpec struct {
...
-	// HTTPRoute is the base HTTPRouteSpec (https://gateway-api.sigs.k8s.io/api-types/httproute/) in
-	// the Gateway API on which this LLMRoute will be implemented. AI Gateway controller will generate a HTTPRoute based
-	// on the configuration given here with the additional modifications to achieve the necessary jobs,
-	// notably inserting the AI Gateway external processor filter.
+	// Rules is the list of LLMRouteRule that this LLMRoute will match the traffic to.
+	// Each rule is a subset of the HTTPRoute in the Gateway API (https://gateway-api.sigs.k8s.io/api-types/httproute/).
 	//
-	// In the matching rules in the HTTPRoute here, `x-envoy-ai-gateway-llm-model` header
-	// can be used to describe the routing behavior.
+	// AI Gateway controller will generate a HTTPRoute based on the configuration given here with the additional
```
yuzisun, line 53: How are we going to generate the HTTPRoute as certain feature like traffic split can't be used on the underlying HTTPRoute ? I guess we can only translate a subset of the matching rules from `LLMRouteRules` to `HTTPRouteRules` which does not include weights etc.

mathetake: i am thinking we will perform traffic splitting at the extproc level, and chose the backend

yuzisun: If I understand correctly, the ext proc resolves the llm backend using the matching rules and the HttpRoute resource should be just created as simple as routing to the backends with the backend http header. Is that correct ?

mathetake: exactly, that's what i am thinking to implement this thingy

api/v1alpha1/api.go
mathetake: https://github.com/envoyproxy/ai-gateway/pull/47/commits/d7bf9126ba0e6901679048e8f932f750a1ce7100 looks good?

yuzisun (approved):

Issue #28, opened 2024-12-05
ci: add `codespell` for spell checks


mathetake:
Following the EG style maybe:
https://github.com/envoyproxy/gateway/tree/main/tools/src/codespell
https://github.com/envoyproxy/gateway/tree/main/tools/linter/codespell
https://github.com/envoyproxy/gateway/blob/905405e72ed5da070ef2d041e524efb8511d837f/tools/make/lint.mk#L33
https://github.com/envoyproxy/gateway/blob/905405e72ed5da070ef2d041e524efb8511d837f/tools/make/tools.mk#L35

PR #48, opened 2024-12-18
chore: enable codespell check

zhaohuabing:
Fixes: https://github.com/envoyproxy/ai-gateway/issues/28

### DIFF
--- a/.codespell.skip
+++ b/.codespell.skip
@@ -1,0 +1,1 @@
+./manifests/charts/ai-gateway-helm/crds/*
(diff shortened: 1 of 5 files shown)


.codespell.skip
```diff
@@ -0,0 +1 @@
+./manifests/charts/ai-gateway-helm/crds/*
```
mathetake, line 1: if this is needed, then the spell check seems not running properly on the code comments in api.go? Since the typo `maxmimum` in api.go is the reason why https://github.com/envoyproxy/ai-gateway/actions/runs/12387688325/job/34577531455 is failing i guess?

zhaohuabing: This comes from upstream Gateway HTTPRoute.

mathetake: aha ok, now that makes sense... thank you for clarifying huabing!

mathetake (approved):

PR #50, opened 2024-12-25
extproc: initial implementation

mathetake:
This scaffolds and implements the initial logic based on the 
current API. Notably, I intentionally organized the extproc code 
in a way that it is decoupled from EG concepts as well as the 
AI Gateway API level constructs. The rationale is that it not only
allows us to test the extproc without relying on not only EG/AIG
and even without kubernetes at all, but also some users are asking
for the use of outside the ai-gateway projects.

As for the implementation, there's nothing special going on, and 
it is simply mostly migrated from the MVP with the decoupling in mind.

I will follow up with the end-to-end test in subsequent PRs.

### DIFF
--- a/extprocconfig/extprocconfig.go
+++ b/extprocconfig/extprocconfig.go
@@ -8,6 +8,14 @@ package extprocconfig
 
 import gwapiv1 "sigs.k8s.io/gateway-api/apis/v1"
 
+// Config is the configuration for the external processor.
+type Config struct {
+	InputSchema             VersionedAPISchema `json:"inputSchema"`
+	ModelNameHeaderKey      string             `json:"modelNameHeaderKey"`
+	BackendRoutingHeaderKey string             `json:"backendRoutingHeaderKey"`
+	Rules                   []RouteRule        `json:"rules"`
+}
+
 // VersionedAPISchema corresponds to LLMAPISchema in api/v1alpha1/api.go.
 type VersionedAPISchema struct {
 	Schema  APISchema
@@ -22,14 +30,6 @@ const (
 	APISchemaAWSBedrock APISchema = "AWSBedrock"
 )
 
-// Config is the configuration for the external processor.
-type Config struct {
-	InputSchema             VersionedAPISchema
-	ModelNameHeaderKey      string
-	BackendRoutingHeaderKey string
-	Rules                   []RouteRule
-}
-
 // HeaderMatch is an alias for HTTPHeaderMatch of the Gateway API.
 type HeaderMatch = gwapiv1.HTTPHeaderMatch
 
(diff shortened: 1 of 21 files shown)


mathetake:
the left is routing code, but almost there!

mathetake:
nothing special going on as I described in the PR description, so I need a rubber stamping i guess 😉

mathetake:
FYI @nacx

extprocconfig/extprocconfig.go
zirain, line 1: prefer to move this out of internal.

internal/cmd/extproc/main.go
zirain, line 1: prefer to move this out of internal.

missBerg (approved):

extprocconfig/extprocconfig.go
```diff
@@ -0,0 +1,121 @@
...
+	// InputSchema specifies the API schema of the input format of requests to the external processor.
+	InputSchema VersionedAPISchema `yaml:"inputSchema"`
+	// ModelNameHeaderKey is the header key to be populated with the model name by the external processor.
+	ModelNameHeaderKey string `yaml:"modelNameHeaderKey"`
+	// BackendRoutingHeaderKey is the header key to be populated with the backend name by the external processor
+	// **after** the routing decision is made by the external processor using Rules.
+	BackendRoutingHeaderKey string `yaml:"backendRoutingHeaderKey"`
+	// Rules is the routing rules to be used by the external processor to make the routing decision.
+	// Inside the routing rules, the header ModelNameHeaderKey may be used to make the routing decision.
+	Rules []RouteRule `yaml:"rules"`
```
yuzisun, line 65: Why the rules are configured in the extproc configmap ?

yuzisun: these rules are expect to be updated quite frequently as users add or remove models endpoints, or we are translating the LLMroute to the config for extproc ?

extprocconfig/extprocconfig.go
```diff
@@ -0,0 +1,121 @@
...
+// at Envoy Gateway level to a simple name.
+type RouteRule struct {
+	// Headers is the list of headers to match for the routing decision.
+	// Currently, only exact match is supported.
+	Headers []HeaderMatch `yaml:"headers"`
+	// Backends is the list of backends to which the request should be routed to when the headers match.
+	Backends []Backend `yaml:"backends"`
+}
+
+// Backend corresponds to LLMRouteRuleBackendRef in api/v1alpha1/api.go
```
yuzisun, line 98: If I understand correctly the goal here for the config is not intended to be tied with AI concepts.

internal/extproc/router/router.go
```diff
@@ -0,0 +1,69 @@
...
+		return rule.Backends[0].Name, rule.Backends[0].OutputSchema
+	}
+	selected := r.rng.Intn(totalWeight)
+	for _, b := range rule.Backends {
+		if selected < b.Weight {
+			return b.Name, b.OutputSchema
+		}
+		selected -= b.Weight
+	}
+	return rule.Backends[0].Name, rule.Backends[0].OutputSchema
```
yuzisun, line 68: should return the selected backend

yuzisun (approved):
left some comments but good to merge to iterate based on this initial implementation

extprocconfig/extprocconfig.go
mathetake: well, I think not exactly - I assume this will have a semantics-caching, prompt guard etc happening in extproc

internal/extproc/router/router.go
mathetake: the selection happened inside the for-loop

extprocconfig/extprocconfig.go
mathetake: > or we are translating the LLMroute to the config for extproc ?

yes, this is because of the requirement about not restarting the extproc on the LLMRoute/etc updates

PR #54, opened 2024-12-27
Moves cmd/* out of internal

mathetake:
Follow up on https://github.com/envoyproxy/ai-gateway/pull/50#discussion_r1897313528


mathetake:
cc @zirain

zirain (approved):

PR #58, opened 2025-01-02
Scaffolds the initial extproc e2e test 

mathetake:
This scaffolds the initial extproc e2e tests using the secrets.
The GHA workflow is defined in a separate file so that we can run
the tests for PRs coming from fork repos as well under the condition that
a maintainer verifies the code and adds the label "safe-to-test".

The actual test for AWS bedrock is commented out now as we will do the
request signing inside the extproc.

PR #60, opened 2025-01-02
Fixes GHA

mathetake:
follow up on #58


mathetake:
self merging to fix the CI on the main branch!

PR #61, opened 2025-01-02
Implements AWS Request Signing inside extproc 

mathetake:
This commit implements the request signing logic in extproc
for AWS Bedrock support. Previously, in our PoC, we had used
Envoy's native AWS signing filter, but we faced a couple of issues,
so we decided to implement the signing in extproc which provides 
us much more flexibility.

The implementation here is pretty naive, and will be refactored 
after/during the implementation phase of the BackendSecurityPolicy
API #43. For example, it's currently just using env var credential provider, 
but this will change and become flexible soon.

This also updates the existing use of v1 AWS Go SDK to v2.


extprocconfig/extprocconfig.go
```diff
@@ -105,8 +105,19 @@ type Backend struct {
...
 }
 
+// BackendAuth ... TODO: refactor after https://github.com/envoyproxy/ai-gateway/pull/43.
+type BackendAuth struct {
+	AWSAuth *AWSAuth `yaml:"aws,omitempty"`
+}
+
+// AWSAuth ... TODO: refactor after https://github.com/envoyproxy/ai-gateway/pull/43.
+type AWSAuth struct{}
+
```
mathetake, line 120: Note: this will be refactored according to the final API of BackendSecurityPolicy in #43

internal/extproc/backendauth/aws.go
```diff
@@ -0,0 +1,89 @@
...
+	signer *v4.Signer
+	region string
+}
+
+func newAWSHandler(_ *extprocconfig.AWSAuth) (*awsHandler, error) {
+	cfg, err := config.NewEnvConfig()
+	if err != nil {
+		return nil, fmt.Errorf("cannot create AWS config: %w", err)
+	}
+	signer := v4.NewSigner()
```
mathetake, line 34: I _believe_ signer is Goroutine safe, but let's get back to this later

internal/extproc/backendauth/aws.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+
+	var body []byte
+	if _body := bodyMut.GetBody(); len(_body) > 0 {
+		body = _body
+	}
+
+	payloadHash := sha256.Sum256(body)
+	req, err := http.NewRequest(method,
+		fmt.Sprintf("https://bedrock-runtime.%s.amazonaws.com%s", a.region, path),
```
mathetake, line 69: I _believe_ we always use "bedrock-runtime.*.amazonaws.com so hardcoded - let's see if someone complains

mathetake (commented):
notes

internal/extproc/backendauth/aws.go
```diff
@@ -0,0 +1,89 @@
...
+
+// awsHandler implements [Handler] for AWS Bedrock authz.
+type awsHandler struct {
+	envCfg config.EnvConfig
+	signer *v4.Signer
+	region string
+}
+
+func newAWSHandler(_ *extprocconfig.AWSAuth) (*awsHandler, error) {
+	cfg, err := config.NewEnvConfig()
```
mathetake, line 30: currently it's just fixed to use env var, but this will change after #43

mathetake:
Hope someone will have left a rubber stamp by the time when I wake up tomorrow morning in Japan time zone

internal/extproc/backendauth/aws_test.go
```diff
@@ -0,0 +1,43 @@
...
+	t.Setenv("AWS_SECRET_ACCESS_KEY", "secret")
+
+	handler, err := newAWSHandler(nil)
+	require.NoError(t, err)
+	require.NotNil(t, handler)
+}
+
+func TestAWSHandler_Do(t *testing.T) {
+	t.Setenv("AWS_ACCESS_KEY_ID", "test")
+	t.Setenv("AWS_SECRET_ACCESS_KEY", "secret")
```
yuzisun, line 22: Can we support credential file as well like aws filter? We need to refresh the aws credential when it expires. @aabchoo can move his cred refreshing code here.

mathetake: Yes of course. We will do after #43

mathetake:
could someone take a look

yuzisun (approved):
/lgtm

PR #62, opened 2025-01-03
extproc: sets token usage into filter metadata

mathetake:
This implements the only necessary logic to achieve 
the token-based rate limit in the external processing code.
The only thing it needs to do is to set the filter dynamic metadata
which contains the extracted used token.

The set number can be used to reduce the rate limit budget thanks
to the new feature introduced in the upstream Envoy to be released
with the next version 1.33. 

See 
* https://github.com/envoyproxy/envoy/pull/37548
* https://github.com/envoyproxy/envoy/pull/37684
* https://github.com/envoyproxy/ratelimit/pull/802

### DIFF
--- a/extprocconfig/extprocconfig.go
+++ b/extprocconfig/extprocconfig.go
@@ -23,6 +23,9 @@ import (
 //	  schema: OpenAI
 //	backendRoutingHeaderKey: x-backend-name
 //	modelNameHeaderKey: x-model-name
+//	tokenUsageMetadata:
+//	  namespace: ai_gateway_llm_ns
+//	  key: token_usage_key
 //	rules:
 //	- backends:
 //	  - name: kserve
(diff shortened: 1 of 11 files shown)


extprocconfig/extprocconfig.go
```diff
@@ -65,6 +73,18 @@ type Config struct {
 	Rules []RouteRule `yaml:"rules"`
 }
 
+// TokenUsageMetadata is the namespace and key to be used in the filter metadata to store the usage token.
```
zirain, line 76: As https://github.com/envoyproxy/envoy/pull/37567 merged, we can support this with `%DYNAMIC_METADATA(NAMESPACE:KEY*):Z%`, is that right?

mathetake: exactly!

mathetake: this is so cool right - we can remove the whole bunch of hacky code from MVP

zirain: Should we only support metadata? it's fine for now, but it's worth thinking more.

mathetake: yeah i guess? token extraction happens *after* response headers are sent to downstream, so we cannot set it in response headers. That makes metadata is the only position to store this.

mathetake:
after this finally i will start adding the controller code and things will become more tangible!

zirain (approved):
LGTM

yuzisun (approved):
/lgtm

missBerg (approved):
lgtm

aabchoo:
Will we need a follow up PR to pass the dynamic key + namespace to envoy so it can determine rate limiting upon request?

PR #68, opened 2025-01-07
ci: skips tests using secrets for pull_request event from forks

mathetake:


.github/workflows/tests_using_secrets.yaml
```diff
@@ -17,8 +17,9 @@ on:
 jobs:
   extproc_tests:
     name: External Processor
-    if: ( github.event.pull_request.head.repo.fork == false ) ||
-      ( contains(github.event.pull_request.labels.*.name, 'safe to test') )
+    # Skip the pull_request event from forks as it cannot access secrets even if the PR is labeled with 'safe to test'.
+    if: (github.event.pull_request.head.repo.fork == false) ||
```
aabchoo, line 21: Should this be `&&` instead of `||` since you want to skip the pull_request event from fork?

PR #71, opened 2025-01-08
Adds initial controller implementation 

mathetake:
This commit adds the initial implementation of the resource 
management controllers. As per the official recommendation of
controller-runtime, the controller interface is implemented per
CRD. Therefore, this adds two typical "controller" one for LLMRoute
and another for LLMBackend.

In addition, this implements "configuration sink" which is the sync
mechanism across multiple resource events via Go channel. It has
the global view over all AI Gateway resources and hence have the 
full context required to create a final extproc configuration as well
as HTTPRoute. The following is the (very) simple diagram for the
relationship among k8s events, controllers and configSink.

k8s events --async--> {llmRouteCtrl, llmBackendCtrl, ...} --sinkEvent (sync)--> configSink

The followup PR will add the end to end tests.

### DIFF
--- a/extprocconfig/extprocconfig.go
+++ b/extprocconfig/extprocconfig.go
@@ -14,6 +14,15 @@ import (
 	gwapiv1 "sigs.k8s.io/gateway-api/apis/v1"
 )
 
+// DefaultConfig is the default configuration for the external processor
+// that can be used as a fallback when the configuration is not explicitly provided.
+const DefaultConfig = `
+inputSchema:
+  schema: OpenAI
+selectedBackendHeaderKey: x-envoy-ai-gateway-selected-backend
+modelNameHeaderKey: x-envoy-ai-gateway-llm-model
+`
+
 // Config is the configuration for the external processor.
 //
 // The configuration is loaded from a file path specified via the command line flag -configPath to the external processor.
@@ -22,8 +31,8 @@ import (
 //
 //	inputSchema:
 //	  schema: OpenAI
-//	backendRoutingHeaderKey: x-backend-name
-//	modelNameHeaderKey: x-model-name
+//	selectedBackendHeaderKey: x-envoy-ai-gateway-selected-backend
+//	modelNameHeaderKey: x-envoy-ai-gateway-llm-model
 //	tokenUsageMetadata:
 //	  namespace: ai_gateway_llm_ns
 //	  key: token_usage_key
@@ -38,24 +47,24 @@ import (
 //	    outputSchema:
 //	      schema: AWSBedrock
 //	  headers:
-//	  - name: x-model-name
+//	  - name: x-envoy-ai-gateway-llm-model
 //	    value: llama3.3333
 //	- backends:
 //	  - name: openai
 //	    outputSchema:
 //	      schema: OpenAI
 //	  headers:
-//	  - name: x-model-name
+//	  - name: x-envoy-ai-gateway-llm-model
 //	    value: gpt4.4444
 //
-// where the input of the external processor is in the OpenAI schema, the model name is populated in the header x-model-name,
-// The model name header `x-model-name` is used in the header matching to make the routing decision. **After** the routing decision is made,
-// the selected backend name is populated in the header `x-backend-name`. For example, when the model name is `llama3.3333`,
+// where the input of the external processor is in the OpenAI schema, the model name is populated in the header x-envoy-ai-gateway-llm-model,
+// The model name header `x-envoy-ai-gateway-llm-model` is used in the header matching to make the routing decision. **After** the routing decision is made,
+// the selected backend name is populated in the header `x-envoy-ai-gateway-selected-backend`. For example, when the model name is `llama3.3333`,
 // the request is routed to either backends `kserve` or `awsbedrock` with weights 1 and 10 respectively, and the selected
-// backend, say `awsbedrock`, is populated in the header `x-backend-name`.
+// backend, say `awsbedrock`, is populated in the header `x-envoy-ai-gateway-selected-backend`.
 //
-// From Envoy configuration perspective, configuring the header matching based on `x-backend-name` is enough to route the request to the selected backend.
-// That is because the matching decision is made by the external processor and the selected backend is populated in the header `x-backend-name`.
+// From Envoy configuration perspective, configuring the header matching based on `x-envoy-ai-gateway-selected-backend` is enough to route the request to the selected backend.
+// That is because the matching decision is made by the external processor and the selected backend is populated in the header `x-envoy-ai-gateway-selected-backend`.
 type Config struct {
 	// TokenUsageMetadata is the namespace and key to be used in the filter metadata to store the usage token, optional.
 	// If this is provided, the external processor will populate the usage token in the filter metadata at the end of the
@@ -65,9 +74,9 @@ type Config struct {
 	InputSchema VersionedAPISchema `yaml:"inputSchema"`
 	// ModelNameHeaderKey is the header key to be populated with the model name by the external processor.
 	ModelNameHeaderKey string `yaml:"modelNameHeaderKey"`
-	// BackendRoutingHeaderKey is the header key to be populated with the backend name by the external processor
+	// SelectedBackendHeaderKey is the header key to be populated with the backend name by the external processor
 	// **after** the routing decision is made by the external processor using Rules.
-	BackendRoutingHeaderKey string `yaml:"backendRoutingHeaderKey"`
+	SelectedBackendHeaderKey string `yaml:"selectedBackendHeaderKey"`
 	// Rules is the routing rules to be used by the external processor to make the routing decision.
 	// Inside the routing rules, the header ModelNameHeaderKey may be used to make the routing decision.
 	Rules []RouteRule `yaml:"rules"`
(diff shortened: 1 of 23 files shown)


mathetake:
the implementation is mechanical so nothing special here. I hope someone approves by the time when I wake up tomorrow then I can unblock the implementation part of #43 as well as the rate limit stuff

extprocconfig/extprocconfig.go
```diff
@@ -22,8 +31,8 @@ import (
 //
 //	inputSchema:
 //	  schema: OpenAI
-//	backendRoutingHeaderKey: x-backend-name
-//	modelNameHeaderKey: x-model-name
+//	selectedBackendHeaderKey: x-envoy-ai-gateway-selected-backend
```
yuzisun, line 34: not sure if we want to call it `selected-backend`, the backend may not be necessarily selected by gateway. User can still specify the backend in the request

extprocconfig/extprocconfig.go
```diff
@@ -38,24 +47,24 @@ import (
...
 //	      schema: OpenAI
 //	  headers:
-//	  - name: x-model-name
+//	  - name: x-envoy-ai-gateway-llm-model
 //	    value: gpt4.4444
 //
-// where the input of the external processor is in the OpenAI schema, the model name is populated in the header x-model-name,
-// The model name header `x-model-name` is used in the header matching to make the routing decision. **After** the routing decision is made,
-// the selected backend name is populated in the header `x-backend-name`. For example, when the model name is `llama3.3333`,
+// where the input of the external processor is in the OpenAI schema, the model name is populated in the header x-envoy-ai-gateway-llm-model,
```
yuzisun, line 60: Since user can configure the model routing header name, we can say configured modelNameHeaderKey

extprocconfig/extprocconfig.go
```diff
@@ -14,6 +14,15 @@ import (
 	gwapiv1 "sigs.k8s.io/gateway-api/apis/v1"
 )
 
+// DefaultConfig is the default configuration for the external processor
+// that can be used as a fallback when the configuration is not explicitly provided.
+const DefaultConfig = `
+inputSchema:
+  schema: OpenAI
+selectedBackendHeaderKey: x-envoy-ai-gateway-selected-backend
+modelNameHeaderKey: x-envoy-ai-gateway-llm-model
```
yuzisun, line 23: consider dropping the llm wording, there are other lvm(large vision model).

yuzisun: We might also need to consider renaming LLMRoute

extprocconfig/extprocconfig.go
mathetake: well, the backend will be "calculated" and "selected" following the matching rule of LLMRoute, so the word "selected" makes sense regardless of who actually "select".

mathetake: this is the internal use only, and not user facing, so it should be fine

extprocconfig/extprocconfig.go
mathetake: well, no, users are not required to choose the model routing header name at this point. https://github.com/envoyproxy/ai-gateway/blob/cb6b2e0f12fb6bc2e23941eff6039cf78d5693ca/api/v1alpha1/api.go#L206

It should be good to make this part of the LLMRoute resource, but the comment here matches what it is now

extprocconfig/extprocconfig.go
mathetake: > We might also need to consider renaming LLMRoute

good point

mathetake: https://github.com/envoyproxy/ai-gateway/pull/71/commits/7d7bd0dc2715901678d9fe214d2362e6d2d7ce32

internal/controller/sink.go
```diff
@@ -0,0 +1,302 @@
...
+// which requires to be done in a single goroutine since we need to
+// consolidate the information from both objects to generate the ExtProcConfig
+// and HTTPRoute objects.
+type configSink struct {
+	client client.Client
+	kube   kubernetes.Interface
+	logger logr.Logger
+
+	eventChan                   chan configSinkEvent
+	httpRoutes                  map[string]*gwapiv1.HTTPRoute
```
yuzisun, line 40: Seems like we are maintaining the local cache here, we can consider using controller runtime client which supports cached client to primarily retrieve data from in-memory cache that's automatically updated by Kubernetes informers., write operations still directly interact with APIServer.
https://github.com/kubernetes-sigs/controller-runtime/tree/main/pkg/client

mathetake: yeah intentionally manually maintaining the cache here - we can iterate on this here or in another PR later. I will take a look if i get a time tomorrow but this can be a TODO i guess

extprocconfig/extprocconfig.go
yuzisun: This is not true, you might not be able to select the backend just based on the model.. For example there is case where the same anthropic models are supported by both google and aws. In this case user needs to set the backend header to determine where to route to. 

see https://aws.amazon.com/bedrock/claude/ and https://cloud.google.com/solutions/anthropic?hl=en.

mathetake: the header `selectedBackendHeaderKey` here is only used to store the backend name calculated in the extproc user based on the LLMRoute.Rules. This is nothing to do with model name or whatever, and completely up to users.

mathetake: Say  you havve
```yaml
kind: LLMRoute
metadata:
  name: some-route
  namespace: default
spec:
  inputSchema:
    schema: OpenAI
  rules:
    - matches:
      - headers:
        - type: Exact
          name: some-random-header-that-can-be-sent-directly-by-clients
          value: foo
      backendRefs:
        - name: some-backend-name
......
```

and clients send `curl -H "some-random-header-that-can-be-sent-directly-by-clients: foo"` then **internally** extproc sets the header `$selectedBackendHeaderKey: some-backend-name`. That's what this does and this is the completely implementation detail. This package is not CRD but a configuration of extproc itself.

yuzisun: So you are saying we are not providing the standard ai gateway backend routing header to user, they MUST define the matching rules on LLMRoute for each backend user creates. Is that correct ?

mathetake: yes exactly currently, though we can provide some "canonical-backend-choosing-header-key" defined in the api/v1alpha package and prioritize the header value when present, which effectively ignores LLMRoute.Rules if the header exists.

mathetake: I will raise an issue tomorrow!

yuzisun: That is what exactly I was thinking to create those rules for user, if this routing header exists then we ignore the rules on llm route. I think this can provide better user experience.

yuzisun (approved):

internal/controller/sink.go
```diff
@@ -0,0 +1,295 @@
...
+	}
+
+	for i := range llmRoutes.Items {
+		llmRoute := &llmRoutes.Items[i]
+		namspacedLLMRouteName := fmt.Sprintf("%s.%s", llmRoute.Name, llmRoute.Namespace)
+		c.llmRoutes[namspacedLLMRouteName] = llmRoute
+
+		for _, rule := range llmRoute.Spec.Rules {
+			for _, backend := range rule.BackendRefs {
+				namspacedBackendName := fmt.Sprintf("%s.%s", backend.Name, llmRoute.Namespace)
```
aabchoo, line 88: nit: the variable is called namespacedBackendName but the order of the variable is backend Name then llmroute namespace. Could we reverse it?

aabchoo (approved):

internal/controller/llmroute.go
```diff
@@ -0,0 +1,252 @@
...
+// This only creates the external process deployment and service for the LLMRoute as well as the extension policy.
+// The actual HTTPRoute and the extproc configuration will be created in the config sink since we need
+// not only the LLMRoute but also the LLMBackend and other resources to create the full configuration.
+func (c *llmRouteController) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
+	c.logger.Info("Reconciling LLMRoute")
+
+	var llmRoute aigv1a1.LLMRoute
+	if err := c.client.Get(ctx, req.NamespacedName, &llmRoute); err != nil {
+		if client.IgnoreNotFound(err) == nil {
+			c.eventChan <- configSinkEventLLMRouteDeleted{namespace: req.Namespace, name: req.Name}
```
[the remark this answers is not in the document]
mathetake: yeah that's the convention of controller-runtime afaik

internal/controller/llmroute.go
```diff
@@ -0,0 +1,252 @@
...
+	if err := c.client.Get(ctx, client.ObjectKey{Name: extProcName(llmRoute), Namespace: llmRoute.Namespace}, &existingPolicy); err == nil {
+		existingPolicy.Spec.PolicyTargetReferences.TargetRefs = llmRoute.Spec.TargetRefs
+		if err := c.client.Update(ctx, &existingPolicy); err != nil {
+			return fmt.Errorf("failed to update extension policy: %w", err)
+		}
+	} else if client.IgnoreNotFound(err) != nil {
+		return fmt.Errorf("failed to get extension policy: %w", err)
+	}
+	pm := egv1a1.BufferedExtProcBodyProcessingMode
+	port := gwapiv1.PortNumber(1063)
```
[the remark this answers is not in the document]
mathetake: yeah we can make this configurable later

mathetake:
@yuzisun @aabchoo thanks for the review!

PR #72, opened 2025-01-08
Add trivy image scan workflow

Krishanx92:
Introduced Trivy image vulnerabilities scan workflow

- This workflow is triggered on merges to the main branch and runs once daily (schedule can be adjusted if needed).

- It currently scans both the controller and extproc images.

### DIFF
--- a/.github/workflows/trivy.yaml
+++ b/.github/workflows/trivy.yaml
@@ -18,7 +18,7 @@ jobs:
     runs-on: ubuntu-22.04
     steps:
       - name: Checkout code
-        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683  
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683
 
       - name: Build Controller Docker image
         run: |
@@ -37,7 +37,7 @@ jobs:
           vuln-type: 'os,library'
           exit-code: '1'  # Fail workflow on detected vulnerabilities
           ignore-unfixed: true  # Ignore unfixed vulnerabilities
-      
+
       - name: Run Trivy vulnerability scanner for Ext-Proc
         uses: aquasecurity/trivy-action@master
         with:
--- a/.github/workflows/trivy.yaml
+++ b/.github/workflows/trivy.yaml
@@ -14,34 +14,25 @@ jobs:
   image-scan:
     permissions:
       contents: read  # Required for actions/checkout to fetch code
-    name: Image Scan
+    name: Scan ${{ matrix.target.command_name }} Image
     runs-on: ubuntu-22.04
+    strategy:
+      matrix:
+        target:
+          - command_name: "controller"
+          - command_name: "extproc"
     steps:
       - name: Checkout code
-        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683
+        uses: actions/checkout@v4
 
-      - name: Build Controller Docker image
+      - name: Build ${{ matrix.target.command_name }} Docker image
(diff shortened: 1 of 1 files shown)


mathetake:
@aabchoo @wengyao04 can either of you review?

.github/workflows/trivy.yaml
```diff
@@ -0,0 +1,49 @@
...
+
+      - name: Run Trivy vulnerability scanner for Ext-Proc
+        uses: aquasecurity/trivy-action@master
+        with:
+          image-ref: ghcr.io/envoyproxy/ai-gateway/extproc:${{ github.sha }}
+          format: 'table'
+          severity: 'CRITICAL,HIGH,MEDIUM,LOW'
+          vuln-type: 'os,library'
+          exit-code: '1'  # Fail workflow on detected vulnerabilities
+          ignore-unfixed: true  # Ignore unfixed vulnerabilities
```
mathetake, line 49: maybe adding this part of the existing docker build flow? like adding this after https://github.com/envoyproxy/ai-gateway/blob/cb6b2e0f12fb6bc2e23941eff6039cf78d5693ca/.github/workflows/docker_builds_template.yaml#L47 ?

mathetake:
well i am assigning myself :D

.github/workflows/trivy.yaml
mathetake: is that costly? (i have never used trivy before)

Krishanx92: trivy is commonly use in os projects for container security and vulnerability scanning.

adding to docker_builds_template.yaml fine, but the problem is it runs on every pull request/commit and it will block all new prs if new vulnerability detect in existing components. so i would suggest to keep it as separate wf.
scheduling is commonly used when cost factor is considered so if needed we can schedule it per week rather that per day.

Krishanx92:
Screenshot 2025-01-09 at 08 28 15

btw i ran it in locally and there's no vulnerability atm which is great. :)

.github/workflows/trivy.yaml
```diff
@@ -0,0 +1,49 @@
...
+
+jobs:
+  image-scan:
+    permissions:
+      contents: read  # Required for actions/checkout to fetch code
+    name: Image Scan
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683
```
mathetake, line 21: why the specific commit? can you use v4 in line with other workflows

.github/workflows/trivy.yaml
```diff
@@ -0,0 +1,49 @@
...
+
+      - name: Run Trivy vulnerability scanner for Controller
+        uses: aquasecurity/trivy-action@master
+        with:
+          image-ref: ghcr.io/envoyproxy/ai-gateway/controller:${{ github.sha }}
+          format: 'table'
+          severity: 'CRITICAL,HIGH,MEDIUM,LOW'
+          vuln-type: 'os,library'
+          exit-code: '1'  # Fail workflow on detected vulnerabilities
+          ignore-unfixed: true  # Ignore unfixed vulnerabilities
```
mathetake, line 39: can you make "controller" and "extproc" as part of matrix so that we could remove the duplication? https://github.com/envoyproxy/ai-gateway/blob/8aa1b4b197f69aef74c8b7c4abc16f6947ee5ff3/.github/workflows/docker_builds_template.yaml#L11-L14

mathetake (approved):
👍

PR #76, opened 2025-01-09
api: changes LLM prefix -> AI*

mathetake:
This comes up in a thread in #71 with @yuzisun, and we might
want to remove the LLM prefix. 

The rationale is that the current functionality is nothing to do 
with "LLM" but more about the general routing and authn/z
with "AI providers" where the input is OpenAI format.

On the other hand, there "will be" the LLM specific CRD such
as the configurations for prompt guard, semantics caching etc.

I would like to have inputs from the community members!

cc @envoyproxy/ai-gateway-assignable


api/v1alpha1/api.go
```diff
@@ -9,57 +9,57 @@ import (
...
+// receive. And then the Gateway will route the traffic to the appropriate AIBackend based
+// on the output schema of the AIBackend while doing the other necessary jobs like
 // upstream authentication, rate limit, etc.
 //
-// LLMRoute generates a HTTPRoute resource based on the configuration basis for routing the traffic.
-// The generated HTTPRoute has the owner reference set to this LLMRoute.
-type LLMRoute struct {
+// AIRoute generates a HTTPRoute resource based on the configuration basis for routing the traffic.
+// The generated HTTPRoute has the owner reference set to this AIRoute.
+type AIRoute struct {
```
yuzisun, line 24: How about AIGatewayRoute? The reason is to differentiate from HTTPRoute which normally indicates “ingress” routes, for our case it is routing to external providers.

yuzisun: Also we have a single gateway route with multiple http matching rules. Without “gateway” it sounds to user that you can have many such routes.

mathetake: sounds good +1

api/v1alpha1/api.go
```diff
@@ -124,42 +124,42 @@ type LLMRouteRuleMatch struct {
...
-// A LLMBackend is "attached" to a Backend which is either a k8s Service or a Backend resource of the Envoy Gateway.
+// A AIBackend is "attached" to a Backend which is either a k8s Service or a Backend resource of the Envoy Gateway.
 //
-// When a backend with an attached LLMBackend is used as a routing target in the LLMRoute (more precisely, the
-// HTTPRouteSpec defined in the LLMRoute), the ai-gateway will generate the necessary configuration to do
+// When a backend with an attached AIBackend is used as a routing target in the AIRoute (more precisely, the
+// HTTPRouteSpec defined in the AIRoute), the ai-gateway will generate the necessary configuration to do
 // the backend specific logic in the final HTTPRoute.
-type LLMBackend struct {
+type AIBackend struct {
```
yuzisun, line 135: AIBackend is too broad, suggests AIServiceBackend.

api/v1alpha1/api.go
Krishanx92: +1 for the AIGatewayRoute, btw @yuzisun 

> we have a single gateway route
what this means is we don't have any other route CR in this gateway, and we don't bring any in future as well. correct me if im wrong.

api/v1alpha1/api.go
Krishanx92: +1 for AIServiceBackend

mathetake: how about AIProvider (we tend to use in discussion)?

mathetake: AIServiceBackend wfm as well

Krishanx92: > how about AIProvider (we tend to use in discussion)?

we are still not supporting provider specific configurations here such as supported models, supported authenticated types, transformation etc. for now i feel like this more towards to managing backend configuration

mathetake:
Applied the change according to the discussion so far:
* AIRoute -> AIGatewayRoute
* AIBackend -> AIServiceBackend


anything else?

mathetake:
PING ALL - this is blocking others

Krishanx92:
LGTM

Krishanx92 (approved):

mathetake:
others thoughts? if not, I will polish this PR as a proper one not just for discussion

mathetake:
ok i will work on it

mathetake:
@yuzisun can you take a look?

yuzisun (approved):

PR #88, opened 2025-01-13
Scaffolds the initial helm chart

mathetake:


mathetake:
need a rubber stamping

aabchoo (approved):

.github/workflows/commit.yaml
```diff
@@ -93,5 +93,23 @@ jobs:
 
   docker_builds:
     name: Docker Build and/or Push
-    needs: [unittest, test_cel_validation, test_controller]
+    needs: [style, unittest, test_cel_validation, test_controller]
     uses: ./.github/workflows/docker_builds_template.yaml
+
+  push_helm:
+    name: Push Helm chart
+    # Push the Helm chart to the GitHub Container Registry on the main branch.
```
aabchoo, line 101: Shouldn't we only push to helm when the PR is merged to main?

mathetake: that's what the line below does `if: github.event_name == 'push`

mathetake: let me clarify the comment more

aabchoo: Didn't realize that is what push meant 🙇 Thanks!

Issue #90, opened 2025-01-13
Migrate extproc to dynamic modules when available in Envoy


mathetake:
As discussed in the community meeting (though I was not able to attend), the upcoming Envoy v1.34 (to be released in April) will have support for Dynamic Modules with all the necessary features to rewrite the extproc in dynamic modules. A dynamic module is a shared object that is written in any language implementing [the ABI specified in a single pure C header](https://github.com/envoyproxy/envoy/blob/main/source/extensions/dynamic_modules/abi.h). It's actively being developed since the last summer (see [the PRs so far](https://github.com/search?q=repo%3Aenvoyproxy%2Fenvoy+dynamic_modules&type=pullrequests)) but almost ready to be used by the real end users like this project.

I am the author of the new feature in Envoy upstream, and am sure that we should do that due to the multiple reasons:
* Dynamic module is faster than ExtProc: it's running almost as identical as native C++ filters. For example, It has zero copy in the sense that it has direct access to Envoy buffer (body and header bytes) vs ExtProc needs to send the entire buffer over the network which has the significant cost.
* Dynamic module makes the deployment really simpler, which results in both simple control plane code and less hassles on user end e.g. no need to manage extproc by themselves.

In addition, the dynamic module has already been worked on by multiple (large) companies like Netflix, Linkedin, etc, so I believe the possible only concern on "maturity" is not what blocks us from adopting it vs the benefits stated above.

Since the "official" language supported by Envoy initially will be Rust (and I think C++ later), I would recommend to rewrite it in Rust but Go is also supported with some care which is no not critical, but the code must be written in a way that won't break Go runtime. As long as I write the initial code, simply porting the Go existing code to dynamic module would be fine as well. 

On that note, the public configuration API package, currently named "extprocconfig", will be the "migration interface" that will be accepted by both extproc and dynamic module versions, so I will rename it to something like "aifilterconfig" or something like that.

mathetake:
cc @envoyproxy/ai-gateway-assignable @missBerg @wengyao04 @aabchoo

mathetake:
since Envoy won't be ready by the end of the month (or the next Envoy v1.33 to be released in a few days),  unfortunately we have to use the extproc in the initial version

mathetake:
or if  it's ok to use the main branch Envoy in our initial release (and possible delay of 1-3 weeks), then we could use the dynamic modules with the initial release. wdyt?

zirain:
[...]

PR #92, opened 2025-01-13
Renames extprocconfig to filterconfig

mathetake:
Per #90, we will migrate to dynamic modules in the second or third version of
AI Gateway, the public configuration package must not be coupled with
extproc (see the description in the issue). This renames `extprocconfig`
to more generic (and accurate) name as `filterconfig`.


aabchoo (approved):
Looks good to me assuming the test passes!

mathetake:
@aabchoo thanks for the look!

PR #98, opened 2025-01-14
Adds deployment configuration for extproc

### LINKED ISSUE #82: Support scaling up ext proc deployment
Currently replica number is not set in reconcileExtProcDeployment, we need to make it configurable or attach HPA. https://github.com/envoyproxy/ai-gateway/blob/a45438b3472626e9704096e959cf13b1b7cf1bda/internal/controller/llmroute.go#L168

mathetake:
Fixes #82


api/v1alpha1/api.go
```diff
@@ -122,6 +132,45 @@ type LLMRouteRuleMatch struct {
 	Headers []gwapiv1.HTTPHeaderMatch `json:"headers,omitempty"`
 }
 
+type LLMRouteFilterConfig struct {
```
mathetake, line 135: This indirection will help us deprecate and keep the API clean before/after extproc removal per #90 

fyi I will work on the compatibility policy doc next

api/v1alpha1/api.go
```diff
@@ -122,6 +132,45 @@ type LLMRouteRuleMatch struct {
...
+type LLMRouteFilterConfigExternalProcess struct {
+	// Replicas is the number of desired pods of the external process deployment.
+	//
+	// +optional
+	Replicas *int32 `json:"replicas,omitempty"`
+	// Resources required by the external process container.
+	// More info: https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/
+	//
+	// +optional
+	Resources *corev1.ResourceRequirements `json:"resources,omitempty"`
```
mathetake, line 169: i could've used either the embedding of appv1.DeploymentSpec or EG's KubernetesDeploymentSpec but both seems too complex for this purpose in addition to the impl cost. we can add additional fields on demand.

api/v1alpha1/api.go
mathetake: Also, this could be `AIGatewayFilterConfig` per #76

mathetake (commented):
Notes

api/v1alpha1/api.go
yuzisun: right, let's change it all together.

yuzisun (approved):

yuzisun:
@mathetake there is a conflict to resolve

mathetake:
aha ok

PR #99, opened 2025-01-14
docs: adds RELEASES.md

### LINKED ISSUE #21: Compatibility promise / Breaking change policy
We must explicitly state about these policies before the initial release

mathetake:
This commit adds RELEASES.md documentation
that describes the release cycles and the support
policy especially around the compatibility promises.

fixes #21

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -1,4 +1,24 @@
-# Support Policy
+# Notes on Releases
+
+## Release Cycles
+
+Since Envoy AI Gateway depends on the Envoy Gateway and Envoy Proxy, we will follow the release cycle of the Envoy Gateway.
+In other words, we aim to cut the release of the Envoy AI Gateway a few days or a week after the new version of the Envoy Gateway 
+is released. Therefore, the release cycle of the Envoy AI Gateway will be approximately every 2-3 months.
+
+We do not distinguish between major and patch releases. We will increment the version number by one for each release,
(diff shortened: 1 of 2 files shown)


mathetake:
note to self: add release cycle doc

mathetake:
ok ready!

RELEASES.md
```diff
@@ -0,0 +1,67 @@
...
+no _un-migrated_ resources including breaking change left in the k8s API server. In other words, after the
+proper use of the API and migration path described above, the user should be able to upgrade the controller
+without any issue. However, this does mean that we do NOT guarantee that the existing configuration will work
+across more than two version of the controller. For example if you are using the version N of the controller,
+and you want to upgrade to the version N+2, you should first upgrade to the version N+1 while following the
+migration path if applicable, and then upgrade to the version N+2.
+
+### Envoy Gateway vs Envoy AI Gateway compatibility
+
+Since Envoy AI Gateway is built on top of Envoy Gateway, the compatibility between the two is important.
```
yuzisun, line 64: There are people asking if envoy ai gateway is compatible with Istio

yuzisun (approved):

RELEASES.md
mathetake: that's where the public `filterconfig` package, which is decoupled from EG/k8s specifics, comes into play! In fact my colleagues are working on the extproc integration in another control plan :+1:

PR #102, opened 2025-01-16
controller: removes unnecessary manual caches

mathetake:


mathetake:
cc @aabchoo - you might need to rebase your code on the impl of BackendSecurity API... i will go ahead and merge this to unblock you

Issue #97, opened 2025-01-14
Allow customizing token cost calculation


yuzisun:
When reducing the token usage limit in external proc, we need a way to allow customizing the token cost calculation such as `input_tokens*weight + output_tokens`

mathetake:
ok so the left is CEL implementation left in #103

PR #103, opened 2025-01-16
api: RequestCost configurations

mathetake:
This adds the RequestCost field to AIGatewayRoute,
which will allows users to do the rate limiting etc
based on the calculated "token usage". 

This is based on the new feature introduced in 
* https://github.com/envoyproxy/gateway/pull/4957
* https://github.com/envoyproxy/gateway/pull/5035

and because of the feature, the only thing we have to do
from AI Gateway side is to set a dynamic metadata as per 
the comment in the API.


api/v1alpha1/api.go
```diff
@@ -378,3 +426,34 @@ type AWSOIDCExchangeToken struct {
...
+	// +kubebuilder:validation:Enum=default
+	Type RequestCostType `json:"type"`
+
+	// CELExpression is the CEL expression to calculate the cost of the request.
+	// The CEL expression must return an integer value. The CEL expression should be
+	// able to access the request headers, model name, backend name, input/output tokens etc.
+	//
+	// +optional
+	// +notImplementedHide https://github.com/envoyproxy/ai-gateway/issues/97
+	CELExpression *string `json:"celExpression"`
```
mathetake, line 444: @yuzisun  I think this should be enough for  https://github.com/envoyproxy/ai-gateway/issues/97? also I will add some extensibility to extproc code so that users can insert their own custom extproc logic (like custom routing etc) - that's where also the custom weight calculation can be done.

mathetake:
cc @envoyproxy/ai-gateway-assignable

mathetake:
will do the e2e tests in another PR

api/v1alpha1/api.go
```diff
@@ -80,6 +80,54 @@ type AIGatewayRouteSpec struct {
...
+	//	            number: 0
+	//	          # This specifies the cost of the response retrieved from the dynamic metadata set by the AI Gateway filter.
+	//	          # The extracted value will be used to consume the rate limit budget, and subsequent requests will be rate limited
+	//	          # if the budget is exhausted.
+	//	          response:
+	//	            from: Metadata
+	//	            metadata:
+	//	              namespace: envoy.filters.http.ai_gateway.token_usage
+	//	              key: ai_gateway_route_request_cost
+	RequestCost *RequestCost `json:"captureTokenUsage,omitempty"`
```
mathetake, line 130: @envoyproxy/ai-gateway-assignable @envoyproxy/ai-gateway-maintainers so the question i have is should this be a part of AIServiceBackend vs per Route. any thoughts?

aabchoo: I believe per Route would be ideal for our use case. We will be using the same request cost type for all our backends 

cc @sukumargaonkar

api/v1alpha1/api.go
```diff
@@ -378,3 +426,34 @@ type AWSOIDCExchangeToken struct {
...
+	// +optional
+	// +notImplementedHide https://github.com/envoyproxy/ai-gateway/issues/97
+	CELExpression *string `json:"celExpression"`
+}
+
+// RequestCostType specifies the type of the RequestCost.
+type RequestCostType string
+
+const (
+	RequestCostTypeDefault RequestCostType = "default"
```
yuzisun, line 451: I think it can be more explicit about the type `RequestCostTypeOutputToken`

api/v1alpha1/api.go
yuzisun: I think we can provide both global and route level override options, if there is no route level cost calculation then we use the global cost calculation.

mathetake: cool - then I think i will leave some TODO comments about allowing the route-specific cost later

api/v1alpha1/api.go
wengyao04: Are we going to validate cel expression in validation webhook ?

mathetake: yeah if it's possible - not sure if we can do the cel expr validation inside the cel validation rule of k8s...

mathetake:
marking as draft until all ready including tests

mathetake:
ok ready - I would like to unblock #111 so I will defer the end to end tests in a subsequent PR.

mathetake:
also CEL implementation is TODO as it will be a relatively large change

tests/testupstream/main.go
```diff
@@ -114,6 +114,9 @@ func sseHandler(w http.ResponseWriter, r *http.Request) {
 }
 
 func handler(w http.ResponseWriter, r *http.Request) {
+	for k, v := range r.Header {
+		fmt.Printf("header %q: %s\n", k, v)
+	}
 	if v := r.Header.Get(expectedHeadersKey); v != "" {
```
mathetake, line 120: note: intentionally left as the testupstream does this elsewhere everywhere

internal/extproc/processor.go
```diff
@@ -75,13 +77,15 @@ func (p *Processor) ProcessRequestBody(_ context.Context, rawBody *extprocv3.Htt
...
 	p.logger.Info("Processing request", "path", path, "model", model)
+	fmt.Println("Processing request", "path", path, "model", model)
 
 	p.requestHeaders[p.config.ModelNameHeaderKey] = model
 	b, err := p.config.router.Calculate(p.requestHeaders)
 	if err != nil {
 		return nil, fmt.Errorf("failed to calculate route: %w", err)
 	}
 	p.logger.Info("Selected backend", "backend", b.Name)
+	fmt.Println("Selected backend", "backend", b.Name)
```
yuzisun, line 88: Remove the prints

internal/extproc/processor.go
```diff
@@ -75,13 +77,15 @@ func (p *Processor) ProcessRequestBody(_ context.Context, rawBody *extprocv3.Htt
 		return nil, fmt.Errorf("failed to parse request body: %w", err)
 	}
 	p.logger.Info("Processing request", "path", path, "model", model)
+	fmt.Println("Processing request", "path", path, "model", model)
```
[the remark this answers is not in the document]
mathetake: lol thanks

internal/extproc/processor.go
mathetake: lol thanks

api/v1alpha1/api.go
```diff
@@ -378,3 +429,36 @@ type AWSOIDCExchangeToken struct {
...
 	AwsRoleArn string `json:"awsRoleArn"`
 }
+
+// RequestCost configures the request cost.
+type RequestCost struct {
+	// Type specifies the type of the request cost. The default is "OutputToken",
+	// and it uses "output token" as the cost. The other types are "InputToken" and "TotalToken".
+	//
+	// +kubebuilder:validation:Enum=OutputToken;InputToken;TotalToken
+	Type RequestCostType `json:"type"`
```
yuzisun, line 439:
```suggestion
	Type RequestTokenCostType `json:"type"`
```

api/v1alpha1/api.go
```diff
@@ -378,3 +429,36 @@ type AWSOIDCExchangeToken struct {
 	// which maps to the temporary AWS security credentials exchanged using the authentication token issued by OIDC provider.
 	AwsRoleArn string `json:"awsRoleArn"`
 }
+
+// RequestCost configures the request cost.
+type RequestCost struct {
+	// Type specifies the type of the request cost. The default is "OutputToken",
```
yuzisun, line 435: This is where we may need to qualify the name with LLM

mathetake: like LLMOutputToken?

api/v1alpha1/api.go
mathetake: so you want me to change RequestCost as well to RequestTokenCost? feels a bit redundant though, works for me

api/v1alpha1/api.go
yuzisun: How about LLMRequestCost and LLMRequestCostType

mathetake: works for me!

api/v1alpha1/api.go
```diff
@@ -80,6 +80,54 @@ type AIGatewayRouteSpec struct {
...
+	//	            number: 0
+	//	          # This specifies the cost of the response retrieved from the dynamic metadata set by the AI Gateway filter.
+	//	          # The extracted value will be used to consume the rate limit budget, and subsequent requests will be rate limited
+	//	          # if the budget is exhausted.
+	//	          response:
+	//	            from: Metadata
+	//	            metadata:
+	//	              namespace: io.envoy.ai_gateway
+	//	              key: ai_gateway_route_request_cost
+	LLMRequestCost *LLMRequestCost `json:"llmRequestCost,omitempty"`
```
aabchoo, line 130: Should we consider allowing multiple LLMRequestCosts? Ie users wanting to limit input token in addition to total tokens

mathetake: That's a very valid point

mathetake: @arkodg i think EG needs to have a support for a generic key in rate limit API? https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#config-route-v3-ratelimit-action-generickey ?

If a user wants to do a rate limit on multiple statics (in this case input token, output token, and total token, for example), currently they have no way to distinguish these costs. For example, if they want to do that per user-id, without a generic key, there's no way to have multiple costs for a user-id.

mathetake: or i think EG could implicitly add a generic key there’s identical rules modulo cost

mathetake: ok confirmed that this can be achieved with the current EG implmenetation. Thank you for the comment @aabchoo i will make it plural

mathetake:
I made it plural so that i can be used to have a separate rate limits per metrics. ptal @aabchoo @yuzisun

api/v1alpha1/api.go
```diff
@@ -80,6 +80,101 @@ type AIGatewayRouteSpec struct {
 	// Currently, the filter is only implemented as an external process filter, which might be
 	// extended to other types of filters in the future. See https://github.com/envoyproxy/ai-gateway/issues/90
 	FilterConfig *AIGatewayFilterConfig `json:"filterConfig,omitempty"`
+
+	// LLMRequestCosts specifies how to capture the cost of the LLM-related request, notably the token usage.
+	// The AI Gateway filter will capture each specified number and store it in the Envoy's dynamic
+	// metadata per HTTP request. The namespaced key is "io.envoy.ai_gateway",
+	//
+	// For example, let's say we have the following LLMRequestCosts configuration:
```
mathetake, line 88: @arkodg appreciate it if you could double check this doc comment

arkodg: looks good, do you want to highlight `Distinct` match so individual limits are applies to unique users - `one`, `two` .... defined in `x-user-id` ?

mathetake: good catch! fixed in https://github.com/envoyproxy/ai-gateway/pull/103/commits/e6370a88cbf651ef63fec1981f22adeaf5b1f3b8

mathetake:
ready 🎵

filterconfig/filterconfig.go
```diff
@@ -66,11 +66,12 @@ modelNameHeaderKey: x-envoy-ai-gateway-model
...
-	// TokenUsageMetadata is the namespace and key to be used in the filter metadata to store the usage token, optional.
-	// If this is provided, the filter will populate the usage token in the filter metadata at the end of the
-	// response body processing.
-	TokenUsageMetadata *TokenUsageMetadata `yaml:"tokenUsageMetadata,omitempty"`
-	// Schema specifies the API schema of the input format of requests to the filter.
+	// MetadataNamespace is the namespace of the dynamic metadata to be used by the filter.
+	MetadataNamespace string `yaml:"namespace"`
+	// LLMRequestCost configures the cost of each LLM-related request. Optional. If this is provided, the filter will populate
+	// the "calculated" cost in the filter metadata at the end of the response body processing.
+	LLMRequestCosts []LLMRequestCost `yaml:"llmRequestCost,omitempty"`
```
mathetake, line 73:
```suggestion
	LLMRequestCosts []LLMRequestCost `yaml:"llmRequestCosts,omitempty"`
```

yuzisun (approved):

PR #104, opened 2025-01-16
extproc: adds ability to insert custom router

mathetake:


internal/extproc/main/main.go
```diff
@@ -0,0 +1,78 @@
+package maincmd
```
mathetake, line 1: this needs to be moved to public package - i will take care of that once it's given a go

mathetake:
@wengyao04 as discussed i will go ahead and merge this. Feel free to raise any question

PR #106, opened 2025-01-16
Implementation of API Key Auth

aabchoo:
This PR implements APIKey Auth for backendSecurityPolicy. The api key stored in a secret will be mounted to the extproc's pod, and later extracted by the appropriate AuthHandler.

Moved extprocDeployment into the `sink` as we need to tinker with the deployment if backend/backendSecurityPolicy is updated.

Removed the envoy config yaml's field `add_request_header` and replaced it with new API Key Auth implementation. CI passes which means that API Key mounted to extproc + specified on backend auth works.

Will work on adding AWS Credential file after this.


mathetake:
could you merge the main branch?

internal/controller/sink.go
```diff
@@ -263,3 +320,159 @@ func (c *configSink) newHTTPRoute(dst *gwapiv1.HTTPRoute, aiGatewayRoute *aigv1a
 	dst.Spec.CommonRouteSpec.ParentRefs = parentRefs
 	return nil
 }
+
+// syncExtProcDeployment syncs the external processor's Deployment and Service.
+func (c *ConfigSink) syncExtProcDeployment(ctx context.Context, aiGatewayRoute *aigv1a1.AIGatewayRoute) error {
```
aabchoo, line 325: This part feels slightly messy. I am hoping to find a way to cleanly get all related backendSecurityPolicies without looping, but this is what I've come up with for so far.

internal/controller/sink.go
```diff
@@ -263,3 +320,159 @@ func (c *configSink) newHTTPRoute(dst *gwapiv1.HTTPRoute, aiGatewayRoute *aigv1a
...
+		},
+	}
+	if _, err = c.kube.CoreV1().Services(aiGatewayRoute.Namespace).Create(ctx, service, metav1.CreateOptions{}); client.IgnoreAlreadyExists(err) != nil {
+		return fmt.Errorf("failed to create Service %s.%s: %w", name, aiGatewayRoute.Namespace, err)
+	}
+	return nil
+}
+
+// mountBackendSecurityPolicySecrets will mount secrets based on backendSecurityPolicies attached to AIServiceBackend.
+func (c *ConfigSink) mountBackendSecurityPolicySecrets(spec *corev1.PodSpec, aiGatewayRoute *aigv1a1.AIGatewayRoute) (*corev1.PodSpec, error) {
```
aabchoo, line 427: This part feels slightly messy. I am hoping to find a way to cleanly get all related backendSecurityPolicies without looping+our map, but this is what I've come up with for so far.

aabchoo:
cc @envoyproxy/assignable

mathetake:
aaron that is a different assignable team (for the entire envoyproxy/envoy repo)... @envoyproxy/ai-gateway-assignable is the correct one :)

internal/controller/sink.go
```diff
@@ -18,50 +20,69 @@ import (
...
-	client client.Client
-	kube   kubernetes.Interface
-	logger logr.Logger
+//
+// Exported for internal testing purposes.
+type ConfigSink struct {
+	client              client.Client
+	kube                kubernetes.Interface
+	logger              logr.Logger
+	defaultExtProcImage string
```
wengyao04, line 44: Could we also add `defaultExtProcImagePullPolicy corev1.PullPolicy` and value `corev1.PullPolicy(IfNoPresent)`

PR #111, opened 2025-01-16
feat: support option for continuous monitoring token usage in streaming response

sukumargaonkar:
Currently only total_tokens usage from response body are pushed to dynamicMetadata. This PR updates that logic to include input and output token usage as well.

This PR also introduces `monitorContinuousUsageStats` flag in config for external process
The flag controls if external process monitors every response-body chunk for usage stats
when true, it will monitor for token metadata usage in every response-body chunk received during request in streaming mode (compatible with vllm's 'continuous_usage_stats' flag)
when false, it will stop monitoring after detecting token metadata usage after finding it for the first time.
(compatible with OpenAI's streaming response (https://platform.openai.com/docs/api-reference/chat/streaming#chat/streaming-usage))
Only affects request in streaming mode

### DIFF
--- a/filterconfig/filterconfig.go
+++ b/filterconfig/filterconfig.go
@@ -71,6 +71,13 @@ type Config struct {
 	// LLMRequestCost configures the cost of each LLM-related request. Optional. If this is provided, the filter will populate
 	// the "calculated" cost in the filter metadata at the end of the response body processing.
 	LLMRequestCosts []LLMRequestCost `json:"llmRequestCosts,omitempty"`
+	// MonitorContinuousUsageStats flag controls if external process monitors every response-body chunk for usage stats
+	// when true, it will monitor for token metadata usage in every response-body chunk received during request in streaming mode
+	// compatible with vllm's 'continuous_usage_stats' flag
+	// when false, it will stop monitoring after detecting token metadata usage after finding it for the first time.
+	// compatible with OpenAI's streaming response (https://platform.openai.com/docs/api-reference/chat/streaming#chat/streaming-usage)
+	// Only affects request in streaming mode
+	MonitorContinuousUsageStats bool `yaml:"monitorContinuousUsageStats,omitempty"`
 	// InputSchema specifies the API schema of the input format of requests to the filter.
 	Schema VersionedAPISchema `json:"schema"`
 	// ModelNameHeaderKey is the header key to be populated with the model name by the filter.
--- a/internal/extproc/translator/translator.go
+++ b/internal/extproc/translator/translator.go
@@ -35,12 +35,12 @@ func isGoodStatusCode(code int) bool {
 type Factory func(path string) (Translator, error)
 
 // NewFactory returns a callback function that creates a translator for the given API schema combination.
-func NewFactory(in, out filterconfig.VersionedAPISchema) (Factory, error) {
+func NewFactory(in, out filterconfig.VersionedAPISchema, monitorContinuousUsageStats bool) (Factory, error) {
 	if in.Name == filterconfig.APISchemaOpenAI {
 		// TODO: currently, we ignore the LLMAPISchema."Version" field.
 		switch out.Name {
 		case filterconfig.APISchemaOpenAI:
-			return newOpenAIToOpenAITranslator, nil
+			return newOpenAIToOpenAITranslatorFactory(monitorContinuousUsageStats), nil
 		case filterconfig.APISchemaAWSBedrock:
 			return newOpenAIToAWSBedrockTranslator, nil
 		}
--- a/filterconfig/filterconfig_test.go
+++ b/filterconfig/filterconfig_test.go
@@ -74,6 +74,7 @@ rules:
 	require.Equal(t, "OpenAI", string(cfg.Schema.Name))
 	require.Equal(t, "x-ai-eg-selected-backend", cfg.SelectedBackendHeaderKey)
 	require.Equal(t, "x-ai-eg-model", cfg.ModelNameHeaderKey)
+
 	require.Len(t, cfg.Rules, 2)
 	require.Equal(t, "llama3.3333", cfg.Rules[0].Headers[0].Value)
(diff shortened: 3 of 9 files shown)


filterconfig/filterconfig.go
```diff
@@ -66,6 +68,13 @@ modelNameHeaderKey: x-envoy-ai-gateway-model
 // From Envoy configuration perspective, configuring the header matching based on `x-envoy-ai-gateway-selected-backend` is enough to route the request to the selected backend.
 // That is because the matching decision is made by the filter and the selected backend is populated in the header `x-envoy-ai-gateway-selected-backend`.
 type Config struct {
+	// MonitorContinuousUsageStats flag controls if external process monitors every response-body chunk for usage stats
+	// when true, it will monitor for token metadata usage in every response-body chunk received during request in streaming mode
+	// compatible with vllm's 'continuous_usage_stats' flag
+	// when false, it will stop monitoring after detecting token metadata usage after finding it for the first time.
+	// compatible with OpenAI's streaming response (https://platform.openai.com/docs/api-reference/chat/streaming#chat/streaming-usage)
+	// Only affects request in streaming mode
+	MonitorContinuousUsageStats bool `yaml:"monitorContinuousUsageStats,omitempty"`
```
mathetake, line 77: could you remove the change related to this? I think this is another issue and metadata is not cumulative so basically it's overriding previous ones if it's emitted in the middle.

internal/extproc/translator/translator.go
```diff
@@ -65,7 +71,7 @@ type Translator interface {
 	ResponseBody(body io.Reader, endOfStream bool) (
 		headerMutation *extprocv3.HeaderMutation,
 		bodyMutation *extprocv3.BodyMutation,
-		usedToken uint32,
+		tokenUsage *TokenUsage,
```
mathetake, line 74: i don't think this three uint64 would be worth the allocations, so could you return the value
```suggestion
		tokenUsage ToenUsage,
```

internal/extproc/translator/translator.go
```diff
@@ -12,19 +12,25 @@ import (
 	"github.com/envoyproxy/ai-gateway/internal/extproc/router"
 )
 
+type TokenUsage struct {
+	InputTokens  uint32
+	OutputTokens uint32
+	TotalTokens  uint32
+}
+
```
mathetake, line 20: Any public structs/fields/methods would need comments.

.gitignore
```diff
@@ -29,3 +29,4 @@ site/yarn-debug.log*
 site/yarn-error.log*
 site/static/.DS_Store
 site/temp
+/.idea/
```
mathetake, line 32: unnecessary change - already in here

mathetake: also you can have a global gitignore in your os... that's where people usually place the gitignore for something that's not generated by the project

mathetake:
sorry this feels a conflict with https://github.com/envoyproxy/ai-gateway/pull/103 i would appreciate it if you could stop it until it lands - I think the PR will supersede this PR besides vllm stuff part

internal/extproc/translator/translator.go
sukumargaonkar: kept it as a pointer so that `nil` value would indicate absence of tokenUsage data in the responseBody
`if tokenUsage == nil {...}`
with struct will have to check zero value for one of the member
`if tokenUsage.TotalTokens == 0 {...}`

thoughts?

mathetake:
@sukumargaonkar thank you for waiting - #103 has landed so could you rework the PR and focus on the vllm stuff?

mathetake:
ping

mathetake:
@sukumargaonkar do you still want continue the PR here? I will close this in a few days if there's no response as there's no reason to keep it open

sukumargaonkar:
yes, will rebase and include only the vllm specific changes

mathetake:
great!

mathetake:
checking - how long do you need to rework the pr here? @sukumargaonkar it should be pretty straightforward right? i wonder if this can get in the initial release.

filterconfig/filterconfig.go
yuzisun: I think this should be a property on the `AIServiceBackend` as only certain backend supports this, e.g vLLM service backend.

mathetake (commented):
a couple of question: 
* do we really want to make this as an option? even if so, adding option only to filterconfig.go doesn't make sense.
* how does this apply to aws?

yuzisun:
> a couple of question:
> 
> * do we really want to make this as an option? even if so, adding option only to filterconfig.go doesn't make sense.
> * how does this apply to aws?

I made a comment above, it should be a flag for `AIServiceBackend` if continuous token usage monitoring is supported.

mathetake:
@yuzisun I think i should've rephrased the question: why don't we just enable this logic by default (== removing the .bufferingDone flag)? In anyways, we parse all event by default (since the used token event is always the last one), so this option doesn't help us at all in terms of the computation. (On that note, i think i shouldn't have introduced `bufferingDone` in the firs tplace)

sukumargaonkar:
i agree, its better to parse every msg chunk while streaming to check for presence of usage data
also noticed that in case of vllm, when usage data in included in each chunk it includes usage of all chunks before as well.

i don't think we need to do the aggregation here
https://github.com/envoyproxy/ai-gateway/blob/2c628746cdd25e795ecb62a0045f5e1f82048e4b/internal/extproc/processor.go#L203-L205
thoughts @mathetake ?

mathetake:
> also noticed that in case of vllm, when usage data in included in each chunk it includes usage of all chunks before as well.

i see that's a good finding. is there any documentation about that? then let's change the aggregation as well as removing the  `bufferingDone` flag

mathetake:
ping

mathetake:
@sukumargaonkar hey are you still interested in this ? I would prefer the fast turnaround of a single PR and avoid keeping the context for a long time, so i would appreciate if you could rework the PR soon. Otherwise, i will close and redo it by myself. Thanks !

mathetake:
ping @sukumargaonkar

PR #112, opened 2025-01-16
test: adds initial translation e2e tests with testupstream

mathetake:


internal/controller/ai_gateway_route.go
```diff
@@ -59,7 +58,7 @@ func NewAIGatewayRouteController(
 	return &aiGatewayRouteController{
 		client:              client,
 		kube:                kube,
-		logger:              logger.WithName("ai-gateway-route-controller"),
+		logger:              logger.WithName("eaig-route-controller"),
```
yuzisun, line 61: should be ai-eg?

mathetake: haha yeah up to you mate - feel free to fix it

PR #117, opened 2025-01-17
Add support for the developer role in chat completion messages

nacx:
With o1 models and newer, `developer` messages replace the previous `system` messages.


yuzisun:
thanks @nacx ! Can you help add a test?

nacx:
What kind of additional test do you mean? I already modified the unit test that unmarshals the request to make sure it it properly handled, so I think that should be enough for the scope of the change? I mean, by adding, say, an e2e test we would be jsu testing that the backend understands that `developer` value, but wouldn't be testing anything specific to the ai-gateway project.

wengyao04:
How does it translated into Bedrock request ? In [OpenAIMessageToBedrockMessage](https://github.com/envoyproxy/ai-gateway/blob/main/internal/extproc/translator/openai_awsbedrock.go#L156C56-L156C85), if the role is `ChatMessageRoleDeveloper`, it will print error.

mathetake:
> [ommit / External Processor Test (pull_request)](https://github.com/envoyproxy/ai-gateway/actions/runs/12828614052/job/35778110401?pr=117)Failing after 2s

fyi this is a bug in ci setting - i am working on it

mathetake:
i think the existing unit test should be good. Now the question is how to translate this in aws bedrock request. Erring seems reasonable though

mathetake:
leaving it as a TODO with a new issue would be good as well

mathetake:
@nacx can you merge the main? I think we could merge as-is

mathetake (approved):

PR #124, opened 2025-01-17
Removes Image field from AIGatewayFilterConfigExternalProcess

mathetake:


manifests/charts/ai-gateway-helm/values.yaml
```diff
@@ -1,13 +1,4 @@
 # Default values for ai-gateway-helm.
-# This is a YAML-formatted file.
-# Declare variables to be passed into your templates.
-
-# ExtProc resource will be configured at CRD level, so we only need to specify the image here.
-# TODO: this also might be moved to a CRD level as in https://github.com/envoyproxy/ai-gateway/issues/82
-extProcImage:
```
yuzisun, line 7: so this is now specified on filterConfig?

yuzisun: it is a bit odd to add this on the user facing CR

mathetake: yeah odd but i think it's the same level of oddness having scaling config i guess? both are config for deployment

mathetake: maybe we should remove https://github.com/envoyproxy/ai-gateway/blob/f4ba5ccb2e17758daf0408ed296f1b0f15b18965/api/v1alpha1/api.go#L254-L273 and move it to helm values?

mathetake: i think it seems better

yuzisun: these usually should be in helm values, but each AIGatewayRoute has its own deployment. I guess the question is that if we can assume they have the same replica, resource, for image it makes sense that they should be all the same.

mathetake: ok makes sense... i will rework the pr to remove the imageref from crd

mathetake: done

yuzisun (approved):

PR #126, opened 2025-01-18
ci: do matrix over Envoy and EG versions for integration tests

mathetake:


mathetake:
on second thought, i should make this a matrix with the latest stable and main

PR #132, opened 2025-01-19
translator: toolCalls support in OpenAI to Bedrock

wengyao04:
In this PR, I improve the translator to handle toolCalls. If toolCalls is not in the OpenAI message we append toolResultBlock for assistant role message. In the Response body, we add toolCalls if bedrock content has toolUse

I also refactor `openAIMessageToBedrockMessage` to have separate functions to convert message for different roles `user`, `assistant`, `system` and `tool`.

When dealing with image content in the roles message, add a function to parse data uri and add proper error handling.


mathetake:
Could you shorten the title so that it fits in the commit title

internal/extproc/translator/openai_awsbedrock.go
```diff
@@ -28,6 +33,14 @@ func newOpenAIToAWSBedrockTranslator(path string) (Translator, error) {
...
 }
 
+func containsClaude(modelID string) bool {
+	return strings.Contains(modelID, claude)
+}
+
+func containsAnthropic(modelID string) bool {
+	return strings.Contains(modelID, anthropic)
+}
+
```
mathetake, line 43: you don't need these two seems unnecessary

internal/extproc/translator/openai_awsbedrock.go
```diff
@@ -19,6 +19,11 @@ import (
 	"github.com/envoyproxy/ai-gateway/internal/extproc/router"
 )
 
+const (
+	anthropic = "anthropic"
+	claude    = "claude"
+)
+
```
mathetake, line 26:
```suggestion
```

internal/extproc/translator/openai_awsbedrock.go
```diff
@@ -128,14 +146,24 @@ func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) openAIToolsToBedrockToolC
...
 				bedrockReq.ToolConfig.ToolChoice = &awsbedrock.ToolChoice{
 					Auto: &awsbedrock.AutoToolChoice{},
 				}
-			} else {
+			case "required":
 				bedrockReq.ToolConfig.ToolChoice = &awsbedrock.ToolChoice{
 					Any: &awsbedrock.AnyToolChoice{},
 				}
+			default:
+				if containsClaude(openAIReq.Model) && containsAnthropic(openAIReq.Model) {
```
mathetake, line 160: what is this? why is this something to do with claude?

mathetake: also why anthropic is something to do with here?

internal/extproc/translator/openai_awsbedrock.go
```diff
@@ -128,14 +146,24 @@ func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) openAIToolsToBedrockToolC
 	if openAIReq.ToolChoice != nil {
 		switch reflect.TypeOf(openAIReq.ToolChoice).Kind() {
 		case reflect.String:
-			if openAIReq.ToolChoice.(string) == "auto" {
+			toolChoice := openAIReq.ToolChoice.(string)
```
mathetake, line 149: are you 100% sure this will be a string? then why the toolchoice is interface in the first place?

internal/extproc/translator/openai_awsbedrock.go
```diff
@@ -152,113 +180,194 @@ func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) openAIToolsToBedrockToolC
...
+// openAIMessageToBedrockMessageRoleUser converts openai user role message
+// returns a list of bedrock message and error
+// the tool content is appended to the list of bedrock message list if tool_call is in openai message
+func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) openAIMessageToBedrockMessageRoleUser(
+	openAiMessage *openai.ChatCompletionMessageParamUnion,
+) ([]*awsbedrock.Message, error) {
+	message := openAiMessage.Value.(openai.ChatCompletionUserMessageParam)
+	bedrockMessages := make([]*awsbedrock.Message, 0)
+
+	if _, ok := message.Content.Value.(string); ok {
```
mathetake, line 192: you don't need to type assertion twice
```suggestion
	if v, ok := message.Content.Value.(string); ok {
```

internal/extproc/translator/openai_awsbedrock.go
```diff
@@ -152,113 +180,194 @@ func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) openAIToolsToBedrockToolC
...
 
-// OpenAIMessageToBedrockMessage converts openai ChatCompletion messages to aws bedrock messages
-func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) OpenAIMessageToBedrockMessage(openAIReq *openai.ChatCompletionRequest,
+// openAIMessageToBedrockMessageRoleUser converts openai user role message
+// returns a list of bedrock message and error
+// the tool content is appended to the list of bedrock message list if tool_call is in openai message
+func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) openAIMessageToBedrockMessageRoleUser(
+	openAiMessage *openai.ChatCompletionMessageParamUnion,
+) ([]*awsbedrock.Message, error) {
+	message := openAiMessage.Value.(openai.ChatCompletionUserMessageParam)
```
mathetake, line 189: same comment.

internal/extproc/translator/openai_awsbedrock.go
```diff
@@ -152,113 +180,194 @@ func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) openAIToolsToBedrockToolC
...
+				} else {
+					return nil, fmt.Errorf("unexpected image data url")
+				}
+			}
+		}
+		bedrockMessages = append(bedrockMessages, chatMessage)
+	} else {
+		return nil, fmt.Errorf("unexpected content type")
+	}
+	// process tool_calls
```
mathetake, line 231:
```suggestion
	// Process tool_calls.
```

internal/extproc/translator/openai_awsbedrock.go
```diff
@@ -152,113 +180,194 @@ func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) openAIToolsToBedrockToolC
...
+	return bedrockSystem, nil
+}
+
+// openAIMessageToBedrockMessageRoleTool converts openai tool role message
+func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) openAIMessageToBedrockMessageRoleTool(
+	openAiMessage *openai.ChatCompletionMessageParamUnion,
+) (*awsbedrock.Message, error) {
+	message := openAiMessage.Value.(openai.ChatCompletionToolMessageParam)
+	return &awsbedrock.Message{
+		// bedrock does not support tool role, merging to the user role
```
mathetake, line 323:
```suggestion
		// Bedrock does not support tool role, merging to the user role.
```

internal/extproc/translator/openai_awsbedrock.go
```diff
@@ -95,26 +96,43 @@ func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) RequestBody(body router.R
...
+	}
+	if openAIReq.Stop != nil {
+		bedrockReq.InferenceConfig.StopSequences = openAIReq.Stop
+	}
+	if openAIReq.Temperature != nil {
+		bedrockReq.InferenceConfig.Temperature = openAIReq.Temperature
+	}
+	if openAIReq.TopP != nil {
+		bedrockReq.InferenceConfig.TopP = openAIReq.TopP
+	}
```
mathetake, line 123: you can use `cmp.Or`

mathetake, line 123: why do you need this ifs

internal/apischema/openai/openai.go
```diff
@@ -255,7 +255,8 @@ type ChatCompletionUserMessageParam struct {
 	Role string `json:"role"`
 	// An optional name for the participant. Provides the model information to
 	// differentiate between participants of the same role.
-	Name string `json:"name,omitempty"`
+	Name      string                               `json:"name,omitempty"`
+	ToolCalls []ChatCompletionMessageToolCallParam `json:"tool_calls,omitempty"`
```
yuzisun, line 259: I do not see this defined on OpenAI schema
Screenshot 2025-01-19 at 1 01 30 AM

internal/extproc/translator/openai_awsbedrock.go
```diff
@@ -152,113 +180,194 @@ func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) openAIToolsToBedrockToolC
...
+						Image: &awsbedrock.ImageBlock{
+							Format: format,
+							Source: awsbedrock.ImageSource{
+								Bytes: []byte(parts[1]),
+							},
+						},
+					})
+				} else {
+					return nil, fmt.Errorf("unexpected image data url")
+				}
```
mathetake, line 224: how can the callsite know how the url was unexpected

internal/extproc/translator/openai_awsbedrock.go
```diff
@@ -152,113 +180,194 @@ func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) openAIToolsToBedrockToolC
...
+		for _, contentPart := range contents {
+			if contentPart.TextContent != nil {
+				textContentPart := contentPart.TextContent
+				chatMessage.Content = append(chatMessage.Content, &awsbedrock.ContentBlock{
+					Text: &textContentPart.Text,
+				})
+			} else if contentPart.ImageContent != nil {
+				imageContentPart := contentPart.ImageContent
+				parts := strings.Split(imageContentPart.ImageURL.URL, ",")
+				if len(parts) == 2 {
```
mathetake, line 211: at least please leave "detailed" code comment about what the expected image url and how this split is expected to work

internal/extproc/translator/openai_awsbedrock.go
```diff
@@ -152,113 +180,194 @@ func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) openAIToolsToBedrockToolC
...
+	if _, ok := message.Content.Value.(string); ok {
+		bedrockMessages = append(bedrockMessages, &awsbedrock.Message{
+			Role: openAiMessage.Type,
+			Content: []*awsbedrock.ContentBlock{
+				{Text: ptr.To(message.Content.Value.(string))},
+			},
+		})
+	} else if contents, ok := message.Content.Value.([]openai.ChatCompletionContentPartUserUnionParam); ok {
+		chatMessage := &awsbedrock.Message{Role: openAiMessage.Type}
+		chatMessage.Content = make([]*awsbedrock.ContentBlock, 0, len(contents))
```
mathetake, line 201: quick q: why does this need to be a pointer? values slice can be used right?

internal/extproc/translator/openai_awsbedrock.go
```diff
@@ -152,113 +180,194 @@ func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) openAIToolsToBedrockToolC
...
+
+// openAIMessageToBedrockMessageRoleSystem converts openai system role message
+// returns a list of bedrock system content and error
+func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) openAIMessageToBedrockMessageRoleSystem(
+	openAiMessage *openai.ChatCompletionMessageParamUnion,
+) ([]*awsbedrock.SystemContentBlock, error) {
+	message := openAiMessage.Value.(openai.ChatCompletionSystemMessageParam)
+	bedrockSystem := make([]*awsbedrock.SystemContentBlock, 0)
+
+	if _, ok := message.Content.Value.(string); ok {
```
mathetake, line 300: same comment

internal/extproc/translator/openai_awsbedrock.go
```diff
@@ -152,113 +180,194 @@ func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) openAIToolsToBedrockToolC
...
+			Role:    openAiMessage.Type,
+			Content: []*awsbedrock.ContentBlock{{Text: message.Content.Refusal}},
+		})
+	} else {
+		bedrockMessages = append(bedrockMessages, &awsbedrock.Message{
+			Role:    openAiMessage.Type,
+			Content: []*awsbedrock.ContentBlock{{Text: message.Content.Text}},
+		})
+	}
+	// process tool_calls
```
mathetake, line 275: please complete the comment line (always starting with capital and ending with period.

internal/apischema/awsbedrock/awsbedrock.go
```diff
@@ -299,6 +302,14 @@ type ToolResultBlock struct {
...
 }
 
+// ToolUseBlock A tool use block contains information about a tool that the model is requesting be run.
+// The model uses the result from the tool to generate a response.
+type ToolUseBlock struct {
+	Name      *string `json:"name"`
+	Input     *string `json:"input"`
+	ToolUseID *string `json:"toolUseId"`
+}
+
```
mathetake, line 312: can you put the url?

internal/apischema/awsbedrock/awsbedrock.go
```diff
@@ -299,6 +302,14 @@ type ToolResultBlock struct {
 	ToolUseID *string `json:"toolUseId"`
 }
 
+// ToolUseBlock A tool use block contains information about a tool that the model is requesting be run.
+// The model uses the result from the tool to generate a response.
+type ToolUseBlock struct {
+	Name      *string `json:"name"`
+	Input     *string `json:"input"`
+	ToolUseID *string `json:"toolUseId"`
```
mathetake, line 310: null or not ?

internal/extproc/translator/openai_awsbedrock.go
```diff
@@ -152,113 +180,194 @@ func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) openAIToolsToBedrockToolC
...
+	message := openAiMessage.Value.(openai.ChatCompletionToolMessageParam)
+	return &awsbedrock.Message{
+		// bedrock does not support tool role, merging to the user role
+		Role: awsbedrock.ConversationRoleUser,
+		Content: []*awsbedrock.ContentBlock{
+			{
+				ToolResult: &awsbedrock.ToolResultBlock{
+					Content: []*awsbedrock.ToolResultContentBlock{
+						{
+							Text: message.Content.Value.(*string),
```
mathetake, line 330: same comment

internal/apischema/openai/openai.go
```diff
@@ -581,7 +582,6 @@ const (
 	ChatCompletionChoicesFinishReasonLength        ChatCompletionChoicesFinishReason = "length"
 	ChatCompletionChoicesFinishReasonToolCalls     ChatCompletionChoicesFinishReason = "tool_calls"
 	ChatCompletionChoicesFinishReasonContentFilter ChatCompletionChoicesFinishReason = "content_filter"
-	ChatCompletionChoicesFinishReasonFunctionCall  ChatCompletionChoicesFinishReason = "function_call"
```
mathetake, line 584: why is this getting deleted

internal/extproc/translator/openai_awsbedrock.go
mathetake: ok the reflection above.... that is totally overkill and unnecessary

internal/extproc/translator/openai_awsbedrock.go
```diff
@@ -152,113 +180,194 @@ func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) openAIToolsToBedrockToolC
...
-							} else {
-								return fmt.Errorf("unexpected image data url")
-							}
-						}
-					}
-					bedrockReq.Messages = append(bedrockReq.Messages, chatMessage)
-				} else {
-					return fmt.Errorf("unexpected content type for user message")
-				}
+			bedrockMessages, err := o.openAIMessageToBedrockMessageRoleUser(&msg)
```
mathetake, line 348: this escapes the message... instead make it the pointer at the beginning of the loop in the first place.

internal/extproc/translator/openai_awsbedrock.go
```diff
@@ -152,113 +180,194 @@ func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) openAIToolsToBedrockToolC
...
 }
 
-// OpenAIMessageToBedrockMessage converts openai ChatCompletion messages to aws bedrock messages
-func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) OpenAIMessageToBedrockMessage(openAIReq *openai.ChatCompletionRequest,
+// openAIMessageToBedrockMessageRoleUser converts openai user role message
+// returns a list of bedrock message and error
+// the tool content is appended to the list of bedrock message list if tool_call is in openai message
+func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) openAIMessageToBedrockMessageRoleUser(
+	openAiMessage *openai.ChatCompletionMessageParamUnion,
+) ([]*awsbedrock.Message, error) {
```
mathetake, line 188: this returned slice allocation is wasted as at the callsite the results are appended into another slice and this return value is discarded immediately

internal/extproc/translator/openai_awsbedrock.go
```diff
@@ -152,113 +180,194 @@ func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) openAIToolsToBedrockToolC
...
+				textContentPart := contentPart.TextContent
+				chatMessage.Content = append(chatMessage.Content, &awsbedrock.ContentBlock{
+					Text: &textContentPart.Text,
+				})
+			} else if contentPart.ImageContent != nil {
+				imageContentPart := contentPart.ImageContent
+				parts := strings.Split(imageContentPart.ImageURL.URL, ",")
+				if len(parts) == 2 {
+					formatPart := strings.Split(parts[0], ";")[0]
+					format := strings.TrimPrefix(formatPart, "data:image/")
```
mathetake, line 213: also this really needs a comment

internal/extproc/translator/openai_awsbedrock.go
```diff
@@ -152,113 +180,194 @@ func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) openAIToolsToBedrockToolC
...
+	return bedrockMessages, nil
+}
+
+// openAIMessageToBedrockMessageRoleAssistant converts openai assistant role message
+// returns a list of bedrock message and error
+// the tool content is appended to the list of bedrock message list if tool_call is in openai message
+func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) openAIMessageToBedrockMessageRoleAssistant(
+	openAiMessage *openai.ChatCompletionMessageParamUnion,
+) ([]*awsbedrock.Message, error) {
+	bedrockMessages := make([]*awsbedrock.Message, 0)
```
mathetake, line 261: same comment (ignore if not applicable) why is this pointer slice instead of value?)

internal/extproc/translator/openai_awsbedrock.go
```diff
@@ -152,113 +180,194 @@ func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) openAIToolsToBedrockToolC
...
+		})
+	}
+	return bedrockMessages, nil
+}
+
+// openAIMessageToBedrockMessageRoleSystem converts openai system role message
+// returns a list of bedrock system content and error
+func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) openAIMessageToBedrockMessageRoleSystem(
+	openAiMessage *openai.ChatCompletionMessageParamUnion,
+) ([]*awsbedrock.SystemContentBlock, error) {
```
mathetake, line 296: same comment on the returned value slice allocation being wasted

internal/extproc/translator/openai_awsbedrock.go
```diff
@@ -286,6 +395,55 @@ func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) ResponseHeaders(headers m
...
+	case awsbedrock.StopReasonToolUse:
+		return openai.ChatCompletionChoicesFinishReasonToolCalls
+	default:
+		return openai.ChatCompletionChoicesFinishReasonStop
+	}
+}
+
+func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) bedrockToolUseToOpenAICalls(
+	toolUse *awsbedrock.ToolUseBlock,
+) []openai.ChatCompletionMessageToolCallParam {
```
mathetake, line 421: why do you need slice

internal/extproc/translator/openai_awsbedrock.go
```diff
@@ -337,6 +495,7 @@ func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) ResponseBody(body io.Read
 		Object:  "chat.completion",
 		Choices: make([]openai.ChatCompletionResponseChoice, 0, len(bedrockResp.Output.Message.Content)),
 	}
+	// convert token usage
```
mathetake, line 498:
```suggestion
	// Convert token usage.
```

internal/extproc/translator/openai_awsbedrock.go
yuzisun: This is my bad, should have added the comment in my original code as this is is parsing as image data url
https://platform.openai.com/docs/guides/vision#uploading-base64-encoded-images

internal/extproc/translator/openai_awsbedrock.go
yuzisun: Is this to check only Anthropic Claude support tooling ? Might be better to leave such validation to the provider.

internal/extproc/translator/openai_awsbedrock.go
yuzisun: I think this is following the code in the aws sdk, for large content block it might be more efficient to use pointer, otherwise it needs to allocate a large chunk of continuous memory.

internal/extproc/translator/openai_awsbedrock.go
yuzisun: True this was my original code, ifs can be removed as they are all pointer types

internal/extproc/translator/openai_awsbedrock.go
wengyao04: Yes, Claude supports [tool_choice](https://docs.anthropic.com/en/docs/build-with-claude/tool-use). When working with the tool_choice parameter, there three possible options:

- `auto` allows Claude to decide whether to call any provided tools or not.
- `any` tells Claude that it must use one of the provided tools, but doesn't force a particular tool.
- `tool` allows us to force Claude to always use a particular tool. When use `tool`, the format is
   ```
   {"tool": "tool_name"}
   ```
   I will put a comment

internal/apischema/openai/openai.go
wengyao04: It is not used anywhere in the code, we can add it if we need it in the future

yuzisun: function_call is deprecated and replaced by tool_calls

internal/extproc/translator/openai_awsbedrock.go
wengyao04: I will update the how we parse data uri and add comments

wengyao04:
- address comments
- fix assistant role message with toolCalls, it should return one bedrock message with a slice of content
- add `parseDataURI` to parse data uri

internal/apischema/openai/openai.go
```diff
@@ -646,8 +645,9 @@ type ChatCompletionResponseChoice struct {
 // ChatCompletionResponseChoiceMessage is described in the OpenAI API documentation:
 // https://platform.openai.com/docs/api-reference/chat/object#chat/object-choices
 type ChatCompletionResponseChoiceMessage struct {
-	Content *string `json:"content,omitempty"`
-	Role    string  `json:"role,omitempty"`
+	Content   *string                              `json:"content,omitempty"`
+	Role      string                               `json:"role,omitempty"`
+	ToolCalls []ChatCompletionMessageToolCallParam `json:"tool_calls,omitempty"`
```
yuzisun, line 650:
```suggestion
        // The contents of the message.
	Content   *string                              `json:"content,omitempty"`
	// The role of the author of this message.
	Role      string                               `json:"role,omitempty"`
	// The tool calls generated by the model, such as function calls.
	ToolCalls []ChatCompletionMessageToolCallParam `json:"tool_calls,omitempty"`
```

internal/extproc/translator/openai_awsbedrock.go
mathetake: again, this is not needed. how does this help developers read code? it's literally the same despite the unnecessary constants.

internal/extproc/translator/openai_awsbedrock.go
```diff
@@ -28,6 +42,10 @@ func newOpenAIToAWSBedrockTranslator(path string) (Translator, error) {
 	}
 }
 
+func isAnthropicClaude(modelID string) bool {
+	return strings.Contains(modelID, anthropic) && strings.Contains(modelID, claude)
+}
```
mathetake, line 47: you don't need this

internal/extproc/translator/openai_awsbedrock.go
```diff
@@ -152,113 +182,184 @@ func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) openAIToolsToBedrockToolC
...
+
+// parseDataURI parse data uri example: data:image/jpeg;base64,/9j/4AAQSkZJRgABAgAAZABkAAD
+func parseDataURI(uri string) (string, []byte, bool) {
+	matches := regDataURI.FindStringSubmatch(uri)
+	if len(matches) != 3 {
+		return "", nil, false
+	}
+	l := len(matches[0])
+	contentType := matches[1]
+	bin, _ := base64.StdEncoding.DecodeString(uri[l:])
```
mathetake, line 196: error handling

internal/extproc/translator/openai_awsbedrock.go
mathetake: well that's only true if that's the scale of 100MB or 1Gb, but i don't believe that's the case. i would say this is a bad code of AWS

internal/extproc/translator/openai_awsbedrock.go
```diff
@@ -152,113 +182,184 @@ func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) openAIToolsToBedrockToolC
...
+func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) openAIMessageToBedrockMessageRoleAssistant(
+	openAiMessage *openai.ChatCompletionMessageParamUnion,
+) (*awsbedrock.Message, error) {
+	var bedrockMessage *awsbedrock.Message
+	message := openAiMessage.Value.(openai.ChatCompletionAssistantMessageParam)
+
+	if message.Content.Type == openai.ChatCompletionAssistantMessageParamContentTypeRefusal {
+		bedrockMessage = &awsbedrock.Message{
+			Role:    openAiMessage.Type,
+			Content: []*awsbedrock.ContentBlock{{Text: message.Content.Refusal}},
```
mathetake, line 262: can you not preallocate the slice space

internal/extproc/translator/openai_awsbedrock.go
mathetake: well that's generally untrue. if the size is that big then allocation per item gets a real bottleneck (believe me i wrote a native compiler in wazero project and this is real one) vs if the size is small allocation of the entire contiguous region is much faster and less GC pressure. i would say this is a bad code of AWS and I would suggest not follow their code and fix it (can do this later).

internal/extproc/translator/openai_awsbedrock.go
```diff
@@ -152,113 +182,184 @@ func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) openAIToolsToBedrockToolC
...
+				},
+			})
+	}
+	return bedrockMessage, nil
+}
+
+// openAIMessageToBedrockMessageRoleSystem converts openai system role message
+// returns a list of bedrock system content and error
+func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) openAIMessageToBedrockMessageRoleSystem(
+	openAiMessage *openai.ChatCompletionMessageParamUnion, bedrockSystem *[]*awsbedrock.SystemContentBlock,
```
mathetake, line 287: why do you need to pass slice

mathetake:
@wengyao04 @yuzisun i guess the items list PR description might be a BB style but could you from now write messages in sentences? (i will add a PR template later, but for now could you follow my PR like #135)

internal/extproc/translator/openai_awsbedrock.go
wengyao04: remove the constant

internal/extproc/translator/openai_awsbedrock.go
```diff
@@ -152,113 +182,184 @@ func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) openAIToolsToBedrockToolC
...
+				Text: textContentPart,
+			})
+		}
+	} else {
+		return fmt.Errorf("unexpected content type for system message")
+	}
+	return nil
+}
+
+// openAIMessageToBedrockMessageRoleTool converts openai tool role message
```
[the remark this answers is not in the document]
mathetake: ok then let's fix them all after follow up

internal/extproc/translator/openai_awsbedrock.go
```diff
@@ -128,14 +121,29 @@ func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) openAIToolsToBedrockToolC
...
-			} else {
+			case "required":
 				bedrockReq.ToolConfig.ToolChoice = &awsbedrock.ToolChoice{
 					Any: &awsbedrock.AnyToolChoice{},
 				}
+			default:
+				// Anthropic Claude supports tool_choice parameter with three options.
+				// auto: allows Claude to decide whether to call any provided tools or not.
+				// any tells Claude that it must use one of the provided tools, but doesn't force a particular tool.
+				// tool allows us to force Claude to always use a particular tool.
```
mathetake, line 138:
```suggestion
				// Anthropic Claude supports tool_choice parameter with three options.
				// * `auto` allows Claude to decide whether to call any provided tools or not.
				// * `any` tells Claude that it must use one of the provided tools, but doesn't force a particular tool.
				// * `tool` allows us to force Claude to always use a particular tool.
```

internal/extproc/translator/openai_awsbedrock.go
```diff
@@ -128,14 +121,29 @@ func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) openAIToolsToBedrockToolC
...
+				// The tool option is only applied to Anthropic Claude
```
mathetake, line 139:
```suggestion
				// The tool option is only applied to Anthropic Claude.
```

internal/extproc/translator/openai_awsbedrock.go
wengyao04: return named slice

internal/extproc/translator/openai_awsbedrock.go
```diff
@@ -152,113 +160,196 @@ func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) openAIToolsToBedrockToolC
...
+			})
+	}
+	return bedrockMessage, nil
+}
+
+// openAIMessageToBedrockMessageRoleSystem converts openai system role message
+// returns a list of bedrock system content and error
+func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) openAIMessageToBedrockMessageRoleSystem(
+	openAiMessage *openai.ChatCompletionSystemMessageParam,
+) (bedrockSystem []*awsbedrock.SystemContentBlock, err error) {
```
mathetake, line 275: same, wasted slice.

internal/extproc/translator/openai_awsbedrock.go
```diff
@@ -111,10 +105,9 @@ func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) openAIToolsToBedrockToolC
 	bedrockReq.ToolConfig = &awsbedrock.ToolConfiguration{}
 	tools := make([]*awsbedrock.Tool, 0, len(openAIReq.Tools))
 	for _, toolDefinition := range openAIReq.Tools {
-		toolType := (string)(toolDefinition.Type)
 		tool := &awsbedrock.Tool{
 			ToolSpec: &awsbedrock.ToolSpecification{
-				Name:        &toolType,
+				Name:        &toolDefinition.Function.Name,
```
mathetake, line 110: i believe this is dangerous as this points to a loop var. you should extract the name in the local variable inside loop

mathetake: basically this feels unnecessary change

wengyao04:
> @wengyao04 @yuzisun i guess the items list PR description might be a BB style but could you from now write messages in sentences? (i will add a PR template later, but for now could you follow my PR like #135)

Update the commit message with readable paragraph.

internal/extproc/translator/openai_awsbedrock.go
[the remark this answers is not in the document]
wengyao04: It is all fixed with cased value as argument

internal/extproc/translator/openai_awsbedrock.go
```diff
@@ -152,113 +160,196 @@ func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) openAIToolsToBedrockToolC
...
+	bin, err := base64.StdEncoding.DecodeString(uri[l:])
+	if err != nil {
+		return "", nil, err
+	}
+	return contentType, bin, nil
+}
+
+// openAIMessageToBedrockMessageRoleUser converts openai user role message
+// returns a list of bedrock message and error
+// the tool content is appended to the list of bedrock message list if tool_call is in openai message.
```
mathetake, line 183: can you please follow the same style as elsewhere and per my previous comments? 
```suggestion
// openAIMessageToBedrockMessageRoleUser converts openai user role message
// returns a list of bedrock message and error.
// The tool content is appended to the list of bedrock message list if tool_call is in openai message.
```

internal/extproc/translator/openai_awsbedrock.go
```diff
@@ -152,113 +160,196 @@ func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) openAIToolsToBedrockToolC
...
+		}
+		return chatMessage, nil
+	} else {
+		return nil, fmt.Errorf("unexpected content type")
+	}
+}
+
+// openAIMessageToBedrockMessageRoleAssistant converts openai assistant role message
+// returns a list of bedrock message and error
+// the tool content is appended to the list of bedrock message list if tool_call is in openai message
```
mathetake, line 242: same comment

internal/extproc/translator/openai_awsbedrock.go
mathetake: maybe my previous comment was bad - accepting pointer to the slice here would be better

wengyao04: revert it back

mathetake (approved):
thanks for the multiple iterations. style wise LGTM, so deferring to @yuzisun for the actual logic final approval

wengyao04:
> thanks for the multiple iterations. style wise LGTM, so deferring to @yuzisun for the actual logic final approval

Thanks you for helping review, I will send another PR to add ToolCalls support in converse stream

internal/extproc/translator/openai_awsbedrock.go
```diff
@@ -152,95 +166,183 @@ func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) openAIToolsToBedrockToolC
 	return nil
 }
 
-// OpenAIMessageToBedrockMessage converts openai ChatCompletion messages to aws bedrock messages
-func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) OpenAIMessageToBedrockMessage(openAIReq *openai.ChatCompletionRequest,
+// regDataURI follows the web uri regex definition.
+// https://developer.mozilla.org/en-US/docs/Web/URI/Schemes/data#syntax
+var regDataURI = regexp.MustCompile(`\Adata:(.+?)?(;base64)?,`)
```
yuzisun, line 171: @wengyao04 can you help qualify the content type with `image/` ? we are only supporting image data url here.

wengyao04: @yuzisun we have image format validation in https://github.com/envoyproxy/ai-gateway/pull/132/files#diff-6da52452dbe8387dcea4abc098618db9120d7d8b3071bf502d60530912158737R213-R225. Prefer to use data uri syntax for regex https://en.wikipedia.org/wiki/Data_URI_scheme

internal/extproc/translator/openai_awsbedrock.go
```diff
@@ -152,95 +166,183 @@ func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) openAIToolsToBedrockToolC
...
+		bedrockMessage.Content = append(bedrockMessage.Content,
+			&awsbedrock.ContentBlock{
+				ToolUse: &awsbedrock.ToolUseBlock{
+					Name:      toolCall.Function.Name,
+					ToolUseID: toolCall.ID,
+					Input:     toolCall.Function.Arguments,
+				},
+			})
+	}
+	return bedrockMessage, nil
```
yuzisun, line 272: According to https://docs.aws.amazon.com/bedrock/latest/userguide/tool-use-inference-call.html, seems like the ToolUse is only used in the ConverseOutput, so the translation should be the other way around from Bedrock response to OpenAI response?

yuzisun: nvm, missed the code in responseBody, but I think the test case is missing for converting the ToolUse in `ConverseOutput` to tool_calls in OpenAI response.

internal/extproc/translator/openai_awsbedrock.go
```diff
@@ -378,26 +509,18 @@ func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) ResponseBody(body io.Read
...
-			case awsbedrock.StopReasonStopSequence, awsbedrock.StopReasonEndTurn:
-				choice.FinishReason = openai.ChatCompletionChoicesFinishReasonStop
-			case awsbedrock.StopReasonMaxTokens:
-				choice.FinishReason = openai.ChatCompletionChoicesFinishReasonLength
-			case awsbedrock.StopReasonContentFiltered:
-				choice.FinishReason = openai.ChatCompletionChoicesFinishReasonContentFilter
-			case awsbedrock.StopReasonToolUse:
-				choice.FinishReason = openai.ChatCompletionChoicesFinishReasonToolCalls
-			}
+		if toolCall := o.bedrockToolUseToOpenAICalls(output.ToolUse); toolCall != nil {
```
yuzisun, line 514: Can you help add a test case for this ?

wengyao04: test is added

internal/extproc/translator/openai_awsbedrock_test.go
```diff
@@ -395,6 +423,250 @@ func TestOpenAIToAWSBedrockTranslatorV1ChatCompletion_RequestBody(t *testing.T)
...
+					ToolChoice: &awsbedrock.ToolChoice{
+						Tool: &awsbedrock.SpecificToolChoice{
+							Name: ptr.To("some-tools"),
+						},
+					},
+				},
+			},
+		},
+		{
+			name: "test structured tool choices",
```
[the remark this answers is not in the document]
wengyao04: It is for anthropic Claude tool choice, I rename it to `test tool choices for anthropic claude model`

internal/extproc/translator/openai_awsbedrock_test.go
```diff
@@ -596,6 +871,49 @@ func TestOpenAIToAWSBedrockTranslatorV1ChatCompletion_ResponseBody(t *testing.T)
...
+						},
+					},
+				},
+			},
+			output: openai.ChatCompletionResponse{
+				Object: "chat.completion",
+				Choices: []openai.ChatCompletionResponseChoice{
+					{
+						Index:        0,
+						FinishReason: openai.ChatCompletionChoicesFinishReasonStop,
```
yuzisun, line 898: The finish reason here should be “tool_calls” as the stop reason on ConverseOutput is set to toolUse in this case.

wengyao04: update the unit test, for input bedrock converse response, the stop reason is `tool_use` for content with toolUseBlock

yuzisun (approved):
Thanks @wengyao04 !!

PR #135, opened 2025-01-20
e2e: ensures the route recalculated

mathetake:
Previously, internally `x-envoy-ai-gateway-selected-backend`
was used for routing. However, it has some conflicts with the 
EG default setting named `suprressEnvoyHeaders: true` and 
that caused the route recalculation ignored. This fixes it.
In general, we shouldn't use the Envoy internal headers 
x-envoy-* for headers we generate in this project.

Thanks @yuzisun for finding this out.


tests/extproc/envoy.yaml
```diff
@@ -59,8 +59,6 @@ static_resources:
                     typed_config:
                       "@type": type.googleapis.com/envoy.extensions.filters.http.ext_proc.v3.ExternalProcessor
                       allow_mode_override: true
-                      mutation_rules:
-                        allow_all_routing: true
```
mathetake, line 63: this was unneeded and makes the configuration diff between EG and raw envoy here smaller

api/v1alpha1/api.go
```diff
@@ -61,7 +61,7 @@ type AIGatewayRouteSpec struct {
 	// modifications to achieve the necessary jobs, notably inserting the AI Gateway filter responsible for
 	// the transformation of the request and response, etc.
 	//
-	// In the matching conditions in the AIGatewayRouteRule, `x-envoy-ai-gateway-model` header is available
+	// In the matching conditions in the AIGatewayRouteRule, `x-eaig-model` header is available
```
yuzisun, line 64: tbh, `eaig` looks like a typo unfortunately, suggest `x-ai-eg-model` which reads better?

mathetake: fixed

mathetake: let's stick to ai-eg-* from now!

PR #144, opened 2025-01-20
extproc: fix filter config marshalling

nacx:
The YAML library being used does not honor the `yaml` tags. It just considers the `json` ones and the extproc filter config was not unmarshaleld according to the defined values.
The `llmRequestCosts.metadataKey` and `MetadataNamespace` properties had wrong tags.


yuzisun:
If we use json tags, shall we reuse the types defined from the control plane API? I feel like we are duplicating the types and it is hard to maintain going forward.

mathetake (approved):
💯

mathetake:
Re: reusing - that's intentional and this is to decouple EG specific doc comment etc. for example backend here is a mere string vs the real Backend resource. If this gets really hard to maintain let's switch back to it, but for now let me take care of it:🙇

PR #149, opened 2025-01-21
extproc: allow configuring an UDS listen address

nacx:
Allows configuring extproc to listen on an Unix domain socket. Note that this PR just enables it in the extproc filter. Follow-up PRs will be needed to accommodate that in the controller as well, but that would require further changes to how the Gateway itself (not only the filter) is deployed, to mount a common volume for the UDS.

### DIFF
--- a/cmd/extproc/mainlib/main.go
+++ b/cmd/extproc/mainlib/main.go
@@ -8,6 +8,7 @@ import (
 	"net"
 	"os"
 	"os/signal"
+	"strings"
 	"syscall"
 	"time"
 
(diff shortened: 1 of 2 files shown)


cmd/extproc/mainlib/main.go
```diff
@@ -19,12 +20,16 @@ import (
...
+	defaultLogLevel = "info"
+)
+
 var (
 	configPath = flag.String("configPath", "", "path to the configuration file. "+
 		"The file must be in YAML format specified in extprocconfig.Config type. The configuration file is watched for changes.")
-	// TODO: unix domain socket support.
-	extProcPort = flag.String("extProcPort", ":1063", "gRPC port for the external processor")
-	logLevel    = flag.String("logLevel", "info", "log level")
+	extProcAddr = flag.String("extProcAddr", defaultAddress, "gRPC address for the external processor")
```
nacx, line 31: This could be a breaking change for users using the extproc directly. The controller doesn't explicitly set the flag and relies on the default value so that should not break.

mathetake (approved):

cmd/extproc/mainlib/main.go
mathetake: yeah

PR #151, opened 2025-01-21
translator: add toolUse support in streaming

wengyao04:
In this PR, I convert `toolUse` in bedrock converse streaming response to openai chat completions `tool_calls`. `ToolUseBlockDelta` is added to`ConverseStreamEventContentBlockDelta` as documented in [ContentBlockDelta](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_ContentBlockDelta.html). `ContentBlockStart` is added as documented in [ContentBlockStart](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_ContentBlockStart.html)

The role in `ChatCompletionResponseChunkChoiceDelta` is not nullable as documented in https://platform.openai.com/docs/api-reference/chat/streaming

Unit test is updated to cover streaming response with `toolUse`. The base64 encoded streaming test data if from calling bedrock anthropic Claude model with `toolConfig`.


mathetake:
looks good to me in general, so deferring to Dan for the rest

mathetake:
so i made it easy to add the end-to-end transformation testing with testupstream here: https://github.com/envoyproxy/ai-gateway/blob/b1b23e2abf707e01e93b706dcbee3fd60db46ed7/tests/extproc/testupstream_test.go#L54

feel free to add more cases to cover this PR (maybe some fix might be necessary for the test as well as testupstream)

wengyao04:
> so i made it easy to add the end-to-end transformation testing with testupstream here:
> 
> https://github.com/envoyproxy/ai-gateway/blob/b1b23e2abf707e01e93b706dcbee3fd60db46ed7/tests/extproc/testupstream_test.go#L54
> 
> feel free to add more cases to cover this PR (maybe some fix might be necessary for the test as well as testupstream)

Thank you, I will add more test cases to cover this PR and previous one for toolCalls

yuzisun (approved):

Issue #133, opened 2025-01-19
Error handling for aws bedrock and openai response


yuzisun:
* Add logic to handle error response from upstream aws bedrock or openai, currently it is assuming the response is 200 and fails on response body unmarshalling.
* Add tests for error responses from upstream 
```
time=2025-01-19T18:34:38.104Z level=INFO msg="starting external processor" version=8056bb8bcfe484ca6f6007b021ea69448b3073f4
time=2025-01-19T18:34:38.104Z level=INFO msg="loading a new config" path=/etc/ai-gateway/extproc/extproc-config.yaml
time=2025-01-19T18:34:38.104Z level=INFO msg="start watching the config file" path=/etc/ai-gateway/extproc/extproc-config.yaml interval=5s
time=2025-01-19T18:34:42.422Z level=DEBUG msg="request headers processing" request_headers="headers:{key:\":authority\"  raw_value:\"localhost:8080\"}  headers:{key:\":path\"  raw_value:\"/v1/chat/completions\"}  headers:{key:\":method\"  raw_value:\"POST\"}  headers:{key:\":scheme\"  raw_value:\"http\"}  headers:{key:\"user-agent\"  raw_value:\"curl/8.7.1\"}  headers:{key:\"accept\"  raw_value:\"*/*\"}  headers:{key:\"content-type\"  raw_value:\"application/json\"}  headers:{key:\"x-ai-gateway-llm-backend\"  raw_value:\"kserve-llama-backend\"}  headers:{key:\"content-length\"  raw_value:\"113\"}  headers:{key:\"x-forwarded-for\"  raw_value:\"10.244.1.40\"}  headers:{key:\"x-forwarded-proto\"  raw_value:\"http\"}  headers:{key:\"x-envoy-external-address\"  raw_value:\"127.0.0.1\"}  headers:{key:\"x-request-id\"  raw_value:\"41605e8c-7ef8-4fbc-b1f9-a8f01740746e\"}"
time=2025-01-19T18:34:42.422Z level=DEBUG msg="request headers processed" response=request_headers:{}
time=2025-01-19T18:34:42.422Z level=DEBUG msg="request body processing" request="request_body:{body:\"{\\\"model\\\": \\\"dsp.llama-3.3-70b-instruct\\\", \\\"prompt\\\": \\\"Write a poem about colors\\\", \\\"stream\\\":false, \\\"max_tokens\\\": 100}\"  end_of_stream:true}  metadata_context:{}"
[...]

PR #152, opened 2025-01-21
translator: handling for OpenAI and AWSBedrock error response

yuzisun:
Add handling for the OpenAI and Bedrock error response. Currently when error response is returned it fails on decoding the `ConverseOutput` or `OpenAIChatCompletionResponse` as it is assuming a 200 response.

For AWSBedrock the error response is converted to OpenAIError type when returning to the user. For OpenAI error response we return directly back without translation. In case of connection errors to the AWS Bedrock or OpenAI backend, we translate to OpenAIError type based on the status code and message body. 

fixes https://github.com/envoyproxy/ai-gateway/issues/133


mathetake:
can you fix the e2e

internal/extproc/translator/openai_awsbedrock.go
```diff
@@ -385,6 +386,9 @@ func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) openAIMessageToBedrockMes
 func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) ResponseHeaders(headers map[string]string) (
 	headerMutation *extprocv3.HeaderMutation, err error,
 ) {
+	if v, ok := headers[statusHeaderName]; ok && v != string(typev3.StatusCode_OK) {
+		return nil, nil
+	}
```
mathetake, line 391: I think this is not necessary or should be panic... if we reach this, that means a bug in extproc code

internal/extproc/translator/openai_openai.go
```diff
@@ -48,10 +49,54 @@ func (o *openAIToOpenAITranslatorV1ChatCompletion) RequestBody(body router.Reque
...
+// For OpenAI based backend we return the OpenAI error type as is.
+// If connection fails the error body is translated to OpenAI error type for events such as HTTP 503 or 504.
+func (o *openAIToOpenAITranslatorV1ChatCompletion) ResponseError(respHeaders map[string]string, body io.Reader) (
+	headerMutation *extprocv3.HeaderMutation, bodyMutation *extprocv3.BodyMutation, err error,
+) {
+	var statusCode string
+	var ok bool
+	if statusCode, ok = respHeaders[statusHeaderName]; ok && statusCode == string(typev3.StatusCode_OK) {
+		return nil, nil, nil
+	}
```
mathetake, line 62: same comment

internal/extproc/translator/openai_openai.go
```diff
@@ -48,10 +49,54 @@ func (o *openAIToOpenAITranslatorV1ChatCompletion) RequestBody(body router.Reque
...
+	if v, ok := respHeaders[contentTypeHeaderName]; ok {
```
mathetake, line 63: i wonder in practice when does this if block is not entered

internal/extproc/translator/translator.go
```diff
@@ -12,6 +12,15 @@ import (
 	"github.com/envoyproxy/ai-gateway/internal/extproc/router"
 )
 
+var (
+	statusHeaderName       = ":status"
+	contentTypeHeaderName  = "content-type"
+	awsErrorTypeHeaderName = "x-amzn-errortype"
+	jsonContentType        = "application/json"
+	OpenAIBackendError     = "OpenAIBackendError"
+	AWSBedrockBackendError = "AWSBedrockBackendError"
```
mathetake, line 21:
```suggestion
	openAIBackendError     = "OpenAIBackendError"
	awsBedrockBackendError = "AWSBedrockBackendError"
```

extprocapi/extproc.go
mathetake, line 1: unnecessary change?

mathetake:
so i made it easy to add the end-to-end transformation testing with testupstream here: https://github.com/envoyproxy/ai-gateway/blob/b1b23e2abf707e01e93b706dcbee3fd60db46ed7/tests/extproc/testupstream_test.go#L54

feel free to add more cases to cover this case (maybe some fix might be necessary for the test as well as testupstream)

internal/extproc/translator/openai_openai.go
yuzisun: removed the check

internal/extproc/translator/translator.go
```diff
@@ -12,6 +12,19 @@ import (
...
+	contentTypeHeaderName  = "content-type"
+	awsErrorTypeHeaderName = "x-amzn-errortype"
+	jsonContentType        = "application/json"
+	openAIBackendError     = "OpenAIBackendError"
+	awsBedrockBackendError = "AWSBedrockBackendError"
+)
+
+func isGoodStatusCode(code int) bool {
+	return code >= 200 && code < 300
+}
```
mathetake, line 26: This needs a detailed comments as the meaning is not obvious without any specific knowledge about how is this used. Like which status codes will be returned by who and when these things happen

internal/extproc/translator/openai_awsbedrock.go
```diff
@@ -385,6 +386,13 @@ func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) openAIMessageToBedrockMes
 func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) ResponseHeaders(headers map[string]string) (
 	headerMutation *extprocv3.HeaderMutation, err error,
 ) {
+	if v, ok := headers[statusHeaderName]; ok {
+		if v, err := strconv.Atoi(v); err == nil {
+			if !isGoodStatusCode(v) {
+				return nil, nil
+			}
+		}
```
mathetake, line 394: also i am still not sure why if this is a bad status code we can continue and when good we return early. could you put the detailed code comments. This will be hard to understand by anyone without the extensive knowledge. I believe the comments will be helpful for future contributors (including me!)

mathetake (commented):
LGTM modulo i need code comments!

internal/extproc/translator/openai_awsbedrock.go
yuzisun: I rewrote the code in a way that it does not require this logic. Before it incorrectly returns a content type check error on error response. I am not a fan of writing too much code comments which means the code is not self-explainable.

yuzisun:
> so i made it easy to add the end-to-end transformation testing with testupstream here:
> 
> https://github.com/envoyproxy/ai-gateway/blob/b1b23e2abf707e01e93b706dcbee3fd60db46ed7/tests/extproc/testupstream_test.go#L54
> 
> feel free to add more cases to cover this case (maybe some fix might be necessary for the test as well as testupstream)

Added the tests all ready to go

mathetake:
Will take a look with an hour!

mathetake (approved):

PR #153, opened 2025-01-21
feat: implements CEL expression API for costs

### LINKED ISSUE #97: Allow customizing token cost calculation

mathetake:
This commit implements the CEL expression API
left as a TODO in #103. Closes #97.


mathetake:
@yuzisun PTAL

Makefile
```diff
@@ -103,7 +103,7 @@ editorconfig: editorconfig-checker
 .PHONY: test
 test:
 	@echo "test => ./..."
-	@go test -v ./...
+	@go test -v -race ./...
```
mathetake, line 106: note: this was to enhance the concurrency CEL test (we should've added before anyways)

internal/extproc/processor.go
```diff
@@ -11,25 +11,32 @@ import (
...
+	"github.com/envoyproxy/ai-gateway/internal/llmcostcel"
 )
 
 // processorConfig is the configuration for the processor.
 // This will be created by the server and passed to the processor when it detects a new configuration.
 type processorConfig struct {
 	bodyParser                                   router.RequestBodyParser
 	router                                       extprocapi.Router
-	ModelNameHeaderKey, selectedBackendHeaderKey string
+	modelNameHeaderKey, selectedBackendHeaderKey string
```
mathetake, line 30: note: shouldn't have been exported

Makefile
nacx: [nit] Now that you're changing this, what about this?
```make
TEST_OPTS ?= -v -race
.PHONY: test
test:
	@echo "test => ./..."
	@go test $(TEST_OPTS) ./...
```
This would allow users to use custom options like `-count`, that are usually handy during development, while keeping the desired defaults.

internal/extproc/processor.go
```diff
@@ -212,6 +220,19 @@ func (p *Processor) maybeBuildDynamicMetadata() (*structpb.Struct, error) {
...
+			costU64, err := llmcostcel.EvaluateProgram(
+				c.celProg,
+				p.requestHeaders[p.config.modelNameHeaderKey],
+				p.requestHeaders[p.config.selectedBackendHeaderKey],
+				p.costs.InputTokens,
+				p.costs.OutputTokens,
+				p.costs.TotalTokens,
+			)
+			if err != nil {
+				return nil, fmt.Errorf("failed to evaluate CEL expression: %w", err)
```
nacx, line 233: Out of curiosity, what does it mean for ext_proc to fail at this stage, when the response headers may have been sent? Does it make sense to make it fail?

mathetake: this will result in connection termination i guess

Makefile
mathetake: i don't want to enable -race by default especially for the e2e tests which makes it unnecessary & slow that's why i added -race only here but open to suggestions

mathetake: but yeah will do that

mathetake:
@envoyproxy/ai-gateway-maintainers ping

mathetake:
@envoyproxy/ai-gateway-maintainers ping

internal/controller/sink.go
```diff
@@ -311,6 +312,13 @@ func (c *configSink) updateExtProcConfigMap(aiGatewayRoute *aigv1a1.AIGatewayRou
 			fc.Type = filterconfig.LLMRequestCostTypeTotalToken
 		case aigv1a1.LLMRequestCostTypeCEL:
 			fc.Type = filterconfig.LLMRequestCostTypeCELExpression
+			expr := *cost.CELExpression
+			// Sanity check the CEL expression.
+			_, err := llmcostcel.NewProgram(expr)
+			if err != nil {
+				return fmt.Errorf("invalid CEL expression: %w", err)
+			}
+			fc.CELExpression = expr
```
yuzisun, line 321: Could we validate this in a webhook instead of controller as user is not going to get the error feedback if they specify the invalid cel expression ?

mathetake: even if that's possible, still we need to do the type validation.

mathetake: even if that's possible, still we need to do the type validation.

mathetake: let me see how much k8s API can validate a CEL expression (not sure even if it's doable)

yuzisun: It is doable, need to setup a validating web hook to achieve this, we can do this post 0.1 though.

internal/llmcostcel/cel.go
```diff
@@ -0,0 +1,79 @@
...
+		err := issues.Err()
+		return nil, fmt.Errorf("cannot compile CEL expression: %w", err)
+	}
+	prog, err = env.Program(ast)
+	if err != nil {
+		return nil, fmt.Errorf("cannot create CEL program: %w", err)
+	}
+
+	// Sanity check by evaluating the expression with some dummy values.
+	_, err = EvaluateProgram(prog, "dummy", "dummy", 0, 0, 0)
```
yuzisun, line 44: This seems like getting called on the data path as well for each request, can we only initialize once?

mathetake: `prog` is initialized on the config load time which is the compiled CEL program. so i don't think we could do beyond that (the other args are depending on the request context)

internal/controller/sink.go
mathetake: hmm, i did a quick research but is it really possible to write a k8s validation to verify a string field content itself being a valid CEL expression?

internal/llmcostcel/cel.go
mathetake: btw to be clear, this function itself `NewProgram` is only called on config path and `prog` is initialized once here. And then on the request path only `EvaluateProgram` is called using the precompiled `prog`

internal/controller/sink.go
yuzisun: Essentially you add the same code in a webhook validation handler and then user can get the error feedback, thought we still need this code to deal with the standalone setup with configmap.

mathetake: ah ok i see not the CRD CEL validation, but instead k8s validation webhook. let's do that later after v0.1.0

internal/llmcostcel/cel.go
yuzisun: You are right I read the code wrong.

yuzisun (approved):

mathetake:
Thanks for the review Dan!

PR #158, opened 2025-01-21
controller: set allowModeOverride=true

mathetake:
This allows us to set allowModeOverride=true
for our external processor. The flag is needed
to deal with the stream=true requests.
This is introduced to EG today: envoyproxy/gateway#5099


go.sum
```diff
@@ -40,8 +40,8 @@ github.com/davecgh/go-spew v1.1.2-0.20180830191138-d8f796af33cc h1:U9qPSI2PIWSS1
 github.com/davecgh/go-spew v1.1.2-0.20180830191138-d8f796af33cc/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
 github.com/emicklei/go-restful/v3 v3.12.1 h1:PJMDIM/ak7btuL8Ex0iYET9hxM3CI2sjZtzpL63nKAU=
 github.com/emicklei/go-restful/v3 v3.12.1/go.mod h1:6n3XBCmQQb25CM2LCACGz8ukIrRry+4bhvbpWn3mrbc=
-github.com/envoyproxy/gateway v0.5.0-rc.1.0.20250115172926-8b89dadfbd2c h1:LsdzL2BAbgQOEAkelca7RzxPu5EP3p/Jacl0SUVBkGg=
-github.com/envoyproxy/gateway v0.5.0-rc.1.0.20250115172926-8b89dadfbd2c/go.mod h1:RV0iaXk9jyAaC8U5okx0Il3Wer5B0ttXrr/PXUS86Zo=
+github.com/envoyproxy/gateway v0.5.0-rc.1.0.20250121222313-d0d2aa718f7e h1:YQO3NgKOD/0Cq44IvJI0RHuO9nJma/ISVxUL54R+a9U=
+github.com/envoyproxy/gateway v0.5.0-rc.1.0.20250121222313-d0d2aa718f7e/go.mod h1:KO4JwUanODBOvOigUygRz78+pmJ4QcxZrtHSwK4qStk=
```
mathetake, line 44: @zirain @arkodg @zhaohuabing somehow `go get github.com/envoyproxy/gateway@latest` or whatever untagged version results in this weird v0.5.0-rc.1....... would you be able to fix it? dependabot tries to update it to the latest tagged version (v1.2.5 for now) as this "looks" like old but it is indeed the latest main branch

mathetake: ack i can add this to ignore dependabot.yaml but weird so i figured i wanted to ask

mathetake:
@envoyproxy/ai-gateway-maintainers ping

mathetake:
@envoyproxy/ai-gateway-maintainers ping

go.sum
zirain: because we didn't tag on master after that.

mathetake: i see... ok so whats' the best way to do deal with trivy and dependabot trying to "downgrade" to the latest tagged version? can you add any tagged created on the main branch back then into `retract` section of go.mod?

mathetake: @arkodg can you cut the tag on the main branch? this makes it really annoying to use the latest EG as a library.

mathetake: well that won't solve the issue.. i guess i will simply make it ignored

zirain: can you use 1.3.0-rc instead? it will release soon.

zirain (approved):

go.sum
mathetake: ok that's fantastic

mathetake: let me try

mathetake:
thanks @zirain

PR #160, opened 2025-01-22
controller: fix watch resources that are created by the controller

### LINKED ISSUE #134: Delete HTTPRoute does not trigger reconciliation
If I manually delete underlying HTTPRoute resource owned by the `AIGatewayRoute`, the reconciliation is not triggered to recreate it.

sivanantha321:
Fixes #134

This PR makes the AIGatewayRoute controller to watch the resources that are created by it and trigger reconcilation.

### DIFF
--- a/internal/controller/sink.go
+++ b/internal/controller/sink.go
@@ -154,15 +154,13 @@ func (c *configSink) syncAIGatewayRoute(aiGatewayRoute *aigv1a1.AIGatewayRoute) 
 		// This means that this AIGatewayRoute is a new one.
 		httpRoute = gwapiv1.HTTPRoute{
 			ObjectMeta: metav1.ObjectMeta{
-				Name:            aiGatewayRoute.Name,
-				Namespace:       aiGatewayRoute.Namespace,
-				OwnerReferences: ownerReferenceForAIGatewayRoute(aiGatewayRoute),
+				Name:      aiGatewayRoute.Name,
+				Namespace: aiGatewayRoute.Namespace,
 			},
 			Spec: gwapiv1.HTTPRouteSpec{},
 		}
-		err := ctrlutil.SetControllerReference(aiGatewayRoute, &httpRoute, c.client.Scheme())
-		if err != nil {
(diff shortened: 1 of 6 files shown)


mathetake:
sorry i can't accept basically any fix without proper tests

mathetake:
checking - are you up for writing test or we have to take this over

sivanantha321:
> checking - are you up for writing test or we have to take this over

Yeah I can write a test for this.

mathetake:
great, simple one would be enough

sivanantha321:
@mathetake Added a test. Let me know if anything else needs to be done.

internal/controller/sink.go
```diff
@@ -159,6 +160,10 @@ func (c *configSink) syncAIGatewayRoute(aiGatewayRoute *aigv1a1.AIGatewayRoute)
 			},
 			Spec: gwapiv1.HTTPRouteSpec{},
 		}
+		err := ctrlutil.SetControllerReference(aiGatewayRoute, &httpRoute, c.client.Scheme())
```
sivanantha321, line 163: Should I remove the owner reference as we have controller reference ?

mathetake: http route is not the only thing we create, so i think the change here will be a partial solution. so you could either delete this line or make changes to all resources not only the http route in this pr.

sivanantha321: This is required for the fix to work, so i will update all other resources to use controller reference.

sivanantha321:
@mathetake I noticed the controller creating HTTPRouteFilter without owner reference. I believe this because the resource is created per namespace. This resource is not watched by the controller. So If it gets deleted accidentally, It won't be recreated until the next reconcilation triggered by some other resource. The controller reference based approach used in this PR may not be appropriate for this resource.

mathetake:
> the controller creating HTTPRouteFilter without owner reference. I believe this because the resource is created per namespace. 

yeah the HTTPRouteFilter is another issue for sure.

mathetake (approved):
excellent! thank you for working on this

mathetake:
@sivanantha321 thank you for your contribution, nice work!

PR #163, opened 2025-01-22
controller: fixes volume mounts and add e2e with real backends

mathetake:
fyi @aabchoo


mathetake:
i will follow up the AWS example into this in another pr

PR #168, opened 2025-01-23
extproc: do not accumulate costs across requests

nacx:
**Commit Message**:

This commit changes the ext-proc filter so that the costs do not accumulate across requests.

The costs are populated in the filter metadata and used later as `hits_addend` in the Envoy rate limit filters. If we populate the accumulated values there, requests over time would have more "weight" than they should, and the rate limits will not be enforced as expected, as each time all the past tokens will be considered.

**Checklist**:
- [x] The PR title follows the same format as the merged PRs.
- [x] I have read the [CONTRIBUTING.md](../CONTRIBUTING.md) (for first-time contributors).
- [x] I have made sure that `make precommit test` passes locally.
- [x] I have written tests for any new line of code unless there's existing tests that cover the changes.
- [ ] I have updated the documentation accordingly.
- [x] I am sure the coding style is consistent with the rest of the codebase.

### DIFF
--- a/internal/extproc/processor.go
+++ b/internal/extproc/processor.go
@@ -58,8 +58,6 @@ type Processor struct {
 	responseHeaders  map[string]string
 	responseEncoding string
 	translator       translator.Translator
-	// cost is the cost of the request that is accumulated during the processing of the response.
-	costs translator.LLMTokenUsage
 }
 
 // ProcessRequestHeaders implements [Processor.ProcessRequestHeaders].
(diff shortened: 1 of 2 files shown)


mathetake:
No the code path here (processor) is per request already, not shared across requests as opposed to the server. Or are you sure that is not the case? @nacx

mathetake:
Also note that the dynamic metadata used here is different from the static per filter_metadata. The former is per request

nacx:
Got it. You're right, I just saw we instantiate a new processor on each call to `Process`. Let me close this. I had some custom metrics built on top of this and I may have a bug there :D

mathetake:
hope you can find the bug soon!

PR #125, opened 2025-01-17
feat: Implementation of API Key Auth

aabchoo:
This PR implements APIKey Auth for backendSecurityPolicy. The api key stored in a secret will be mounted to the extproc's pod, and later extracted by the appropriate AuthHandler.

Moved extprocDeployment into the sink as we need to tinker with the deployment if backend/backendSecurityPolicy is updated.

Removed the envoy config yaml's field add_request_header and replaced it with new API Key Auth implementation. CI passes which means that API Key mounted to extproc + specified on backend auth works.

Will work on adding AWS Credential file after this.

PR #169, opened 2025-01-23
extproc: load APIKey on load time

mathetake:
**Commit Message**:

Previously, the API was read on the request path,
which is left to be fixed and left as TODO in #125.
This fixes it and makes extproc load the APIKey
at the config load time.

Follow up on #125


mathetake:
cc @aabchoo

PR #170, opened 2025-01-23
e2e: creating e2e agent test for using tools

alexagriffith:
**Commit Message**:

This commit adds a real provider test for testing out the aws bedrock tool call conversion. The test uncovered a bug for tool conversion where the bedrock tool call input type should be JSON not string, so we need to unmarshall to unstructured map instead of string. 


tests/extproc/real_providers_test.go
```diff
@@ -205,4 +205,56 @@ func TestWithRealProviders(t *testing.T) {
...
 	})
+
+	t.Run("Bedrock calls tool get_weather function", func(t *testing.T) {
+		client := openai.NewClient(option.WithBaseURL(listenerAddress + "/v1/"))
+		for _, tc := range []struct {
+			testCaseName,
+			modelName string
+		}{
+			// {testCaseName: "openai", modelName: "gpt-4o-mini"},                            // This will go to "openai"
+			{testCaseName: "aws-bedrock", modelName: "us.meta.llama3-2-1b-instruct-v1:0"}, // This will go to "aws-bedrock" using credentials file.
```
yuzisun, line 216: @alexagriffith we need to switch to use the anthropic claude which supports tooling `bedrock.us.anthropic.claude-3-5-sonnet-20240620-v1:0`

mathetake: FYI 
image

mathetake: unfortunately this is not doable via terraform i am doing this manually (it saying taking a few minutes to be available)

mathetake: ok granted

yuzisun: Thanks @mathetake !

yuzisun:
@alexagriffith Plan to get this one merged in first as it contains a bug discovered by adding the real provider test, let's continue adding more tests in follow up PR using this tutorial.
https://platform.openai.com/docs/guides/function-calling

yuzisun (approved):

PR #173, opened 2025-01-23
test: fixes flake of examples/basic e2e

mathetake:
**Commit Message**:

This fixes or mitigates a flake of examples/basic
end to end test by re-creating the connection inside
the require-eventually-loop.


mathetake:
i will re-run 10 times on this

PR #174, opened 2025-01-24
deps: upgrades go dependencies, ignore EG CVE

mathetake:
**Commit Message**:

This pull request includes updates to dependencies and 
configuration files for better security and compatibility. 
Envoy Gateway dependency is kept to the latest, whose 
tag is a weird v0.5.0-rc2, and because of that Trivy and
Dependabot tries to "downgrade" to the latest tagged 
version. This fixes that by adding EG to the ignore list of
dependency as well as add the latest EG CVE to the 
trivyignore file.


Krishanx92:
@mathetake any reason to keep EG to v0.5.0-rc.1.0.20250115172926-8b89dadfbd2c in go.sum instead of update to newest release version 1.2.6?

mathetake:
As I described it's the latest hash and newer than 1.26 despite it's prefixed with 0.5 (so it's basically causing false positives both from trivy and dependabot)

PR #185, opened 2025-01-27
extproc: retrieve AWS credentials for every request

wengyao04:
**Commit Message**:

Instead of retrieving AWS credentials in `newAWSHandler`, could we get the AWS credentials for every request in first release and optimize it later with `AssumeRoleWithWebIdentity` API ?

I propose this change because we have temporary AWS credentials in the credential secrets, and we have our own token-refresher to refresh the credentials.

Retrieve credentials for each request 

### DIFF
--- a/internal/extproc/backendauth/aws.go
+++ b/internal/extproc/backendauth/aws.go
@@ -22,39 +22,27 @@ import (
 
 // awsHandler implements [Handler] for AWS Bedrock authz.
 type awsHandler struct {
-	credentials aws.Credentials
-	signer      *v4.Signer
-	region      string
+	credentials        aws.Credentials
+	credentialFileName string
+	signer             *v4.Signer
+	region             string
 }
 
 func newAWSHandler(awsAuth *filterconfig.AWSAuth) (*awsHandler, error) {
 	var credentials aws.Credentials
 	var region string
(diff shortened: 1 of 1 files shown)


wengyao04:
@yuzisun, @aabchoo and @mathetake , do you think if we can retrieve the AWS credentials for every request in first release and optimize it later with `AssumeRoleWithWebIdentity` API?

yuzisun:
> @yuzisun, @aabchoo and @mathetake , do you think if we can retrieve the AWS credentials for every request in first release and optimize it later with `AssumeRoleWithWebIdentity` API?

We can implement a secret watcher and update the credential in cache which is orthogonal to AssumeRoleWithWebIdentity, it is still a useful feature for user who rotates the aws credential manually.

mathetake:
yes, secret watcher is definitely the way to go and this reading-file-on-request-path is unacceptable (see #169). That's exactly what I suggested to @aabchoo in #125 : https://github.com/envoyproxy/ai-gateway/pull/125#discussion_r1920996139

mathetake:
Btw you can already emulate the refreshing behavior by using a unique secret resource name in your refresher. Anyways this is exactly the opposite of whwre we should go and I don't see any reason to have this now.

PR #202, opened 2025-01-28
feat: print diff on config change

leseb:
**Commit Message**:

feat: print diff on config change

When the configuration file is updated, the logger will print the
updated sections at the DEBUG level. The implementation is
straightforward and does not rely on any external libraries, making it
particularly useful for troubleshooting.

Output:

```
time=2025-01-28T11:46:03.911+01:00 level=INFO msg="loading a new config" path=docs/extproc-config.yaml
time=2025-01-28T11:46:03.911+01:00 level=DEBUG msg="config line changed" line=14 path=docs/extproc-config.yaml old="weight: 1" new="weight: 0"
```

Signed-off-by: Sébastien Han <seb@redhat.com>

### DIFF
--- a/internal/extproc/watcher.go
+++ b/internal/extproc/watcher.go
@@ -69,15 +69,15 @@ func (cw *configWatcher) loadConfig(ctx context.Context) error {
 	cw.lastMod = stat.ModTime()
 	cw.l.Info("loading a new config", slog.String("path", cw.path))
 
-	// Re-hydrate the current config file for later diffing.
-	previous := cw.current
-	current, err := cw.getConfig()
-	if err != nil {
-		return fmt.Errorf("failed to read the config file: %w", err)
-	}
-
 	// Print the diff between the old and new config.
 	if cw.l.Enabled(ctx, slog.LevelDebug) {
+		// Re-hydrate the current config file for later diffing.
+		previous := cw.current
+		current, err := cw.getConfigString()
+		if err != nil {
+			return fmt.Errorf("failed to read the config file: %w", err)
+		}
+
 		cw.diff(previous, current)
 	}
 
@@ -88,7 +88,9 @@ func (cw *configWatcher) loadConfig(ctx context.Context) error {
 	return cw.rcv.LoadConfig(cfg)
 }
 
-func (cw *configWatcher) getConfig() (string, error) {
+// getConfigString gets a string representation of the current config
+// read from the path. This is only used for debug log path for diff prints.
+func (cw *configWatcher) getConfigString() (string, error) {
 	currentByte, err := os.ReadFile(cw.path)
 	if err != nil {
 		return "", err
--- a/internal/extproc/watcher.go
+++ b/internal/extproc/watcher.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"log/slog"
 	"os"
+	"strings"
 	"time"
 
 	"github.com/envoyproxy/ai-gateway/filterconfig"
@@ -21,6 +22,7 @@ type configWatcher struct {
 	path    string
 	rcv     ConfigReceiver
 	l       *slog.Logger
+	current string
 }
 
 // StartConfigWatcher starts a watcher for the given path and Receiver.
@@ -28,7 +30,7 @@ type configWatcher struct {
 func StartConfigWatcher(ctx context.Context, path string, rcv ConfigReceiver, l *slog.Logger, tick time.Duration) error {
 	cw := &configWatcher{rcv: rcv, l: l, path: path}
 
-	if err := cw.loadConfig(); err != nil {
+	if err := cw.loadConfig(ctx); err != nil {
 		return fmt.Errorf("failed to load initial config: %w", err)
 	}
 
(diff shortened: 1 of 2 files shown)


internal/extproc/watcher.go
```diff
@@ -66,9 +68,54 @@ func (cw *configWatcher) loadConfig() error {
...
+	if err != nil {
+		return "", err
+	}
+	current := string(currentByte)
+	cw.current = current
+
+	return current, nil
+}
+
+func (cw *configWatcher) diff(oldConfig, newConfig string) {
```
mathetake, line 100: could you write a very simple unit test for this function so that we can what the expected output looks like?

mathetake (commented):
I like this - minor comment but LGTM

internal/extproc/watcher.go
```diff
@@ -66,9 +68,54 @@ func (cw *configWatcher) loadConfig() error {
...
+
+	// Re-hydrate the current config file for later diffing.
+	previous := cw.current
+	current, err := cw.getConfig()
+	if err != nil {
+		return fmt.Errorf("failed to read the config file: %w", err)
+	}
+
+	// Print the diff between the old and new config.
+	cw.diff(previous, current)
```
mathetake, line 80: can you avoid calling this function when it's not debugging mode by using  `cw.l.Enabled(ctx, slog.LevelDebug)`

internal/extproc/watcher.go
leseb: I’m happy to write a new unit test, but I’m unsure if it’s necessary since TestStartConfigWatcher already has a section that updates the config, triggering the diff. Additionally, TestStartConfigWatcher has been updated to reflect this behavior. My concern is that adding another test might result in an overlap with what TestStartConfigWatcher already covers. However, if you think it’s worth it, I’m open to writing the test. Please let me know your thoughts. Thanks!

internal/extproc/watcher.go
```diff
@@ -66,9 +68,56 @@ func (cw *configWatcher) loadConfig() error {
...
+
 	cfg, err := filterconfig.UnmarshalConfigYaml(cw.path)
 	if err != nil {
 		return err
 	}
 	return cw.rcv.LoadConfig(cfg)
 }
+
+func (cw *configWatcher) getConfig() (string, error) {
+	currentByte, err := os.ReadFile(cw.path)
```
mathetake, line 92: this overlaps with `filterconfig.UnmarshalConfigYaml(cw.path)` above. can you do the deduplication?

internal/extproc/watcher.go
```diff
@@ -66,9 +68,56 @@ func (cw *configWatcher) loadConfig() error {
 	}
 	cw.lastMod = stat.ModTime()
 	cw.l.Info("loading a new config", slog.String("path", cw.path))
+
+	// Re-hydrate the current config file for later diffing.
+	previous := cw.current
+	current, err := cw.getConfig()
+	if err != nil {
+		return fmt.Errorf("failed to read the config file: %w", err)
+	}
```
mathetake, line 77: this should also be within the if-block below

internal/extproc/watcher.go
mathetake: yeah please go ahead and write a test

internal/extproc/watcher.go
```diff
@@ -66,9 +68,56 @@ func (cw *configWatcher) loadConfig() error {
...
+
 	cfg, err := filterconfig.UnmarshalConfigYaml(cw.path)
 	if err != nil {
 		return err
 	}
 	return cw.rcv.LoadConfig(cfg)
 }
+
+func (cw *configWatcher) getConfig() (string, error) {
```
mathetake, line 91:
```suggestion
// getConfigString gets a string representation of the current config
// read from the path. This is only used for debug log path for diff prints.
func (cw *configWatcher) getConfigString() (string, error) {
```

mathetake:
@leseb could you not force-push as per CONTRIBUTING.md?

internal/extproc/watcher_test.go
```diff
@@ -33,6 +35,15 @@ func (m *mockReceiver) getConfig() *filterconfig.Config {
 	return m.cfg
 }
 
+// NewTestLoggerWithBuffer creates a new logger with a buffer for testing and asserting the output.
+func NewTestLoggerWithBuffer() (*slog.Logger, *bytes.Buffer) {
```
mathetake, line 39: do not need to be exported
```suggestion
// newTestLoggerWithBuffer creates a new logger with a buffer for testing and asserting the output.
func newTestLoggerWithBuffer() (*slog.Logger, *bytes.Buffer) {
```

internal/extproc/watcher.go
mathetake: especially to ensure the logic in `log.Int("line", i+1), slog.String("path", cw.path), slog.String("old", oldLine), slog.String("new", newLine)` which i believe is not checked in the current test.

mathetake:
also, could you modify the PR description so that it follows the template

leseb:
I had to do another force-push on the second commit since DCO was complaining...

mathetake (approved):
Thank you!!

leseb:
> Thank you!!

Thanks! So when CI is green I'll amend it to have a single final commit and force-push, is that alright?

mathetake:
well, all PRs will be squash-merged, so you don't need to do that

PR #204, opened 2025-01-28
api: allows extproc log level

mathetake:
**Commit Message**:

This adds `logLevel` field nob to `AIGatewayFilterConfigExternalProcess`
to allows fine-tune the log level of external processing.

### DIFF
--- a/api/v1alpha1/api.go
+++ b/api/v1alpha1/api.go
@@ -255,6 +255,12 @@ const (
 )
 
 type AIGatewayFilterConfigExternalProcess struct {
+	// LogLevel is the log level of the external process filter.
+	//
+	// +kubebuilder:default=info
+	// +kubebuilder:validation:Enum=debug;info;warn;error
+	LogLevel string `json:"logLevel,omitempty"`
+
 	// Replicas is the number of desired pods of the external process deployment.
 	//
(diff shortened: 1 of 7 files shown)


api/v1alpha1/api.go
```diff
@@ -255,6 +255,12 @@ const (
 )
 
 type AIGatewayFilterConfigExternalProcess struct {
+	// LogLevel is the log level of the external process filter.
+	//
+	// +kubebuilder:default=info
+	// +kubebuilder:validation:Enum=debug;info;warn;error
+	LogLevel string `json:"logLevel,omitempty"`
```
mathetake, line 262: @envoyproxy/ai-gateway-maintainers I think we could instead have this as a helm config but i find this it useful to have it here to be able to debug faster. wdyt

mathetake: ping

aabchoo: Is it possible to define a global logLevel in helm, and allow these to override the default? Would be helpful if we only want to debug a particular ai-gateway-route

mathetake: yeah possible and make sense - let me add the global helm config first instead of API like this

PR #206, opened 2025-01-28
ci: set ups dependabot for /site 

mathetake:
**Commit Message**:

This configures the dependabot for the site directory.


mathetake:
FYI! @missBerg

PR #217, opened 2025-01-29
feat: redact sensitive headers and body content in debug logs

leseb:
**Commit Message**:

This commit introduces functions to redact sensitive headers and body content before logging, ensuring that sensitive information such as authorization tokens is not exposed in debug logs. The new filterSensitiveHeaders function redacts specified headers, replacing their values with [REDACTED], while filterSensitiveBody ensures that sensitive content in request bodies is also redacted before being logged.

Additionally, the logging behavior has been updated so that request headers and body content are only logged when the DEBUG level is enabled, further improving security and performance.

Unit tests have been added to verify the functionality of both filterSensitiveHeaders and filterSensitiveBody, ensuring that sensitive data is properly redacted and that the logging behavior works as expected.

Signed-off-by: Sébastien Han <seb@redhat.com>

### DIFF
--- a/internal/extproc/server.go
+++ b/internal/extproc/server.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"log/slog"
 	"slices"
+	"strings"
 
 	corev3 "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
 	extprocv3 "github.com/envoyproxy/go-control-plane/envoy/service/ext_proc/v3"
@@ -26,6 +27,8 @@ import (
 const (
 	redactedKey = "[REDACTED]"
 )
+
+var sensitiveHeaderKeys = []string{"authorization"}
 
 // Server implements the external process server.
 type Server[P ProcessorIface] struct {
@@ -141,7 +144,7 @@ func (s *Server[P]) processMsg(ctx context.Context, p P, req *extprocv3.Processi
 		requestHdrs := req.GetRequestHeaders().Headers
 		// If DEBUG log level is enabled, filter sensitive headers before logging.
 		if s.logger.Enabled(ctx, slog.LevelDebug) {
-			filteredHdrs := filterSensitiveHeaders(requestHdrs, s.logger, []string{"authorization"})
+			filteredHdrs := filterSensitiveHeaders(requestHdrs, s.logger, sensitiveHeaderKeys)
 			s.logger.Debug("request headers processing", slog.Any("request_headers", filteredHdrs))
 		}
 		resp, err := p.ProcessRequestHeaders(ctx, requestHdrs)
@@ -155,7 +158,7 @@ func (s *Server[P]) processMsg(ctx context.Context, p P, req *extprocv3.Processi
 		resp, err := p.ProcessRequestBody(ctx, value.RequestBody)
 		// If DEBUG log level is enabled, filter sensitive body before logging.
 		if s.logger.Enabled(ctx, slog.LevelDebug) {
-			filteredBody := filterSensitiveBody(resp, s.logger, []string{"Authorization"})
+			filteredBody := filterSensitiveBody(resp, s.logger, sensitiveHeaderKeys)
 			s.logger.Debug("request body processed", slog.Any("response", filteredBody))
 		}
 		if err != nil {
@@ -205,7 +208,8 @@ func filterSensitiveHeaders(headers *corev3.HeaderMap, logger *slog.Logger, sens
 	}
 	filteredHeaders := &corev3.HeaderMap{}
 	for _, header := range headers.Headers {
-		if slices.Contains(sensitiveKeys, header.GetKey()) {
+		// We convert the header key to lowercase to make the comparison case-insensitive but we don't modify the original header.
+		if slices.Contains(sensitiveKeys, strings.ToLower(header.GetKey())) {
 			logger.Debug("filtering sensitive header", slog.String("header_key", header.Key))
 			filteredHeaders.Headers = append(filteredHeaders.Headers, &corev3.HeaderValue{
 				Key:   header.Key,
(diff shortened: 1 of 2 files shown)


mathetake:
hmm, i am not sure if this is worth it - at the end of the day, Envoy will log all headers when the debugging log enabled just like this extproc. So the caveat is that anyone can see the gateway containers logs should be able to have enough privileges to see the secrets. If that assumption is not true, this will only be a partial solution.

mathetake:
see https://github.com/envoyproxy/envoy/issues/27820 for example

arkodg:
also relates to https://github.com/envoyproxy/envoy/issues/37656

leseb:
> hmm, i am not sure if this is worth it - at the end of the day, Envoy will log all headers when the debugging log enabled just like this extproc. So the caveat is that anyone can see the gateway containers logs should be able to have enough privileges to see the secrets. If that assumption is not true, this will only be a partial solution.

While it is true that enabling debug logging in Envoy (or any similar system) may expose all headers, including sensitive ones, this does not mean that redacting sensitive information is unnecessary. Security should always be approached in layers, and even partial solutions contribute to reducing risk. Making a first step in extproc reduces the risk of accidental exposure, such as when logs are shared publicly or sent to support teams. Even if high-privilege users can access logs, it’s still a best practice to minimize the surface area of sensitive data. Redacting secrets helps maintain security, aligns with compliance standards, and is a simple layer of defense against potential mistakes or leaks.

internal/extproc/server_test.go
```diff
@@ -259,3 +260,55 @@ func TestServer_Process(t *testing.T) {
...
+	for _, h := range filtered.Headers {
+		if h.Key == "authorization" {
+			require.Equal(t, "[REDACTED]", h.Value)
+		} else {
+			require.Equal(t, "bar", h.Value)
+		}
+	}
+	require.Eventually(t, func() bool {
+		return strings.Contains(buf.String(), "filtering sensitive header")
+	}, 1*time.Second, 100*time.Millisecond, buf.String())
```
mathetake, line 278: why do you need eventually?

internal/extproc/server_test.go
```diff
@@ -259,3 +260,55 @@ func TestServer_Process(t *testing.T) {
...
+	filtered := filterSensitiveBody(resp, logger, []string{"Authorization"})
+	require.NotNil(t, filtered)
+	for _, h := range filtered.Response.(*extprocv3.ProcessingResponse_RequestBody).RequestBody.Response.GetHeaderMutation().GetSetHeaders() {
+		if h.Header.Key == "Authorization" {
+			require.Equal(t, "[REDACTED]", string(h.Header.RawValue))
+		}
+	}
+	require.Eventually(t, func() bool {
+		return strings.Contains(buf.String(), "filtering sensitive header")
+	}, 1*time.Second, 100*time.Millisecond, buf.String(), filtered)
```
mathetake, line 313: same question

internal/extproc/server.go
```diff
@@ -131,7 +137,11 @@ func (s *Server[P]) processMsg(ctx context.Context, p P, req *extprocv3.Processi
 	switch value := req.Request.(type) {
 	case *extprocv3.ProcessingRequest_RequestHeaders:
 		requestHdrs := req.GetRequestHeaders().Headers
-		s.logger.Debug("request headers processing", slog.Any("request_headers", requestHdrs))
+		// If DEBUG log level is enabled, filter sensitive headers before logging.
+		if s.logger.Enabled(ctx, slog.LevelDebug) {
+			filteredHdrs := filterSensitiveHeaders(requestHdrs, s.logger, []string{"authorization"})
```
mathetake, line 142: why here is the "authorization" (lowercase) and "authorization"?  can you extract this []string as a global slice? and share it here and in requestbody handling?

internal/extproc/server.go
```diff
@@ -178,3 +192,56 @@ func (s *Server[P]) Check(context.Context, *grpc_health_v1.HealthCheckRequest) (
...
+// Specifically, it redacts the value of the "authorization" header and logs this action.
+// The function returns a new HeaderMap with the filtered headers.
+func filterSensitiveHeaders(headers *corev3.HeaderMap, logger *slog.Logger, sensitiveKeys []string) *corev3.HeaderMap {
+	if headers == nil {
+		logger.Debug("received nil HeaderMap, returning empty HeaderMap")
+		return &corev3.HeaderMap{}
+	}
+	filteredHeaders := &corev3.HeaderMap{}
+	for _, header := range headers.Headers {
+		if slices.Contains(sensitiveKeys, header.GetKey()) {
```
mathetake, line 206: if we want to be extra cautious (which I believe is what we want), can we use the strings.ToLower(maybe the name is different ;) to ensure the case-insensitivity

mathetake (commented):
yeah, thank you for sharing your opinion @leseb. That makes sense as in anyways this is only enabled when -logLevel=debug. I left some comments

internal/extproc/server_test.go
leseb: I don't, logging is synchronous and the log message should be written immediately to the buffer. Thanks!

internal/extproc/server_test.go
leseb: ditto

internal/extproc/server.go
leseb: I believe "Authorization" comes from: https://github.com/envoyproxy/ai-gateway/blob/4b198313d763dabbc9b8ae0cf2b5e318f6597815/internal/extproc/backendauth/api_key.go#L33

I believe "authorization" comes from Envoy here is the log:

```
time=2025-01-29T15:26:59.446+01:00 level=DEBUG msg="request headers processing" request_headers="headers:{key:\":authority\"  raw_value:\"localhost:1062\"}  headers:{key:\":path\"  raw_value:\"/v1/chat/completions\"}  headers:{key:\":method\"  raw_value:\"POST\"}  headers:{key:\":scheme\"  raw_value:\"http\"}  headers:{key:\"accept-encoding\"  raw_value:\"gzip, deflate\"}  headers:{key:\"accept\"  raw_value:\"application/json\"}  headers:{key:\"content-type\"  raw_value:\"application/json\"}  headers:{key:\"user-agent\"  raw_value:\"OpenAI/Python 1.60.2\"}  headers:{key:\"x-stainless-lang\"  raw_value:\"python\"}  headers:{key:\"x-stainless-package-version\"  raw_value:\"1.60.2\"}  headers:{key:\"x-stainless-os\"  raw_value:\"MacOS\"}  headers:{key:\"x-stainless-arch\"  raw_value:\"arm64\"}  headers:{key:\"x-stainless-runtime\"  raw_value:\"CPython\"}  headers:{key:\"x-stainless-runtime-version\"  raw_value:\"3.8.20\"}  headers:{key:\"authorization\"  value:\"[REDACTED]\"}  headers:{key:\"x-stainless-async\"  raw_value:\"false\"}  headers:{key:\"x-stainless-retry-count\"  raw_value:\"0\"}  headers:{key:\"content-length\"  raw_value:\"84\"}  headers:{key:\"x-forwarded-proto\"  raw_value:\"http\"}  headers:{key:\"x-request-id\"  raw_value:\"f9c40848-d9ca-4cac-a866-467b3777e5bd\"}"
```

I'll use a global slice with both. Another question is whether we should change the `api_key.go` to use `authorization` to stay consistent with Envoy's format? (in a future PR). Thanks!

mathetake (approved):
:100:

PR #222, opened 2025-01-29
ci: deletes doctests

mathetake:
**Commit Message**:

This deletes the doc test as it turns out that simply
using a markdown parser seems difficult to makintain.
As discussed offline, we decided to remove it for now.
If any necessity comes up in the future, we revisit this
and reevaluate what is the best way to ensure the
correctness of the documentations with code blocks.


missBerg:
We can think about ways to test docs instructions in the future :)

PR #219, opened 2025-01-29
controller: adds secret watcher to trigger extproc config updates

mathetake:
**Commit Message**:

This adds a secret watcher controller that enables the 
hot reload of any secret referenced by backendTrafficPolicy.

**Related Issues/PRs (if applicable)**:

Follow up on #43  #106 #161 
Supersede #185

PR #224, opened 2025-01-29
controller: annotate extproc pods with uuid for faster refresh

mathetake:
**Commit Message**:

Without triggering pods into the reconcile loop of k8s server,
the config map updates will take a few minutes to be picked up
and reflected on the actual file of the pod [^1].

This commit changes the config sink so that it will add the config
uuid to the extproc pods annotations.

[^1]: https://neonmirrors.net/post/2022-12/reducing-pod-volume-update-times/

Follow up on #219

### DIFF
--- a/internal/controller/sink.go
+++ b/internal/controller/sink.go
@@ -11,6 +11,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 	uuid2 "k8s.io/apimachinery/pkg/util/uuid"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/utils/ptr"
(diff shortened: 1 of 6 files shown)


internal/controller/sink.go
```diff
@@ -422,6 +430,28 @@ func (c *configSink) newHTTPRoute(dst *gwapiv1.HTTPRoute, aiGatewayRoute *aigv1a
...
+		c.logger.Info("annotating pod", "namespace", pod.Namespace, "name", pod.Name)
+		if pod.Annotations == nil {
+			pod.Annotations = make(map[string]string)
+		}
+
+		pod.Annotations["aigateway.envoyproxy.io/extproc-config-uuid"] = uuid
+		if _, err := c.kube.CoreV1().Pods(aiGatewayRoute.Namespace).Update(ctx, &pod, metav1.UpdateOptions{}); err != nil {
+			return fmt.Errorf("failed to update pod %s: %w", pod.Name, err)
+		}
+	}
```
mathetake, line 451: note to self - use patch instead

PR #232, opened 2025-01-30
feat: implementation of AWS OIDC

aabchoo:
**Commit Message**

This implements the API for AWS OIDC within the extproc.

The main changes breaks down into:
1. OIDC Handler which can be used for any providers
2. AWS handles calling STS for temporary credentials
3. Passes Proxy + configmap CERTs

Related PR: #43

Need to add tests


internal/extproc/backendauth/aws.go
```diff
@@ -90,6 +104,17 @@ func (a *awsHandler) Do(requestHeaders map[string]string, headerMut *extprocv3.H
 		return fmt.Errorf("cannot create request: %w", err)
 	}
 
+	if a.oidcHandler != nil {
+		err := a.oidcHandler.updateOIDCTokenIfExpired(context.TODO())
```
mathetake, line 108: so you have to have some sync mechanism as a handler is hared across multiple goroutines/requests.

mathetake: one idea is to do this upgrades on the background groutine and on the request path do not make any blocking calls like this if block right now

mathetake:
fyi i renamed `filterconfig` to `filterapi` so the conflict is because of that 🙇

internal/extproc/backendauth/aws.go
aabchoo: hmm fair, I wanted to put it within the request path to avoid unnecessary queries to the SSO/STS. But the race condition is also unideal, so I'll add a goroutine to handle updating creds

mathetake (commented):
High level question: it seems better to me that we should do the credentials update **inside** the controller and store the exchanged credentials in a controller-managed secret. Then I _guess_ no change would be needed to extproc? Not only that, then the controller pod (one pod) will be the only one making requests to sts vs with the current impl, it will be tens/hundreds  of pods making requests to sts?

missBerg:
I agree with @mathetake; the code to rotate and refresh should happen in the controller, and the ext proc should use the credentials.

mathetake:
Another great advantages is that you won't need these certificates API then - you have full control over the deployment of controller.

aabchoo:
I believe the reason we didn't put the OIDC code within the controller is because we need our certs mounted to the pod calling our SSO and we thought it would be better not share the cert with the entire controller.

cc @yuzisun could you confirm

mathetake:
> we need our certs mounted to the pod calling our SSO and we thought it would be better not share the cert with the entire controller.


ok so what you are saying is 

* Mounting the certificate with the "less privileged" exptorc (many containers, tens/hundreds, almost zero privilege) is OK
* Mounting the certificate with the "highly privileged" controller (only exists one container at a time usually) is Not OK 


Controller is already watching all secrets (by default though), so I don't think that makes sense to have this complex API/configurations. It can already have the enough privilege to read/write/delete the certificate.

aabchoo:
Discussed with @yuzisun offline, and it doesn't seem to be an issue. I will rework the OIDC changes to be apart of the controller. Will aim to get it in by tonight 🙇

mathetake:
great! thank you for the work anyways!

missBerg (changes requested):
Will take a bit more of a look, but there are definitely things to address :)

api/v1alpha1/api.go
```diff
@@ -471,6 +471,9 @@ type AWSOIDCExchangeToken struct {
 	// AwsRoleArn is the AWS IAM Role with the permission to use specific resources in AWS account
 	// which maps to the temporary AWS security credentials exchanged using the authentication token issued by OIDC provider.
 	AwsRoleArn string `json:"awsRoleArn"`
+
+	// ProxyURL can be used when communication with STS.
+	ProxyURL string `json:"proxyUrl,omitempty"`
```
mathetake, line 476: i don't think you will need ?
https://github.com/golang/go/blob/ce7ea0a6a529ce91327900e29afc3abd620030b4/src/net/http/transport.go#L42-L44

internal/controller/oidc/oidc.go
```diff
@@ -0,0 +1,176 @@
...
+func (j IdentityTokenValue) GetIdentityToken() ([]byte, error) {
+	return []byte(j), nil
+}
+
+type oauth2TokenWithExp struct {
+	token   *oauth2.Token
+	expTime time.Time
+}
+
+type Handler struct {
```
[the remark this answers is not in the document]
aabchoo: `TokenRefresher` makes sense if the "handler" only managed refreshing OIDC tokens. What do you think about `CredentialRefresher` since it updates the OIDC token and updates/creates a secret file with new logins

internal/controller/oidc/oidc.go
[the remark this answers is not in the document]
aabchoo: Which specific functions do you suspect? I might be missing something

internal/controller/oidc/oidc.go
```diff
@@ -0,0 +1,176 @@
...
+	}
+	oauth2Config.EndpointParams = url.Values{"audience": []string{aud}}
+	t, err := oauth2Config.Token(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("fail to refresh oauth2 token %w", err)
+	}
+	return t, nil
+}
+
+func (o *Handler) oauth2TokenExpireTime(accessToken *oauth2.Token) (*time.Time, error) {
```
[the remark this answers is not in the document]
aabchoo: So for the two castings:
1. We set the "implementation" of `Claim` to be type `MapClaims` when parsing the access token via `ParseUnverified`. We later cast it because we want to use its properties of being a map (we essentially are checking the type this way).
2. The value of claims is an interface, so we want to type cast to an integer for time.Unix to understand it. This is essentially checking the types + casting all at once.

internal/controller/oidc/oidc.go
```diff
@@ -0,0 +1,176 @@
...
+
+import (
+	"context"
+	"fmt"
+	"log/slog"
+	"net/url"
+	"time"
+
+	"github.com/coreos/go-oidc/v3/oidc"
+	egv1a1 "github.com/envoyproxy/gateway/api/v1alpha1"
```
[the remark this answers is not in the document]
mathetake: yes and this is the standard pattern and naming consistency is ensured by the linter in CI (btw, this is not only here but also across the whole bunch of k8s related projects. (you see the API package name `v1alpha1` exists across tons of  projects so not renaming is terrible.

internal/controller/oidc/oidc.go
```diff
@@ -0,0 +1,176 @@
...
+		logger:              logger,
+		k8sClient:           k8sClient,
+		awsCredentialCache:  make(map[string]time.Time),
+		oidcCredentialCache: make(map[string]*oauth2TokenWithExp),
+		interval:            time.Minute,
+	}
+	return handler, nil
+}
+
+func (o *Handler) UpdateCredentials(ctx context.Context) {
```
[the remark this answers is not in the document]
aabchoo: I'm assuming you're referring to this [process](https://gobyexample.com/spawning-processes)? 

Correct me if I'm misunderstanding you, but would this mean we create an executable for the OIDC credential update and have the process handle updating?

internal/controller/oidc/oidc.go
```diff
@@ -0,0 +1,176 @@
...
+)
+
+// OIDC expects the SecretKey to be "client-secret".
+const secretKey = "client-secret"
+
+// IdentityTokenValue is for retrieving an identity token
+type IdentityTokenValue string
+
+// GetIdentityToken retrieves the JWT and returns the contents as a []byte
+func (j IdentityTokenValue) GetIdentityToken() ([]byte, error) {
```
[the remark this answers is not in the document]
aabchoo: `GetIdentityToken` is the official function for IdentityTokenRetriever's interface

```
// IdentityTokenRetriever is an interface for retrieving a JWT
type IdentityTokenRetriever interface {
	GetIdentityToken() ([]byte, error)
}
```

But I do agree it's not very clear

api/v1alpha1/api.go
aabchoo: Wouldn't this mean that all HTTP requests will have the proxy which may unideal?

mathetake: no, it will return nil if not configured. as i suggested you can do the if switch based on whether or not it is nil

mathetake: you can always check the source code of the standard library and right a simple test btw. 

```go
	t.Setenv("HTTP_PROXY", "http://localhost:8080") // See the result with or without this line.
	req, _ := http.NewRequest("GET", "http://example.com", nil)
	url, err2 := http.ProxyFromEnvironment(req)
	require.NoError(t, err2)
	fmt.Println(url)
```

mathetake: > like you can use ProxyFromEnvironment and conditionally use that

i meant conditionally use that for that.

mathetake: or, maybe aws sdk already maybe already be the env var aware / or you can use `httpproxy.FromEnvironment()` directly. in any case I am strongly against this proxy api field at this api

mathetake: * Option 1.: Use the default "PROXY_URL" env var + you might need to tweak the transport config of k8s clients 
* Option 2.: Define a special "AI_GATEWY_STS_PROXY_URL" and check the existence and define the transport accordingly

internal/controller/oidc/oidc.go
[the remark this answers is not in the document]
yuzisun: It does both token exchange and refresh, we discussed to have a generic interface to support both aws and gcp which we are going to add very soon. The steps are the same to get the oidc token to exchange the cloud credential, then update the credential file. The token refresher go routine updates the credential before it expires.

internal/controller/oidc/oidc.go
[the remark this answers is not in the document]
yuzisun: Simple go routine based timer should work, we do not need a process.

internal/controller/oidc/oidc.go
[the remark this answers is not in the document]
missBerg: Why not rename to egv1alpha1?

internal/controller/oidc/aws.go
```diff
@@ -0,0 +1,155 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	aigv1a1 "github.com/envoyproxy/ai-gateway/api/v1alpha1"
+)
+
+const OidcAwsPrefix = "oidc-aws-"
+
+type AWSSpec struct {
```
yuzisun, line 26: Suggest AWSOIDCTokenExchange

yuzisun, line 26: This name is a bit unclear of the purpose, maybe we should name the other struct AWSCredentialHandler and this one AWSCredentialExchange

internal/controller/controller.go
```diff
@@ -115,6 +116,15 @@ func StartControllers(ctx context.Context, config *rest.Config, logger logr.Logg
...
 	}
+
+	// Have a token exchange handler per provider type.
+	handler, err := oidc.NewOIDCTokenExchange(&logger, c, aigv1a1.BackendSecurityPolicyTypeAWSCredentials)
+	if err != nil {
+		return fmt.Errorf("failed to create OIDC handler: %w", err)
+	}
+
+	go handler.RefreshCredentials(ctx)
+
```
mathetake, line 127: these code are not called since mgr.Start is blocking.

mathetake (commented):
I didn't look into the detail but can't this be a part of existing backendSecurityPolicyController? Even though I don't have a concrete answer, I think there must be a clear way to do that. the current struct seems totally different than the existing style and this doesn't feel right (but again i am just skimming the code)

internal/controller/controller.go
wengyao04: I think you use let the manager Add your runable
https://github.com/kubernetes-sigs/controller-runtime/blob/v0.20.1/pkg/manager/manager.go#L52-L60

wengyao04: I think you can use Manager to Add a Runable 
https://github.com/kubernetes-sigs/controller-runtime/blob/v0.20.1/pkg/manager/manager.go#L56-L60

and you don't need 
```
go handler.RefreshCredentials(ctx)
```

Manager can filter by
```
//reconciliation is only triggered by a GenericEvent generated by a time ticker
		WithEventFilter(predicate.Funcs{
			GenericFunc: func(event.GenericEvent) bool { return true },
````
We can write a tick event source to generate tick time event. I can find our internal example for you

PR #248, opened 2025-01-30
docs: fixes CODE_OF_CONDUCT.md

mathetake:
**Commit Message**:

This was copied from Envoy Gateway as-is
and should be "Envoy AI Gateawy" instead of
Envoy Gateway.


missBerg (approved):
Thanks! Lgtm

PR #252, opened 2025-01-31
controller: mark all volume mounts ReadOnly=true

### LINKED ISSUE #250: controller: mark all volume mounts readOnly=true
*Description*:

The AI Gateway ExtProc deployment/containers mount some secrets/configmaps. However, it only needs to read the contents, and so the write permission is unnecessary.

[optional *Relevant Links*:]

I would like the `corev1.VolumeMount`'s ReadOnly flag to always true in https://github.com/envoyproxy/ai-gateway/blob/main/internal/controller/sink.go

daixiang0:
**Commit Message**

Mark all volume mounts readonly to ensure security.

Fix #250


internal/controller/sink.go
```diff
@@ -489,7 +489,11 @@ func (c *configSink) syncExtProcDeployment(ctx context.Context, aiGatewayRoute *
 										"-logLevel", c.extProcLogLevel,
 									},
 									VolumeMounts: []corev1.VolumeMount{
```
mathetake, line 491: I believe this is not the only volume mount

mathetake: Sorry never mind

mathetake:
Yep, I need a test!

mathetake (approved):

internal/controller/sink_test.go
```diff
@@ -821,6 +818,10 @@ func TestConfigSink_MountBackendSecurityPolicySecrets(t *testing.T) {
 	require.Equal(t, "rule0-backref0-some-other-backend-security-policy-2", updatedSpec.Volumes[1].Name)
 	require.Equal(t, "rule0-backref0-some-other-backend-security-policy-2", updatedSpec.Containers[0].VolumeMounts[1].Name)
 	require.Equal(t, "/etc/backend_security_policy/rule0-backref0-some-other-backend-security-policy-2", updatedSpec.Containers[0].VolumeMounts[1].MountPath)
+
+	for _, v := range updatedSpec.Containers[0].VolumeMounts {
```
leseb, line 822: Are we sure that all the mounts will always be ro? Should we precisely look up for the name?

mathetake: yeah all mounts will be read only for sure - looking up by name works for me if you want

daixiang0:
@leseb @mathetake thanks for review.

internal/controller/sink_test.go
leseb: you seem confident that they'll always be read-only so I'm fine with that!

PR #256, opened 2025-01-31
feat: Add routing processor service design proposal

rootfs:


docs/proposals/002-routing-processor-proposal/routing-processor.proto
```diff
@@ -0,0 +1,188 @@
+syntax = "proto3";
+
+package routing_processor.v1alpha1;
+
+import "google/protobuf/duration.proto";
+import "google/protobuf/struct.proto";
+import "envoy/service/ext_proc/v3/external_processor.proto";
```
mathetake, line 7: well, please do not expose/use the Envoy Extproc implementation detail at this layer - if you use that, then what's the point of abstraction we are introducing in this project vs you can setup your own extproc by yourself.

docs/proposals/002-routing-processor-proposal/routing-processor.proto
```diff
@@ -0,0 +1,188 @@
+syntax = "proto3";
+
+package routing_processor.v1alpha1;
+
+import "google/protobuf/duration.proto";
+import "google/protobuf/struct.proto";
```
mathetake, line 6: let's not use the third party protobuf dependency and start with a really simple self contained definition.

mathetake:
cc @envoyproxy/ai-gateway-maintainers @envoyproxy/ai-gateway-assignable

docs/proposals/002-routing-processor-proposal/routing-processor.proto
rootfs: dependency removed now

rootfs:
@mathetake while i am working on the PoC, i found it is not a scalable solution to forward the request and response back and forth between extproc and python service. It makes more sense to use another extproc based on [cachefilter](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/cache_filter). WDYT?

So for this proposal, I am going to focus on the model and backend selection use case.

rootfs:
The semantic processor protocol is updated. I also have an semantic caching PoC using this protocol. Please find it [here](https://asciinema.org/a/704430). 

The extproc integration PoC is implemented [here](https://github.com/envoyproxy/ai-gateway/commit/a622ffcd366915b774115f1d4ab72b83a70cb653)

docs/proposals/002-routing-processor-proposal/routing-processor.proto
```diff
@@ -0,0 +1,232 @@
...
+// ===== Services =====
+
+service RoutingProcessor {
+  rpc ExternalProcess(stream ProcessingRequest) returns (stream ProcessingResponse);
+  rpc GetCapabilities(CapabilitiesRequest) returns (CapabilitiesResponse);
+}
+
+service SemanticCacheService {
+  rpc SearchCache(SearchRequest) returns (SearchResponse) {}
+  rpc StoreChat(StoreChatRequest) returns (StoreChatResponse) {}
```
rootfs, line 231: In developing the PoC, I also found we can provide yet another service to first store the request message, gets a unique ID for the entry, and then later use that unique ID along with the response message to persist the complete request/response pair, reducing the traffic and the overhead of tracking request/response in extproc.

mathetake:
ok now i am convinced that we could do this general direction. Could you delete all Python and Go code and focus on the protobuf and documentation?

mathetake:
or you could just close this and raise a new PR focusing on the protobuf API sitting in https://github.com/envoyproxy/ai-gateway/tree/main/filterapi/x

rootfs:
@mathetake sure, let me close this one and update the x.go with the proposal.

PR #266, opened 2025-02-01
chore: fixes PR style check

mathetake:
**Commit Message**

This makes the PR style check stricter than before:
* The description's first line must be `**Commit Message**`
* The PR title length must be less than 53 characters

### DIFF
--- a/.github/workflows/pr_title_check.yaml
+++ b/.github/workflows/pr_title_check.yaml
@@ -28,9 +28,9 @@ jobs:
               echo "PR description contains '-->'. Please remove all the comment out lines in the template after carefully reading them."
               exit 1
           fi
-          if [[ ! $TITLE =~ "**Commit Message**:" ]]; then
-              echo "PR description must begin with '**Commit Message**:'."
-              exit 1
+          if [[ ! $TITLE =~ ^\*\*Commit Message\*\* ]]; then
+            echo "The first line of the PR description should be '**Commit Message**'"
+            exit 1
           fi
 
   title:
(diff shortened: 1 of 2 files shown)


.github/workflows/pr_title_check.yaml
```diff
@@ -63,3 +66,16 @@ jobs:
...
             didn't match the configured pattern. Please ensure that the subject
             doesn't start with an uppercase character.
+
+      - name: Check length of PR title
+        env:
+          # Do not use ${{ github.event.pull_request.title }} directly in run command.
+          TITLE: ${{ github.event.pull_request.title }}
+        # We want to make sure that each commit "subject" is under 53 characters not to
+        # be truncated in the git log as well as in the GitHub UI.
+        # https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/process/submitting-patches.rst?id=bc7938deaca7f474918c41a0372a410049bd4e13#n664
```
leseb, line 76: This mentions between 70 and 75 in length, so where is 53 coming from?

.github/workflows/pr_title_check.yaml
```diff
@@ -63,3 +66,16 @@ jobs:
...
+        run: |
+          if (( ${#TITLE} >= 53 )); then
+            echo "The PR title is too long. Please keep it under 52 characters."
```
leseb, line 79: The message doesn’t accurately describe what the check is doing. It should state 53, as “under 52” implies less than 52 to me.

PR #272, opened 2025-02-02
chore: more clarify what to do on Style check failure

mathetake:
**Commit Message**

Apparently, the message was clear enough


missBerg (approved):
Thanks for adding more detail and fix instructions

PR #284, opened 2025-02-03
ci: adjust issue/pr style checks

mathetake:
**Commit Message**

This adjusts the issue title check as well as the PR style check. For issue title, I adopt the suggestion by @leseb as indeed 60 was too strict. For the PR stuff, it simply changes it to forbid >60.


.github/workflows/pr_style_check.yaml
```diff
@@ -79,7 +79,7 @@ jobs:
         # be truncated in the git log as well as in the GitHub UI.
```
aabchoo, line 79: nit: update the comment on line 78 to match your echo

aabchoo (approved):

.github/workflows/pr_style_check.yaml
mathetake: good catch aaron!

PR #286, opened 2025-02-03
feat: backend auth credentials management and rotation [WIP]

missBerg:
**Commit Message**

Introduces a new authentication management system for handling backend service credentials with automatic rotation support. 

This is a draft PR for a proposed solution for handling the rotation of access credentials for upstream targets.
See the [Design Documentation](https://github.com/missBerg/ai-gateway/blob/aws-credentials-rotation/docs/design/backend-auth.md) for details of the solution design.

There is notable test coverage that needs to be fixed.


missBerg:
@yuzisun @aabchoo - FYI to take a look. As mentioned in the description, please see as mentioned this code does not have enough test coverage.

The goals of the design is to be able to easily add other rotators in the near future, like for GCP and Azure.

api/v1alpha1/api.go
```diff
@@ -435,6 +435,50 @@ type BackendSecurityPolicyAWSCredentials struct {
...
 	OIDCExchangeToken *AWSOIDCExchangeToken `json:"oidcExchangeToken,omitempty"`
+
+	// Rotation specifies the configuration for rotating AWS credentials.
+	// Only applies when CredentialsFile is used.
+	// +optional
+	Rotation *AWSCredentialsRotation `json:"rotation,omitempty"`
+}
+
+// AWSCredentialsRotation specifies whether and how to rotate AWS credentials
+type AWSCredentialsRotation struct {
```
yuzisun, line 446: I feel we do not necessarily need to expose this API to user, this is controller implementation details for rotating credentials and it is also not specific to AWS.

api/v1alpha1/api.go
```diff
@@ -435,6 +435,50 @@ type BackendSecurityPolicyAWSCredentials struct {
...
+type AWSCredentialsRotationConfig struct {
+	// RotationInterval specifies how often to rotate the credentials.
+	// If not specified, defaults to 24 hours.
+	// The minimum allowed value is 1 hour.
+	//
+	// +optional
+	// +kubebuilder:validation:Type=string
+	// +kubebuilder:validation:Pattern=^([0-9]+h)?([0-9]+m)?$
+	// +kubebuilder:default="24h"
+	RotationInterval string `json:"rotationInterval,omitempty"`
```
yuzisun, line 471: Do we need to specify the rotate interval if we know when the credential is going to expire and controller schedules an event to rotate before the expiration ? Or we are rotating regardless of the expiration with the specified interval.

missBerg: This is if you want to override the defaults, sometimes people want to rotate sooner than expiry, but the window to rotate before expiry is also optional but allows you to say how much earlier than expiry to rotate.

missBerg:
@yuzisun let me know if I can remove this draft PR or if you still want to be able to look at this in this view :)

PR #300, opened 2025-02-06
chore: fixes CEL validation test target name

mathetake:
**Commit Message**

This renames `make test-cel` to `make test-crdcel` to be precise 
about what it is testing since we are also using CEL for LLM costs.

Also, this renames the tests/cel-validation directory to tests/crdcel
since it was previously incorrect and didn't match the make target.


.github/workflows/tests.yaml
```diff
@@ -51,12 +51,11 @@ jobs:
             ~/go/pkg/mod
             ~/go/bin
           key: unittest-${{ hashFiles('**/go.mod', '**/go.sum', '**/Makefile') }}
-      - name: Run unit tests
-        run: make test-coverage
+      - run: make test-coverage
```
mathetake, line 54: note: i figured the naming for the make target doesn't provide any helpful information so i rather have the raw make command exposed - that way anyone can see what they are supposed to do locally when this and elsewhere is failing

PR #305, opened 2025-02-06
e2e: update/fix extproc test

alexagriffith:
**Commit Message**

make fixes to current test file 
- remove TEST_ in aws config name
- skip if openai key is empty (this is for running locally for us, as we dont have that key)


tests/extproc/real_providers_test.go
```diff
@@ -108,6 +110,9 @@ func TestWithRealProviders(t *testing.T) {
 			{testCaseName: "openai", modelName: "gpt-4o-mini"},                            // This will go to "openai"
 			{testCaseName: "aws-bedrock", modelName: "us.meta.llama3-2-1b-instruct-v1:0"}, // This will go to "aws-bedrock" using credentials file.
 		} {
+			if tc.testCaseName == "openai" && openAIAPIKey == "" {
+				t.Skip("Skipping openai test case as TEST_OPENAI_API_KEY is not set")
+			}
```
mathetake, line 115: please do not specialize openai. what happens when we are adding more providers? what about someone wanting to test openai but not aws? Can you simply add a env var requirement check and skip each case accordingly?

mathetake (changes requested):

tests/extproc/real_providers_test.go
alexagriffith: what do you suggest? rn we skip all if no aws key, which makes sense, and we check openai key because its the only other one to check at the moment, if we add a new key we can check it too. can you help me understand better how it could be improved now? open to changing it

alexagriffith: im not sure if an env var req check would look much different than this? maybe im not understanding

mathetake:
i went ahed and implemented the generic version in #326 - hope this works for you :)

alexagriffith:
I know it takes a little time, but I was def willing to update it just wanted to understand what it was you were suggesting. I understand the change though thanks!

mathetake:
yeah sorry i just wanted to fix it by myself as in hindsight i should've made it optional in the first place so i wanted to avoid wasting your time. Thanks for your help anyways!

PR #316, opened 2025-02-10
feat: add extproc prometheus metrics

rootfs:
**Commit Message**
Add prometheus metrics to extproc to track data plan metrics. These metrics help understand requests or token level usage on models and backends.

**Commit Message**

Below is an example output from extproc:
```console
# curl localhost:9190/metrics  
# HELP aigateway_first_token_latency_seconds Time to receive first token in streaming responses
# TYPE aigateway_first_token_latency_seconds histogram
aigateway_first_token_latency_seconds_bucket{backend="ollama",model="phi4",le="0.1"} 0
aigateway_first_token_latency_seconds_bucket{backend="ollama",model="phi4",le="0.25"} 0
aigateway_first_token_latency_seconds_bucket{backend="ollama",model="phi4",le="0.5"} 0
aigateway_first_token_latency_seconds_bucket{backend="ollama",model="phi4",le="1"} 1
aigateway_first_token_latency_seconds_bucket{backend="ollama",model="phi4",le="2.5"} 1
aigateway_first_token_latency_seconds_bucket{backend="ollama",model="phi4",le="5"} 1
aigateway_first_token_latency_seconds_bucket{backend="ollama",model="phi4",le="10"} 1
aigateway_first_token_latency_seconds_bucket{backend="ollama",model="phi4",le="+Inf"} 1
aigateway_first_token_latency_seconds_sum{backend="ollama",model="phi4"} 0.566842292
aigateway_first_token_latency_seconds_count{backend="ollama",model="phi4"} 1
# HELP aigateway_inter_token_latency_seconds Time between consecutive tokens in streaming responses
# TYPE aigateway_inter_token_latency_seconds histogram
aigateway_inter_token_latency_seconds_bucket{backend="ollama",model="phi4",le="0.1"} 7
aigateway_inter_token_latency_seconds_bucket{backend="ollama",model="phi4",le="0.25"} 7
aigateway_inter_token_latency_seconds_bucket{backend="ollama",model="phi4",le="0.5"} 7
aigateway_inter_token_latency_seconds_bucket{backend="ollama",model="phi4",le="1"} 7
aigateway_inter_token_latency_seconds_bucket{backend="ollama",model="phi4",le="2.5"} 7
aigateway_inter_token_latency_seconds_bucket{backend="ollama",model="phi4",le="5"} 7
aigateway_inter_token_latency_seconds_bucket{backend="ollama",model="phi4",le="10"} 7
aigateway_inter_token_latency_seconds_bucket{backend="ollama",model="phi4",le="+Inf"} 7
aigateway_inter_token_latency_seconds_sum{backend="ollama",model="phi4"} 4.837e-05
aigateway_inter_token_latency_seconds_count{backend="ollama",model="phi4"} 7
# HELP aigateway_model_tokens_total Total number of tokens processed by model and type
# TYPE aigateway_model_tokens_total counter
aigateway_model_tokens_total{backend="ollama",model="phi4",type="completion"} 1048
aigateway_model_tokens_total{backend="ollama",model="phi4",type="prompt"} 34
aigateway_model_tokens_total{backend="ollama",model="phi4",type="total"} 1082
# HELP aigateway_requests_total Total number of requests processed
# TYPE aigateway_requests_total counter
aigateway_requests_total{backend="ollama",model="phi4",status="success"} 2
# HELP aigateway_total_latency_seconds Time spent processing request
# TYPE aigateway_total_latency_seconds histogram
aigateway_total_latency_seconds_bucket{backend="ollama",model="phi4",status="success",le="0.1"} 0
aigateway_total_latency_seconds_bucket{backend="ollama",model="phi4",status="success",le="0.5"} 0
aigateway_total_latency_seconds_bucket{backend="ollama",model="phi4",status="success",le="1"} 1
aigateway_total_latency_seconds_bucket{backend="ollama",model="phi4",status="success",le="2.5"} 1
aigateway_total_latency_seconds_bucket{backend="ollama",model="phi4",status="success",le="5"} 1
aigateway_total_latency_seconds_bucket{backend="ollama",model="phi4",status="success",le="10"} 1
aigateway_total_latency_seconds_bucket{backend="ollama",model="phi4",status="success",le="20"} 2
aigateway_total_latency_seconds_bucket{backend="ollama",model="phi4",status="success",le="30"} 2
aigateway_total_latency_seconds_bucket{backend="ollama",model="phi4",status="success",le="60"} 2
aigateway_total_latency_seconds_bucket{backend="ollama",model="phi4",status="success",le="+Inf"} 2
aigateway_total_latency_seconds_sum{backend="ollama",model="phi4",status="success"} 14.449215641
aigateway_total_latency_seconds_count{backend="ollama",model="phi4",status="success"} 2
```

### DIFF
--- a/cmd/extproc/mainlib/main.go
+++ b/cmd/extproc/mainlib/main.go
@@ -35,7 +35,7 @@ type extProcFlags struct {
 	configPath  string     // path to the configuration file.
 	extProcAddr string     // gRPC address for the external processor.
 	logLevel    slog.Level // log level for the external processor.
-	promPort    string     // Prometheus port
+	promAddr    string     // Prometheus address
 }
 
 // parseAndValidateFlags parses and validates the flas passed to the external processor.
@@ -63,10 +63,10 @@ func parseAndValidateFlags(args []string) (extProcFlags, error) {
 		"log level for the external processor. One of 'debug', 'info', 'warn', or 'error'.",
 	)
 
-	fs.StringVar(&flags.promPort,
-		"promPort",
+	fs.StringVar(&flags.promAddr,
+		"promAddr",
 		":9190",
-		"port for prometheus metrics, default is 9190",
+		"address for prometheus metrics, default is :9190",
 	)
 
 	if err := fs.Parse(args); err != nil {
@@ -97,7 +97,7 @@ func Main() {
 		slog.String("version", version.Version),
 		slog.String("address", flags.extProcAddr),
 		slog.String("configPath", flags.configPath),
-		slog.String("promPort", flags.promPort),
+		slog.String("promAddr", flags.promAddr),
 	)
 
 	ctx, cancel := context.WithCancel(context.Background())
@@ -127,29 +127,31 @@ func Main() {
 	s := grpc.NewServer()
 	extprocv3.RegisterExternalProcessorServer(s, server)
 	grpc_health_v1.RegisterHealthServer(s, server)
-	go func() {
-		<-ctx.Done()
-		s.GracefulStop()
-	}()
 
 	// Add prometheus metrics
 	metricsInstance := metrics.GetOrCreate()
 
 	// Setup prometheus handler
 	http.Handle("/metrics", promhttp.HandlerFor(metricsInstance.Registry, promhttp.HandlerOpts{}))
+	promServer := &http.Server{
+		Addr:              flags.promAddr,
+		Handler:           nil,
+		ReadHeaderTimeout: 5 * time.Second,
+		ReadTimeout:       10 * time.Second,
+		WriteTimeout:      10 * time.Second,
+		IdleTimeout:       15 * time.Second,
+	}
+	go func() {
+		if err := promServer.ListenAndServe(); err != nil && err != http.ErrServerClosed {
+			log.Fatalf("Metrics server failed: %v", err)
+		}
+	}()
 
 	go func() {
-		server := &http.Server{
-			Addr:              flags.promPort,
-			Handler:           nil,
-			ReadHeaderTimeout: 5 * time.Second,
-			ReadTimeout:       10 * time.Second,
-			WriteTimeout:      10 * time.Second,
-			IdleTimeout:       15 * time.Second,
-		}
-
-		if err := server.ListenAndServe(); err != nil && err != http.ErrServerClosed {
-			log.Fatalf("Metrics server failed: %v", err)
+		<-ctx.Done()
+		s.GracefulStop()
+		if err := promServer.Shutdown(ctx); err != nil {
+			log.Fatalf("Metrics server failed to shutdown: %v", err)
 		}
 	}()
 
--- a/cmd/extproc/mainlib/main.go
+++ b/cmd/extproc/mainlib/main.go
@@ -13,6 +13,7 @@ import (
 	"log"
 	"log/slog"
 	"net"
+	"net/http"
 	"os"
 	"os/signal"
 	"strings"
@@ -20,10 +21,12 @@ import (
 	"time"
 
 	extprocv3 "github.com/envoyproxy/go-control-plane/envoy/service/ext_proc/v3"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/health/grpc_health_v1"
 
 	"github.com/envoyproxy/ai-gateway/internal/extproc"
+	"github.com/envoyproxy/ai-gateway/internal/metrics"
 	"github.com/envoyproxy/ai-gateway/internal/version"
 )
 
@@ -32,6 +35,7 @@ type extProcFlags struct {
 	configPath  string     // path to the configuration file.
 	extProcAddr string     // gRPC address for the external processor.
 	logLevel    slog.Level // log level for the external processor.
+	promPort    string     // Prometheus port
 }
 
 // parseAndValidateFlags parses and validates the flas passed to the external processor.
@@ -57,6 +61,12 @@ func parseAndValidateFlags(args []string) (extProcFlags, error) {
 		"logLevel",
 		"info",
 		"log level for the external processor. One of 'debug', 'info', 'warn', or 'error'.",
+	)
+
+	fs.StringVar(&flags.promPort,
+		"promPort",
+		":9190",
+		"port for prometheus metrics, default is 9190",
 	)
 
 	if err := fs.Parse(args); err != nil {
@@ -87,6 +97,7 @@ func Main() {
 		slog.String("version", version.Version),
 		slog.String("address", flags.extProcAddr),
 		slog.String("configPath", flags.configPath),
+		slog.String("promPort", flags.promPort),
 	)
 
 	ctx, cancel := context.WithCancel(context.Background())
@@ -120,6 +131,28 @@ func Main() {
 		<-ctx.Done()
 		s.GracefulStop()
 	}()
+
+	// Add prometheus metrics
+	metricsInstance := metrics.GetOrCreate()
+
+	// Setup prometheus handler
+	http.Handle("/metrics", promhttp.HandlerFor(metricsInstance.Registry, promhttp.HandlerOpts{}))
+
+	go func() {
+		server := &http.Server{
+			Addr:              flags.promPort,
+			Handler:           nil,
+			ReadHeaderTimeout: 5 * time.Second,
+			ReadTimeout:       10 * time.Second,
+			WriteTimeout:      10 * time.Second,
+			IdleTimeout:       15 * time.Second,
+		}
+
+		if err := server.ListenAndServe(); err != nil && err != http.ErrServerClosed {
+			log.Fatalf("Metrics server failed: %v", err)
+		}
+	}()
+
 	_ = s.Serve(lis)
 }
 
--- a/internal/extproc/chatcompletion_processor.go
+++ b/internal/extproc/chatcompletion_processor.go
@@ -14,7 +14,6 @@ import (
 	"fmt"
 	"io"
 	"log/slog"
-	"time"
 
 	corev3 "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
 	extprocv3 "github.com/envoyproxy/go-control-plane/envoy/service/ext_proc/v3"
@@ -38,7 +37,7 @@ func NewChatCompletionProcessor(config *processorConfig, requestHeaders map[stri
 		config:         config,
 		requestHeaders: requestHeaders,
 		logger:         logger,
-		metrics:        metrics.GetOrCreate(),
+		metrics:        metrics.NewTokenMetrics(),
 	}, nil
 }
 
@@ -53,10 +52,9 @@ type chatCompletionProcessor struct {
 	// cost is the cost of the request that is accumulated during the processing of the response.
 	costs translator.LLMTokenUsage
 	// for metrics
-	requestStart time.Time
-	modelName    string
-	backendName  string
-	metrics      *metrics.Metrics
+	metrics     *metrics.TokenMetrics
+	backendName string
+	modelName   string
 }
 
 // selectTranslator selects the translator based on the output schema.
@@ -86,18 +84,20 @@ func (c *chatCompletionProcessor) ProcessRequestHeaders(_ context.Context, _ *co
 
 // ProcessRequestBody implements [Processor.ProcessRequestBody].
 func (c *chatCompletionProcessor) ProcessRequestBody(ctx context.Context, rawBody *extprocv3.HttpBody) (res *extprocv3.ProcessingResponse, err error) {
-	c.requestStart = time.Now() // Track the start time of the request
+	c.backendName = "unknown"
+	c.modelName = "unknown"
 	model, body, err := parseOpenAIChatCompletionBody(rawBody)
 	if err != nil {
-		c.metrics.RequestsTotal.WithLabelValues("unknown", "unknown", "error").Inc()
+		c.metrics.UpdateRequestMetrics(c.backendName, c.modelName, "error")
 		return nil, fmt.Errorf("failed to parse request body: %w", err)
 	}
 	c.logger.Info("Processing request", "path", c.requestHeaders[":path"], "model", model)
 
 	c.requestHeaders[c.config.modelNameHeaderKey] = model
+	c.modelName = model
 	b, err := c.config.router.Calculate(c.requestHeaders)
 	if err != nil {
-		c.metrics.RequestsTotal.WithLabelValues("unknown", model, "error").Inc()
+		c.metrics.UpdateRequestMetrics(c.backendName, c.modelName, "error")
 		if errors.Is(err, x.ErrNoMatchingRule) {
 			return &extprocv3.ProcessingResponse{
 				Response: &extprocv3.ProcessingResponse_ImmediateResponse{
@@ -112,15 +112,15 @@ func (c *chatCompletionProcessor) ProcessRequestBody(ctx context.Context, rawBod
 		return nil, fmt.Errorf("failed to calculate route: %w", err)
 	}
 	c.logger.Info("Selected backend", "backend", b.Name)
-
+	c.backendName = b.Name
 	if err = c.selectTranslator(b.Schema); err != nil {
-		c.metrics.RequestsTotal.WithLabelValues(b.Name, model, "error").Inc()
+		c.metrics.UpdateRequestMetrics(c.backendName, c.modelName, "error")
 		return nil, fmt.Errorf("failed to select translator: %w", err)
 	}
 
 	headerMutation, bodyMutation, override, err := c.translator.RequestBody(body)
 	if err != nil {
-		c.metrics.RequestsTotal.WithLabelValues(b.Name, model, "error").Inc()
+		c.metrics.UpdateRequestMetrics(c.backendName, c.modelName, "error")
 		return nil, fmt.Errorf("failed to transform request: %w", err)
 	}
 
@@ -134,9 +134,9 @@ func (c *chatCompletionProcessor) ProcessRequestBody(ctx context.Context, rawBod
 		Header: &corev3.HeaderValue{Key: c.config.selectedBackendHeaderKey, RawValue: []byte(b.Name)},
 	})
 
-	if authHandler, ok := c.config.backendAuthHandlers[b.Name]; ok {
+	if authHandler, ok := c.config.backendAuthHandlers[c.backendName]; ok {
 		if err := authHandler.Do(ctx, c.requestHeaders, headerMutation, bodyMutation); err != nil {
-			c.metrics.RequestsTotal.WithLabelValues(b.Name, model, "error").Inc()
+			c.metrics.UpdateRequestMetrics(c.backendName, c.modelName, "error")
 			return nil, fmt.Errorf("failed to do auth request: %w", err)
 		}
 	}
@@ -155,8 +155,7 @@ func (c *chatCompletionProcessor) ProcessRequestBody(ctx context.Context, rawBod
 	}
 
 	// Track the backend and model name for metrics
-	c.backendName = b.Name
-	c.modelName = model
+	c.metrics.StartRequest(c.backendName, c.modelName)
 	return resp, nil
 }
 
@@ -223,18 +222,16 @@ func (c *chatCompletionProcessor) ProcessResponseBody(_ context.Context, body *e
 	c.costs.OutputTokens += tokenUsage.OutputTokens
 	c.costs.TotalTokens += tokenUsage.TotalTokens
 	// Track the token usage for metrics
-	c.metrics.TokensTotal.WithLabelValues(c.modelName, "total").Add(float64(tokenUsage.TotalTokens))
-	c.metrics.TokensTotal.WithLabelValues(c.modelName, "prompt").Add(float64(tokenUsage.InputTokens))
-	c.metrics.TokensTotal.WithLabelValues(c.modelName, "completion").Add(float64(tokenUsage.OutputTokens))
+	c.metrics.UpdateTokenMetrics(c.backendName, c.modelName, tokenUsage.OutputTokens, tokenUsage.InputTokens, tokenUsage.TotalTokens)
 
 	if body.EndOfStream && len(c.config.requestCosts) > 0 {
 		resp.DynamicMetadata, err = c.maybeBuildDynamicMetadata()
 		if err != nil {
+			c.metrics.UpdateRequestMetrics(c.backendName, c.modelName, "error")
 			return nil, fmt.Errorf("failed to build dynamic metadata: %w", err)
 		}
-		c.metrics.RequestsTotal.WithLabelValues(c.backendName, c.modelName, "success").Inc()
-		c.metrics.BackendLatency.WithLabelValues(c.backendName, c.modelName, "success").Observe(time.Since(c.requestStart).Seconds())
 	}
+	c.metrics.UpdateRequestMetrics(c.backendName, c.modelName, "success")
 	return resp, nil
 }
 
--- a/internal/extproc/chatcompletion_processor.go
+++ b/internal/extproc/chatcompletion_processor.go
@@ -14,6 +14,7 @@ import (
 	"fmt"
 	"io"
 	"log/slog"
+	"time"
 
 	corev3 "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
 	extprocv3 "github.com/envoyproxy/go-control-plane/envoy/service/ext_proc/v3"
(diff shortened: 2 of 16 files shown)


internal/metrics/metrics.go
```diff
@@ -0,0 +1,89 @@
...
+	TokensTotal *prometheus.CounterVec
+
+	// RequestsTotal tracks total requests by backend, model and status
+	RequestsTotal *prometheus.CounterVec
+
+	// FirstTokenLatency measures time to first token in streaming responses
+	FirstTokenLatency *prometheus.HistogramVec
+
+	// InterTokenLatency measures time between tokens in streaming responses
+	InterTokenLatency *prometheus.HistogramVec
```
mathetake, line 26: let's avoid the global state, but instantiate them in the main routine and pass them through via a struct field for testability.

internal/extproc/translator/openai_openai.go
```diff
@@ -151,5 +162,26 @@ func (o *openAIToOpenAITranslatorV1ChatCompletion) extractUsageFromBufferEvent()
...
+				// And in some cases, the number of tokens can be 0, so we need to check for that.
+				div := tokenUsage.OutputTokens
+				if div == 0 {
+					div = 1
+				}
+				itl := now.Sub(o.lastTokenTime).Seconds() / float64(div)
+				metrics.InterTokenLatency.WithLabelValues(o.backendName, o.modelName).
+					Observe(itl)
+				o.lastTokenTime = now
+			}
```
mathetake, line 184: i don't think this is unit tested?

internal/extproc/translator/openai_openai.go
```diff
@@ -151,5 +162,26 @@ func (o *openAIToOpenAITranslatorV1ChatCompletion) extractUsageFromBufferEvent()
 			o.buffered = nil
 			return
 		}
+		if o.backendName != "" && o.modelName != "" {
```
mathetake, line 165: when exactly this condition satisfies?

internal/extproc/translator/openai_awsbedrock.go
```diff
@@ -551,6 +562,27 @@ func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) ResponseBody(respHeaders
 		if endOfStream {
 			mut.Body = append(mut.Body, []byte("data: [DONE]\n")...)
 		}
+		if o.backendName != "" && o.modelName != "" {
```
mathetake, line 565: same question

internal/extproc/translator/openai_awsbedrock.go
```diff
@@ -551,6 +562,27 @@ func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) ResponseBody(respHeaders
...
+				// And in some cases, the number of tokens can be 0, so we need to check for that.
+				div := tokenUsage.OutputTokens
+				if div == 0 {
+					div = 1
+				}
+				itl := now.Sub(o.lastTokenTime).Seconds() / float64(div)
+				metrics.InterTokenLatency.WithLabelValues(o.backendName, o.modelName).
+					Observe(itl)
+				o.lastTokenTime = now
+			}
```
mathetake, line 584: also here is not tested?

internal/extproc/processor.go
```diff
@@ -204,11 +224,28 @@ func (p *Processor) ProcessResponseBody(_ context.Context, body *extprocv3.HttpB
...
+	if tokenUsage.TotalTokens > 0 {
+		metrics.TokensTotal.WithLabelValues(p.modelName, "total").Add(float64(tokenUsage.TotalTokens))
+	}
+
 	if body.EndOfStream && len(p.config.requestCosts) > 0 {
 		resp.DynamicMetadata, err = p.maybeBuildDynamicMetadata()
 		if err != nil {
 			return nil, fmt.Errorf("failed to build dynamic metadata: %w", err)
 		}
+		if p.backendName != "" && p.modelName != "" {
```
mathetake, line 244: also here - when exactly does this happen? if it ever happens then we need a code comment here as it's not obvious without the context vs if not let's remove. Either way this is not tested

mathetake (commented):
thanks! Left comments mostly about the testability

mathetake:
> I have only tested on OpenAI compatible platforms. If anybody can test aws bedrock, that'll be great

this is why at least you need unit tests ;) anyways the coverage check is failing as expected

internal/metrics/metrics.go
rootfs: sounds good, will follow up later to get these done

internal/extproc/processor.go
```diff
@@ -81,9 +87,13 @@ func (p *Processor) ProcessRequestHeaders(_ context.Context, headers *corev3.Hea
 
 // ProcessRequestBody implements [Processor.ProcessRequestBody].
 func (p *Processor) ProcessRequestBody(ctx context.Context, rawBody *extprocv3.HttpBody) (res *extprocv3.ProcessingResponse, err error) {
+	p.requestStart = time.Now() // Track request metrics
+
 	path := p.requestHeaders[":path"]
 	model, body, err := p.config.bodyParser(path, rawBody)
 	if err != nil {
+		// Add to metrics tracker
+		metrics.RequestsTotal.WithLabelValues("unknown", "unknown", "error").Inc()
```
nacx, line 96: I see that we're setting some metrics when there are errors, but not on all of them. Is it intentional? If not, this looks a bit error-prone, and easy to miss setting the error metric when adding new conditionals to the code int he future...

WDYT about refactoring the metrics functionality out to a processor that could wrap this one? something like an `InstrumentedProcessor` that wraps this one and emits the metrics when needed? This way it's easier to always emit the error metrics (when the inner processor returned an error), and don't miss that.
Also, the backend and model are published in the headers, so they should be accessible to the wrapping processor.

I don't know if given the current code and execution flows it is possible, but if it is, it would probably be cleaner

internal/extproc/translator/openai_awsbedrock.go
rootfs: test case updated

internal/extproc/processor.go
rootfs: The metrics is created in its own factory. When the extproc process starts, the metrics is there. If there is a misconstructed body, tracking it in the metrics is more helpful than dropping it silently or burry it in logs.

rootfs:
@mathetake PR updated with metrics builder and more test coverage. PTAL and I'll rebase. Thanks

internal/extproc/processor.go
mathetake: +1

rootfs: @nacx i see what you meant. Yes, all errors should be tracked. The codebase is rapidly changing to catch up these errors :D I'll make sure the metrics are all the places you mentioned during next rebase.

internal/extproc/processor.go
```diff
@@ -184,8 +204,13 @@ func (p *Processor) ProcessResponseBody(_ context.Context, body *extprocv3.HttpB
 		return &extprocv3.ProcessingResponse{Response: &extprocv3.ProcessingResponse_ResponseBody{}}, nil
 	}
 
-	headerMutation, bodyMutation, tokenUsage, err := p.translator.ResponseBody(p.responseHeaders, br, body.EndOfStream)
+	headerMutation, bodyMutation, tokenUsage, err := p.translator.ResponseBody(p.responseHeaders, br, body.EndOfStream, p.backendName, p.modelName)
 	if err != nil {
+		if p.backendName != "" && p.modelName != "" {
```
yuzisun, line 209: We probably should not have this check here, if the model name or backend are not there we should fail upfront

internal/extproc/processor.go
```diff
@@ -184,8 +204,13 @@ func (p *Processor) ProcessResponseBody(_ context.Context, body *extprocv3.HttpB
...
+			p.metrics.RequestsTotal.WithLabelValues(p.backendName, p.modelName, "error").Inc()
```
yuzisun, line 210: I think we should track the error count at the higher level here 
https://github.com/envoyproxy/ai-gateway/blob/9d4fc925b81249231f2b0a5ba394ed8e8536e336/internal/extproc/server.go#L185

internal/extproc/translator/openai_awsbedrock.go
```diff
@@ -97,6 +113,7 @@ func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) RequestBody(body router.R
 		return nil, nil, nil, fmt.Errorf("failed to marshal body: %w", err)
 	}
 	setContentLength(headerMutation, mut.Body)
+	o.requestStart = time.Now()
```
yuzisun, line 116: Move this line upfront as there are early returns, also request starts is counted from requestHeader not body @mathetake ?

internal/extproc/translator/openai_awsbedrock.go
```diff
@@ -551,6 +570,28 @@ func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) ResponseBody(respHeaders
...
+				o.metrics.FirstTokenLatency.WithLabelValues(o.backendName, o.modelName).
+					Observe(now.Sub(o.requestStart).Seconds())
+				o.firstTokenSent = false
+				o.lastTokenTime = now
+			} else {
+				// Calculate the time between tokens.
+				// Since we are only interested in the time between tokens, and openai streaming is by chunk,
+				// we can calculate the time between tokens by the time between the last token and the current token, divided by the number of tokens.
+				// And in some cases, the number of tokens can be 0, so we need to check for that.
+				div := tokenUsage.OutputTokens
```
yuzisun, line 585: need to confirm if the token count reported is accumulative or just the output tokens in the trunk, it may differ between different backends

internal/metrics/metrics.go
```diff
@@ -0,0 +1,89 @@
...
+				Buckets: []float64{.1, .5, 1, 2.5, 5, 10, 20, 30, 60},
+			},
+			[]string{"backend", "model", "status"},
+		),
+		TokensTotal: prometheus.NewCounterVec(
+			prometheus.CounterOpts{
+				Name: "aigateway_model_tokens_total",
+				Help: "Total number of tokens processed by model and type",
+			},
+			[]string{"model", "type"},
```
yuzisun, line 41: missed the backend tag

internal/extproc/translator/openai_openai.go
```diff
@@ -151,5 +179,26 @@ func (o *openAIToOpenAITranslatorV1ChatCompletion) extractUsageFromBufferEvent()
...
+			} else {
+				// Calculate the time between tokens.
+				// Since we are only interested in the time between tokens, and openai streaming is by chunk,
+				// we can calculate the time between tokens by the time between the last token and the current token, divided by the number of tokens.
+				// And in some cases, the number of tokens can be 0, so we need to check for that.
+				div := tokenUsage.OutputTokens
+				if div == 0 {
+					div = 1
+				}
+				itl := now.Sub(o.lastTokenTime).Seconds() / float64(div)
```
yuzisun, line 197: Can extract to an util function as it is used by multiple translators

rootfs: sure, will do

mathetake:
@rootfs would it be possible to resolve the conflicts and address the existing comments? @nacx and I really would like to see this finish

rootfs:
@mathetake sure thing, back to work today, will rebase soon.

cmd/extproc/mainlib/main.go
```diff
@@ -59,6 +63,12 @@ func parseAndValidateFlags(args []string) (extProcFlags, error) {
 		"log level for the external processor. One of 'debug', 'info', 'warn', or 'error'.",
 	)
 
+	fs.StringVar(&flags.promPort,
+		"promPort",
+		":9190",
+		"port for prometheus metrics, default is 9190",
```
nacx, line 69: * Let's rename it to "addr" because it's a listen address, not just a port?
* Also, the flag library already includes the default in the message. If we add it to the description it will show up twice.
```suggestion
	fs.StringVar(&flags.promAddr,
		"promAddr",
		":9190",
		"address where Prometheus metrics are exposed",
```

cmd/extproc/mainlib/main.go
```diff
@@ -120,6 +131,28 @@ func Main() {
...
 	}()
+
+	// Add prometheus metrics
+	metricsInstance := metrics.GetOrCreate()
+
+	// Setup prometheus handler
+	http.Handle("/metrics", promhttp.HandlerFor(metricsInstance.Registry, promhttp.HandlerOpts{}))
+
+	go func() {
+		server := &http.Server{
```
nacx, line 142: nit: You may want to define the `server` outside this goroutine so you can also call `server.Shutdown`, in the goroutine above that waits for the context to be done, this way we gracefully terminate everything.

internal/metrics/metrics.go
```diff
@@ -0,0 +1,89 @@
...
+
+import (
+	"sync"
+
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+// Metrics holds all prometheus metrics
+type Metrics struct {
+	BackendLatency    *prometheus.HistogramVec
```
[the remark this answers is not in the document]
rootfs: renamed to `TotalLatency`

internal/extproc/chatcompletion_processor.go
```diff
@@ -78,15 +86,18 @@ func (c *chatCompletionProcessor) ProcessRequestHeaders(_ context.Context, _ *co
 
 // ProcessRequestBody implements [Processor.ProcessRequestBody].
 func (c *chatCompletionProcessor) ProcessRequestBody(ctx context.Context, rawBody *extprocv3.HttpBody) (res *extprocv3.ProcessingResponse, err error) {
+	c.requestStart = time.Now() // Track the start time of the request
```
nacx, line 89: Should we set the `requestStart` in the OnRequestHeaders method?

internal/extproc/chatcompletion_processor.go
```diff
@@ -78,15 +86,18 @@ func (c *chatCompletionProcessor) ProcessRequestHeaders(_ context.Context, _ *co
...
 	model, body, err := parseOpenAIChatCompletionBody(rawBody)
 	if err != nil {
+		c.metrics.RequestsTotal.WithLabelValues("unknown", "unknown", "error").Inc()
```
nacx, line 92: What about this for a less error-prone code.

Given that we have "named return values" we could have something like this at the beginning of this method:
```go
c.backendName := "unknown"
c.modelName := "unknown"
defer func() {
  if err != nil {
    c.metrics.RequestsTotal.WithLabelValues(c. backendName, c.modelName, "error").Inc()
  }
}()
```
This way you don't have to increment the metric on every conditional path (which is easy to miss), and you can always have the errors properly accounted for.

internal/extproc/chatcompletion_processor.go
```diff
@@ -204,11 +222,18 @@ func (c *chatCompletionProcessor) ProcessResponseBody(_ context.Context, body *e
...
+	c.metrics.TokensTotal.WithLabelValues(c.modelName, "total").Add(float64(tokenUsage.TotalTokens))
+	c.metrics.TokensTotal.WithLabelValues(c.modelName, "prompt").Add(float64(tokenUsage.InputTokens))
+	c.metrics.TokensTotal.WithLabelValues(c.modelName, "completion").Add(float64(tokenUsage.OutputTokens))
+
 	if body.EndOfStream && len(c.config.requestCosts) > 0 {
 		resp.DynamicMetadata, err = c.maybeBuildDynamicMetadata()
 		if err != nil {
 			return nil, fmt.Errorf("failed to build dynamic metadata: %w", err)
 		}
+		c.metrics.RequestsTotal.WithLabelValues(c.backendName, c.modelName, "success").Inc()
```
nacx, line 235: Shouldn't we increment the errors on every error returned by this method and any other processor method as well?
Otherwise, depending on "when" during the lifecycle of the request the failure is, you get error metrics or just nothing.
You can use the same "defer" pattern as suggested above.

internal/extproc/processor.go
```diff
@@ -37,6 +43,9 @@ type processorConfigRequestCost struct {
 	celProg cel.Program
 }
 
+type processorMetrics struct {
+}
```
nacx, line 47: What is this type used for?

internal/extproc/translator/openai_awsbedrock.go
nacx: I agree with @yuzisun. The main issue is that there is a mix of lifecycles here:

* The request lifecycle is owned by the processor. That is the one that has the OnRequestHeaders, OnResponseBody, etc.
* This is "translator code", which basically transforms request/responses from/to the specific selected backend formats.

With this, I think there a re a couple options:

* If we want the translator code to be recording metrics, then some context about the initial request (such as the request creation time) should be passed to it when it's being instantiated.
* If we want to keep the translator focused on just translating (this is probably what I think would be cleaner), we should think about a way the translator could provide information about internal processing to the calling processor, so that the processor can emit the right metrics.

internal/extproc/translator/openai_awsbedrock.go
```diff
@@ -646,6 +680,9 @@ func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) convertEvent(event *awsbe
 			PromptTokens:     event.Usage.InputTokens,
 			CompletionTokens: event.Usage.OutputTokens,
 		}
+		o.metrics.TokensTotal.WithLabelValues(o.modelName, "total").Add(float64(event.Usage.TotalTokens))
+		o.metrics.TokensTotal.WithLabelValues(o.modelName, "prompt").Add(float64(event.Usage.InputTokens))
+		o.metrics.TokensTotal.WithLabelValues(o.modelName, "completion").Add(float64(event.Usage.OutputTokens))
```
nacx, line 685: Aren't these metrics already recorded by the processor?

internal/extproc/translator/openai_openai.go
```diff
@@ -120,6 +134,10 @@ func (o *openAIToOpenAITranslatorV1ChatCompletion) ResponseBody(respHeaders map[
 		OutputTokens: uint32(resp.Usage.CompletionTokens), //nolint:gosec
 		TotalTokens:  uint32(resp.Usage.TotalTokens),      //nolint:gosec
 	}
+	// add metrics
+	o.metrics.TokensTotal.WithLabelValues(o.modelName, "total").Add(float64(tokenUsage.TotalTokens))
+	o.metrics.TokensTotal.WithLabelValues(o.modelName, "prompt").Add(float64(tokenUsage.InputTokens))
+	o.metrics.TokensTotal.WithLabelValues(o.modelName, "completion").Add(float64(tokenUsage.OutputTokens))
```
nacx, line 140: Same comment about recording metrics that are already recorded by the processor.

This is something that should be covered by unit tests, but the current setup makes it hard to test.

One approach to make the whole code more testable would be to create an interface in the `metrics` package like:
```go
type MetricsRecorder interface{
    RecordTokensTotal(...)
    RecordTimeToFirstToken(...)
    ...
}
```
That implements the calls to the real metrics backend.

* This would decouple the processor code from the metrics backend of choice.
* This would allow you to plug a mock implementation of the `MetricsRecorder` in the unit tests, so that you can have proper assertions on the expected values and number of times each method is called. This will help make sure that:
  * Each method is called just when needed.
  * We can properly account for all error paths being properly accounted for in metrics.

internal/metrics/metrics.go
```diff
@@ -0,0 +1,89 @@
...
+	Registry          *prometheus.Registry
+}
+
+var (
+	instance *Metrics
+	once     sync.Once
+)
+
+// New creates a new Metrics instance
+func New() *Metrics {
```
nacx, line 25: Make this private so everyone calls `GetOrCreate()`

internal/metrics/metrics.go
rootfs: backend label added

cmd/extproc/mainlib/main.go
rootfs: updated

cmd/extproc/mainlib/main.go
rootfs: updated

internal/extproc/chatcompletion_processor.go
rootfs: i don't see this method yet

internal/extproc/chatcompletion_processor.go
rootfs: adopted

internal/extproc/chatcompletion_processor.go
rootfs: updated

internal/extproc/processor.go
rootfs: removed

internal/extproc/translator/openai_awsbedrock.go
rootfs: all the metrics are now encapsulated.

internal/metrics/metrics.go
rootfs: updated

internal/extproc/processor.go
rootfs: `processMsg` doesn't have model or backend names yet.

internal/extproc/processor.go
rootfs: removed

internal/extproc/processor.go
rootfs: removed

internal/extproc/translator/openai_awsbedrock.go
rootfs: removed

internal/extproc/translator/openai_openai.go
rootfs: they are tested now

internal/extproc/translator/openai_openai.go
rootfs: all metrics are encapsulated now.

rootfs:
@mathetake @nacx @yuzisun at the moment, there are still questions on entrypoint of recording the start of the request header. I don't see request header entrypoint in main branch yet. But the current code is flexible to move to new entrypoints when you have them.

mathetake:
* I would strongly suggest to not have a metrics context tied to translator. The metric "sink" should be coupled with the kind of Processor, not translator. To do so,
    1. Define `ChatCompletionsMetrics` and make it owned by ChatCompletionsProcessor. 
    2. Pass it from processor to the OpenAIChatCompletionTranslator methods as an argument or pass it to `NewChatCompletionOpenAIToOpenAITranslator` and `NewChatCompletionOpenAIToAWSBedrockTranslator`. Not initialize the metric stuff in translator's constructor. (#425 will be merged in a sec)
* Then, the request time initialization should start at RequestHeaders method

internal/extproc/chatcompletion_processor.go
nacx: Sorry, I meant `ProcessRequestHeaders`

internal/extproc/translator/openai_awsbedrock.go
```diff
@@ -603,6 +618,8 @@ func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) ResponseBody(respHeaders
 	}
 	headerMutation = &extprocv3.HeaderMutation{}
 	setContentLength(headerMutation, mut.Body)
+	o.tokenMetrics.UpdateTokenMetrics(o.backendName, o.modelName, tokenUsage.OutputTokens, tokenUsage.InputTokens, tokenUsage.TotalTokens)
```
nacx, line 621: As mentioned in the previous review, these metrics are already recorded by the calling processor in its ProcessResponseBody method, once the translator returns.

It looks like the values will be incremened twice, which is not correct?

This is why it is important to have the tests I mention above, to make sure we're only recording metrics when needed, and not doing it N times by mistake.

This also applies to the other translators.

internal/extproc/processor.go
mathetake: This is not resolved

rootfs:
> * I would strongly suggest to not have a metrics context tied to translator. The metric "sink" should be coupled with the kind of Processor, not translator. To do so,
>   
>   1. Define `ChatCompletionsMetrics` and make it owned by ChatCompletionsProcessor.
>   2. Pass it from processor to the OpenAIChatCompletionTranslator methods as an argument or pass it to `NewChatCompletionOpenAIToOpenAITranslator` and `NewChatCompletionOpenAIToAWSBedrockTranslator`. Not initialize the metric stuff in translator's constructor. ([extproc: renames translator interface #425](https://github.com/envoyproxy/ai-gateway/pull/425) will be merged in a sec)
> * Then, the request time initialization should start at RequestHeaders method

@mathetake ok, that'll be a major refactor. I am going to do it in a clean way then. Closing this PR and will submit a new one based on the updated interface.

mathetake:
No reason to close the PR though ...

nacx:
Agree. No need to close the PR. You're doing an amazing job on this one, keeping up with all the reviews, etc. We can merge this and tackle that comment in a follow-up PR!

PR #260, opened 2025-01-31
docs: added a usage-based ratelimiting section in the docs

melsal13:
**Commit Message**
Added a usage-based ratelimiting section in the docs. Making it easier for people to understand how to configure rate-limiting.

**Related Issues/PRs (if applicable)**
Fixes #242

PR #318, opened 2025-02-10
docs: tweaks rate limit doc on model

mathetake:
**Commit Message**

The model name is extracted by AI Gateway filter, not the one explicitly added by downstream clients.

Follow up on #260


site/docs/capabilities/usage-based-ratelimiting.md
```diff
@@ -151,18 +151,17 @@ When configuring rate limits:
...
 ```shell
 curl --fail \
     -H "x-user-id: user123" \
-    -H "x-ai-eg-model: gpt-4" \    # Both user ID and model are required
     -d '{
         "messages": [
             {
                 "role": "user",
+                "model": "gpt-4",
```
mathetake, line 164: ah this is not the right place

PR #323, opened 2025-02-11
chore: add helm install

### LINKED ISSUE #302: build: install helm command as a part of Makefile.tools.mk
*Description*:

Currently, `helm` is being used inside the e2e test and we assume developers install the command locally. Ideally, we want to install it as a part of Makefile.tools.mk like the other tools such as kind.

daixiang0:
**Commit Message**

Add helm install.

Fixes #302

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -251,7 +251,7 @@ HELM_DIR := ./manifests/charts/ai-gateway-helm
 
 # This lints the helm chart, ensuring that it is for packaging.
 .PHONY: helm-lint
-helm-lint:
+helm-lint: helm
 	@echo "helm-lint => .${HELM_DIR}"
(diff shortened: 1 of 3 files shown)


mathetake:
sorry what i meant in #302 is to use `go install`. is it possible?

daixiang0:
@mathetake changed.

mathetake (commented):
I think you need to make change to e2e test as well where spawning helm command so that it will actually point to .bin/ directory (see the case of kind). Then we can verify it's working correctly

mathetake:
@daixiang0 hey could you stop force pushing like i said before? thanks for the change anyways

daixiang0:
@mathetake I will keep it in mind, sorry.

Issue #186, opened 2025-01-27
discussion: serving common AI features


nacx:
I would like to discuss whether serving common AI features would be within this project's scope.

A good example would be the `/v1/models` endpoint. This is not implemented by every AI provider, but it is very commonly used by applications that allow users to choose their desired model. Right now, the ext-proc filter would fail for requests that go to any endpoint that is not the chat completions endpoint (https://github.com/envoyproxy/ai-gateway/issues/115 was created to address this), but probably the project could do more to facilitate the adoption to existing apps that rely on such APIs.

In the case of the `/v1/models` endpoint, for example, it would make a lot of sense that the ai-gateway could serve the response for such requests based on what has been configured in the ConfigMap, returning those models that have been configured and are allowed to be used.

What is the general feeling about ai-gateway directly implementing common AI features?

mathetake:
cc @envoyproxy/ai-gateway-assignable @envoyproxy/ai-gateway-maintainers

wengyao04:
We currently have this requirement. In our POC solution, we have a service to return /v1/models and we create a envoyproxy backend pointing to this service. Wondering if AIGatewayRoute allow both AIServiceBckend and Backend ?

mathetake:
yeah maybe adding a field of a list of available models to AIServiceBackend or AIGatewayRoute makes sense

wengyao04:
@yuzisun and I discuss about it and propose that gateway can aggregate this model information from AIServiceBackend + AIGatewayRoute and return immediate response.

In our use case, we can enrich the response from our internal DB to add more information

mathetake:
yep, let's do this. @nacx want to raise an API PR for `/models` endpoints?

mathetake:
and i wonder what other endpoints will fall into the similar style

missBerg:
Yeah probably worth investigating. @nacx @wengyao04 @Krishanx92 - got any thoughts on what other information type endpoints would make sense here?

mathetake:
@nacx do you have any update?

Rutledge:
Sharing from the community meeting today:

This is a key use case for users. A common need would be:
1. User adds their API key for a provider to the gateway
2. User is provided the list of models they have access to via that provider
3. Per model information is available include per token pricing, input/output window sizes, capabilities of these models (tool use, vision, audio, etc)

mathetake:
> User adds their API key for a provider to the gateway

@Rutledge this is not relevant here as already implemented and working!

[...]

PR #325, opened 2025-02-11
extproc: custom processors per path and serve /v1/models

nacx:
**Commit Message**

extproc: custom processors per path and serve /v1/models

Refactors the server processing to allow registering custom Processors for different request paths,
and adds a custom processor for requests to `/v1/models` that returns an immediate response based
on the models that are configured in the filter configuration.

Related discussion: https://github.com/envoyproxy/ai-gateway/issues/186

~~Opening as a draft to discuss on the proposed approach.~~

In order to return a direct response for some paths, we need to be able to customize the behaviour. Instead of messing
up with the translator, this PR proposes a change to refactor the Server and allow it to define custom processors per request path. This would make it easier to extend the functionality of the filter in the future, when more endpoints are supported.

This is used in the current PR to serve the `/v1/models` endpoint and return a computed list of models based on the configured routes. Right now we take advantage of exact header matching being the only option, but should be good enough for a first implementation.

~~The current PR doesn't add a feature flag to enable/disable this endpoint handling, but it should be easy to add if we feel the approach proposed in this PR is good.~~


cmd/extproc/mainlib/main.go
```diff
@@ -89,10 +89,12 @@ func Main() {
 		log.Fatalf("failed to listen: %v", err)
 	}
 
-	server, err := extproc.NewServer[*extproc.Processor](l, extproc.NewProcessor)
+	server, err := extproc.NewServer(l)
```
nacx, line 92: The generics were tying the server to a concrete processor type, but now it allows defining custom processors per request path, so the generics made no sense anymore. I've removed them and kept the code using the `ProcessorIface`

internal/extproc/chatcompletion_processor.go
```diff
@@ -0,0 +1,238 @@
...
+	extprocv3 "github.com/envoyproxy/go-control-plane/envoy/service/ext_proc/v3"
+	"google.golang.org/protobuf/types/known/structpb"
+
+	"github.com/envoyproxy/ai-gateway/filterapi"
+	"github.com/envoyproxy/ai-gateway/internal/extproc/translator"
+	"github.com/envoyproxy/ai-gateway/internal/llmcostcel"
+)
+
+// NewChatCompletionProcessor creates a new processor for the chat completino requests
+func NewChatCompletionProcessor(config *processorConfig, logger *slog.Logger) ProcessorIface {
```
nacx, line 22: This is the `Processor` object that I've renamed to ChatCompletionProcessor as now we can have different processors per request path. The contents are just a verbatim copy (modulo the type name change).

internal/extproc/server.go
```diff
@@ -125,6 +160,17 @@ func (s *Server[P]) process(p P, stream extprocv3.ExternalProcessor_ProcessServe
 			return status.Errorf(codes.Unknown, "cannot receive stream request: %v", err)
 		}
 
+		// If we're processing the request headers, read the :path header to instantiate the
+		// right processor.
+		if headers := req.GetRequestHeaders().GetHeaders(); headers != nil {
```
nacx, line 165: This is the main change in this PR.

In order to allow having custom processors per-request path we need to access the `:path` header. We can't instantiate the `ProcessorIface` object until we've read the first request headers message. This is why I instantiate the processor here, and reuse it for the following messages.

mathetake (commented):
I think I like the general direction! One question: then we don't need translator.Factory but instead maybe need concrete translator.NewChatCompletionTranslator, etc since it's currently translator.Factory takes path as an input?

nacx:
I think that's correct. We don't need the factory as it is today, but we still need to let the processor decide what translator to use, depending on the selected backend. I can refactor and clean that up in a follow-up PR if that makes sense?

mathetake:
yeah let's do a follow up refactor - that makes sense

internal/extproc/processor.go
```diff
@@ -33,13 +26,18 @@ type processorConfig struct {
...
 }
 
+// processorConfigRequestCost is the configuration for the request cost.
 type processorConfigRequestCost struct {
 	*filterapi.LLMRequestCost
 	celProg cel.Program
 }
 
+// ProcessorFactory is the factory function used to create new instances of a processor.
+type ProcessorFactory func(*processorConfig, *slog.Logger) ProcessorIface
```
nacx, line 39: The processor config is private and in the `internal` package.
Following the idea of the `Router` in the `x` experimental API package, would it make sense (not for this PR), to move the processor interfaces there too, to allow to easily extend the processors and let users build an extproc registering their own ones?

cmd/extproc/mainlib/main.go
```diff
@@ -21,59 +22,88 @@ import (
...
 		"logLevel",
-		"info", "log level for the external processor. One of 'debug', 'info', 'warn', or 'error'.",
+		"info",
+		"log level for the external processor. One of 'debug', 'info', 'warn', or 'error'.",
+	)
+	processorsPtr := fs.String(
+		"processors",
+		"chatcompletions,models",
+		"comma-separated list of processors to enable. Available processors: chatcompletions, models",
 	)
```
mathetake, line 70: instead of exposing this in this PR, can we just use the static list for now until someone requests for exposing this as a flag? I don't think anyone wants to expose so shall we start small for now?

internal/extproc/processor.go
mathetake: as per my comment above let's wait for someone to request for this being exposed for now...

cmd/extproc/mainlib/main.go
nacx: Done. I've removed the flag.

internal/extproc/processor.go
nacx: sounds good!

cmd/extproc/mainlib/main.go
```diff
@@ -21,59 +22,66 @@ import (
...
 
-// parseAndValidateFlags parses and validates the flags passed to the external processor.
-func parseAndValidateFlags(args []string) (configPath, addr string, logLevel slog.Level, err error) {
-	fs := flag.NewFlagSet("AI Gateway External Processor", flag.ContinueOnError)
-	configPathPtr := fs.String(
+// extProcFlags is the struct that holds the flags passed to the external processor.
+type extProcFlags struct {
+	configPath  string     // path to the configuration file
+	extProcAddr string     // gRPC address for the external processor
+	logLevel    slog.Level // log level for the external processor
```
mathetake, line 29:
```suggestion
	configPath  string     // path to the configuration file.
	extProcAddr string     // gRPC address for the external processor.
	logLevel    slog.Level // log level for the external processor.
```

internal/apischema/openai/openai.go
```diff
@@ -723,3 +725,41 @@ type ErrorType struct {
...
+	// Object is the object type, which is always "model".
+	Object string `json:"object"`
+	// OwnedBy is the organization that owns the model.
+	OwnedBy string `json:"owned_by"`
+}
+
+// JSONUNIXTime is a helper type to marshal/unmarshal time.Time UNIX timestamps.
+type JSONUNIXTime time.Time
+
+func (t JSONUNIXTime) MarshalJSON() ([]byte, error) {
```
mathetake, line 754:
```suggestion

// MarshalJSON implements [json.Marshaler].
func (t JSONUNIXTime) MarshalJSON() ([]byte, error) {
```

internal/apischema/openai/openai.go
```diff
@@ -723,3 +725,41 @@ type ErrorType struct {
...
+	return []byte(strconv.FormatInt(time.Time(t).Unix(), 10)), nil
+}
+
+func (t *JSONUNIXTime) UnmarshalJSON(s []byte) error {
```
mathetake, line 758:
```suggestion
// UnmarshalJSON implements [json.Unmarshaler].
func (t *JSONUNIXTime) UnmarshalJSON(s []byte) error {
```

internal/apischema/openai/openai_test.go
```diff
@@ -234,3 +235,30 @@ func TestOpenAIChatCompletionMessageUnmarshal(t *testing.T) {
...
+	require.JSONEq(t, raw, string(b))
+
+	var out ModelList
+	require.NoError(t, json.Unmarshal([]byte(raw), &out))
+	require.Len(t, out.Data, 1)
+	require.Equal(t, "list", out.Object)
+	require.Equal(t, model.ID, out.Data[0].ID)
+	require.Equal(t, model.Object, out.Data[0].Object)
+	require.Equal(t, model.OwnedBy, out.Data[0].OwnedBy)
+	// Unmarshalling initializes other fields in time.Time we're not interested with. Just compare the actual time
```
mathetake, line 262:
```suggestion
	// Unmarshalling initializes other fields in time.Time we're not interested with. Just compare the actual time.
```

internal/extproc/chatcompletion_processor.go
```diff
@@ -0,0 +1,238 @@
...
+	corev3 "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
+	extprocv3 "github.com/envoyproxy/go-control-plane/envoy/service/ext_proc/v3"
+	"google.golang.org/protobuf/types/known/structpb"
+
+	"github.com/envoyproxy/ai-gateway/filterapi"
+	"github.com/envoyproxy/ai-gateway/internal/extproc/translator"
+	"github.com/envoyproxy/ai-gateway/internal/llmcostcel"
+)
+
+// NewChatCompletionProcessor creates a new processor for the chat completino requests
```
mathetake, line 21:
```suggestion
// NewChatCompletionProcessor implements [ProcessorIface] for the /chat/completions endpoint.
```

internal/extproc/models_processor.go
```diff
@@ -0,0 +1,97 @@
...
+
+	corev3 "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
+	extprocv3 "github.com/envoyproxy/go-control-plane/envoy/service/ext_proc/v3"
+	typev3 "github.com/envoyproxy/go-control-plane/envoy/type/v3"
+	"google.golang.org/grpc/codes"
+
+	"github.com/envoyproxy/ai-gateway/internal/apischema/openai"
+)
+
+// ModelsProcessor handles the processing of requests to the `/v1/models` endpoint.
```
mathetake, line 19:
```suggestion
// ModelsProcessor implements [ProcessorIface] for the `/v1/models` endpoint.
```

internal/extproc/models_processor.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+// ProcessRequestHeaders implements [ProcessorIface.ProcessRequestHeaders].
+func (m *ModelsProcessor) ProcessRequestHeaders(_ context.Context, _ *corev3.HeaderMap) (*extprocv3.ProcessingResponse, error) {
+	m.logger.Info("Serving list of declared models")
+
+	body, err := json.Marshal(m.models)
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal body: %w", err)
+	}
```
mathetake, line 55: nit: i think we could marshal during the config setup and simply have []byte in the ModelsProcessor struct vs create the byte slice each time?

internal/extproc/server.go
```diff
@@ -251,3 +298,17 @@ func filterSensitiveBody(resp *extprocv3.ProcessingResponse, logger *slog.Logger
 	}
 	return filteredResp
 }
+
+func getHeader(headers *corev3.HeaderMap, name string) string {
```
mathetake, line 302: nit: so basically i see we (unnecessarily) iterate through all headers here and headersToMap. would it be possible to use headersToMap before the processor selection and pass the constructed map to `processorForPath ` and eventually to `newProcessor`?

internal/extproc/models_processor.go
```diff
@@ -0,0 +1,97 @@
...
+
+	"github.com/envoyproxy/ai-gateway/internal/apischema/openai"
+)
+
+// ModelsProcessor handles the processing of requests to the `/v1/models` endpoint.
+// This processor returns an immediate response with the list of models that are declared in the filter
+// configuration.
+// Since it returns an immediate response after processing the headers, the rest of the methods of the
+// ProcessorIface are not implemented. Those should never be called.
+type ModelsProcessor struct {
```
mathetake, line 24: nit does this need to be exported

internal/extproc/models_processor.go
nacx: I though about that, but we're instantiating the processor once per-request, and this will be executed only once per request anyway.
If we want to save computation probably we should do this in the LoadConfig method, after loading the config, so that's available when the processors are instantiated. I didn't want to to this, though, because it couples some concrete processor behaviour (the models processor) to the Server's load config.

I also think this is a negligible overhead when returning the immediate response.

WDYT?

internal/extproc/server.go
nacx: I thought about this, but was unsure about modifying the processor factory methods to accept a request headers map. I can change that if you feel it is OK

internal/extproc/models_processor.go
mathetake: yeah let's leave it as is

nacx:
OK, I've addressed all the review comments and rebased to the latest `main`. I squashed the commits because otherwise there were too many conflicts to resolve. I hope it doesn't complicate much the final review.

internal/extproc/chatcompletion_processor.go
```diff
@@ -0,0 +1,226 @@
...
+			},
+		},
+		ModeOverride: override,
+	}
+	return resp, nil
+}
+
+// ProcessResponseHeaders implements [ProcessorIface.ProcessResponseHeaders].
+func (p *chatCompletionProcessor) ProcessResponseHeaders(_ context.Context, headers *corev3.HeaderMap) (res *extprocv3.ProcessingResponse, err error) {
+	p.responseHeaders = headersToMap(headers)
```
mathetake, line 114: maybe make `[ProcessorIface.ProcessResponseHeaders]` accept map[string]string ?

internal/extproc/server.go
```diff
@@ -125,6 +162,16 @@ func (s *Server[P]) process(p P, stream extprocv3.ExternalProcessor_ProcessServe
 			return status.Errorf(codes.Unknown, "cannot receive stream request: %v", err)
 		}
 
+		// If we're processing the request headers, read the :path header to instantiate the
+		// right processor.
+		if headers := req.GetRequestHeaders().GetHeaders(); headers != nil {
+			p, err = s.processorForPath(headersToMap(headers))
```
mathetake, line 168: I think this will create a new `p` each time the grpc stream receives a new message? maybe this block should be conditioned on `if p != nil { ... }` ?

mathetake: ah `headers := req.GetRequestHeaders().GetHeaders(); headers != nil` implicitly does that - maybe I would like a comment at least on this

nacx: The `req.GetRequestHeaders()` will only be non-nil when it is a RequestHeaders message, so I think this will be executed only once?
```go
func (x *ProcessingRequest) GetRequestHeaders() *HttpHeaders {
	if x, ok := x.GetRequest().(*ProcessingRequest_RequestHeaders); ok {
		return x.RequestHeaders
	}
	return nil
}
```

nacx: I'll add a comment :)

nacx: Done

mathetake (approved):
LGTM!

PR #334, opened 2025-02-12
Remove the path from the translator factory

nacx:
**Commit Message**

Remove the path from the translator factory

Removes the path from the translator factory, now that there is a dedicated processor for the chat completion endpoint.

Follow-up for: https://github.com/envoyproxy/ai-gateway/pull/325#pullrequestreview-2609658329

Note that I don't remove the `Factory` type completely so that only the right translator is instantiated and only when needed.


internal/extproc/translator/translator.go
```diff
@@ -28,11 +28,8 @@ func isGoodStatusCode(code int) bool {
...
-// Factory creates a [Translator] for the given API schema combination and request path.
-//
-//   - `path`: the path of the request.
-//   - `l`: the logger.
-type Factory func(path string) (Translator, error)
+// Factory creates a [Translator] for the given API schema combination.
+type Factory func() Translator
 
 // NewFactory returns a callback function that creates a translator for the given API schema combination.
 func NewFactory(in, out filterapi.VersionedAPISchema) (Factory, error) {
```
mathetake, line 35: should we now rename this to `NewChatCompletionTranslatorFactory` ?

nacx: I've completely removed the Factory now, and moved the translator instantiation to the chat completion processor, as it does only make sense for that processor.

WDYT?

mathetake:
coverage!

internal/extproc/translator/translator.go
mathetake: sg

internal/extproc/translator/openai_openai.go
```diff
@@ -14,12 +14,9 @@ import (
...
 )
 
-// newOpenAIToOpenAITranslator implements [Factory] for OpenAI to OpenAI translation.
-func newOpenAIToOpenAITranslator(path string) (Translator, error) {
-	if path == "/v1/chat/completions" {
-		return &openAIToOpenAITranslatorV1ChatCompletion{}, nil
-	}
-	return nil, fmt.Errorf("unsupported path: %s", path)
+// NewOpenAIToOpenAITranslator implements [Factory] for OpenAI to OpenAI translation.
+func NewOpenAIToOpenAITranslator() Translator {
```
mathetake, line 18: let's call it more precisely 
```suggestion
func NewChatCompletionOpenAIToOpenAITranslator() Translator {
```

nacx: Done

internal/extproc/translator/openai_awsbedrock.go
```diff
@@ -21,12 +21,9 @@ import (
...
 )
 
-// newOpenAIToAWSBedrockTranslator implements [Factory] for OpenAI to AWS Bedrock translation.
-func newOpenAIToAWSBedrockTranslator(path string) (Translator, error) {
-	if path == "/v1/chat/completions" {
-		return &openAIToAWSBedrockTranslatorV1ChatCompletion{}, nil
-	}
-	return nil, fmt.Errorf("unsupported path: %s", path)
+// NewOpenAIToAWSBedrockTranslator implements [Factory] for OpenAI to AWS Bedrock translation.
+func NewOpenAIToAWSBedrockTranslator() Translator {
```
mathetake, line 25: here as well

internal/extproc/chatcompletion_processor.go
```diff
@@ -38,42 +38,52 @@ type chatCompletionProcessor struct {
...
 
-	factory, ok := p.config.factories[b.Schema]
-	if !ok {
-		return nil, fmt.Errorf("failed to find factory for output schema %q", b.Schema)
-	}
-
-	t, err := factory(path)
-	if err != nil {
-		return nil, fmt.Errorf("failed to create translator: %w", err)
+	if err = c.selectTranslator(b.Schema); err != nil {
```
mathetake, line 82: i feel like this could be called when calling NewProcessor instead of inside ProcessRequestBody? Then we won't need `.translator != nil` inside selectTranslator i guess?

mathetake (approved):
Thank you for a nice clean up!

PR #337, opened 2025-02-13
chore: apply license headers to all source files

nacx:
**Commit Message**

chore: apply license headers to all source files

Apply license headers to all source files, and add a check in the `format` make target to make sure they're not missed.

N/A

N/A


mathetake:
hah you can ignore the terraform stuff or ignore .github/** directory maybe

manifests/charts/ai-gateway-helm/templates/serviceaccount.yaml
```diff
@@ -1,4 +1,9 @@
-{{- if .Values.controller.serviceAccount.create -}}
+# Copyright Envoy AI Gateway Authors
+# SPDX-License-Identifier: Apache-2.0
+# The full text of the Apache license is available in the LICENSE file at
+# the root of the repo.
+
+{{ if .Values.controller.serviceAccount.create -}}
```
nacx, line 6: Note that I've removed the leading `-` in the conditional; otherwise the next line was appended directly to the comment instead of having a newline

mathetake (approved):

PR #341, opened 2025-02-13
test: use envoy docker to ensure it's runnable

mathetake:
**Commit Message**


mathetake:
i think i will give up for now

PR #363, opened 2025-02-18
docs: update AWS Bedrock next steps

davidxia:
**Commit Message**

to link to next page for "Connect Local Model"
instead of linking to previous page "Connect OpenAI."


missBerg (approved):
Thanks!

PR #376, opened 2025-02-19
extproc: load default config if file does not exist

nacx:
**Commit Message**

extproc: load default config if file does not exist

If the configuration file does not exist, let extproc run with the default configuration. This prevents the extproc process from terminating if the config does not exist, which is a valid use case when the extproc filter is used in conjunction with other control planes.

N/A

N/A

### DIFF
--- a/internal/extproc/watcher.go
+++ b/internal/extproc/watcher.go
@@ -13,8 +13,6 @@ import (
 	"strings"
 	"time"
 
-	"k8s.io/apimachinery/pkg/util/yaml"
-
 	"github.com/envoyproxy/ai-gateway/filterapi"
 )
 
@@ -76,7 +74,7 @@ func (cw *configWatcher) loadConfig(ctx context.Context) error {
 	if err != nil && os.IsNotExist(err) {
 		// If the file does not exist, do not fail (which could lead to the extproc process to terminate)
 		// Instead, load the default configuration and keep running unconfigured
-		cfg, raw, err = loadDefaultConfig()
+		cfg, raw = filterapi.MustLoadDefaultConfig()
 	}
 	// If reading the file fails, or loading the default config fails, abort
 	if err != nil {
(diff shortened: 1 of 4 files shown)


internal/extproc/watcher_test.go
```diff
@@ -40,9 +42,30 @@ func (m *mockReceiver) getConfig() *filterapi.Config {
...
 }
 
+var _ io.Writer = (*syncBuffer)(nil)
+
+// syncBuffer is a bytes.Buffer that is safe for concurrent read/write access.
+// used just in the tests to safely read the logs in assertions without data races.
+type syncBuffer struct {
+	mu sync.RWMutex
+	b  *bytes.Buffer
+}
```
nacx, line 52: I'm not sure why this didn't fail before, but when running the unit tests I got data races in the buffer, between the message logs and the call to `buf.String()`. This should fix that .

internal/extproc/watcher_test.go
```diff
@@ -54,7 +77,24 @@ func TestStartConfigWatcher(t *testing.T) {
 	path := tmpdir + "/config.yaml"
 	rcv := &mockReceiver{}
 
-	require.NoError(t, os.WriteFile(path, []byte{}, 0o600))
+	ctx, cancel := context.WithCancel(t.Context())
+	defer cancel()
```
mathetake, line 81: i think this cancel is unnecessary anymore so can you remove it 🙏

internal/extproc/watcher.go
```diff
@@ -126,3 +131,12 @@ func (cw *configWatcher) diff(oldConfig, newConfig string) {
...
 }
+
+// loadDefaultConfig loads the default configuration when the given config file path is not found.
+func loadDefaultConfig() (*filterapi.Config, []byte, error) {
+	var cfg filterapi.Config
+	if err := yaml.Unmarshal([]byte(filterapi.DefaultConfig), &cfg); err != nil {
+		return nil, nil, err
+	}
+	return &cfg, []byte(filterapi.DefaultConfig), nil
+}
```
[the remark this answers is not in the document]
nacx: Done

internal/extproc/watcher_test.go
nacx: Done!

internal/extproc/watcher.go
```diff
@@ -65,45 +65,48 @@ func (cw *configWatcher) watch(ctx context.Context, tick time.Duration) {
...
+	var (
+		cfg *filterapi.Config
+		raw []byte
+	)
+
 	stat, err := os.Stat(cw.path)
+	if err != nil && os.IsNotExist(err) {
+		// If the file does not exist, do not fail (which could lead to the extproc process to terminate)
+		// Instead, load the default configuration and keep running unconfigured
+		cfg, raw = filterapi.MustLoadDefaultConfig()
```
mathetake, line 77: need to clear `err = nil`  now? not exist will end up captured in if err ! = nil below

nacx: yes. when this method returned an error it made sense; not anymore. fixed

mathetake (approved):
Fantastic!

PR #377, opened 2025-02-19
extproc: do not modify original headers when debug enabled

mathetake:
**Commit Message**
 When debug level enabled, extproc logs the headers on request paths, and it reduces the sensitive value, especially authorization header values as a security measure. However, previously it modifies the original object accidentally hence it had caused a hard-to-debug unexpected behavior. This fixes the bug and ensures that the reduction is not applied to the original objects.

Thanks for the report by Bloomberg team @yuzisun 

Fixes a bug in #217


mathetake:
@nacx PTAL

PR #380, opened 2025-02-19
controller: fixes extproc config update on removal

mathetake:
**Commit Message**

Previously, the deployment config is populated with the values specified in the AIGatewayRoute resource and not removed if the values is not specified. For example, after a replica value is specified, if a user removes it from AIGatewayFilterConfig, then the replica is not reverted to a default value. This fixes applyExtProcDeploymentConfigUpdate so that it will remove the previously specified value.

Thanks for pointing out @sanposhiho !


mathetake:
PTAL somebody!

aabchoo (approved):

PR #375, opened 2025-02-19
test: error message from upstream provider is passed thru

davidxia:
**Commit Message**

TODO

PR #389, opened 2025-02-20
docs: removes --fail flags

mathetake:
**Commit Message**

At the first iteration of the "getting started guide", I added `--fail` in all commands as we had an automated testing of the site documentation and it allowed us to detect if the curl command not is working. However, now we don't have any tests like that + `--fail` prevents users from seeing the actual error response from the upstream which makes it difficult to troubleshoot since it suppresses the response body due to the early cancel of requests. 

The root cause of #379 #375


mathetake:
cc @davidxia

aabchoo (approved):

PR #390, opened 2025-02-20
ci: enable tests on release/** branches

mathetake:
**Commit Message**

This is enabling the tests workflow on release/** branches for the future backports etc.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -2,7 +2,11 @@ name: Tests
 on:
   pull_request:
     branches:
-      - main
+      - 'main'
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yaml
```diff
@@ -2,15 +2,20 @@ name: Tests
 on:
   pull_request:
     branches:
-      - main
+      - 'main'
```
aabchoo, line 5: Do we have branch protection rules to prevent branches named `release/**` from being pushed?

mathetake: that's exactly what i am looking at now :) i think once the branch is created we can enforce the rule

PR #392, opened 2025-02-20
extproc: remove noise when reloading config

nacx:
**Commit Message**

extproc: remove noise when reloading config

Reduce the noise when configuration is reloaded. Only print logs when the configuration actually changes and do not reload the default config on every tick.

Related to https://github.com/envoyproxy/ai-gateway/pull/376

N/A

### DIFF
--- a/internal/extproc/watcher.go
+++ b/internal/extproc/watcher.go
@@ -24,11 +24,12 @@ type ConfigReceiver interface {
 }
 
 type configWatcher struct {
-	lastMod time.Time
-	path    string
-	rcv     ConfigReceiver
-	l       *slog.Logger
-	current string
+	lastMod         time.Time
+	path            string
+	rcv             ConfigReceiver
+	l               *slog.Logger
+	current         string
+	usingDefaultCfg bool
 }
 
 // StartConfigWatcher starts a watcher for the given path and Receiver.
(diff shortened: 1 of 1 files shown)


internal/extproc/watcher.go
```diff
@@ -81,13 +82,18 @@ func (cw *configWatcher) loadConfig(ctx context.Context) error {
 	}
 
 	if cfg != nil {
+		if cw.usingDefaultCfg { // Do not re-reload the same thing on every tick
```
mathetake, line 85:
```suggestion
		if cw.usingDefaultCfg { // Do not re-reload the same thing on every tick.
```

mathetake (commented):
test?

internal/extproc/watcher_test.go
```diff
@@ -151,18 +158,22 @@ rules:
...
-	// Verify the buffer contains the config line changed
+	// Verify the buffer contains the config line changed.
 	require.Eventually(t, func() bool {
 		return strings.Contains(buf.String(), "config line changed")
-	}, 1*time.Second, 100*time.Millisecond, buf.String())
+	}, 1*time.Second, tickInterval, buf.String())
+
+	// Wait for a couple ticks to verify config is not reloaded if file does not change.
+	time.Sleep(2 * tickInterval)
+	require.Equal(t, 3, rcv.loadCount)
```
mathetake, line 176: can you use require.Eventual like the above?

mathetake (approved):

PR #396, opened 2025-02-20
fix: return HTTP 404 when route not found

davidxia:


internal/extproc/chatcompletion_processor.go
```diff
@@ -84,6 +84,25 @@ func (c *chatCompletionProcessor) ProcessRequestBody(ctx context.Context, rawBod
 	c.requestHeaders[c.config.modelNameHeaderKey] = model
 	b, err := c.config.router.Calculate(c.requestHeaders)
 	if err != nil {
+		// if err.Error() == "no matching rule found" {
```
nacx, line 87: Instead of doing this and looking the error tree like you do, let's create a proper `ErrRouteNotFound` in the `filterapi/x` package. This was all implementations of Calculate can return that error consistently.
Then you can modify the default Calculate method to return that error, and here can cleanly just use `errors.Is` and return the immediate response directly.

internal/extproc/server.go
```diff
@@ -168,6 +170,25 @@ func (s *Server) Process(stream extprocv3.ExternalProcessor_ProcessServer) error
 
 		resp, err := s.processMsg(ctx, p, req)
 		if err != nil {
+			s.logger.Info("AAAAA")
```
nacx, line 173: The fact that routes are calculated is an implementation detail of the chat completion processor, because it's based on the request body for requests to that endpoint. Therefore, I would not have this logic here as it couples the `server` to semantics of a particular processor.

Let's keep this  unchanged and just have the chat completion processor return the immediate response.

internal/extproc/server.go
```diff
@@ -168,6 +170,25 @@ func (s *Server) Process(stream extprocv3.ExternalProcessor_ProcessServer) error
...
+			if checkErrorTreeForMessage(err, "no matching rule found") {
+				s.logger.Info("BBBBBB")
+				// return status.Errorf(codes.NotFound, err.Error())
+				resp := &extprocv3.ImmediateResponse{
+					Status: &v32.HttpStatus{
+						Code: http.StatusNotFound,
+					},
+				}
+
+				if err := stream.Send(resp); err != nil {
```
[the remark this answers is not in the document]
nacx: In the `model_processor.go` you have an example of how to return an immediate response that you can use in the chat completion procesor.

davidxia (commented):
@nacx thanks for the very helpful guidance! Closing and will open another PR with your suggestions.

davidxia:
new one https://github.com/envoyproxy/ai-gateway/pull/399

PR #397, opened 2025-02-21
bugfix: aws oidc secret creation and update process

xiaolin593:
**Commit Message**

This commit fixes the bug when creating secret for AWS OIDC rotation and streamlines secret upsert operation into Kubernetes.

**Detailed explanation**:
In aws_oidc_rotator.go [line 162 to 168](https://github.com/envoyproxy/ai-gateway/blob/1192716fce86c4a66e275884bbf3d528d5f99ac2/internal/controller/rotators/aws_oidc_rotator.go#L162-L168), existing buggy code snippet:
```
err = r.client.Create(ctx, secret)
if err != nil {
    if !apierrors.IsAlreadyExists(err) { 
	    return r.client.Update(ctx, secret)
    }
    return fmt.Errorf("failed to create secret: %w", err)
}
```

The intention of above code is to create secret regardless of its existence, then check if k8s returns `AlreadyExist` error. 

However, k8s returns `BadRequest` error with message `resourceVersion can not be set for Create requests` if debug through and inspect err at [line 164](https://github.com/envoyproxy/ai-gateway/blob/1192716fce86c4a66e275884bbf3d528d5f99ac2/internal/controller/rotators/aws_oidc_rotator.go#L164)

Discussed with original commit author and fixed the bug by changing secret upsert process with a control flag to improve readability instead of a quick fix to switch condition from `errors.IsAlreadyExists` to `errors.IsBadRequest`. 

Related PR: #306


mathetake:
Can you write a test?

internal/controller/rotators/aws_oidc_rotator.go
```diff
@@ -159,15 +163,21 @@ func (r *AWSOIDCRotator) Rotate(ctx context.Context, token string) error {
...
-	}
+	if secretFound {
+		r.logger.Info(fmt.Sprintf("updating aws secret %s for backend security policy namespace: %s, name: %s",
+			secret.Name,
+			r.backendSecurityPolicyNamespace,
+			r.backendSecurityPolicyName))
+
+		return r.client.Update(ctx, secret)
+	} else {
+		r.logger.Info(fmt.Sprintf("creating aws secret %s for backend security policy namespace: %s, name: %s",
```
aabchoo, line 174: Could you change the log to specify `new aws credential secret`?

internal/controller/rotators/aws_oidc_rotator.go
```diff
@@ -159,15 +163,21 @@ func (r *AWSOIDCRotator) Rotate(ctx context.Context, token string) error {
...
 
-	err = r.client.Create(ctx, secret)
-	if err != nil {
-		if !apierrors.IsAlreadyExists(err) {
-			return r.client.Update(ctx, secret)
-		}
-		return fmt.Errorf("failed to create secret: %w", err)
-	}
+	if secretFound {
+		r.logger.Info(fmt.Sprintf("updating aws secret %s for backend security policy namespace: %s, name: %s",
```
aabchoo, line 167: similar to below

internal/controller/rotators/aws_oidc_rotator.go
```diff
@@ -130,11 +130,15 @@ func (r *AWSOIDCRotator) Rotate(ctx context.Context, token string) error {
...
 	}
 
+	secretFound := true
 	secret, err := LookupSecret(ctx, r.client, r.backendSecurityPolicyNamespace, GetBSPSecretName(r.backendSecurityPolicyName))
+
 	if err != nil {
 		if !apierrors.IsNotFound(err) {
 			return err
 		}
+
```
aabchoo, line 140: Consider adding a log here?

aabchoo (commented):
Thank you for putting up a fix for this bug! Just minor comments added 🙇

internal/controller/rotators/aws_oidc_rotator.go
yuzisun: How about rewrite the code to

if err == nil {
   updateSecret
} else if apierrors.IsNotFound(err) {
   createSecret
}

internal/controller/rotators/aws_oidc_rotator.go
```diff
@@ -116,58 +116,53 @@ func (r *AWSOIDCRotator) GetPreRotationTime(ctx context.Context) (time.Time, err
...
 	}
 
-	secret, err := LookupSecret(ctx, r.client, r.backendSecurityPolicyNamespace, GetBSPSecretName(r.backendSecurityPolicyName))
-	if err != nil {
-		if !apierrors.IsNotFound(err) {
-			return err
-		}
+	secret, err := LookupSecret(ctx, r.client, bspNamespace, secretName)
+
+	if err != nil && apierrors.IsNotFound(err) {
```
yuzisun, line 150: Need to cover the case for errors other than `IsNotFound` which we can simply return the error.

internal/controller/backend_security_policy_test.go
mathetake, line 1: i would like a test to TestAWS_OIDCRotator instead of here. if you cannot and have to do it here, then the abstraction seems wrong

xiaolin593: Sure, I'll add test cases in `aws_oidc_rotator_test.go`. Would it make sense keep updated test case in `backend_security_policy_test.go` in-place  or you prefer to rip it off?

mathetake: yeah you can keep it! thanks for enhancing the existing one anyways 👍

internal/controller/rotators/aws_oidc_rotator.go
```diff
@@ -116,58 +116,56 @@ func (r *AWSOIDCRotator) GetPreRotationTime(ctx context.Context) (time.Time, err
...
+					Namespace: bspNamespace,
+				},
+				Type: corev1.SecretTypeOpaque,
+				Data: make(map[string][]byte),
+			}
+			populateSecretWithAwsIdentity(secret, awsIdentity, r.region)
+			return r.client.Create(ctx, secret)
 		}
-		return fmt.Errorf("failed to create secret: %w", err)
+		return err
```
aabchoo, line 164: please add a log line to notify users that lookup secret failed (keep it consistent with line 145)

internal/controller/rotators/aws_oidc_rotator.go
```diff
@@ -116,58 +116,56 @@ func (r *AWSOIDCRotator) GetPreRotationTime(ctx context.Context) (time.Time, err
...
-	}}
-
-	updateAWSCredentialsInSecret(secret, &credsFile)
-
-	err = r.client.Create(ctx, secret)
+	secret, err := LookupSecret(ctx, r.client, bspNamespace, secretName)
 	if err != nil {
-		if !apierrors.IsAlreadyExists(err) {
-			return r.client.Update(ctx, secret)
+		if apierrors.IsNotFound(err) {
```
aabchoo, line 151: Missing call to `updateAWSCredentialsInSecret` / logic to add expiration time

aabchoo (changes requested):

internal/controller/rotators/aws_oidc_rotator.go
xiaolin593: `updateAWSCredentialsInSecret` and `updateExpirationSecretAnnotation` moved into `populateSecretWithAwsIdentity` function which will be called after secret creation/update.

aabchoo: My bad

yuzisun (approved):

aabchoo (approved):
thanks for adding this change

PR #398, opened 2025-02-21
test: add TestBackendSecurityPolicyController

aabchoo:
**Commit Message**

Adding test TestBackendSecurityPolicyController and exporting BackendSecurityPolicy + NewBackendSecurityPolicy

### DIFF
--- a/tests/controller/controller_test.go
+++ b/tests/controller/controller_test.go
@@ -490,7 +490,135 @@ func TestAIGatewayRouteController(t *testing.T) {
 	})
 }
 
-func TestBackendSecurityPolicyController(t *testing.T) { t.Skip("TODO") }
+func TestBackendSecurityPolicyController(t *testing.T) {
+	c, cfg, k := testsinternal.NewEnvTest(t)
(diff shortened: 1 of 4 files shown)


tests/controller/controller_test.go
```diff
@@ -490,7 +490,135 @@ func TestAIGatewayRouteController(t *testing.T) {
...
+	require.NoError(t, c.Create(t.Context(), &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "secret",
+			Namespace: backendSecurityPolicyNamespace,
+		},
+		Immutable: nil,
+		Data: map[string][]byte{
+			"apiKey": []byte("apiKey"),
+		},
+	}))
```
mathetake, line 522: i don't think you would need to create secret? 
```suggestion
```

aabchoo: Hmm good catch 😄 we won't need it for this workflow

mathetake (approved):

PR #399, opened 2025-02-21
extproc: return HTTP 404 when no matching route

davidxia:
**Commit Message**

extproc: return HTTP 404 when no matching route
in `AIGatewayRoute` instead of a 500 status code.

## Before

```console
$ curl \
    -d '{
    "model": "DEADBEEF",
    "messages": [
        {
            "role": "user",
            "content": "Who should be king of the world? Be direct, decisive, and opinionated."
        }
    ]
}' $GATEWAY_URL/v1/chat/completions -i
HTTP/1.1 500 Not Found
content-length: 22
content-type: text/plain
date: Fri, 21 Feb 2025 12:51:06 GMT
```

## After

```console
$ curl \
    -d '{
    "model": "DEADBEEF",
    "messages": [
        {
            "role": "user",
            "content": "Who should be king of the world? Be direct, decisive, and opinionated."
        }
    ]
}' $GATEWAY_URL/v1/chat/completions -i
HTTP/1.1 404 Not Found
content-length: 22
content-type: text/plain
date: Fri, 21 Feb 2025 12:51:06 GMT

no matching rule found
```

### DIFF
--- a/internal/extproc/router/router.go
+++ b/internal/extproc/router/router.go
@@ -15,6 +15,9 @@ import (
 	"github.com/envoyproxy/ai-gateway/filterapi/x"
 )
 
+// ErrNoMatchingRule is the error when no matching rule is found.
+var ErrNoMatchingRule = errors.New("no matching rule found")
+
 // router implements [x.Router].
 type router struct {
 	rules []filterapi.RouteRule
@@ -46,7 +49,7 @@ func (r *router) Calculate(headers map[string]string) (backend *filterapi.Backen
 		}
 	}
 	if rule == nil {
-		return nil, errors.New("no matching rule found")
+		return nil, ErrNoMatchingRule
 	}
 	return r.selectBackendFromRule(rule), nil
 }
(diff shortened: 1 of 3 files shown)


mathetake (approved):

mathetake:
💯💯💯💯💯

internal/extproc/router/router.go
```diff
@@ -15,6 +15,9 @@ import (
 	"github.com/envoyproxy/ai-gateway/filterapi/x"
 )
 
+// ErrNoMatchingRule is the error when no matching rule is found.
+var ErrNoMatchingRule = errors.New("no matching rule found")
```
nacx, line 19: I think this should be moved to the `filterapi/x` package (see my comment: https://github.com/envoyproxy/ai-gateway/pull/396#discussion_r1964826373).
Reason is that the concrete error is now part of the business logic of the chat completion processor and unless we move it there custom implementations of `Calculate` can't return it.

cc @mathetake

nacx: https://github.com/envoyproxy/ai-gateway/pull/401

Issue #53, opened 2024-12-27
Polish the routing source code


mathetake:
I added the **very** naive implementation in #50 

https://github.com/envoyproxy/ai-gateway/blob/a77bfc28a7c4400c02e8fcbe744222f5c83daf37/internal/extproc/router/router.go#L30-L69

which I think we should refactor before v0.1.0

theBeginner86:
Hi @mathetake,
May I help in improving this?

mathetake:
nah i am sorry but i would rather one of maintainers to take care of this, but thanks for the offer anyways

liangyuanpeng:
I'm click into here because the `help wanted` label.

> but i would rather one of maintainers to take care of this

And seems like this probably not need to add the `help wanted` label (:

mathetake:
fair enough - removed the label

wengyao04:
@mathetake does any maintainer takes care of this issue ? If not, I can help on it

mathetake:
Yeah go for it

mathetake:
ping @wengyao04 any updates?

wengyao04:
> ping [@wengyao04](https://github.com/wengyao04) any updates?

I am testing out the ai-gateway within BB for all the integration, since there are lots of change after we switch to open source version, we have to fix our components to fit this change. Sorry I don't have time this week, and pick it up next week.

PR #402, opened 2025-02-21
extproc: pick random backend when multiple unweighted ones

davidxia:
**Commit Message**

Right now if there are multiple, unweighted `backendRefs` in a [AIGatewayRouteRule], the first one is always returned. The correct behavior is to return a random one since they're essentially all weighted the same.

[AIGatewayRouteRule]: https://envoy-ai-gateway.netlify.app/docs/api/#aigatewayrouterule

contributes to #53

### DIFF
--- a/internal/extproc/router/router.go
+++ b/internal/extproc/router/router.go
@@ -57,9 +57,12 @@ func (r *router) selectBackendFromRule(rule *filterapi.RouteRule) (backend *filt
 	for _, b := range rule.Backends {
 		totalWeight += b.Weight
 	}
+
+	// Pick a random backend if none of them have a weight.
 	if totalWeight == 0 {
-		return &rule.Backends[0]
+		return &rule.Backends[r.rng.Intn(len(rule.Backends))]
 	}
+
 	selected := r.rng.Intn(totalWeight)
 	for i := range rule.Backends {
 		b := &rule.Backends[i]
--- a/internal/extproc/router/router_test.go
+++ b/internal/extproc/router/router_test.go
@@ -53,6 +53,15 @@ func TestRouter_Calculate(t *testing.T) {
 			},
 			{
 				Backends: []filterapi.Backend{
+					{Name: "baz", Schema: outSchema},
+					{Name: "qux", Schema: outSchema},
+				},
+				Headers: []filterapi.HeaderMatch{
+					{Name: "x-model-name", Value: "o1"},
+				},
+			},
+			{
+				Backends: []filterapi.Backend{
 					{Name: "openai", Schema: outSchema},
 				},
 				Headers: []filterapi.HeaderMatch{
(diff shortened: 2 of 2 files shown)


internal/extproc/router/router.go
```diff
@@ -33,13 +33,12 @@ func New(config *filterapi.Config, newCustomFn x.NewCustomRouterFn) (x.Router, e
 // Calculate implements [x.Router.Calculate].
 func (r *router) Calculate(headers map[string]string) (backend *filterapi.Backend, err error) {
 	var rule *filterapi.RouteRule
-	for i := range r.rules {
-		_rule := &r.rules[i]
+	for _, _rule := range r.rules {
```
davidxia, line 36: This change saves us one line of getting the element by index.

internal/extproc/router/router.go
```diff
@@ -57,9 +56,12 @@ func (r *router) selectBackendFromRule(rule *filterapi.RouteRule) (backend *filt
 	for _, b := range rule.Backends {
 		totalWeight += b.Weight
 	}
+
+	// Pick a random backend if none of them have a weight.
 	if totalWeight == 0 {
-		return &rule.Backends[0]
+		return &rule.Backends[r.rng.Intn(len(rule.Backends))]
```
davidxia, line 62: the important change

mathetake:
FWIW at the control plane API level, the "unweighted" backend should result in having 1 as a weight (same as the k8s GW API) https://github.com/envoyproxy/ai-gateway/blob/05043a29a20573f6e364fe5faed0b321c4d99ab5/api/v1alpha1/api.go#L233

internal/extproc/router/router.go
mathetake: this `_rule := &r.rules[i]` was intentional to avoid copying the rule struct when ranging over

davidxia: ah I see. reverted

davidxia:
> FWIW at the control plane API level, the "unweighted" backend should result in having 1 as a weight (same as the k8s GW API)

I think [AIGatewayRouteRuleBackendRef.Weight's default of 1](https://github.com/envoyproxy/ai-gateway/blob/05043a29a20573f6e364fe5faed0b321c4d99ab5/api/v1alpha1/api.go#L229-L234) isn't carried over during the [conversion here](https://github.com/envoyproxy/ai-gateway/blob/05043a29a20573f6e364fe5faed0b321c4d99ab5/filterapi/filterconfig.go#L195) or maybe the default isn't set in this code path at all so it's just the zero-value of `int`. But I wanted to keep this change minimal. Lmk if we should do the defaulting to 1 in this PR.

mathetake:
it's carried over from here https://github.com/envoyproxy/ai-gateway/blob/05043a29a20573f6e364fe5faed0b321c4d99ab5/internal/controller/ai_gateway_route.go#L305 so as long as we use AI Gateway is used normally, it will always be set to either 1 or explicitly specified value.

However, my colleague @nacx is using extproc directly without going through the control plane API, so having the defaulting-to-1 logic in `filterapi.UnmarshalConfigYaml` also makes sense anyways as you suspect.


Anyways the current change makes sense on its own so i will approve the change

mathetake (approved):
Thank you!

PR #417, opened 2025-02-24
extproc: fixes race condition with RNG

mathetake:
**Commit Message**

Previously, router instance shared the rng among multiple goroutines which is not a thread safe. This fixes it by having a goroutine local RNG. 

Contributes to #53

### DIFF
--- a/internal/extproc/router/router_test.go
+++ b/internal/extproc/router/router_test.go
@@ -7,6 +7,7 @@ package router
 
 import (
 	"sync"
+	"sync/atomic"
 	"testing"
 
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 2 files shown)


aabchoo (approved):
lgtm!

internal/extproc/router/router_test.go
```diff
@@ -110,6 +112,32 @@ func TestRouter_Calculate(t *testing.T) {
...
+
+				if b.Name == "foo" {
+					foo.Add(1)
+				} else {
+					bar.Add(1)
+				}
+			}()
+		}
+		wg.Wait()
+		require.Greater(t, bar.Load(), foo.Load())
```
aabchoo, line 137: Should we have a case where bar.load() + foo.load() === 1000?

mathetake: It's implicitly ensured via `wg.Wait` but good point!

PR #421, opened 2025-02-25
site: add release 0.1

missBerg:
**Commit Message**

Blog announcing 0.1 Release, merge when 0.1 release is published.

PR #422, opened 2025-02-25
docs: pin yaml urls to tags

mathetake:
**Commit Message**

In versioned docs, the URL of a yaml must point to the tagged permalink.

Follow up on #421


aabchoo (approved):
awesome

PR #425, opened 2025-02-25
extproc: renames translator interface

mathetake:
**Commit Message**

After #325, the translator is not where the abstraction over :path is implemented, but it moved to the choice of processor. As a result, the translator interface has become effectively tied to a specific endpoint, notably /v1/chat/completion. This renames the `translator.Translator` interface accordingly not only to remove the unnecessary code path but also facilitates the implementation of additional features like metrics. 

Follow up on #325 and contribute to #316


mathetake:
FYI @nacx

aabchoo (approved):

PR #430, opened 2025-02-26
chore: check force push

### LINKED ISSUE #429: Add a job in pr_style_check.yaml to check force pushes
As per the guideline CONTRIBUTING.md, we require all PRs not to do force push following the same rule as envoyproxy/envoy.

I would like to have a GHA job to check in [pr_style_check.yaml](https://github.com/envoyproxy/ai-gateway/blob/main/.github/workflows/pr_style_check.yaml) if the PR received a force push. If it finds it, then it should put a automated comment that references CONTRIBUTING.md and ask the author not to do so.

daixiang0:
**Commit Message**

Add a check for force push in CI.

Fixes #429


daixiang0:
Example: https://github.com/daixiang0/ai-gateway/pull/2

mathetake (changes requested):
I would like to see three changes :

* Do not check when the PR is draft
* Do not fail the workflow even if the force push is detected.
* Instead, put a bot comment on the PR

.github/workflows/pr_style_check.yaml
```diff
@@ -11,6 +11,24 @@ permissions:
   contents: read
```
mathetake, line 11: to put a comment on the PR, the permission might need a change

daixiang0:
Example: https://github.com/daixiang0/ai-gateway/pull/3

Skip draft and WIP as well.

mathetake (commented):
thanks, looking good! one last request: would it be possible to delete the previous comment if exists?

daixiang0:
> thanks, looking good! one last request: would it be possible to delete the previous comment if exists?

Hard to do this :(

.github/workflows/force_push_check.yaml
```diff
@@ -0,0 +1,49 @@
+name: PR Force Push Check
+
+on:
+  pull_request_target:
+    types:
```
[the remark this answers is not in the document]
daixiang0: You are right, we can delete those.

daixiang0:
@mathetake please review again :)

Sorry for slow response due to busy job.

.github/workflows/force_push_check.yaml
```diff
@@ -0,0 +1,46 @@
...
+
+
+jobs:
+  detect-force-push:
+    name: Detect Force push
+    runs-on: ubuntu-latest
+    if: github.actor != 'dependabot[bot]' &&!github.event.pull_request.draft &&!contains(github.event.pull_request.title, 'WIP')
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
```
mathetake, line 19:
```suggestion
        uses: actions/checkout@v4
```

mathetake (approved):

mathetake:
@daixiang0 thanks for your work!!

PR #431, opened 2025-02-26
extproc: wip proposal for chat completion instrumentation

nacx:
**Commit Message**

exproc: instrument requests tot eh chat completion endpoint

Related to: https://github.com/envoyproxy/ai-gateway/pull/316

Opening as a draft to get early feedback on the direction.

This takes the good work that was done in https://github.com/envoyproxy/ai-gateway/pull/316 around metrics as the starting point.

The approach in this PR, however, is to just create an instrumentation decorator for the chat completion processor, to keep the metrics logic independent. This allows us to:
* Fully enable/disable instrumentation if desired.
* Properly separate concerns and responsibilities. With this we're able to test independently the processor and the instrumentation logic, keeping the processor and translator logic unmodified.
* Allow to inject mocks for the metrics provider so that we can properly unit test the instrumented processor and make sure we record metrics when needed.

@mathetake @rootfs WDYT? 

If this looks OK, I'll proceed with adding unit tests, etc, but want to get some validation on direction first, as there are things that could be discussed (such as having logic not inside and int he middle of some methods, but after they return), or this approach not be the preferred one.

### DIFF
--- a/cmd/extproc/mainlib/main.go
+++ b/cmd/extproc/mainlib/main.go
@@ -13,6 +13,7 @@ import (
 	"log"
 	"log/slog"
 	"net"
+	"net/http"
 	"os"
 	"os/signal"
 	"strings"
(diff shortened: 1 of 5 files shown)


rootfs:
@nacx please see #432. The difference between the two is the instrumentation wrapper in this PR vs the inline metrics update in #432. As noted in #432, instrumenting processor doesn't get us all the info. Additional works have to be happen at the translator to measure inter token latency.

mathetake:
i don't think functionality wise this cannot do anything 432 can do, but i guess 432 is already getting closer to what i wanted to see. so let's iterate on there first and later at the refactoring phase, revisit this approach (thanks @nacx!!)

Issue #233, opened 2025-01-30
[Discussion] Enable Contribution of Experimental Processing Extensions


missBerg:
*Description*:
This Issue is for discussing how we can enable experimental data plane processing steps and control plane APIs

*Relevant Links*:
See notes from community meeting: https://docs.google.com/document/d/10e1sfsF-3G3Du5nBHGmLjXw5GVMqqCvFDqp_O65B0_w/edit?tab=t.0#heading=h.h7jtr1uxwvpi 

cc @envoyproxy/ai-gateway-maintainers @envoyproxy/ai-gateway-assignable

mathetake:
So I did a small refactoring of the Go packages before the v0.1.0rc release: #234. Based on that, the automatic model selection stuff proposed by @rootfs can live inside the experimental package https://github.com/envoyproxy/ai-gateway/blob/main/filterapi/x/x.go just like the custom router stuff (which requires the rebuild though).


I think the semantics caching (#30) will also fall into the same style of "calling out from the extproc"? cc @Krishanx92 
Then let's start with the API design of "AI Gateway Semantics Processor" i guess ... which I believe will be a gRPC service

rootfs:
Sounds good! I'll rebase to the latest code and create a PR.

mathetake:
yeah the initial PR should be only focused on the API of the external gRPC sever - no other stuff needed yet !

rootfs:
@mathetake it would be counterproductive to review an API protocol without a supporting PoC :D

mathetake:
ok up to you - it would be helpful for reviews anyways!

arkodg:
would be great to also see a PoC, which could be as simple as a `kubectl patch` cmd that enhances the generated `EnvoyExtensionPolicy` with a new `extProc` and documents this, and based off user needs, can be incorporated [into a first class API ](https://github.com/envoyproxy/ai-gateway/pull/256)

rootfs:
@mathetake @arkodg  #256 is updated. I also have a semantic cache PoC in the PR. You can see the demo [here](https://asciinema.org/a/704430)

mathetake:
[...]

PR #433, opened 2025-02-26
feat: add external routing processor proposal

rootfs:
**Commit Message**
This proposal is to replace #256.
Following up the discussion #233 on add external routing processor and gRPC interface.

@mathetake @aabchoo @arkodg  PTAL


mathetake:
i think the discussion was that we should see the concrete number to prove the benefit for bringing in the callouts from our extproc vs the complexity in the project since users are able to insert their own extproc in anyways. let's close this one for now as this needs rework due to the merge conflicts

PR #438, opened 2025-02-27
controller: adds comments on default / rule

mathetake:
**Commit Message**

This adds a code comment on why each translated HTTPRoute has a default route / on the first backend. This was brought up in the conversation in #384 and it is definitely worth a code comment on why it exists.


mathetake:
ok passing

aabchoo (approved):
Changes make sense to me!

mathetake:
Thanks for the look!

PR #442, opened 2025-02-28
backport: e2e: add agent test using tool (#426)

aabchoo:
**Commit Message**

* adding e2e test making sure that we can use a tool and return the proper response from the LLM

This PR also contains a few fixes that allow this test to work 
* added validateToolCallID as its needed for the bedrock message or it fails
* Validate and cast the openai content value into bedrock content block
- before we assumed it was always a string but it can also be an array
* Merge the content blocks if one has text and the other has toolcall but no text - bedrockResp.Output.Message.Content is not 1:1 with openai choice.Message.Content.. we get the result in chunks that should be merged, for example the text and it's toolconfig were in seperate elements but openai expects them to be in the same
* in openAIMessageToBedrockMessageRoleAssistant we assume that either there is a refusal or a content text, but we actually dont pass in a text. this was causing an error as the length of the array was set to 1 so the first was empty and there must be a key specified in each content element.

note: i think this is another bug that im trying to look into/create unit test for, but since there is already a lot in this PR, maybe its best to follow up with that one the next one. basically, the assistant openai param message's content doesn't seem to be translating to this method, and we have no content. this doesn't prevent anything else from working though, we are just missing a text field like `{"content":[{"text":"Certainly! I can help you get the weather information for New York City. To do that, I'll use the available weather tool. Let me fetch that information for you right away."}`

the rests are tests

### DIFF
--- a/.github/workflows/pr_style_check.yaml
+++ b/.github/workflows/pr_style_check.yaml
@@ -80,6 +80,7 @@ jobs:
             examples
             blog
             site
+            backport
           subjectPattern: ^(?![A-Z]).+$
           subjectPatternError: |
             The subject "{subject}" found in the pull request title "{title}"
--- a/internal/apischema/awsbedrock/awsbedrock.go
+++ b/internal/apischema/awsbedrock/awsbedrock.go
@@ -277,10 +277,10 @@ type ToolResultContentBlock struct {
 	Image *ImageBlock `json:"image,omitempty"`
 
 	// A tool result that is text.
-	Text *string `json:"text" type:"string,omitempty"`
+	Text *string `json:"text,omitempty"`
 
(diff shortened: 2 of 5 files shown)


aabchoo:
Backporting commit into release branch

mathetake:
hehe yeah i think we needed the title into the base release/v0.1 (PR targer)... let's ignore the title check for now - should be no problem. But backporting the title check here anyways make sense

mathetake (approved):

mathetake:
image


huh - what's this

aabchoo:
The original commits are verified. I wonder if cherry picking commits alters that 🤔

mathetake:
i think all resolved now (i disable some checks on the protection rule)

PR #447, opened 2025-03-02
controller: fix syncing extproc image tag

### LINKED ISSUE #446: ExtProc deployment image tag is not updated after upgrading from 0.1.0 to 0.1.1
*Description*:
After upgrading the ai gateway helm chart version from 0.1.0 to 0.1.1, the extproc deployment image tag is not updated which stays on 0.1.0. The upgrade should make sure both ai gateway controller and extproc are on the same version.

*Repro steps*:
- Upgrade ai gateway helm chart from 0.1.0 to 0.1.1
- The ai gateway controller has the correct extproc image tag passed `- --extProcImage=artifactory.inf.bloomberg.com/ds/ext/docker.io/envoyproxy/ai-gateway-extproc:v0.1.1`
- However the ai gateway extproc deployment image tag is not updated and stays on 0.1.0 because if the deployment already exists the update logic does not handle image tag change.
https://github.com/envoyproxy/ai-gateway/blob/beec3dd8b4871f7b775dcca424a965b1c55ea693/internal/controller/ai_gateway_route.go#L556

yuzisun:
**Commit Message**
Fix syncing extproc image tag in ai gateway routing reconciliation. 

Fixes #446

### DIFF
--- a/internal/controller/ai_gateway_route.go
+++ b/internal/controller/ai_gateway_route.go
@@ -153,7 +153,7 @@ func extProcName(route *aigv1a1.AIGatewayRoute) string {
 	return fmt.Sprintf("ai-eg-route-extproc-%s", route.Name)
 }
 
-func applyExtProcDeploymentConfigUpdate(d *appsv1.DeploymentSpec, filterConfig *aigv1a1.AIGatewayFilterConfig) {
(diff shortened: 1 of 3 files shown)


mathetake (approved):
good catch!

PR #455, opened 2025-03-04
controller: update rotate function to return expiration time

aabchoo:
**Commit Message**

Update `Rotate` to return the expiration time in addition to the error. This will allow us to get the expiration time without having to make a `GET` API request to Kubernetes.

https://github.com/envoyproxy/ai-gateway/blob/8f38c672ee11e6e724fff7b1e430979417fd3ce6/internal/controller/backend_security_policy.go#L106-L108

```
2025-03-02T20:12:23Z    INFO    controller.backend-security-policy      successfully rotated credentials for credentials in namespace default of auth type AWSCredentials, renewing in -0.009733 minutes
```

The requeue time after rotating a set of credentials (AWS in this case) returns a negative number of minutes occasionally.  It requires a manual action/global sync to restart the rotation logic.

Upon further investigation, the issue found lies with `rotator.GetPreRotationTime` returning the incorrect time. GetPreRotation gets the expiration time from the secret's annotation, and the assumption is it's getting a cached previous secret's annotation instead of the new expiration time.

https://github.com/envoyproxy/ai-gateway/blob/8f38c672ee11e6e724fff7b1e430979417fd3ce6/internal/controller/backend_security_policy.go#L143-L147

### DIFF
--- a/internal/controller/backend_security_policy.go
+++ b/internal/controller/backend_security_policy.go
@@ -140,10 +140,10 @@ func (c *BackendSecurityPolicyController) rotateCredential(ctx context.Context, 
 		return time.Minute, err
 	}
 	rotationTime := expiration.Add(-preRotationWindow)
-	if rotationTime.Before(time.Now()) {
-		return time.Minute, fmt.Errorf("newly rotate credentials is already expired (%v) for policy %s in %s", rotationTime, policy.Name, policy.Namespace)
+	if requeue := time.Until(rotationTime); requeue > 0 {
+		return requeue, nil
 	}
-	return time.Until(rotationTime), nil
+	return time.Minute, fmt.Errorf("newly rotate credentials is already expired (%v) for policy %s in %s", rotationTime, policy.Name, policy.Namespace)
 }
 
 // getBackendSecurityPolicyAuthOIDC returns the backendSecurityPolicy's OIDC pointer or nil.
--- a/internal/controller/backend_security_policy.go
+++ b/internal/controller/backend_security_policy.go
@@ -140,6 +140,9 @@ func (c *BackendSecurityPolicyController) rotateCredential(ctx context.Context, 
 		return time.Minute, err
 	}
 	rotationTime := expiration.Add(-preRotationWindow)
+	if rotationTime.Before(time.Now()) {
+		return time.Minute, fmt.Errorf("newly rotate credentials is already expired (%v) for policy %s in %s", rotationTime, policy.Name, policy.Namespace)
+	}
 	return time.Until(rotationTime), nil
 }
 
--- a/internal/controller/backend_security_policy.go
+++ b/internal/controller/backend_security_policy.go
@@ -119,7 +119,6 @@ func (c *BackendSecurityPolicyController) reconcile(ctx context.Context, backend
 // rotateCredential rotates the credentials using the access token from OIDC provider and return the requeue time for next rotation.
 func (c *BackendSecurityPolicyController) rotateCredential(ctx context.Context, policy *aigv1a1.BackendSecurityPolicy, oidcCreds egv1a1.OIDC, rotator rotators.Rotator) (time.Duration, error) {
 	bspKey := backendSecurityPolicyKey(policy.Namespace, policy.Name)
-
 	var err error
 	c.oidcTokenCacheMutex.RLock()
 	validToken, ok := c.oidcTokenCache[bspKey]
@@ -136,14 +135,11 @@ func (c *BackendSecurityPolicyController) rotateCredential(ctx context.Context, 
 	}
 
 	token := validToken.AccessToken
-	err = rotator.Rotate(ctx, token)
+	expiration, err := rotator.Rotate(ctx, token)
 	if err != nil {
 		return time.Minute, err
 	}
-	rotationTime, err := rotator.GetPreRotationTime(ctx)
-	if err != nil {
-		return time.Minute, err
-	}
+	rotationTime := expiration.Add(-preRotationWindow)
 	return time.Until(rotationTime), nil
 }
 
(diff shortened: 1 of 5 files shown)


internal/controller/backend_security_policy.go
```diff
@@ -136,14 +135,11 @@ func (c *BackendSecurityPolicyController) rotateCredential(ctx context.Context,
...
-	if err != nil {
-		return time.Minute, err
-	}
-	rotationTime, err := rotator.GetPreRotationTime(ctx)
+	expiration, err := rotator.Rotate(ctx, token)
 	if err != nil {
 		return time.Minute, err
 	}
+	rotationTime := expiration.Add(-preRotationWindow)
 	return time.Until(rotationTime), nil
```
xiaolin593, line 143: Can we add negative check for `time.Until(rotationTime)` here just in case?

internal/controller/rotators/aws_oidc_rotator.go
```diff
@@ -159,14 +160,14 @@ func (r *AWSOIDCRotator) Rotate(ctx context.Context, token string) error {
 				Data: make(map[string][]byte),
 			}
 			populateSecretWithAwsIdentity(secret, awsIdentity, r.region)
-			return r.client.Create(ctx, secret)
+			return *awsIdentity.Credentials.Expiration, r.client.Create(ctx, secret)
```
xiaolin593, line 163: Can we add a log line to print out actual expiration time for debugging purpose?

internal/controller/rotators/aws_oidc_rotator.go
```diff
@@ -159,14 +160,14 @@ func (r *AWSOIDCRotator) Rotate(ctx context.Context, token string) error {
...
 		}
 		r.logger.Error(err, "failed to lookup aws credentials secret", "namespace", bspNamespace, "name", bspName)
-		return err
+		return time.Time{}, err
 	}
 	r.logger.Info("updating existing aws credential secret", "namespace", bspNamespace, "name", bspName)
 	populateSecretWithAwsIdentity(secret, awsIdentity, r.region)
-	return r.client.Update(ctx, secret)
+	return *awsIdentity.Credentials.Expiration, r.client.Update(ctx, secret)
```
xiaolin593, line 170: ditto

internal/controller/rotators/aws_oidc_rotator_test.go
```diff
@@ -160,11 +163,47 @@ func TestAWS_OIDCRotator(t *testing.T) {
...
 		}
-		err := awsOidcRotator.Rotate(t.Context(), newOidcToken)
+
+		expiration, err := awsOidcRotator.Rotate(t.Context(), newOidcToken)
 		require.Error(t, err)
+		require.True(t, expiration.IsZero())
 		assert.Contains(t, err.Error(), "failed to assume role")
 	})
 
+	t.Run("error handling - invalid expiration failure", func(t *testing.T) {
```
xiaolin593, line 173: Can you add test case to cover scenario when current system time passes rotation time?

internal/controller/rotators/aws_oidc_rotator_test.go
```diff
@@ -192,8 +231,10 @@ func TestAWS_OIDCRotator(t *testing.T) {
 			region:                         awsRegion,
 			roleArn:                        awsRoleArn,
 		}
-		err := rotator.Rotate(t.Context(), newOidcToken)
+		expiration, err := rotator.Rotate(t.Context(), newOidcToken)
 		require.NoError(t, err)
+		require.NotNil(t, expiration)
+		require.WithinRange(t, expiration, time.Now(), time.Now().Add(1*time.Hour))
```
xiaolin593, line 237: nit: It'd be better to store `startTime := time.Now()` at line 221, then `require.WithinRange(t, expiration, startTime, startTime.Add(1*time.Hour))`

xiaolin593 (changes requested):

wengyao04:
Hi, @mathetake:
In our DEV setup, we find that a negative minutes is used in the requeueAfter filed. In the controller-runtime [implementation](https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/internal/controller/controller.go#L311-L360), if the requeqAfter is `false`, the event will be dropped to the reconciliation loop. We have to wait till the next global reconcile period (default is 10 hours)

We use mgr.Client() to get the secret, where the default read client read the secret from the informer cache and it is common that the informer cache sometimes is async with the k8s etcd. 

Beside this change, I think maybe we can use controller generic events with a time ticker to reconcile the backendsecuritypolicy every 10 minutes in case we miss any requeue event.

internal/controller/rotators/aws_oidc_rotator.go
```diff
@@ -143,9 +143,10 @@ func (r *AWSOIDCRotator) Rotate(ctx context.Context, token string) error {
 	awsIdentity, err := r.assumeRoleWithToken(ctx, token)
 	if err != nil {
 		r.logger.Error(err, "failed to assume role", "role", r.roleArn, "access token", token)
-		return err
+		return time.Time{}, err
+	} else if awsIdentity == nil || awsIdentity.Credentials == nil || awsIdentity.Credentials.Expiration == nil {
```
mathetake, line 147: exactly when awsIdentity can be nil and not error?

aabchoo: hmm I will remove it since I don't see a case for that. Will also remove `awsIdentity.Credentials.Expiration == nil` as Expiration is required if `awsIdentity.Credentials` exists.

xiaolin593 (approved):
LGTM, wait overnight integration test done to merge.

mathetake (approved):
Feel free to merge whenever ready

PR #458, opened 2025-03-05
cli: initial implementation of `aigw run`

mathetake:
**Commit Message**

This implements the `aigw run` command that runs the AI Gateway locally. This utilizes the Envoy Gateway's standalone mode [1]. This commit only scaffolds the command and runs the static default configuration. The follow up PR will allows users to pass the real AI Gateway yaml inputs.

1: https://gateway.envoyproxy.io/docs/tasks/operations/standalone-deployment-mode/

**Related Issues/PRs (if applicable)**

Contributes to #412

PR #460, opened 2025-03-05
deps: use latest EG, pin cel-go to 0.22

mathetake:
**Commit Message**

The CLI's standalone mode needs to use Envoy Gateway's latest patch [1] that allows us to use it inside our CLI. This upgrades the EG dependency version to the latest. It is not compatible with the latest cel-go, so this also pins it to 0.22 until the upstream k8s.io/apiserver dependency [2] issue is resolved. This commit is separated from the standalone implementation commit as the upgrade results in a CRD generation which is not relevant to the change.

1: https://github.com/envoyproxy/gateway/pull/5407
2: https://github.com/kubernetes/apiserver/issues/116

This is a preparation for #458


go.mod
```diff
@@ -9,10 +9,10 @@ require (
 	github.com/aws/aws-sdk-go-v2/config v1.29.9
 	github.com/aws/aws-sdk-go-v2/service/sts v1.33.17
 	github.com/coreos/go-oidc/v3 v3.12.0
-	github.com/envoyproxy/gateway v1.3.0
-	github.com/envoyproxy/go-control-plane/envoy v1.32.4
+	github.com/envoyproxy/gateway v0.5.0-rc.1.0.20250305132535-64b753bfa5fd
```
mathetake, line 12: ` v0.5.0-rc` prefix is really weird but that's the last tag cut from their main branch so 🤷‍♂️

nacx (approved):

PR #461, opened 2025-03-05
backport: fix requeue bug and extproc tag sync

aabchoo:
**Commit Message**

Backporting fix for extproc image tag not being sync'd properly and case when requeue is negative.

For more info, please check out:
- controller: fix syncing extproc image tag (https://github.com/envoyproxy/ai-gateway/pull/447)
- controller: update rotate function to return expiration time (https://github.com/envoyproxy/ai-gateway/pull/455)


wengyao04 (approved):

mathetake:
why they are three commits vs two commits mentioned? what about https://github.com/envoyproxy/ai-gateway/pull/461/commits/c35e9c738dbee076c2eaecb28a94166408817685 ?

mathetake:
ah ok the test

PR #468, opened 2025-03-07
extproc: properly stream chat completions

mathetake:
**Commit Message**

This fixes a bug in the extproc when handling stream=true requests. Previously, mode_override was set at the request body handling phase, and it was not set in the response headers phase. That resulted in buffering the entire response body which is clearly not ideal as from clients point of view, they will receive the entire streaming vs line by line. This refactors around the mode override and properly handle it.

### DIFF
--- a/internal/extproc/chatcompletion_processor.go
+++ b/internal/extproc/chatcompletion_processor.go
@@ -16,6 +16,7 @@ import (
 	"log/slog"
 
 	corev3 "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
+	extprocv3http "github.com/envoyproxy/go-control-plane/envoy/extensions/filters/http/ext_proc/v3"
 	extprocv3 "github.com/envoyproxy/go-control-plane/envoy/service/ext_proc/v3"
 	typev3 "github.com/envoyproxy/go-control-plane/envoy/type/v3"
 	"google.golang.org/protobuf/types/known/structpb"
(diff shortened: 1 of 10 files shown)


internal/extproc/chatcompletion_processor.go
mathetake, line 1: this is where the actual change took place: Instead of setting mode override in ProcessRequestBody, this will instead set it in ProcessResponseHeaders

internal/extproc/translator/translator.go
```diff
@@ -56,28 +53,16 @@ type OpenAIChatCompletionTranslator interface {
...
 	//  - This returns `tokenUsage` that is extracted from the body and will be used to do token rate limiting.
-	//
-	// TODO: this is coupled with "LLM" specific. Once we have another use case, we need to refactor this.
 	ResponseBody(respHeaders map[string]string, body io.Reader, endOfStream bool) (
 		headerMutation *extprocv3.HeaderMutation,
 		bodyMutation *extprocv3.BodyMutation,
 		tokenUsage LLMTokenUsage,
 		err error,
 	)
-
```
aabchoo, line 71: Why remove this?

aabchoo (approved):
overall looks good to me

internal/extproc/translator/translator.go
mathetake: Cause it's not used :)

mathetake:
i am trying to write a regression test that actually verifies the streaming behavior

mathetake:
ok the regression test added (verified it's failing on the current main)

PR #477, opened 2025-03-11
controller: refactor rotation interface

yuzisun:
**Commit Message**
Refactor the rotation interface to make it easier to add other OIDC rotation like Azure and GCP.  The main change is to move the oidc token token into the rotate function as we may have different token provider such as Azure token service which needs to call azure specific APIs, we do not want leak this code into the main reconciliation loop. Also we removed the access token cache which is unnecessary as the call is only made whenever the credential needs to be refreshed and we have timeout to not block the controller.


internal/controller/backend_security_policy_test.go
```diff
@@ -241,39 +235,14 @@ func TestBackendSecurityController_RotateCredentials(t *testing.T) {
...
-	_, err = c.rotateCredential(ctx, bsp, oidc, rotator)
+	_, err = rotator.Rotate(ctx)
 	require.NoError(t, err)
 	awsSecret2, err := rotators.LookupSecret(t.Context(), cl, bspNamespace, awsSecretName)
 	require.NoError(t, err)
 	t2 := awsSecret2.Annotations[rotators.ExpirationTimeAnnotationKey]
 	require.NotEqual(t, t1, t2)
 }
 
-func TestBackendSecurityPolicyController_RotateExpiredCredential(t *testing.T) {
```
yuzisun, line 252: discussed with @aabchoo this test case is no longer needed as we now get the expiration time from the token directly

mathetake:
lol force push detection seems to have a false positive cc @daixiang0

mathetake (approved):
I like less code

internal/controller/backend_security_policy.go
```diff
@@ -99,13 +94,21 @@ func (c *BackendSecurityPolicyController) reconcile(ctx context.Context, backend
...
+					if r := time.Until(rotationTime); r > 0 {
+						requeue = r
+						c.logger.Info(
+							fmt.Sprintf("successfully rotated credentials for %s in namespace %s of auth type %s, renewing in %f minutes",
+								backendSecurityPolicy.Name, backendSecurityPolicy.Namespace, backendSecurityPolicy.Spec.Type, requeue.Minutes()))
+					} else {
+						c.logger.Error(fmt.Errorf("newly rotate credentials is already expired %s",
+							rotationTime), "namespace", backendSecurityPolicy.Namespace, "name", backendSecurityPolicy.Name)
+					}
+
```
aabchoo, line 111: Could we add a log on line 114 ie

fmt.Sprintf("credentials has not yet expired for %s in namespace %s of auth type %s, renewing in %f minutes",						backendSecurityPolicy.Name, backendSecurityPolicy.Namespace, backendSecurityPolicy.Spec.Type, requeue.Minutes()))

internal/controller/rotators/aws_oidc_rotator_test.go
```diff
@@ -104,6 +110,30 @@ func (m *mockStsOperations) AssumeRoleWithWebIdentity(ctx context.Context, param
 }
 
 func TestAWS_OIDCRotator(t *testing.T) {
+	discoveryServer := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
```
aabchoo, line 113: discoveryServer isn't used? can either set it to Provider Issuer or remove entirely

aabchoo (approved):

daixiang0:
@mathetake I check all CI logs for each commit, no force-push check action log :(

Do you have any log about it?

mathetake (approved):

PR #478, opened 2025-03-11
ci: revert the force push detection

mathetake:
**Commit Message**

It seems there's a false-positive in the original logic as we see in #477. This commit temporarily reverts the workflow until it's sorted out.

Reverts #430


mathetake:
cc @daixiang0

PR #480, opened 2025-03-12
extproc: do not panic when immediate response in debug mode

mathetake:
**Commit Message**

Previously, `filterSensitiveBodyForLogging` resulted in panic when the response was immediate response due to, for example, the matching route was not found.


mathetake:
Thanks for the look!

PR #483, opened 2025-03-12
extproc: allow building with custom metric implementations

nacx:
**Commit Message**

extproc: allow building with custom metric implementations

**Related Issues/PRs (if applicable)**

N/A

**Special notes for reviewers (if applicable)**

N/A

PR #484, opened 2025-03-12
extproc: allow extra attributes in the chat completion metrics interface

nacx:
**Commit Message**

extproc: allow extra attributes in the chat completion metrics interface

Follow-up of: https://github.com/envoyproxy/ai-gateway/pull/483

This complements the previous pull request by making the ChatCompletionMetrics capable of receiving additional attributes to be added to the generated metrics. This will facilitate custom builds of the completion metrics to enrich metrics with custom attributes.

Note that this would not be immediately leverageable by the default build of extprox, as the chat completion Processor will call the methods with no extra attirbtues, but it facilitates writing cleaner code when building a custom implementation of the metrics. In addition, we will have an interface that already makes sense and is better suited for extension (and if we eventually allow users to customizer metrics via the filter API config, etc, the metrics interface would already be able to nicely accommodate that).

### DIFF
--- a/internal/metrics/chat_completion_metrics.go
+++ b/internal/metrics/chat_completion_metrics.go
@@ -68,12 +68,12 @@ func (c *chatCompletion) SetBackend(backend filterapi.Backend) {
 }
 
 // RecordTokenUsage implements [ChatCompletion.RecordTokenUsage].
-func (c *chatCompletion) RecordTokenUsage(ctx context.Context, inputTokens, outputTokens, totalTokens uint32) {
-	attrs := []attribute.KeyValue{
+func (c *chatCompletion) RecordTokenUsage(ctx context.Context, inputTokens, outputTokens, totalTokens uint32, extraAttrs ...attribute.KeyValue) {
+	attrs := append([]attribute.KeyValue{
 		attribute.Key(genaiAttributeOperationName).String(genaiOperationChat),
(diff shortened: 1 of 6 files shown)


internal/metrics/chat_completion_metrics.go
```diff
@@ -68,12 +68,12 @@ func (c *chatCompletion) SetBackend(backend filterapi.Backend) {
...
 // RecordTokenUsage implements [ChatCompletion.RecordTokenUsage].
-func (c *chatCompletion) RecordTokenUsage(ctx context.Context, inputTokens, outputTokens, totalTokens uint32) {
-	attrs := []attribute.KeyValue{
+func (c *chatCompletion) RecordTokenUsage(ctx context.Context, inputTokens, outputTokens, totalTokens uint32, extraAttrs ...attribute.KeyValue) {
+	attrs := append([]attribute.KeyValue{
 		attribute.Key(genaiAttributeOperationName).String(genaiOperationChat),
 		attribute.Key(genaiAttributeSystemName).String(c.backend),
 		attribute.Key(genaiAttributeRequestModel).String(c.model),
-	}
+	}, extraAttrs...)
```
mathetake, line 76: can you pre allocate the capacity for the base slice ?

mathetake (approved):

internal/metrics/chat_completion_metrics.go
nacx: Done

PR #486, opened 2025-03-13
feat:  openai: allow assistant content to be string

aabchoo:
**Commit Message**

Content type should be able to be string so adding this for compatibility. It's already done for the other types - keeping it consistent.


internal/apischema/openai/openai.go
```diff
@@ -149,6 +149,28 @@ func (c *ChatCompletionContentPartUserUnionParam) UnmarshalJSON(data []byte) err
...
+	}
+
+	var content ChatCompletionAssistantMessageParamContent
+	err = json.Unmarshal(data, &content)
+	if err == nil {
+		s.Value = content
+		return nil
+	}
+
+	return fmt.Errorf("cannot unmarshal JSON data as string or assistant content parts")
```
mathetake, line 171: if there's no format then let's use errors.New
```suggestion
	return errors.New("cannot unmarshal JSON data as string or assistant content parts")
```

internal/extproc/translator/openai_awsbedrock.go
```diff
@@ -245,11 +245,15 @@ func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) openAIMessageToBedrockMes
...
-	} else if openAiMessage.Content.Text != nil {
-		// TODO: we are sometimes missing the content (should fix)
-		contentBlocks = append(contentBlocks, &awsbedrock.ContentBlock{Text: openAiMessage.Content.Text})
+	if v, ok := openAiMessage.Content.Value.(string); ok {
+		contentBlocks = append(contentBlocks, &awsbedrock.ContentBlock{Text: &v})
+	} else if content, ok := openAiMessage.Content.Value.(openai.ChatCompletionAssistantMessageParamContent); ok {
+		if content.Type == openai.ChatCompletionAssistantMessageParamContentTypeRefusal {
+			contentBlocks = append(contentBlocks, &awsbedrock.ContentBlock{Text: content.Refusal})
+		} else if content.Text != nil {
+			// TODO: we are sometimes missing the content (should fix)
```
yuzisun, line 254: Can remove this TODO now, this is the fix

internal/extproc/translator/openai_awsbedrock_test.go
```diff
@@ -82,8 +82,10 @@ func TestOpenAIToAWSBedrockTranslatorV1ChatCompletion_RequestBody(t *testing.T)
 					},
 					{
 						Value: openai.ChatCompletionAssistantMessageParam{
-							Content: openai.ChatCompletionAssistantMessageParamContent{
-								Text: ptr.To("I dunno"),
+							Content: openai.StringOrAssistantRoleContentUnion{
```
yuzisun, line 85: Can you help add the missing unmarshall test which is why this issue is not caught

aabchoo: The issue was content being parsed as string even though it was sent as AssistantContentType. Added a test, but the behavior is unideal since the unmarshalling now sees it as valid:

https://github.com/envoyproxy/ai-gateway/pull/486/files#diff-3387664749ae0034b39372c41c009a68c9b68b7ee9ae5886babe0a4f846e542fR168

mathetake:
LGTM from me

yuzisun (approved):

PR #489, opened 2025-03-13
chore: check force push

daixiang0:
**Commit Message**

Add a check for force push in CI.

### DIFF
--- a/.github/workflows/force_push_check.yaml
+++ b/.github/workflows/force_push_check.yaml
@@ -1,0 +1,49 @@
+name: PR Force Push Check
+
+on:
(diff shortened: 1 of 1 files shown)


daixiang0:
Test PR: https://github.com/daixiang0/ai-gateway/pull/5

I only find that the first push would raise an unexpected error.

@mathetake could you help save the log if any unexpected?

mathetake:
i am so sorry but just as much as i want a single rule but i decided to not have this rule as it seems too much. Thanks for your hard work and I really appreciate it @daixiang0 🙏 🙇

Issue #423, opened 2025-02-25
Support k8s gateway API inference extensions


yanavlasov:
Support [k8s gateway API inference extensions](https://gateway-api-inference-extension.sigs.k8s.io/concepts/api-overview/#api-overview)

- [ ] Translate k8s gateway API inference extensions CRDs into Envoy configuration
- [ ] Add e2e tests
- [ ] TBD

daixiang0:
@yanavlasov @envoyproxy/ai-gateway-maintainers  I see EG also want to implement it refer to https://gateway-api-inference-extension.sigs.k8s.io/implementations/#envoy-gateway, so will we have two implement for this?

arkodg:
@daixiang0 the preference is to implement this in Envoy AI Gateway first, get user feedback, iterate on the API until it moves to standard and then also directly support the implementation in Envoy Gateway

kfswain:
We've worked with Envoy GW quite a bit, it's been great. Happy to continue work with y'all to get these projects workin together! Is it easiest if I just pop in on the community meeting on Th?

arkodg:
thanks for picking this one up @kfswain !  you're familiar with the translation so this should be _easy_ for you 😄 
the community meeting is a great place to bring your questions

hey @mathetake can you point Kellen to the reconciliation and translation areas

mathetake:
@yanavlasov said he will start working from next week, so could you @kfswain coordinate with Yan (i assume you both are working at Google)

kfswain:
SGTM!

mathetake:
Hi @yanavlasov @kfswain Luckily I have cycles to work on this from tomorrow, so I will do the initial work soon and i might ask for some help from you guys whenever I need. SG?

kfswain:
That sounds great. Happy to help

mathetake:
now i get the big picture on how to implement in here... i have one question @kfswain may i ask where's the conformance test suite? i couldn't find one in the repo

yuzisun:
Let's have a design doc first, this certainly changes the scope of the envoy ai gateway project.

mathetake:
@yuzisun certainly, i will write up by noon tomorrow

mathetake:
I opened a PR for the high level doc: #492 and here's a couple note from my end and offline discussion:
* We will allow AIServiceBackend.BackendRef to reference InferencePool 
[...]

PR #492, opened 2025-03-13
docs: adds high level proposal on GAIE support

mathetake:
**Commit Message**

This adds a proposal doc on the support for Gateway API Inference Extension in Envoy AI Gateway project. This involves the change of the project scope as well as we need to make sure that the existing API layer will co-exist nicely with the GAIE.

Preliminary to #423

### DIFF
--- a/docs/proposals/002-inference-gateway-support/proposal.md
+++ b/docs/proposals/002-inference-gateway-support/proposal.md
@@ -1,0 +1,56 @@
+# Inference Gateway API Extension Support
+
+## Background
+
+[Gateway API Inference Extension](https://gateway-api-inference-extension.sigs.k8s.io) (GAIE) is a brand new extension of
+the Gateway API that aims to solve the problem of serving large language models (LLMs) inside Kubernetes. Especially,
+it specifies a way to make an intelligent load balancing decision etc.
+
+Envoy AI Gateway (EAIG) 's initial goal was, in contrast, to provide a way to route traffic to different AI providers.
+In other words, EAIG is closer to the application developers vs GAIE is closer to the AI platform team. Or, EAIG is
+more focused on the "egress" vs GAIE is more focused on the "ingress" (especially k8s).
+
+Even though, there's a clear difference between these two projects, some of the capability GAIE specifying can be useful for EAIG
+users as well. For example, [Endpoint picker with fallback and subset](https://github.com/kubernetes-sigs/gateway-api-inference-extension/pull/445) is a nice feature that can be leveraged and one of the
+oldest feature requests for EAIG.
+
+This document is about two key points; how EAIG project's scope can be changed to support GAIE,
+and how it integrates with the existing API.
+
+## Changes to the scope of the project
+
+Initially, the project was focused on routing traffic to different AI providers, namely "egress" traffic. However, with the
+introduction of GAIE, we need to expand it to support "ingress" traffic as well. This means EAIG will evolve to support
+"all AI traffic" instead of just "egress AI traffic".
+
+However, this doesn't mean that we need to have completely different APIs for "ingress" and "egress". Instead, we will ensure
+that the existing APIs nicely integrate with GAIE APIs.
+
+## How it integrates with the existing API
+
+We propose to allow the existing `AIServiceBackend.BackendRef` to reference an `InferencePool`. This will allow
+the users to leverage the new features of GAIE without changing the existing APIs. Even the translation and security policy
+related functionalities can be applied as is.
+
+```diff
+--- a/api/v1alpha1/api.go
++++ b/api/v1alpha1/api.go
+@@ -334,7 +334,7 @@ type AIServiceBackendSpec struct {
+        APISchema VersionedAPISchema `json:"schema"`
+        // BackendRef is the reference to the Backend resource that this AIServiceBackend corresponds to.
+        //
+-       // A backend can be of either k8s Service or Backend resource of Envoy Gateway.
++       // A backend can be of either k8s Service, Backend resource of Envoy Gateway, or InferencePool of Gateway API Inference Extension.
+        //
+        // This is required to be set.
+        //
+```
+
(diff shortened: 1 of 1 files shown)


mathetake:
@yanavlasov @yuzisun @arkodg @missBerg PTAL 🙏  Meanwhile i will start adding the trivial code like scaffolding the control plane etc

docs/proposals/002-inference-gateway-support/proposal.md
```diff
@@ -0,0 +1,56 @@
...
+and how it integrates with the existing API.
+
+## Changes to the scope of the project
+
+Initially, the project was focused on routing traffic to different AI providers, namely "egress" traffic. However, with the
+introduction of GAIE, we need to expand it to support "ingress" traffic as well. This means EAIG will evolve to support
+"all AI traffic" instead of just "egress AI traffic".
+
+However, this doesn't mean that we need to have completely different APIs for "ingress" and "egress". Instead, we will ensure
+that the existing APIs nicely integrate with GAIE APIs.
```
mathetake, line 27: @missBerg i need your nice edits here

mathetake:
I marked it as draft because there's a couple points that i cannot still get my heads around especially on how to make it play as a part of the existing API...  https://github.com/envoyproxy/ai-gateway/issues/423#issuecomment-2722839817

docs/proposals/002-inference-gateway-support/proposal.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Expanding Project Scope
+
+By evolving EAIG's focus beyond routing traffic to external AI providers (egress) to include internal Kubernetes LLM deployments (ingress) through GAIE implementation we position EAIG as a comprehensive traffic management solution for all AI traffic.
+
+Note that this enhancement is an addition to the existing EAIG APIs, and does not compete with existing functionality for the egress use case.
+
+## API Integration Approach
+
+We propose enabling the existing `AIServiceBackend.BackendRef` to reference GAIE's `InferencePool`. This approach allows users to leverage GAIE's advanced features without changing existing APIs. Translation and security policy functionalities will continue to work as designed.
```
robscott, line 36: Not that familiar with AIServiceBackend, but why not just rely on references from HTTPRoute to InferencePool?

yuzisun: AIServiceBackends are cloud AI providers, aws bedrock, gcp Gemini and azure OpenAI. We want a pool to be able to failover across cloud regions or AI service providers not at the http router level.

robscott: That makes sense, thanks @yuzisun! Just to clarify, we'd also support direct reference from HTTPRoute -> InferencePool right?

missBerg: @robscott This note describes how the Envoy AI GW API would integrate with the Inference Extension and does not impact how the Gateway API would integrate with the Inference Extension. 

So all as laid out in https://gateway-api-inference-extension.sigs.k8s.io/ stays intact

image

missBerg (approved):

docs/proposals/002-inference-gateway-support/proposal.md
mathetake: The short answer is no (atm). AIServiceBackend is a top level construct here in EAIG (I need a acronym btw) which _manages_ HTTPRoute and HTTPRoute is not exposed to the users here. As we discussed offline with @yuzisun and others, and we decided not to introduce multiple API surfaces to do the same thing (which is obviously a bad thing) and hence only expose the inference gateway at `AIServiceBackend.BackendRef`.

missBerg: @mathetake I'd expect 
1. HTTPRoute -> InferencePool
to be supported

2. HTTPRoute -> AIServiceBackend - > InferencePool
I would not expect

missBerg: Note: I'm good for us to start with `AIServiceBackend - > InferencePool` first then if requested/needed we can add `HTTPRoute -> InferencePool`

mathetake: 👍

mathetake:
so i would hold off merging until I completely understand how https://github.com/envoyproxy/envoy/pull/38757 works ...

docs/proposals/002-inference-gateway-support/proposal.md
mathetake: further discussed offline and the consensus is that we start with AIServiceBackend.BackendRef only. Then later if someone really wants it exposed with a valid reason, then the direct ref from HTTPRoute will be supported here.

docs/proposals/002-inference-gateway-support/proposal.md
```diff
@@ -0,0 +1,65 @@
...
+- How it integrates with the existing EAIG APIs
+- Proposed implementation details
+
+## Background
+
+[Gateway API Inference Extension](https://gateway-api-inference-extension.sigs.k8s.io) (GAIE) is a new extension of the Kubernetes Gateway API that addresses serving large language models (LLMs) inside Kubernetes, with particular focus on intelligent load balancing decisions.
+
+Envoy AI Gateway (EAIG) was initially designed to route traffic to different AI providers, serving primarily as an egress solution.
+
+The EAIG MVP features targeted application developers with an egress focus, while the GAIE serves AI platform teams with a Kubernetes ingress focus.
```
askulkarni2, line 23: For someone who is deploying models solely on kubernetes, is there a benefit to use EAIG?

mathetake: I think even right now without GAIE implementation, the rate limit and the metrics/monitoring feature are useful to those without egress plus we are looking to expand the request transformation pluggable so that local hosted models can host whatever endpoint they want. These existing API/features will be available to GAIE used via AIServiceBackend.BackendRef. cc @yuzisun @missBerg for additional comments

mathetake:
ok let's merge this!

Issue #34, opened 2024-12-05
[Design] Failover Logic between Providers and Models


missBerg:
The design proposal should include:

- [ ] Motivation
- [ ] Feature Definition
- [ ] Control Plane API
- [ ] Technical Implementation Proposal

**This issue was created from conversation during Dec 5th Community meeting**
https://docs.google.com/document/d/10e1sfsF-3G3Du5nBHGmLjXw5GVMqqCvFDqp_O65B0_w/edit?tab=t.0#bookmark=id.dz9gpy397ymu

yuzisun:
This can potentially leverage the implementation in envoy gateway though we will to think how to expose on `AIGatewayRoute`
https://gateway.envoyproxy.io/docs/tasks/traffic/failover/

mathetake:
so what we essentially need is to "go through the extproc filter chain again to perform the necessary mutation" to re-mutate auth header, re-perform transformation, etc and that's not possible by either default Envoy or Inference Extension spec. 

One workaround I can think of is to make a request directly from the extproc to the same listener where the extproc is running when the response header for the initially chosen backend is 5xx. When directly sending a request from the extproc, we can add a special header like #73 to specifically choose the backend without running the router's matching. Then, if the response from the second request originating from the extproc is 200, we can use ImmediateResponse to send the response crafted from the contents from the fallback request to the client.


But this comes with a cost; until we receive the first 200 response, the extproc needs to retain the whole request body (and this is inevitable by the nature of this type of fallback regardless of Envoy or whatever proxy we use).


The good thing about this workaround is that all the logic is implementable without any changes in EG or Envoy themselves

yuzisun:
> so what we essentially need is to "go through the extproc filter chain again to perform the necessary mutation" to re-mutate auth header, re-perform transformation, etc and that's not possible by either default Envoy or Inference Extension spec.
> 
> One workaround I can think of is to make a request directly from the extproc to the same listener where the extproc is running when the response header for the initially chosen backend is 5xx. When directly sending a request from the extproc, we can add a special header like [#73](https://github.com/envoyproxy/ai-gateway/issues/73) to specifically choose the backend without running the router's matching. Then, if the response from the second request originating from the extproc is 200, we can use ImmediateResponse to send the response crafted from the contents from the fallback request to the client.
> 
[...]

PR #493, opened 2025-03-14
feat: initial implementation of Inference Extension 

mathetake:
**Commit Message**

This commit scaffolds the foundation for the Inference Extension API [1]. The design documentation was merged in #492. The controller needs to be started with `--enableInferenceExtension=true` to not break the existing controller deployment where the Inference Extension CRDs are not installed.

This commit doesn't implement the actual "metrics-aware" load balancing and instead it just does the random routing out of given (resolved) endpoints. The follow up implementations will add more advanced algorithm while expanding the metrics interface that currently only provides the setter APIs.

The summary of the implementation is:
* Added `kind` field to AIGatewayRouteRuleBackendRef so that it can reference InferencePool.
* InferencePool.Spec.Selector is allowed to specify multiple AIServiceBackend.
* When building up all the extproc config via filterapi.Config, the controller reads the referenced InferencePool and its binding InferenceModels, and group them together into a single filterapi.DynamicLoadBalancing configuration. 
* When the extproc loads the configuration containing DynamicLoadBalancing, it will resolve all the IP addresses for hostnames belonging to the DyanmicLoadbalancing. The presence of DynamicLoadBalancing in th config forces the config watcher reload and refresh the config regardless of the updates. That way, the list of ip addresses will always be updated (eventual consistency anyways) in a non-hot path.
* On the request path, the ChatCompletionProcessor will check the existence of the DynamicLoadBalancing config for the backend selected by the router. If so, it further tries to resolve the ip:port level endpoint selection.
* The selected ip:port will be set to the special header that will be routed to ORIGINAL_DST.
* ORIGINAL_DST cluster will be added by the EG extension sever implementation. Also, the extension server modifies some routes to properly route to that cluster. 

1: https://github.com/kubernetes-sigs/gateway-api-inference-extension

Built on #492
Contributes to #423


mathetake:
cc @yuzisun

mathetake:
obviously not working but i think this shows how this can be implemented here with minimal changes

mathetake:
one big missing piece is to generate a special Envoy cluster that has a new LB policy attached; that can be a global EnvoyPatchPolicy resource used at a global level.

filterapi/filterconfig.go
```diff
@@ -164,6 +164,34 @@ type Backend struct {
 	// Auth is the authn/z configuration for the backend. Optional.
 	// TODO: refactor after https://github.com/envoyproxy/ai-gateway/pull/43.
 	Auth *BackendAuth `json:"auth,omitempty"`
+	// DynamicLoadBalancing is the dynamic backend configuration which forces the AI filter to
+	// route the request to the specified endpoints (ip:port) instead of the cluster.
+	DynamicLoadBalancing *DynamicLoadBalancing `json:"dynamicBackend,omitempty"`
```
yuzisun, line 169: I think this is a level above backend, think about the case you want to failover between different backends GCP, AWS for Anthropic model

filterapi/filterconfig.go
```diff
@@ -164,6 +164,34 @@ type Backend struct {
...
+
+type DynamicLoadBalancingEndpoint struct {
+	// HostName is the hostname of the endpoint.
+	HostName string `json:"hostName,omitempty"`
+	// IP is the IP address of the endpoint.
+	IP string `json:"ip,omitempty"`
+	// Port is the port of the endpoint.
+	Port int `json:"port"`
+	// TODO: metrics, like for example, either concrete metrics or an endpoint to fetch the metrics
+	//	to do the intelligent routing.
```
mathetake, line 194: this will probably not be needed as we already have the otel metrics collection in place and should be able to utilize that (by like for example adding the selected ip:port as a metric label etc)

mathetake:
ok get back to this ......

filterapi/filterconfig.go
mathetake: right!

mathetake: ok so the reality is such failover like "going through filter chain again when the request got 5xx" cannot be implemented with the inference extension's spec. But I agree this should be one level above to allow us to our own failover mechanism based on that. I will put a comment in #34

mathetake:
Instead of allowing AIServiceBackend.BackendRef to reference InferencePool, it seem semantically correct to allow `AIGatewayRouteRuleBackendRef` to reference it, so i will tweak the API

mathetake:
ok the control plane is almost done

mathetake:
monday i will continue backfilling the unit tests for control plane and then extproc part

mathetake:
almos there

mathetake:
~Was thinking it should be better to use the EG extension server (currently not in use) but encountered a few issues (not sure who's to blame), switching to use PatchPolicy instead~ working

mathetake:
controller 100% done 🎉

kfswain:
I can make some time today to look over this from an IGW perspective, as it sounds like certain sections are ready for review. Unless I should hold off?

mathetake:
@kfswain yeah, the left is the implementation details + some tests so you can start reviewing

internal/controller/controller.go
```diff
@@ -152,25 +177,45 @@ const (
 	// k8sClientIndexBackendSecurityPolicyToReferencingAIServiceBackend is the index name that maps from a BackendSecurityPolicy
 	// to the AIServiceBackend that references it.
 	k8sClientIndexBackendSecurityPolicyToReferencingAIServiceBackend = "BackendSecurityPolicyToReferencingAIServiceBackend"
+	// k8sClientIndexInferencePoolToReferencingAIServiceBackend is the index name that maps from an InferencePool to the AIServiceBackend
+	// that references it.
+	k8sClientIndexInferencePoolToReferencingAIGateawyRoute = "InferencePoolToReferencingAIGatewayRoute"
```
kfswain, line 182:
```suggestion
	k8sClientIndexInferencePoolToReferencingAIGatewayRoute = "InferencePoolToReferencingAIGatewayRoute"
```

api/v1alpha1/api.go
```diff
@@ -218,8 +218,30 @@ type AIGatewayRouteRule struct {
...
 type AIGatewayRouteRuleBackendRef struct {
+	// Kind is the kind of the backend, which is either "AIServiceBackend" or "InferencePool" in Gateway API Inference Extension.
+	//
+	// When this references InferencePool, the selector of the InferencePool is used to select (multiple) AIServiceBackend(s)
+	// that can serve the same model sets that the InferencePool binds.
+	//
+	// Default is AIServiceBackend.
+	//
+	// +kubebuilder:validation:Enum=AIServiceBackend;InferencePool
+	// +kubebuilder:default=AIServiceBackend
```
mathetake, line 242: note to self: add cel validation test

mathetake:
OK i editted the description and added the summary of the implementation here. 95% done - the left is to add the end-to-end test case (extproc, controller tests were already added respectively).

mathetake:
i will mark this ready for ready as soon as the e2e test case is added but ready for review anyways

internal/extproc/dynlb/dynlb.go
```diff
@@ -0,0 +1,171 @@
...
+	dlb.logger.Info("selected endpoint", slog.String("endpoint", string(ep.ipPort)))
+
+	selected = ep.backend
+	headers = []*corev3.HeaderValueOption{
+		{Header: &corev3.HeaderValue{Key: originalDstHeaderName, RawValue: ep.ipPort}},
+	}
+	if hn := ep.hostname; hn != "" {
+		// TODO: Set host header if the IP is resolved from a hostname. Without this, it is likely that we cannot
+		// 	route the requests to external services that reject requests with the mismatching host header.
+		// 	Currently, EG API doesn't support allow us to set mutation_rules.
```
mathetake, line 167: TODO for tomorrow

mathetake:
finally ready for review; all tests passing with comprehensive test coverage

mathetake:
sorry that the PR gets larger than I thought, so please refer to the PR description for the overview before diving into it

mathetake:
One thing to consider; i think it also makes sense to do the dns resolution during the reconciliation and return `Requeue: true` (like OIDC refresh) when the dns refresh is necessary rather than letting extproc do that. We can do the refactoring in that way later as in anyways that's an implementation detail

missBerg (approved):
Thanks for all the hard work @mathetake 🙌

yuzisun:
Thanks @mathetake ! I am going to take a look tonight.

yuzisun:
Thanks @mathetake ! I am going to take a look tonight.

mathetake:
No rush !

kfswain:
I also should be able to do a review in the next 24h. Thanks for this!

api/v1alpha1/api.go
```diff
@@ -218,8 +218,30 @@ type AIGatewayRouteRule struct {
...
-// AIGatewayRouteRuleBackendRef is a reference to a AIServiceBackend with a weight.
+// AIGatewayRouteRuleBackendRefKind specifies the kind of the backend reference.
+type AIGatewayRouteRuleBackendRefKind string
+
+const (
+	// AIGatewayRouteRuleBackendRefAIServiceBackend is the kind of the AIServiceBackend.
+	AIGatewayRouteRuleBackendRefAIServiceBackend AIGatewayRouteRuleBackendRefKind = "AIServiceBackend"
+	// AIGatewayRouteRuleBackendRefInferencePool is the kind of the InferencePool in the Gateway API Inference Extension.
+	// https://github.com/kubernetes-sigs/gateway-api-inference-extension
+	AIGatewayRouteRuleBackendRefInferencePool AIGatewayRouteRuleBackendRefKind = "InferencePool"
```
yuzisun, line 229: Currently InferencePool is a list of pods, If I want to create a pool of AIServiceBackends, would that be another CR for AIServiceBackendPool?

mathetake: > Currently InferencePool is a list of pods,

As i commented in the description as well as in the API's comment, InferencePool.Spec.Selector will be used to select AIServiceBackends, not pods. That is allowed in the API spec of InfExt and i think that works as you intended in the comment?

yuzisun: Ah saw that in your examples now, that’s pretty neat way to unify the ingress and egress use cases!

internal/extproc/dynlb/dynlb.go
```diff
@@ -0,0 +1,171 @@
...
+// SelectChatCompletionsEndpoint implements [DynamicLoadBalancer.SelectChatCompletionsEndpoint].
+//
+// TODO: expand x.ChatCompletionMetrics to add getter methods to be able to make a decision based on the metrics.
+// TODO: this might need to return dynamic metadata instead of headers.
+func (dlb *dynamicLoadBalancer) SelectChatCompletionsEndpoint(model string, _ x.ChatCompletionMetrics) (
+	selected *filterapi.Backend, headers []*corev3.HeaderValueOption, err error,
+) {
+	m, ok := dlb.models[model]
+	if !ok {
+		err = fmt.Errorf("model %s is not found in the dynamic load balancer", model)
```
yuzisun, line 149: This may not be necessarily an error as `models` is optional in dlb, it could a list of backend hosts.

internal/controller/ai_gateway_route.go
```diff
@@ -708,3 +778,144 @@ func (c *AIGatewayRouteController) updateAIGatewayRouteStatus(ctx context.Contex
...
+			// Note: do not resolve the (pod) IPs at this level which will be resolved by the external processor since
+			// this is the reconciliation subroutine happening when AIServiceBackend (or its transitive referencing resources)
+			// changes.
+			//
+			// This means that if the service is headless, the external processor will resolve the Pod IPs, otherwise
+			// the static cluster IP.
+			//
+			// TODO: we assume that the cluster domain is "cluster.local" which is the default in Kubernetes.
+			// 	Read /etc/resolv.conf and use the search domain to add the suffix instead of hardcoding it.
+			dynB.Hostnames = append(dynB.Hostnames, fmt.Sprintf("%s.%s.svc.cluster.local", svc.Name, svc.Namespace))
```
MaYuan-02, line 869: How can service FQDN and regular FQDN be distinguished in extproc? They are different because, if it's a service, you need to obtain the pod IP for load balancing.

api/v1alpha1/api.go
yuzisun: if the dynamic load balancing becomes a core feature of the envoy ai gateway, then this InferencePool API will be part of the core AI Gateway API.

internal/controller/ai_gateway_route.go
mathetake: See the comment above this line, users are free to choose the type of service either headless or not, and if they choose headless, the Service DNS resolves the pod ips, which I believe what you want. is there anything I am missing?

api/v1alpha1/api.go
mathetake: yeah that's a good point and my concern as well; I think the point would be like where we enforce the cluster level load balancing (!= endpoint level one as  it cannot do the transformation,auth etc) functionality. If it won't be a part of InfExt API's scope, then we are good to go like we provider cluster level dynamic load balancing at our API and InfPool can only do the endpoint level stuff.

internal/controller/ai_gateway_route.go
MaYuan-02: Sorry for not noticing the above comment. However, could this DNS-based method lead to service endpoints not being updated promptly? I noticed that the current implementation only requests DNS when the configuration changes.

mathetake: It refreshes periodically, not only the configuration changes as you will see in the code (watcher.go). The eventual consistency is inevitable by the nature of DNS regardless of the kubernetes or not. so i don't think that's an real issue

filterapi/filterconfig.go
```diff
@@ -156,15 +156,55 @@ type RouteRule struct {
...
+	// DynamicLoadBalancing is the dynamic backend configuration which forces the AI filter to
+	// route the request to a specific endpoint (ip:port) instead of the normal cluster routing.
+	//
+	// When this is specified, the AI Gateway filter assume that the ORIGINAL_DST cluster is configured
+	// to route the request to an endpoint from `x-ai-eg-original-dst` header.
+	DynamicLoadBalancing *DynamicLoadBalancing `json:"dynamicLoadBalancing,omitempty"`
+}
+
+// DynamicLoadBalancing corresponds to InferencePool and InferenceModels belonging to the same pool.
+type DynamicLoadBalancing struct {
```
kfswain, line 178: OOC, why the indirection of InferenceModel? 

Was `Backend` preexisting and this just the faster method of adoption?

I think it's awesome, I'm actually really excited to see that you all implemented your own extension, just hoping to learn where our API may have some rough edges

api/v1alpha1/api.go
kfswain: Is there a concern with relying on the InfPool API? Our intent is to keep it simple and flexible. (as it was used here, we use it for pods in the reference implementation, but as you have done here, any inference endpoint can work so long as its being selected on.

internal/controller/infext_model.go
```diff
@@ -0,0 +1,81 @@
...
+	}, &inferencePool,
+	); err != nil {
+		if apierrors.IsNotFound(err) {
+			c.logger.Info("InferencePool not found.",
+				"namespace", im.Namespace, "name", im.Name)
+			return nil
+		}
+		return err
+	}
+	if err := c.syncInferencePoolFn(ctx, &inferencePool); err != nil {
```
kfswain, line 77: So these syncs cascade up to the `syncAIGatewayRoute` and eventually call `reconcileExtProcConfigMap` to reconcile on the inferenceModels.

We expected InferenceModels to potentially have high churn, if a user has a high number of LoRA adapters, and rolls out new versions regularly. I'm not sure how expensive the complete reconciliation func is, but just calling that out as a potential issue.

kfswain (commented):
Overall LGTM, I left a few comments. But this is awesome!

I'll run through the guide to set this up myself to try it out

filterapi/filterconfig.go
```diff
@@ -156,15 +156,55 @@ type RouteRule struct {
...
+	// Models that can be served by this backend. If not matched, the 404 is returned to the client.
+	//
+	// If multiple models are provided, the request is routed to the backend based on the weights, criticality, etc.
```
yuzisun, line 181: We also want fallback between multiple models, for example if `provisioned throughout` model does not have extra capacity we fallback to `on-demand` model. In this case routing is based on model name. A concrete example is to route to `bedrock-runtime.us-east-1.aws.com` and fallback using two different model names `model1-pt` and `model1-od`.

yuzisun: We may need to extend InferenceModel API to add a field `fallbackModels`

internal/controller/controller.go
```diff
@@ -152,25 +177,45 @@ const (
...
 		k8sClientIndexBackendToReferencingAIGatewayRoute, aiGatewayRouteIndexFunc)
 	if err != nil {
 		return fmt.Errorf("failed to index field for AIGatewayRoute: %w", err)
 	}
+	if enableInfExt {
+		err = indexer(ctx, &aigv1a1.AIGatewayRoute{},
+			k8sClientIndexInferencePoolToReferencingAIGatewayRoute, aiGatewayRouteIndexFuncForInferencePool)
+		if err != nil {
+			return fmt.Errorf("failed to index field for InferencePool to AIServiceBackend: %w", err)
+		}
```
wengyao04, line 200: `"failed to index field for InferencePool to AIGatewayRoute: %w", err`

internal/controller/controller.go
```diff
@@ -152,25 +177,45 @@ const (
...
+	}
 	err = indexer(ctx, &aigv1a1.AIServiceBackend{},
 		k8sClientIndexBackendSecurityPolicyToReferencingAIServiceBackend, aiServiceBackendIndexFunc)
 	if err != nil {
-		return fmt.Errorf("failed to index field for AIServiceBackend: %w", err)
+		return fmt.Errorf("failed to index field for BackendSecurityPolicy to AIServiceBackend: %w", err)
```
wengyao04, line 205: `"failed to index field for BackendSecurityPolicy to AIGatewayRoute: %w", err`

filterapi/filterconfig.go
wengyao04: Is Models and Backends are 1-1 maping, like
model[i] -> backend[i] ?

or 
[mode0, model1, ... model_N-1] -> backend [0]
...
[mode0, model1, ... model_N-1] -> backend [N-1]

filterapi/filterconfig.go
mathetake: so this `filterapi` is intentionally trying to basically be decouple the extproc component away from k8s/gateway concepts so that's why you see the "indirection" here. The rationale is the ease of testing (ie running the e2e with the extproc) as well as to drive the adoption/use of the core AI feature outside EG (in fact we are using it in our internal Istio environment)

mathetake: to be clear in other words, this is not the configuration/control plane API and more of an implementation detail for almost all the people except for the advanced users using only the core extproc without using the control plane here.

good question anyway!

api/v1alpha1/api.go
mathetake: no concern at the moment but we will see

internal/controller/infext_model.go
mathetake: yeah i am aware of that potential overload to syncAIGatewayRoute function - it should be possible to refactor later (i am not an expert on k8s controller impl to be clear...)

filterapi/filterconfig.go
mathetake: this case N to M is the answer. InferencePool can reference multiple AIServiceBackends via Selector and one pool can have multiple models

mathetake: so the model name fallback is interesting ... I think it's not within the realm of InferencePool at the moment vs currently it only has weight based random picking (optional) cc @kfswain https://github.com/kubernetes-sigs/gateway-api-inference-extension/blob/a13a1239330ffaaafa4d0f948c00cafd106086aa/pkg/epp/handlers/request.go#L71-L72

mathetake:
cool, i think this should be almost ready!

mathetake:
thank you folks!!!!!!1

api/v1alpha1/api.go
nayihz: > InferencePool.Spec.Selector will be used to select AIServiceBackends, not pods.

IIUC, `InferencePool.Spec.Selector` is indeed used for select Pods. https://github.com/kubernetes-sigs/gateway-api-inference-extension/blob/bd9ee36450d68fb4d0d8ac4f9be4db7d1ec4fee3/pkg/epp/datastore/datastore.go#L129

nayihz: It may be that I misunderstood. In this project it did use `InferencePool.Spec.Selector` to select AIServiceBackends. Sorry for that.

Issue #494, opened 2025-03-14
Getting response_payload_too_large error for chat completion streaming response


sivanantha321:
*Description*:
I'm encountering a `response_payload_too_large` error when using AI Gateway with KServe for chat completion streaming. However, if streaming is disabled (stream=false), the response is received correctly. I also tested the KServe InferenceService without AI Gateway, and streaming works as expected.

*Repro steps*:

1. Create InferenceService
```yaml
apiVersion: serving.kserve.io/v1beta1
kind: InferenceService
metadata:
  name: llama3-1b
  namespace: kserve-test
  annotations:
    serving.kserve.io/deploymentMode: RawDeployment
spec:
  predictor:
    model:
      args:
        - '--model_name=llama3-1b'
        - '--model_id=meta-llama/Llama-3.2-1B-Instruct'
      env:
        - name: HF_TOKEN
          valueFrom:
            secretKeyRef:
              key: HF_TOKEN
              name: hf-secret
              optional: false
        - name: KSERVE_OPENAI_ROUTE_PREFIX
          value: ''
      modelFormat:
        name: huggingface
      resources:
        limits:
          cpu: '6'
          memory: 14Gi
        requests:
          cpu: '4'
          memory: 12Gi
```

2. Create AIServiceBackend

```yaml
apiVersion: aigateway.envoyproxy.io/v1alpha1
kind: AIServiceBackend
metadata:
  name: envoy-ai-gateway-openai-kserve
  namespace: default
spec:
  backendRef:
    group: ''
    kind: Service
    name: llama3-1b-predictor
    namespace: kserve-test
    port: 80
  schema:
    name: OpenAI
  timeouts:
    request: 60s
```
3. Create ReferenceGrant
```yaml
apiVersion: gateway.networking.k8s.io/v1beta1
kind: ReferenceGrant
metadata:
  name: envoy-gateway-ref-grant
  namespace: kserve-test
spec:
  from:
    - group: gateway.networking.k8s.io
      kind: HTTPRoute
      namespace: default
  to:
    - group: ''
      kind: Service
```

4. Create AIGatewayRoute
```yaml
apiVersion: aigateway.envoyproxy.io/v1alpha1
kind: AIGatewayRoute
metadata:
  name: envoy-ai-gateway
  namespace: default
spec:
  rules:
    - backendRefs:
        - name: envoy-ai-gateway-openai-kserve
          weight: 100
      matches:
        - headers:
            - name: x-ai-eg-model
[...]

PR #496, opened 2025-03-14
test: adds huge number of events in a stream

mathetake:
**Commit Message**

This adds a new test case in extproc+testupstream where the upstream server sends huge number of events. This at least ensures that the AI Gateway extproc can handle such case at least with the raw Envoy configuration given in the tests/extproc case.

Related to #494


mathetake:
lol with this now the extproc test is taking more than the e2e on a kind cluster

PR #501, opened 2025-03-16
chore: replicas using values.yaml + enable leader election

rajatvig:
**Commit Message**

Set deployment replicas based off the `values.yaml` and enable leader election for the controller.

Signed-off-by: Rajat Vig <rvig@etsy.com>


manifests/charts/ai-gateway-helm/templates/deployment.yaml
```diff
@@ -42,6 +43,7 @@ spec:
             - -logLevel={{ .Values.controller.logLevel }}
             - --extProcImage={{ .Values.extProc.repository }}:{{ .Values.extProc.tag | default .Chart.AppVersion }}
             - --extProcLogLevel={{ .Values.extProc.logLevel }}
+            - --enableLeaderElection=true
```
yuzisun, line 46: Can we make it configurable ?

rajatvig: That is possible though with > 1 replica, having leader election is required. Do you think it would be better to have that as a conditional here, i.e., always turn it on if there is > 1 replica but not add it if there is only 1?

mathetake: I think it won't hurt to enable this by default ?

arkodg: is this for the ai-controller ? then you'd have to do extra work in code, to detect the leader election and based on that, only allow the leader to generate the Gateway API resources

mathetake: yeah controller and i _believe_ it's already handled assuming reader election = true (i don't remember precisely;))

rajatvig: Would I be correct in assuming then that the change is safe to do and this pull request looks fine?

yuzisun: @rajatvig You can set the default to true in values.yaml, otherwise there is no way to disable it if needed.

rajatvig: Have changed the code per suggestion.

mathetake (approved):

PR #514, opened 2025-03-20
extproc: do not list non model header values on /models

mathetake:
**Commit Message**

Previously, /v1/models endpoint returned all header matching values, not limited to the special "x-ai-eg-model" header. This fixes it, and only lists the values from the matching rules using that header.

Follow up on #325


mathetake:
ah

nacx (approved):

PR #515, opened 2025-03-21
extproc: set extProcImage before potentially returning

aabchoo:
**Commit Message**

The scenario where `filterConfig` is nil will result in skipping the code: `d.Template.Spec.Containers[0].Image = c.extProcImage`. The container image should be updated regardless of filterConfig, so moving the logic up.

https://github.com/envoyproxy/ai-gateway/pull/447


mathetake:
can you add a regression test?

mathetake (approved):
nice!

PR #536, opened 2025-03-30
extproc: bugfix router weight negative check

xiaolin593:
**Commit Message**

Avoid negative weight panic error in extproc


yuzisun:
> when does it become negative?
> 
> > // +kubebuilder:validation:Minimum=0
> > // +kubebuilder:default=1
> > Weight int `json:"weight,omitempty"`
> 
> All these weight come from this API which default to 1 and minimum 0, so i don't think it will happen in practice, though the defensive change makes sense

I think it can still happen when running standalone ?

mathetake:
Yeah good point

mathetake (approved):

Issue #509, opened 2025-03-18
Suport for multiple AIGatewayRoutes per Gateway


mathetake:

wengyao04:
Hi, @mathetake do we split the `HTTPRoute` for each `AIServiceBackend` to address k8s gateway api maximum rules restriction ?  `AIGatewayRoute` will encounter the same restriction of k8s resource size (1 MiB) and thus need to set maximum rules.

mathetake:
yeah agree

xiaolin593:
We had a discussion, @wengyao04 will work on this. Please assign to her, thanks!

mathetake:
👍

Krishanx92:
@mathetake @wengyao04 Any update on adding multiple AIGWR in upcoming releases?

mathetake:
personally i am not working on this one - maybe @wengyao04 has some update

mathetake:
@Krishanx92 by any chance do you folks have cycles to work on this if Yao hasn't started the work yet?

mathetake:
https://github.com/envoyproxy/ai-gateway/pull/599 will refactor around the implementation detail so this should be worked on AFTER it lands, otherwise there would be huge conflicts

wengyao04:
Hi @mathetake as we will refactor the code towards sidecar + UDS extproc [https://github.com/envoyproxy/ai-gateway/pull/599](https://github.com/envoyproxy/ai-gateway/pull/599). Do you think if we can still keep one AIGatewayRoute and create mulitple underlying HTTPRoute or we still prefer multiple AIGatewayRoute

PR #537, opened 2025-03-30
controller: create HTTPRoute for each backend

wengyao04:
**Commit Message**

In this PR, I create HTTPRoute for each backend in the AIServiceBackend rules. When AIGatewayRouteController reconciles a AIGatewayRoute, it lists all the HTTPRoutes belongs to that AIGatewayRoute. It also constructs a de-duplicated backends from the rules of that AIGatewayRoute.
The controller creates new HTTRoutes for each backend if they do not exist before, and updates the HTTRoutes if they are already existed. We delete HTTPRoutes if they are not in the backends and only belong to that AIGatewayRoute.
The HTTRoute is created per backend and per AIGatewayRoute, naming convention is `<backendName>-<aiGatewayRouteName>`, it makes the logic of HTTRoute lifecycle much clearer. Otherwise, I need merge owner references during update.

Fix issue #509 

N.A.


internal/controller/ai_gateway_route.go
```diff
@@ -211,44 +211,98 @@ func (c *AIGatewayRouteController) syncAIGatewayRoute(ctx context.Context, aiGat
...
-	existingRoute := err == nil
-	if apierrors.IsNotFound(err) {
-		// This means that this AIGatewayRoute is a new one.
-		httpRoute = gwapiv1.HTTPRoute{
-			ObjectMeta: metav1.ObjectMeta{
-				Name:      aiGatewayRoute.Name,
-				Namespace: aiGatewayRoute.Namespace,
-			},
-			Spec: gwapiv1.HTTPRouteSpec{},
+	// List all HTTPRoutes in the same namespace as this aiGatewayRoute.
```
yuzisun, line 215: We can move to a separate function `reconcileHTTPRoute` as it is getting bigger

internal/controller/ai_gateway_route.go
```diff
@@ -211,44 +211,98 @@ func (c *AIGatewayRouteController) syncAIGatewayRoute(ctx context.Context, aiGat
...
 		}
-	} else if err != nil {
-		return fmt.Errorf("failed to get HTTPRoute: %w", err)
 	}
 
-	// Update the HTTPRoute with the new AIGatewayRoute.
-	if err = c.newHTTPRoute(ctx, &httpRoute, aiGatewayRoute); err != nil {
-		return fmt.Errorf("failed to construct a new HTTPRoute: %w", err)
+	// The naming convention of HTTRoute is <backendName>-<aiGatewayRouteName>,
+	// It is impossible that a HTTRRoute belongs to two AIGatewayRoute.
```
yuzisun, line 254: Can you have two AIGatewayRoute in the same namespace?

internal/controller/ai_gateway_route.go
```diff
@@ -211,44 +211,98 @@ func (c *AIGatewayRouteController) syncAIGatewayRoute(ctx context.Context, aiGat
...
+			}
+			var backend *aigv1a1.AIServiceBackend
+			backend, err = c.backend(ctx, aiGatewayRoute.Namespace, br.Name)
+			if err != nil {
+				return fmt.Errorf("AIServiceBackend %s not found", key)
+			}
+			backends[key] = backend
+			// Choose default backend whose name is alphabetically smallest.
+			if defaultBackend == nil || defaultBackend.Name > backend.Name {
+				defaultBackend = backend
```
yuzisun, line 248: What's the defaultBackend for?

internal/controller/ai_gateway_route.go
wengyao04: Yes, it is possible that there are two AIGatewayRoute in the same namespace, and they have some same backends. In this implementation we create HTTPRoute per backend for each AIGatewayRoute.

mathetake:
i believe we agreed that instead of making HTTPRoute for each Backend, we should support multiple AIGWR attached to one GW that will solve the limitation that this was initially trying to solve

PR #552, opened 2025-04-03
controller: surface IsNotFound error in GetPreRotationTime

### LINKED ISSUE #466: Surface IsNotFound error in GetPreRotationTime
*Description*:

`GetPreRotationTime` should surface `apierrors.IsNotFound` instead of swallowing the error. Callers of the function should be in charge of handling the error. 

https://github.com/envoyproxy/ai-gateway/blob/9efcd32dc541b340aa6b3dff85eaffeb31792a34/internal/controller/rotators/aws_oidc_rotator.go#L106-L107

Context:

For OIDC backendSecurityPolicies, the controller is in charge of storing the temporary credentials to access the provider (AWS is only currently supported). The controller does this by storing the credentials (obtained from STS in AWS) in a Kubernetes secret. The secret will be created by the controller if it was previously deleted OR if a new BackendSecurityPolicy was created (or policy was updated to use OIDC).

[optional *Relevant Links*:]

This line will need to check if err is NotFound, and if so, rotate credentials.
https://github.com/envoyproxy/ai-gateway/blob/main/internal/controller/backend_security_policy.go#L97

johnlanda:
**Commit Message**

This commit ensures that an `apierrors.IsNotFound` is not discarded in the `GetPreRotationTime` method.
Instead, the error is returned to the caller. This commit also updates the `executeRotation` method to
maintain the preexisting behavior by allowing the `apierrors.IsNotFound` case trigger the token rotation.

Fixes #466 

This commit does not add a new test for the `executeRotation` method as the `TestBackendSecurityPolicyController_ReconcileOIDC_Fail` already depends on this behavior.

### DIFF
--- a/internal/controller/backend_security_policy.go
+++ b/internal/controller/backend_security_policy.go
@@ -153,7 +153,7 @@ func (c *BackendSecurityPolicyController) executeRotation(ctx context.Context, r
 	requeue := time.Minute
 	var rotationTime time.Time
 	rotationTime, err = rotator.GetPreRotationTime(ctx)
-	if err != nil {
+	if err != nil && !apierrors.IsNotFound(err) {
 		c.logger.Error(err, "failed to get rotation time, retry in one minute")
 	} else {
 		if rotator.IsExpired(rotationTime) {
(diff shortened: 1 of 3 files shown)


mathetake (approved):
Thank you!

PR #565, opened 2025-04-07
chore: upgrade github.com/openai/openai-go

### LINKED ISSUE #545: test: upgrades OpenAI Go SDK to beta
*Description*:

See https://github.com/envoyproxy/ai-gateway/pull/541#issuecomment-2769644359, basically we need a bit of non trivial tweaks across tests using OpenAI Go SDK

davidxia:
**Commit Message**
to 0.1.0-beta.6 from 0.1.0-alpha.65

closes #545

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -909,8 +909,8 @@ github.com/onsi/gomega v1.8.1/go.mod h1:Ho0h+IUsWyvy1OpqCwxlQ/21gkhVunqlU8fDGcoT
 github.com/onsi/gomega v1.10.1/go.mod h1:iN09h71vgCQne3DLsj+A5owkum+a2tYe+TOCB1ybHNo=
 github.com/onsi/gomega v1.36.2 h1:koNYke6TVk6ZmnyHrCXba/T/MoLBXFjeC1PtvYgw0A8=
 github.com/onsi/gomega v1.36.2/go.mod h1:DdwyADRjrc825LhMEkD76cHR5+pUnjhUN8GlHlRPHzY=
-github.com/openai/openai-go v0.1.0-alpha.65 h1:G12sA6OaL+cVMElMO3m5RVFwKhhg40kmGeGhaYZIoYw=
-github.com/openai/openai-go v0.1.0-alpha.65/go.mod h1:3SdE6BffOX9HPEQv8IL/fi3LYZ5TUpRYaqGQZbyk11A=
+github.com/openai/openai-go v0.1.0-beta.6 h1:JquYDpprfrGnlKvQQg+apy9dQ8R9mIrm+wNvAPp6jCQ=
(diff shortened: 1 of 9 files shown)


go.sum
davidxia, line 1: Should this file be marked as generated with .gitattributes?

mathetake (approved):
Thank you so much!!

mathetake:
I should've kicked the real provider tests on this PR but looks like something going wrong 

https://github.com/envoyproxy/ai-gateway/actions/runs/14315945561/job/40230187725

cc @yuzisun @xiaolin593  looks like I _guess_ there's some bug in OpenAI->Bedrock translation as the change here is only changing the client side code

davidxia:
relevant error logs

```console
real_providers_test.go:264: Appended tool message: {<nil> <nil> <nil> <nil> 0xc000185a40 <nil> {{<nil>}}}
real_providers_test.go:274: error during second response: POST "http://localhost:1062/v1/chat/completions": 400 Bad Request {"type":"ValidationException:http://internal.amazon.com/coral/com.amazon.bedrock/","code":"400","message":"1 validation error detected: Value at 'messages.3.member.content.1.member.toolResult.toolUseId' failed to satisfy constraint: Member must satisfy regular expression pattern: [a-zA-Z0-9_-]+"}
```

So it's probably related to this change. https://github.com/envoyproxy/ai-gateway/pull/565/files#diff-4891808a40887f5064c582db6a34bcc5ccd7aaf766446d4215c7925265493552L264-R263

Is there a way to repro locally? `make test-extproc` on my Macbook succeeded. Do we need to set real keys for `TEST_AWS_ACCESS_KEY_ID`, `TEST_AWS_SECRET_ACCESS_KEY`, and `TEST_OPENAI_API_KEY`?

```shell
$ make test-extproc 
-> Building extproc for darwin/arm64
<- Built out/extproc-darwin-arm64
-> Building extproc_custom_router for darwin/arm64
<- Built out/extproc_custom_router-darwin-arm64
-> Building extproc_custom_metrics for darwin/arm64
<- Built out/extproc_custom_metrics-darwin-arm64
-> Building testupstream for darwin/arm64
<- Built out/testupstream-darwin-arm64
Run ExtProc test
ok      github.com/envoyproxy/ai-gateway/tests/extproc  21.263s
```

mathetake:
Yeah you need real AWS credentials to run that test... (OpenAI is not needed tho)

davidxia:
> Yeah you need real AWS credentials to run that test... (OpenAI is not needed tho)

Ah does the test simply not run if you don't set those three env vars? Is the console output misleading in that it's not actually running the test?

mathetake:
On main branch it's running as well as PR coming directly from main repository (non fork). So usually PRs like this do not run the tests unless i put a special label "safe-to-test", otherwise anyone can steal credentials

mathetake:
By default the extproc tests are running cases not requiring credentials (test upstream cases) so the CI saying its passing is not altogether false

davidxia:
@mathetake PTAL https://github.com/envoyproxy/ai-gateway/pull/566

Issue #586, opened 2025-04-22
helm chart version format inconsistent with envoy-gateway


kerthcet:
*Description*:
>Describe the issue.

When adding helm chat dependencies, I just found the difference here:

```
dependencies
  - name: gateway-helm
    version: 0.0.0-latest
    repository: "oci://registry-1.docker.io/envoyproxy/"
    namespace: envoy-gateway-system
    condition: envoy-gateway.enabled
  - name: ai-gateway-helm
    version: v0.0.0-latest
    repository: "oci://registry-1.docker.io/envoyproxy/"
    namespace: envoy-ai-gateway-system
    condition: envoy-ai-gateway.enabled
```

In envoy-gateway, the version has no `v` but ai gateway does, perhaps we should unify them.

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

mathetake:
huh interesting - thoughts? @arkodg

arkodg:
`v0.0.0-latest` and `0.0.0-latest` are both supported which should point to the same image
https://hub.docker.com/r/envoyproxy/gateway-helm/tags

kerthcet:
From my local env, the 0.0.0-latest seems don't work for ai-gateway-helm, that's why I found the burr, but I will verify tomorrow.

kerthcet:
![Image](https://github.com/user-attachments/assets/4d5d1257-d3d4-4a17-b281-7e81cdefbd82)

As show above, for ai-gateway-helm, we only have `v<version>`  tags. Something different here.

mathetake:
cool, @kerthcet do you mind contributing a patch to push both with and without `v` prefix to the repository?

kerthcet:
Created one here: https://github.com/envoyproxy/ai-gateway/pull/593

PR #593, opened 2025-04-28
ci: push helm chart without v as well on release

kerthcet:
**Commit Message**

Once release, for example the tag is v1.2.3, both `v1.2.3` and ` 1.2.3` helm chart will be pushed.

fix: https://github.com/envoyproxy/ai-gateway/issues/586

None


mathetake (approved):

.github/workflows/release.yaml
```diff
@@ -29,7 +30,9 @@ jobs:
           username: ${{ vars.DOCKERHUB_USERNAME }}
           password: ${{ secrets.DOCKERHUB_PASSWORD }}
       - name: Push Helm chart
-        run: make helm-push HELM_CHART_VERSION=${HELM_CHART_VERSION}
+        run: |
+          make helm-push HELM_CHART_VERSION=${HELM_CHART_VERSION}
+          make helm-push HELM_CHART_VERSION=${HELM_CHART_VERSION_WITHOUT_V}
```
mathetake, line 35: I think to be consistent, we also should make the same change to https://github.com/envoyproxy/ai-gateway/blob/895d8958f05b10d5c0ac81ede94724b36a2f477a/.github/workflows/tests.yaml#L212 for the latest v0.0.0

kerthcet: Fair, updated.

Issue #588, opened 2025-04-24
Configurations for `owned_by` and `created` in /v1/models ep


kerthcet:
*Description*:
>Describe the desired behavior, what scenario it enables and how it
would be used.

We hope to make the fields `owned_by` and `created`  configurable. 

- For `owned_by`, I think we can add a field named like `OwnedBy` to the AIServiceBackendSpec, which is default to `Envoy AI Gateway` as displayed today.

- For `created`, the value from our system is from another CRD, could that possible by referring to a CRD object, and exact the `creationTimestamp`, however, this requires additional RBAC configurations.

Thanks!

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

mathetake:
let me confirm my understanding - are you saying that you want to configure 

https://github.com/envoyproxy/ai-gateway/blob/0f5da46339c36875e5c03f2b30731b169ef49346/internal/extproc/models_processor.go#L45-L48


this TODOs right?

mathetake:
cc @nacx

kerthcet:
Yes @mathetake , also parts of the project plans.

kerthcet:
Any thoughts?

mathetake:
i think we can simply change `Envoy AI Gateway` to use more information from AIServiceBackend like `Envoy AI Gateway (backend=${AIServicebackend.name}` or whatever format and use the stamp at the time of reconciliation as you suggested.

Then you won't need a control plane API for this but we can achieve this by modifying the filterapi (which is the configuration api for extproc): 

https://github.com/envoyproxy/ai-gateway/blob/main/filterapi/filterconfig.go

kerthcet:
For timestamp, I'm ok to use the AIServiceBackend creationTimestamp and I think it's convincing since the model is accessible only when AIServiceBackend is there. But I didn't check the code, would that lead to another query to fetch the AIServiceBackend?

But for the model provider, I don't think the AIServiceBackend can provide any information at this moment, for example, for OpenAI provider, the OwnedBy might be OpenAI, but for self host models, the OwnedBy might be the platform name. We need to extract something either from the field or from the annotation I think.

mathetake:
> For timestamp, I'm ok to use the AIServiceBackend creationTimestamp and I think it's convincing since the model is accessible only when AIServiceBackend is there. But I didn't check the code, would that lead to another query to fetch the AIServiceBackend?

[...]

PR #599, opened 2025-05-02
feat: cross Backend failover/fallback and retry support

mathetake:
**Commit Message**

This commit is a relatively large refactoring of internals to make Envoy AI Gateawy's API more aligned with Envoy Gateway's BackendTrafficPolicy as well as HTTPRoute. Specifically, the main objective here to allow failover and retires to work well across multiple AIServiceBackend. 

One of the most notable changes in this commit is that we split the extproc's logic into two phases; one is executed at the normal router level that selects a route (as opposed to the backend selection previously) and the other as the upstream filter that performs auth and transformation. In other words, Envoy AI Gateway configures two external processing filters.

As a result, users are now able to configure failover as well as the retry/fallback using Envoy Gateway's BackendTrafficPolicy attached to HTTPRoute generated by the Envoy AI Gateway. For example, this allows us to support the case where primary cluster is an Azure OpenAI and when it's failing, the AI Gateway fallbacks to AWS Bedrock with the standard Envoy Gateway configuration.

**Background**
At the Envoy configuration level, Envoy Gateway translates multiple backends in a single HTTPRoute's Rule into a single Envoy cluster whose endpoints consists of multiple Endpoint set (called `LocalityLbEndpoints` in Envoy API [1]) and each set corresponds to a Backend with priority configured. For example, very roughly speaking, the following pseudo HTTPRoute

```yaml
kind: HTTPRoute
metadata
  name: provider-fallback
spec:
  rules:
  - backendRefs:
      kind: Backend
      name: primary-backend
      kind: Backend
      name: secondary-backend
    matches:
    - path:
        type: PathPrefix
        value: /
```

will be translated as, when `secondary-backend` is marked as `fallback: true` in its Backend definition ([2]):

```yaml
- cluster:
  '@type': type.googleapis.com/envoy.config.cluster.v3.Cluster
  loadAssignment:
    clusterName: httproute/default/provider-fallback/rule/0
    endpoints:
    - lbEndpoints:
      - endpoint:
          address:
            socketAddress:
              address: primary.com
              portValue: 443
      priority: 0
    - lbEndpoints:
      - endpoint:
          address:
            socketAddress:
              address: secondary.com
              portValue: 443
      priority: 1
```

where priority is configured 0 and 1 for each primary and secondary backend. When retry or passive health check is configured, Envoy will retry or fallback into the secondary cluster.

In our API, transformation as well as upstream authentication must be performed per Backend so these logic must be inserted after this endpoint set (or LocalityLbEndpoints to be precise) is chosen by Envoy. For example, primary.com and secondary.com might have different API schema, authentication etc. Since Envoy has a specific HTTP filter chain that will be executed at this stage, which is called "upstream filters", if we insert the extproc that performs these logic, we can properly do authn/z and transformation in response to the retry attempts by Envoy natively.

From the upstream filter level external processor's perspective, it needs to know which exactly backend is chosen by the Envoy's cluster load balancing logic. We add some additional metadata information into the endpoint with EG's extension server so that the extproc can retrieve these information. We also use the extension server to insert the upstream extproc filter since currently it's not supported by EG. These logic in our extension server can be eliminated when the corresponding functionality become available in EG ([3],[4]).

**Caveats**
* Due to the limitation of EG's extension server API, AIBackendService that references k8s Service cannot be supported so we have to drop the support for it. Since there's a workaround for it, it should be fine plus EG can be fixed easily so the version after the next release should be able to revive the support.
* `aigw run` temporarily disabled until [5] is resolved
* Infernce Extension support temporarily disabled but will be revived before the next release.

[1] https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/endpoint/v3/endpoint_components.proto
[2] https://gateway.envoyproxy.io/latest/api/extension_types/#backendspec
[3] https://github.com/envoyproxy/gateway/issues/5523
[4] https://github.com/envoyproxy/gateway/issues/5351
[5] https://github.com/envoyproxy/gateway/issues/5918

Partially resolves the provider level fallbacks for #34


mathetake:
note to self: TODOs after this PR:
* Change weight to the optional int32; non-breaking change
* Add timeout at the rule level to match HTTPRoute; deprecate the AIServiceBackend level timeout
* Fix `aigw run` by running the extension server in the standalone mode
* Fix and Redo Inference Extension support
* Document that BackendRef must NOT be Service; not breaking change (there's a workaround)

examples/basic/basic.yaml
```diff
@@ -123,20 +119,6 @@ spec:
...
-  namespace: default
-spec:
-  type: AzureCredentials
-  azureCredentials:
-    clientID: AZURE_CLIENT_ID   # Replace with your Azure Client ID.
-    tenantID: AZURE_TENANT_ID   # Replace with your Azure Tenant ID.
-    clientSecretRef:
-      name: envoy-ai-gateway-basic-azure-client-secret
-      namespace: default
----
```
mathetake, line 137: TODO: revert!

mathetake:
finally passed all tests.....

internal/extproc/router/router.go
```diff
@@ -45,38 +41,8 @@ outer:
...
 		}
 	}
-	if rule == nil || len(rule.Backends) == 0 {
-		return nil, x.ErrNoMatchingRule
-	}
-	return r.selectBackendFromRule(rule), nil
-}
-
-// selectBackendFromRule selects a backend from the given rule. Precondition: len(rule.Backends) > 0.
-func (r *router) selectBackendFromRule(rule *filterapi.RouteRule) (backend *filterapi.Backend) {
```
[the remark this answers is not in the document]
mathetake: yes, all these weights are pushed to the generated HTTRRoute logic (handled by Envoy natively)

filterapi/filterconfig.go
```diff
@@ -82,12 +82,14 @@ type Config struct {
...
-	// SelectedBackendHeaderKey is the header key to be populated with the backend name by the filter
+	// SelectedRouteHeaderKey is the header key to be populated with the route name by the filter
 	// **after** the routing decision is made by the filter using Rules.
-	SelectedBackendHeaderKey string `json:"selectedBackendHeaderKey"`
+	SelectedRouteHeaderKey string `json:"selectedRouteHeaderKey"`
 	// Rules is the routing rules to be used by the filter to make the routing decision.
 	// Inside the routing rules, the header ModelNameHeaderKey may be used to make the routing decision.
 	Rules []RouteRule `json:"rules"`
+	// Backends is the list of backends to which the request should be routed to when the headers match.
+	Backends []*Backend `json:"backends"`
```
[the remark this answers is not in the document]
mathetake: well, not really. AIGatewayRoute: HTTPRoute is one-to-one and each rule in AIGatewayRoute will also one-to-one correspond to HTTRoute's rule. The route level extproc's only responsibility is to choose the matching rule (as opposed to the backend before). And then Envoy will route the requests to the chosen rule (== cluster binding multiple backends).

internal/extensionserver/extensionserver.go
```diff
@@ -60,47 +76,229 @@ const (
...
+	httpRouteRuleIndexStr := parts[4]
+	httpRouteRuleIndex, err := strconv.Atoi(httpRouteRuleIndexStr)
+	if err != nil {
+		s.log.Error(err, "failed to parse HTTPRoute rule index",
+			"cluster_name", cluster.Name, "rule_index", httpRouteRuleIndexStr)
+		return
+	}
+	// Get the HTTPRoute object from the cluster name.
+	var aigwRoute aigv1a1.AIGatewayRoute
+	err = s.k8sClient.Get(context.Background(), client.ObjectKey{Namespace: httpRouteNamespace, Name: httpRouteName}, &aigwRoute)
```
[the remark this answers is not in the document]
mathetake: It won't split into multiple HTTPRoutes as I commented above. The size limit thingy can be fixed with the support for multiple AIGatewayRoutes per Gateway as we discussed in the slack with Yao and others. Keeping AIGatewayRoute: HTTPRoute = 1: 1 is much simpler from UX perspective as well since you can create the retry/fallback policy for the generated one HTTPRoute as in the example https://github.com/envoyproxy/ai-gateway/pull/599/commits/3549245ed132e8e172a87a2ea5ce7f8022241a08

mathetake:
PTAL @yuzisun @arkodg @wengyao04 

cc @yanavlasov

filterapi/filterconfig.go
```diff
@@ -145,13 +143,16 @@ type HeaderMatch = gwapiv1.HTTPHeaderMatch
 // besides the `Backends` field is modified to abstract the concept of a backend
 // at Envoy Gateway level to a simple name.
 type RouteRule struct {
+	// Name is the name of the route rule.
+	Name RouteRuleName `json:"name"`
 	// Headers is the list of headers to match for the routing decision.
 	// Currently, only exact match is supported.
 	Headers []HeaderMatch `json:"headers"`
-	// Backends is the list of backends to which the request should be routed to when the headers match.
```
kerthcet, line 151: I think still possible with https://github.com/envoyproxy/ai-gateway/issues/588#issuecomment-2842370535 here?

internal/controller/ai_gateway_route.go
```diff
@@ -437,47 +435,41 @@ func (c *AIGatewayRouteController) newHTTPRoute(ctx context.Context, dst *gwapiv
...
-				timeouts = backend.Spec.Timeouts
+				backendRefs = append(backendRefs,
+					gwapiv1.HTTPBackendRef{BackendRef: gwapiv1.BackendRef{
+						BackendObjectReference: backend.Spec.BackendRef,
+						// TODO: change the weight's type to int32 in the API.
+						Weight: ptr.To[int32](int32(br.Weight)), // nolint:gosec
+					}},
+				)
+				// For now, if there're multiple backends with different timeouts, we just take the first one.
+				// TODO: deprecate the AIServiceBackend timeout and move it to Rule level.
```
kerthcet, line 461: Do you mean the AIGatewayRoute.rules?

mathetake: https://github.com/envoyproxy/ai-gateway/pull/599/commits/372b671331c8fd2064b81cbdbbb225660b4d98aa clarified

filterapi/filterconfig.go
mathetake: yeah it still should be possible even after this PR, just that the `backends` moved to the top level vs under Rules here.

filterapi/filterconfig.go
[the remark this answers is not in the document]
mathetake: >  we are dropping the capability of making the routing decision in extproc from a list of backend refs if I understand correctly.

Not really. If we want to implement `loadBalancer: latency` something like this after this PR, we can still translate add a original_dst cluster and then set a specific backend in the header from the ext proc. it should be doable.

internal/extensionserver/extensionserver.go
```diff
@@ -60,47 +76,207 @@ const (
...
+		}
+		po.HttpFilters = append(po.HttpFilters, upstreamCodec)
+	}
+	cluster.TypedExtensionProtocolOptions[httpProtocolOptions] = mustToAny(po)
+}
+
+func mustToAny(msg proto.Message) *anypb.Any {
+	b, err := proto.Marshal(msg)
+	if err != nil {
+		panic(fmt.Sprintf("BIG: failed to marshal message: %v", err))
```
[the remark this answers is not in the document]
mathetake: good catch

yuzisun (approved):
Awesome work !!!

filterapi/filterconfig.go
```diff
@@ -82,12 +78,14 @@ type Config struct {
...
-	// SelectedBackendHeaderKey is the header key to be populated with the backend name by the filter
+	// SelectedRouteHeaderKey is the header key to be populated with the route name by the filter
 	// **after** the routing decision is made by the filter using Rules.
-	SelectedBackendHeaderKey string `json:"selectedBackendHeaderKey"`
+	SelectedRouteHeaderKey string `json:"selectedRouteHeaderKey"`
 	// Rules is the routing rules to be used by the filter to make the routing decision.
 	// Inside the routing rules, the header ModelNameHeaderKey may be used to make the routing decision.
 	Rules []RouteRule `json:"rules"`
+	// Backends is the list of backends to which the request should be routed to when the headers match.
+	Backends []*Backend `json:"backends"`
```
kerthcet, line 88: Sorry, How we map the Backends with the headers now? Seems they're separated, no obvisous relationship between them.

mathetake: oh you are right. that's necessary for #588 right? we need to store the route->backends info somewhere in here. would you mind sending a patch?

PR #605, opened 2025-05-10
api: changes the type of Weight to match GWAPI

mathetake:
**Commit Message**

This changes the type of Weight of each Backend so that it matches the corresponding one in GWAPI. This is not a breaking change but more of an internal implementation change, especially in the translator.


mathetake:
PTAL @arkodg @yuzisun

arkodg (approved):

PR #613, opened 2025-05-14
ci: use EG v1.4.0 on e2e test

mathetake:
**Commit Message**

EG v1.4.0 has just been released.

### DIFF
--- a/tests/e2e/e2e_test.go
+++ b/tests/e2e/e2e_test.go
@@ -22,9 +22,9 @@ import (
 )
 
 const (
-	egLatest      = "v0.0.0-latest" // This defaults to the latest dev version.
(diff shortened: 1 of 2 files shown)


tests/e2e/e2e_test.go
```diff
@@ -22,9 +22,9 @@ import (
 )
 
 const (
-	egLatest      = "v0.0.0-latest" // This defaults to the latest dev version.
-	egNamespace   = "envoy-gateway-system"
-	egDefaultPort = 10080
+	egDefaultVersion = "v1.4.0"
```
mathetake, line 25: i switched the default local version to the stable one as it can be a source of flake (if something broken on EG repo)

yuzisun (approved):

PR #618, opened 2025-05-16
ci: create separate crds chart for ai-gateway-helm

### LINKED ISSUE #513: Separate CRDs and Resources Charts for deployment
*Description*:
>Describe the desired behavior, what scenario it enables and how it would be used.
Make new CRDs change take effect even previous CRDs are installed during deployment

Helm doesn't support CRDs upgrade/delete and will not apply new CRDs change - 

Existing structure as below
```
├── charts
│   └── ai-gateway-helm
│       ├── Chart.yaml
│       ├── crds
│       │   ├── aigateway.envoyproxy.io_aigatewayroutes.yaml
│       │   ├── aigateway.envoyproxy.io_aiservicebackends.yaml
│       │   └── aigateway.envoyproxy.io_backendsecuritypolicies.yaml
│       ├── templates
│       │   ├── NOTES.txt
│       │   ├── _helpers.tpl
│       │   ├── deployment.yaml
│       │   ├── service.yaml
│       │   └── serviceaccount.yaml
│       └── values.yaml
```
Proposed Solution:

Separate CRDs and Resources into separate Charts as [Helm doc](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/#method-2-separate-charts.) suggests e.g. aigateway-crd, aigateway-resources similar to what [KServe](https://github.com/kserve/kserve/tree/master/charts) does

[optional *Relevant Links*:]

yuzisun:
**Commit Message**
Separate envoy ai gateway crds to its own chart so helm upgrade can update the crd when there are changes. 

Fixes #513


aabchoo:
You will need to update the output path of apigen to the new chart

https://github.com/envoyproxy/ai-gateway/blob/a17174379aec6d05adc8af7fc7e1278124d4c904/Makefile#L97-L100

Update that line to:
```
	@go tool controller-gen object crd paths="./api/v1alpha1/..." output:dir=./api/v1alpha1 output:crd:dir=./manifests/charts/ai-gateway-crd-helm/templates
```

aabchoo (changes requested):
https://github.com/envoyproxy/ai-gateway/pull/618/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L100

aabchoo (approved):
lgtm!

mathetake:
shouldn't we need to push the new crd chart to docker as well ?

yuzisun:
> shouldn't we need to push the new crd chart to docker as well ?

Good catch! done.

mathetake:
i _guess_ we need to create a new dockerhub repo named `ai-gateway-crd-helm` like https://hub.docker.com/r/envoyproxy/ai-gateway-helm but not 100% sure. i don't have a permission to create one, so let's see how it goes by just merging this. if it fails, i will ask admin to create one

mathetake:
ah one last thing; could you change the below so we upload the gzip crd charts to GH releases

https://github.com/envoyproxy/ai-gateway/blob/7b583f04f9a6bbc693d5cb35c937f4dffe1291bb/.github/workflows/release.yaml#L39

https://github.com/envoyproxy/ai-gateway/blob/7b583f04f9a6bbc693d5cb35c937f4dffe1291bb/.github/workflows/release.yaml#L45

.github/workflows/release.yaml
```diff
@@ -36,12 +36,20 @@ jobs:
 
       - name: Create a release candidate
         if: ${{ contains(github.ref, '-rc') }}
-        run: gh release create $TAG --prerelease --title $TAG --notes "Release candidate" ./out/ai-gateway-helm-${TAG}.tgz
```
mathetake, line 39: i think simply adding `./out/ai-gateway-crd-helm-${TAG}.tgz` to this line should be enough? the current change doesn't quite make sense to me.
```
USAGE
  gh release create [<tag>] [<files>...]
```

mathetake:
this also needs a documentation change i guess...

mathetake:
https://github.com/envoyproxy/ai-gateway/blob/main/site/docs/getting-started/installation.md here

.github/workflows/release.yaml
```diff
@@ -37,19 +37,15 @@ jobs:
       - name: Create a release candidate
         if: ${{ contains(github.ref, '-rc') }}
         run: |
-          gh release create $TAG --prerelease --title $TAG --notes "Release candidate" ./out/ai-gateway-crd-helm-${TAG}.tgz
-          gh release create $TAG --prerelease --title $TAG --notes "Release candidate" ./out/ai-gateway-helm-${TAG}.tgz
-          gh release upload $TAG ./out/ai-gateway-crd-helm-${TAG}.tgz
-          gh release upload $TAG ./out/ai-gateway-helm-${TAG}.tgz
+          gh release create $TAG --prerelease --title $TAG --notes "Release candidate" ./out/ai-gateway-crd-helm-${TAG}.tgz ./out/ai-gateway-helm-${TAG}.tgz
+          gh release upload $TAG ./out/ai-gateway-crd-helm-${TAG}.tgz ./out/ai-gateway-helm-${TAG}.tgz
```
mathetake, line 41: i think `upload` wouldn't be needed but i think harmless so i will go ahead and merge

mathetake (approved):

mathetake:
> i _guess_ we need to create a new dockerhub repo named `ai-gateway-crd-helm` like https://hub.docker.com/r/envoyproxy/ai-gateway-helm but not 100% sure. i don't have a permission to create one, so let's see how it goes by just merging this. if it fails, i will ask admin to create one

yeah the push is failing so now i am asking the admin to create another dockerhub rep...

yuzisun:
> > i _guess_ we need to create a new dockerhub repo named `ai-gateway-crd-helm` like https://hub.docker.com/r/envoyproxy/ai-gateway-helm but not 100% sure. i don't have a permission to create one, so let's see how it goes by just merging this. if it fails, i will ask admin to create one
> 
> yeah the push is failing so now i am asking the admin to create another dockerhub rep...

thanks!

mathetake:
https://github.com/envoyproxy/ai-gateway/pull/619 PTAL folks - after i chatted with the admin, we created the repo but with the plural "ai-gateway-crds-helm" to match the one with EG.

PR #620, opened 2025-05-18
Add ownedBy and createdAt fields to AIGatewayRouteRuleBackendRef

kerthcet:
**Commit Message**

This commit added new fields CreatedAt & OwnedBy to AIGatewayRoute as well as the internal Backend, so people can customize the output of OpenAI-compatible APIs.

Fixes: https://github.com/envoyproxy/ai-gateway/issues/588

Still need to handle the issue of the map between headers and backends.


kerthcet:
API definitions only now. cc @mathetake  for feedbacks.

mathetake:
sorry i've been sick & I needed to focus on the important refactoring in #629 ... will check in tomorrow

mathetake (commented):
i didn't give much thought but am i understanding correctly that this assumes that "model maps to a unique backend? my concern is mainly about this comment https://github.com/envoyproxy/ai-gateway/issues/588#issuecomment-2877264085

kerthcet:
> i didn't give much thought but am i understanding correctly that this assumes that "model maps to a unique backend? my concern is mainly about this comment [#588 (comment)](https://github.com/envoyproxy/ai-gateway/issues/588#issuecomment-2877264085)

~Yes, for different backends, like the platform and OpenAI, the owner may differ. Since the `Owner` works at backend level, for fallback examples, it still works as expected, what we do is just ad different owner for each backend, for example:~
```
 backendRefs: 
   - name: provider-fallback-always-failing-upstream  # This is the primary backend and trying to speak TLS, which always fails. 
     # ownedBy: # default to Envoy AI Gateway if not specified
   - name: provider-fallback-aws 
     ownedBy: "aws"
```

RE: 

For fallback case, what I think a bit tricky here is when querying the model list, we have no idea whether the services(backends) is ready or not, which means we have no idea who's exactly the model owner among all the backends (the same for all kinds of multi-backends, like weight based routing). So mapping the modelOwner:backend = 1:1 seems useless, I may think more about this.

kerthcet:
I think there're three approaches:
- header level: this will match with the owner one by one, but it requires to extend the `AIGatewayRouteRuleMatch.Headers` to include the `Owner` field but seems a little weird
- backend level: as we implemented in this PR, however, we may have same model with different owners in `/models` request. But actually there's no difference between them when querying the chat api since we only need the model name.
- rule level: we'll add these fields to `AIGatewayRouteRule` and different backends will share the same Owner, not a perfect idea but seems not bad.

internal/apischema/openai/openai.go
```diff
@@ -790,6 +790,6 @@ func (t *JSONUNIXTime) UnmarshalJSON(s []byte) error {
 	if err != nil {
 		return err
 	}
-	*(*time.Time)(t) = time.Unix(q, 0)
+	*(*time.Time)(t) = time.Unix(q, 0).UTC()
```
kerthcet, line 793: Change to UTC to make sure the output is comparable in tests.

kerthcet:
Ping @mathetake for review, thanks a lot.

internal/controller/ai_gateway_route.go
```diff
@@ -283,6 +285,19 @@ func (c *AIGatewayRouteController) reconcileExtProcConfigMap(ctx context.Context
...
+		ownedBy := rule.OwnedBy
+		createdAt := rule.CreatedAt
+		ec.Rules[i].OwnedBy = defaultOwnedBy
+		if ownedBy != nil {
+			ec.Rules[i].OwnedBy = *ownedBy
+		}
+		ec.Rules[i].CreatedAt = aiGatewayRoute.CreationTimestamp.Time
+		if createdAt != nil {
+			ec.Rules[i].CreatedAt = createdAt.Time
+		}
```
mathetake, line 297:
```suggestion
		ec.Rules[i].OwnedBy = ptr.Deref(rule.OwnedBy, defaultOwnedBy)
		ec.Rules[i].CreatedAt = ptr.Deref(rule.CreatedAt, aiGatewayRoute.CreationTimestamp.Time)
```

internal/controller/ai_gateway_route.go
```diff
@@ -283,6 +285,19 @@ func (c *AIGatewayRouteController) reconcileExtProcConfigMap(ctx context.Context
...
+		// convert to UTC time in force to avoid timezone issues.
```
mathetake, line 298:
```suggestion
		// Convert to UTC time in force to avoid timezone issues.
```

api/v1alpha1/api.go
```diff
@@ -227,6 +227,25 @@ type AIGatewayRouteRule struct {
...
 	Timeouts *gwapiv1.HTTPRouteTimeouts `json:"timeouts,omitempty"`
+
+	// OwnedBy represents the owner of the running model serving by the backend,
+	// which will be exported as the field of "OwnedBy" in openai-compatible API "/models".
+	//
+	// Default to "Envoy AI Gateway" if not set.
+	//
+	// +optional
+	// +kubebuilder:default="Envoy AI Gateway"
+	OwnedBy *string `json:"ownedBy,omitempty"`
```
mathetake, line 238: I would like to emphasize that this is only for "/models" endpoint. so I would suggest to prefix "models"
```suggestion
	ModelsOwnedBy *string `json:"modelsOwnedBy,omitempty"`
```

api/v1alpha1/api.go
```diff
@@ -227,6 +227,25 @@ type AIGatewayRouteRule struct {
...
+
+	// CreatedAt represents the creation timestamp of the running model serving by the backend,
+	// which will be exported as the field of "Created" in openai-compatible API "/models".
+	// It follows the format of RFC 3339, for example "2024-05-21T10:00:00Z".
+	//
+	// Default to the creation timestamp of the AIGatewayRoute if not set.
+	//
+	// +optional
+	// +kubebuilder:validation:Format=date-time
+	CreatedAt *metav1.Time `json:"createdAt,omitempty"`
```
mathetake, line 248: same suggestion 

```suggestion
	ModelsCreatedAt *metav1.Time `json:"modelsCreatedAt,omitempty"`
```

api/v1alpha1/api.go
```diff
@@ -227,6 +227,25 @@ type AIGatewayRouteRule struct {
 	//
 	// +optional
 	Timeouts *gwapiv1.HTTPRouteTimeouts `json:"timeouts,omitempty"`
+
+	// OwnedBy represents the owner of the running model serving by the backend,
+	// which will be exported as the field of "OwnedBy" in openai-compatible API "/models".
```
mathetake, line 232:
```suggestion
	// which will be exported as the field of "OwnedBy" in openai-compatible API "/models".
	// 
	// This is used only when this rule contains "x-ai-eg-model" in its header matching where the header value will be
	// recognized as a "model" in "/models" endpoint.
```

mathetake (commented):
minor comments, but I think this looks good. 

any further comments ?  @envoyproxy/ai-gateway-maintainers

kerthcet:
Comments addressed, PTAL.

kerthcet:
Kindly ping.

mathetake (approved):

mathetake:
thanks for your work @kerthcet !

kerthcet:
Thanks @mathetake

PR #627, opened 2025-05-19
extproc: fix aigateway typo for filter metadata

yuzisun:
**Commit Message**

Fix typo for endpoint filter metadata from aigateawy to aigateway


aabchoo (approved):
Nice catch - we should look into adding a const file to avoid all these magic strings

PR #628, opened 2025-05-20
api: preparing for new azure credential types

aabchoo:
**Commit Message**
`BackendSecurityPolicyAzureCredentials` needs to be refactored to support different credential types/logic. Moving client credentials used to get azure tokens via the azidentity package [1] to its own type.

The next step will to be to introduce an Azure OIDC type.

1: https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azidentity


aabchoo:
Apologies, I kept messing up the signoff message

manifests/charts/ai-gateway-crds-helm/templates/aigateway.envoyproxy.io_aigatewayroutes.yaml
```diff
@@ -1,8 +1,3 @@
-# Copyright Envoy AI Gateway Authors
-# SPDX-License-Identifier: Apache-2.0
-# The full text of the Apache license is available in the LICENSE file at
```
aabchoo, line 3: Reverting this change

api/v1alpha1/api.go
```diff
@@ -504,6 +504,13 @@ type BackendSecurityPolicyAPIKey struct {
 
 // BackendSecurityPolicyAzureCredentials contains the supported authentication mechanisms to access Azure.
 type BackendSecurityPolicyAzureCredentials struct {
+	// ClientCredentials specifies the client id/secret and tenant id used to obtain azure access token.
+	//
+	// +optional
+	ClientCredentials *AzureClientCredentials `json:"clientCredentials,omitempty"`
```
yuzisun, line 510: This is known as Microsoft Entra ID https://learn.microsoft.com/en-us/azure/ai-services/openai/reference#authentication, not sure what's proper name for this.

It is a breaking change but probably fine before 0.2 is released.

aabchoo: Credentials is associated with App Registration, so what do you think of `ApplicationIdentity`?

missBerg: Client Credentials is the generic name for ClientID and Client Secret normally is that what you are asking @yuzisun? The link you are sharing looks like it shows simply sending a Bearer Token as proof of authorization in the API request which looks like standard OAuth

missBerg: I think ClientCredentials is a good name

api/v1alpha1/api.go
```diff
@@ -504,6 +504,13 @@ type BackendSecurityPolicyAPIKey struct {
 
 // BackendSecurityPolicyAzureCredentials contains the supported authentication mechanisms to access Azure.
 type BackendSecurityPolicyAzureCredentials struct {
+	// ApplicationIdentity specifies the client id/secret and tenant id used to obtain azure access token.
+	//
+	// +optional
+	ApplicationIdentity *AzureApplicationIdentity `json:"applicationIdentity,omitempty"`
```
aabchoo, line 510: May need Client and Tenant ID for OIDC

api/v1alpha1/api.go
aabchoo: Moved TenantID and ClientID up, and this structure is now only the client secret ref. I think it'd make sense to call it clientSecretCredential?

aabchoo:
Client ID and Tenant ID is needed for OIDC token exchange as well, so I am only moving secret one level down

api/v1alpha1/api.go
```diff
@@ -534,6 +534,13 @@ type BackendSecurityPolicyAzureCredentials struct {
 	// +kubebuilder:validation:MinLength=1
 	TenantID string `json:"tenantID"`
 
+	// ClientSecretCredential specifies the client secret used to obtain azure access token.
```
jamescummins, line 537: the OAuh2 flow is Client-credentials, suggestion, the name should be `ClientCredentialSecret`

yuzisun: thanks @jamescummins !

aabchoo:
Talked with @yuzisun, and concluded making all these changes just to move clientSecretRef to its own type is overkill.

PR #630, opened 2025-05-21
api: specific type for oidc related fields

aabchoo:
**Commit Message**

Creating a generic type that includes OIDC field from envoy proxy gateway, and grant_type/aud. Embedding struct into `AWSOIDCExchangeToken` to avoid much refactoring. Reduces repetition when introducing OIDC  for other providers


sukumargaonkar:
lgtm

mathetake:
lgtm from me - assigning @yuzisun for the rest

yuzisun (approved):

Issue #609, opened 2025-05-12
Support GCP models


sukumargaonkar:
*Description*:
Support Models on GCP in ai-gateway. This includes Gemini models and Anthropic models hosted on GCP.
GCP google provided models like gemini as well as models from other provider like anthropic
Each provider has their own specific api with google specific authentication tweaks.

Some useful API features
- Function calling
- Guided Generation
- Thinking/Reasoning related requests

[optional *Relevant Links*:]
Anthropic [core api](https://docs.anthropic.com/en/api/messages) with some [GCP's specific tweaks](https://docs.anthropic.com/en/api/claude-on-vertex-ai)

Gemini Rest API: https://cloud.google.com/vertex-ai/generative-ai/docs/reference/rest/v1/projects.locations.endpoints/generateContent

sukumargaonkar:
Rough work breakdown. will add links to corresponding PRs as and when they are created

GCP Integration
- GCP Auth
    - [x] API change: https://github.com/envoyproxy/ai-gateway/pull/635
    - [x] TokenRotator change and backendauth change: https://github.com/envoyproxy/ai-gateway/pull/752

- Gemini API
    - [x] Basic Request: https://github.com/envoyproxy/ai-gateway/pull/819
    - [x] Image/PDF Support: https://github.com/envoyproxy/ai-gateway/pull/819
    - [x]  Function/Tool Calling: https://github.com/envoyproxy/ai-gateway/pull/869
    - [x]  Streaming Request: https://github.com/envoyproxy/ai-gateway/pull/900

- Anthropic API
    - [x] Basic Request: https://github.com/envoyproxy/ai-gateway/pull/838
    - [x] Image/PDF Support: https://github.com/envoyproxy/ai-gateway/pull/838
    - [x] Function/Tool Calling: https://github.com/envoyproxy/ai-gateway/pull/838
    - [x] Streaming Request
    - [x] Thinking/Reasoning

mathetake:
After all the implementation is done, we need to add additional docs under connect-providers page: https://aigateway.envoyproxy.io/docs/getting-started/connect-providers/

mathetake:
Created a sub issue for the documentation https://github.com/envoyproxy/ai-gateway/issues/1015 which is a blocker for v0.3

mathetake:
The documentation is still TODO for v0.3

yuzisun:
> The documentation is still TODO for v0.3

Will update doc by the end of week then we are good to go !

PR #635, opened 2025-05-22
api: add GCP support for backend security policies and API schemas

sukumargaonkar:
**Commit Message**
This Commit adds required fields on BackendSecurityPolicy for GCP authentication

Related Issue: https://github.com/envoyproxy/ai-gateway/issues/609

This PR only make API changes. Implementation for token rotator and backendauth will follow in future PRs


api/v1alpha1/api.go
```diff
@@ -520,6 +534,17 @@ type BackendSecurityPolicyOIDC struct {
 	Aud string `json:"aud,omitempty"`
 }
 
+// BackendSecurityPolicyGCPCredentials contains the supported authentication mechanisms to access GCP.
+type BackendSecurityPolicyGCPCredentials struct {
+	ProjectID                string `json:"projectID"`
+	ServiceAccountEmail      string `json:"serviceAccountEmail"`
```
xiaolin593, line 540: Would it make sense to use secret to avoid user email exposure?

api/v1alpha1/api.go
```diff
@@ -520,6 +534,17 @@ type BackendSecurityPolicyOIDC struct {
...
+	WorkloadIdentityPoolName string `json:"workloadIdentityPoolName"`
```
xiaolin593, line 541: Can you add a comment for it with example value?

sukumargaonkar: don't have a concrete example or a format. This is an open-ended string that user based on what the pool is named on GCP

api/v1alpha1/api.go
sukumargaonkar: The service-account-email itself need not be secret as its useless without the password/secret along with it.

api/v1alpha1/api.go
```diff
@@ -520,6 +534,32 @@ type BackendSecurityPolicyOIDC struct {
...
+	//
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinLength=1
+	WorkloadIdentityPoolName string `json:"workloadIdentityPoolName"`
+	// GCPIdentityProviderName refers to the external identity provider whose credentials your workload is using to authenticate to Google Cloud.
+	//
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinLength=1
+	GCPIdentityProviderName string `json:"GCPIdentityProviderName"`
+	// BackendSecurityPolicyOIDC is the generic OIDC fields.
```
aabchoo, line 559: Is OIDC the only way to auth against GCP?

aabchoo: Mainly asking to see if we want to add another level for OIDC type

sukumargaonkar: good point. This is not the only way. will refactor accordingly

api/v1alpha1/api.go
```diff
@@ -520,6 +534,65 @@ type BackendSecurityPolicyOIDC struct {
...
+	//
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinLength=1
+	WorkloadIdentityPoolName string `json:"workloadIdentityPoolName"`
+
+	// ServiceAccountImpersonation is the service account impersonation configuration.
+	// This is used to impersonate a service account when getting access token.
+	//
+	// +optional
+	ServiceAccountImpersonation *GCPServiceAccountImpersonationConfig `json:"serviceAccountImpersonation"`
```
aabchoo, line 562: omit empty?

api/v1alpha1/api.go
```diff
@@ -520,6 +534,65 @@ type BackendSecurityPolicyOIDC struct {
...
+}
+
+type GCPWorkloadIdentityProvider struct {
+	// Name of the external identity provider as registered on Google Cloud Platform.
```
aabchoo, line 566: Could you explain that this is a generic provider for various idP which includes OIDC?

yuzisun: Same question if name needs to be set along with OIDCConfig

examples/basic/basic.yaml
```diff
@@ -148,6 +166,28 @@ spec:
...
 ---
+apiVersion: aigateway.envoyproxy.io/v1alpha1
+kind: BackendSecurityPolicy
+metadata:
+  name: envoy-ai-gateway-basic-gcpgemini-credentials
+  namespace: default
+spec:
+  type: GCPCredentials
+  gcpCredentials:
+    workLoadIdentityFederationConfig:
```
yuzisun, line 177: Should this field be part of workloadIdentityProvider?

api/v1alpha1/api.go
yuzisun: I think here we are not only supporting OIDC not other auth provider like aws yet

sukumargaonkar: The name itself can be an arbitrary string that user registers on GCP and is not related to whether the provider is oidc or aws.

For now `GCPWorkloadIdentityProvider` struct only contains OIDCProvider as field, implying that only oidc is supported. 
in future we can add more fields to the struct for each type of provider we want to support

api/v1alpha1/api.go
```diff
@@ -527,6 +541,72 @@ type BackendSecurityPolicyOIDC struct {
 	Aud string `json:"aud,omitempty"`
 }
 
+type GCPWorkLoadIdentityFederationConfig struct {
+	// ProjectID is the GCP project ID.
+	//
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinLength=1
+	ProjectID string `json:"projectID"`
```
aabchoo, line 567: Is ProjectID specific to Federation Config? Would it make sense to bubble this field up if it's needed for all GCP auth

filterapi/filterconfig.go
```diff
@@ -222,6 +224,16 @@ type AzureAuth struct {
...
 }
 
+// GCPAuth defines the file containing GCP credential that will be mounted to the external proc.
+type GCPAuth struct {
+	// AccessToken is the access token as a literal string.
+	AccessToken string `json:"accessToken"`
+	// Region is the GCP region to use for the request.
+	Region string `json:"region"`
+	// ProjectName is the GCP project name to use for the request.
+	ProjectName string `json:"projectName"`
```
aabchoo, line 236: How will region and project be used in the extproc?

tests/crdcel/testdata/backendsecuritypolicies/gcp_oidc.yaml
```diff
@@ -0,0 +1,26 @@
...
+apiVersion: aigateway.envoyproxy.io/v1alpha1
+kind: BackendSecurityPolicy
+metadata:
+  name: envoy-ai-gateway-basic-gcp-credentials
+  namespace: default
+spec:
+  type: GCPCredentials
+  gcpCredentials:
+    workLoadIdentityFederationConfig:
+      projectID: GCP_PROJECT_ID  # Replace with your GCP project ID
```
aabchoo: remove unnecessary comments since users won't be updating this?

aabchoo (commented):
Overall looks good to me -- made a few comments

filterapi/filterconfig.go
sukumargaonkar: The region and project-id are required while transforming request path
`https://<region>-aiplatform.googleapis.com/v1/projects/<project-id>/locations/<region>/publishers/google/models/<model-name>:generateContent`

api/v1alpha1/api.go
sukumargaonkar: not sure if other authentication methods would need project-id, so keeping it within federation config for now

api/v1alpha1/api.go
```diff
@@ -434,6 +434,15 @@ const (
 	//
 	// https://learn.microsoft.com/en-us/azure/ai-services/openai/reference#api-specs
 	APISchemaAzureOpenAI APISchema = "AzureOpenAI"
+	// APISchemaGCPGemini is the schema followed by Gemini models hosted on GCP.
```
mathetake: shouldn't this be "GCPVertex" instead of gemini?

mathetake: The reason I am commenting is that the path you mentioned "https://<region>-aiplatform.googleapis.com/v1/projects/<project-id>/locations/<region>/publishers/google/models/<model-name>:generateContent" is the vertex API format 


vs

"Gemini API" just has a weird prefix: https://ai.google.dev/gemini-api/docs/openai

Such weird prefix is not unique to Gemini API but several others out there have the same one format. I am planning to add support for such providers altogether so I would like to rename this one to "GeminiVertex" for clarity.

edited: I opened the PR https://github.com/envoyproxy/ai-gateway/pull/706

sukumargaonkar: good point. switching to GeminiVertexAI and AnthropicVertexAI

mathetake: This is for API schema/ request format in the first place. Then how do you support any other model in Vertex AI? What happens when google introduces another latest models with new name? The layer you're talking about is different than Anthropic

sukumargaonkar: mybad, had misunderstood your previous comment
updated the naming to GeminiVertexAI

yuzisun: Gemini essentially is VertexAI's model endpoint API, there are other embedding, imagegen APIs according to https://cloud.google.com/vertex-ai/generative-ai/docs/model-reference/inference, so we can just name it GCPVertexAI schema like AWSBedrock.

api/v1alpha1/api.go
```diff
@@ -437,6 +437,15 @@ const (
 	//
 	// https://learn.microsoft.com/en-us/azure/ai-services/openai/reference#api-specs
 	APISchemaAzureOpenAI APISchema = "AzureOpenAI"
+	// APISchemaGeminiVertexAI is the schema followed by Gemini models hosted on GCP's Vertex AI platform.
+	//
+	// https://cloud.google.com/vertex-ai/docs/reference/rest/v1/projects.locations.endpoints/generateContent?hl=en
+	APISchemaGeminiVertexAI APISchema = "GeminiVertexAI"
```
mathetake: This should be `GCPVertexAI` - this is not addressing my previous comment and thats' what Dan suggested https://github.com/envoyproxy/ai-gateway/pull/635#discussion_r2145246822.

filterapi/filterconfig.go
mathetake: then maybe ProjectID instead of ProjectName

api/v1alpha1/api.go
```diff
@@ -437,6 +437,15 @@ const (
 	//
 	// https://learn.microsoft.com/en-us/azure/ai-services/openai/reference#api-specs
 	APISchemaAzureOpenAI APISchema = "AzureOpenAI"
+	// APISchemaGeminiVertexAI is the schema followed by Gemini models hosted on GCP's Vertex AI platform.
+	//
+	// https://cloud.google.com/vertex-ai/docs/reference/rest/v1/projects.locations.endpoints/generateContent?hl=en
```
mathetake, line 457: i think we should add the documentation about this schema that this requires the BackendSecurityPolicy to be configured and attached since this transformation will use project-id from there ?

sukumargaonkar: makes sense, can you point to where i need to add the documentation?

mathetake: sorry i meant the documentation == the code comment here. The comments in the API here will be included in the auto generated documentation

api/v1alpha1/api.go
```diff
@@ -414,7 +414,7 @@ type AIServiceBackendSpec struct {
 type VersionedAPISchema struct {
 	// Name is the name of the API schema of the AIGatewayRoute or AIServiceBackend.
 	//
-	// +kubebuilder:validation:Enum=OpenAI;AWSBedrock;AzureOpenAI
+	// +kubebuilder:validation:Enum=OpenAI;AWSBedrock;AzureOpenAI;GCPGemini;GCPAnthropic
```
yuzisun:
```suggestion
	// +kubebuilder:validation:Enum=OpenAI;AWSBedrock;AzureOpenAI;GCPVertexAI;GCPAnthropic
```

examples/basic/basic.yaml
```diff
@@ -91,6 +91,24 @@ spec:
...
 ---
+# GCP Example
+apiVersion: aigateway.envoyproxy.io/v1alpha1
+kind: AIServiceBackend
+metadata:
+  name: envoy-ai-gateway-basic-gcp
+  namespace: default
+spec:
+  schema:
+    name: GCPGemini
```
yuzisun:
```suggestion
    name: GCPVertexAI
```

manifests/charts/ai-gateway-crds-helm/templates/aigateway.envoyproxy.io_aigatewayroutes.yaml
```diff
@@ -505,6 +505,8 @@ spec:
                     - OpenAI
                     - AWSBedrock
                     - AzureOpenAI
+                    - GCPGemini
```
yuzisun:
```suggestion
                    - GCPVertexAI
```

manifests/charts/ai-gateway-crds-helm/templates/aigateway.envoyproxy.io_aiservicebackends.yaml
```diff
@@ -177,6 +177,8 @@ spec:
                     - OpenAI
                     - AWSBedrock
                     - AzureOpenAI
+                    - GCPGemini
```
yuzisun:
```suggestion
                    - GCPVertexAI
```

filterapi/filterconfig.go
```diff
@@ -224,6 +226,16 @@ type AzureAuth struct {
...
```
yuzisun, line 236:
```suggestion
	// ProjectName is the GCP project id to use for the request.
	ProjectID string `json:"projectID"`
```

sukumargaonkar: Apparently this needs to be project ID and not project name.
[docs](https://cloud.google.com/iam/docs/service-accounts-create#creating)

The service account's email address in gcp follow below format
SERVICE_ACCOUNT_NAME@PROJECT_ID.iam.gserviceaccount.com

we use the service-account email when getting the token

filterapi/filterconfig.go
sukumargaonkar: Tried using project ID in the request url, but kept getting `IAM_PERMISSION_DENIED` error 
not sure if that is an account specific configuration issue or GCP just doesn't like project ID instead of project-name

if future project name is found to be acceptable, we can add that as another field and accept either of them as valid config

yuzisun: looks like it is using the name from the examples

https://aiplatform.googleapis.com/v1/projects/test-project/locations/global/publishers/google/models/gemini-2.0-flash-001:generateContent

yuzisun: if project name is currently required on the url path, then it needs to be added to the BackendSecurityPolicy as well?

mathetake (approved):

PR #644, opened 2025-05-28
refactor: move resp handling to router from upstream level

mathetake:
**Commit Message**


mathetake:
yeah tests passing

internal/extproc/chatcompletion_processor.go
```diff
@@ -70,6 +78,22 @@ type chatCompletionProcessorRouterFilter struct {
 	upstreamFilterCount int
 }
 
+// ProcessResponseHeaders implements [Processor.ProcessResponseHeaders].
+func (c *chatCompletionProcessorRouterFilter) ProcessResponseHeaders(ctx context.Context, headerMap *corev3.HeaderMap) (*extprocv3.ProcessingResponse, error) {
+	if c.upstreamFilter != nil { // See the comment on the "upstreamFilter" field.
+		return c.upstreamFilter.ProcessResponseHeaders(ctx, headerMap)
+	}
+	return c.passThroughProcessor.ProcessResponseHeaders(ctx, headerMap)
```
yuzisun, line 86: Why do we need to call passThroughProcessor ?

internal/extensionserver/extensionserver.go
```diff
@@ -246,11 +247,6 @@ func (s *Server) maybeModifyCluster(cluster *clusterv3.Cluster) {
 		},
 		Timeout: durationpb.New(30 * time.Second),
 	}
-	extProcConfig.MetadataOptions = &extprocv3http.MetadataOptions{
```
yuzisun, line 249: This is not needed?

internal/extproc/chatcompletion_processor.go
mathetake: https://github.com/envoyproxy/ai-gateway/pull/644/commits/67b9e1de6f6249972f9121b72df2fb57cc9ae944 good q and added comment

PR #645, opened 2025-05-28
feat: merge token ratelimit request cost type

Xunzhuo:
**Commit Message**

feat: merge token ratelimit request cost type


mathetake:
i am not sure what you want to achieve with this but generally we wanted to avoid bringing in the control plane level concepts/dependency into extproc and that's why we almost-redefine many things in filterapi. so i am not sure we could approve this unless there's absolute necessity

Xunzhuo:
@mathetake just thought we dont need to redefine everything twice like the cost type, we need to copy that every time we add one type to two places

mathetake:
yeah, i understand the motivation but at the same time not untenable either. so let me close this one for now and let's revisit if it gets really hairy

PR #655, opened 2025-05-30
controller: set extproc limits 

mathetake:
**Commit Message**

This adds a missing cluster-level configuration limit so that the extproc cluster has sane default configurations.


internal/controller/gateway.go
```diff
@@ -156,14 +157,25 @@ func (c *GatewayController) ensureExtensionPolicy(ctx context.Context, gw *gwapi
...
+							Namespace: ptr.To(gwapiv1.Namespace(gw.Namespace)),
+						},
+					}},
+					BackendSettings: &egv1a1.ClusterSettings{
+						Connection: &egv1a1.BackendConnection{
+							// Default is 32768 bytes == 32 KiB which seems small:
+							// https://github.com/envoyproxy/gateway/blob/932b8b55fa562ae917da19b497a4370733478f1/internal/xds/translator/cluster.go#L49
+							//
+							// So, we set it to 50MBi.
+							BufferLimit: ptr.To(resource.MustParse("50Mi")),
```
aabchoo, line 175: Will we need to bump https://github.com/envoyproxy/ai-gateway/pull/655/files#diff-b00c95bb5a08e5d0d7a300a51ae2d9c4bc334f505ee1b1f087f256e92fb276b4R108 if we bump this line?

mathetake: yeah both needed as they are separate clusters at envoy level

mathetake: good point

PR #445, opened 2025-03-01
feat: add Azure OpenAI backend handler

xiaolin593:
**Commit Message**

Implement Azure backend auth.

**Related Issues/PRs (if applicable)**
Related Issue: #347 
Related PR: #424

Issue #531, opened 2025-03-27
site: Connect Azure OpenAI


xiaolin593:
*Description*:
Write **Connect Azure OpenAI** documentation.

@xiaolin593 to create a guide for user to set up Azure OpenAI similar to https://aigateway.envoyproxy.io/docs/getting-started/connect-providers/aws-bedrock

PR #658, opened 2025-05-30
feat: adding support for accessing azure via oidc

aabchoo:
**Commit Message**

Updating support for accessing azure via OIDC in addition to the already implemented access via access secret. Token retrieved from OIDC will be exchanged for an azure access token which will be store the same way as getting token via client secret/id and tenant id.

Key changes
- renamed azure_token_provider -> azure_secret_token_provider (since it's client secret specific)
- generalized new azure_token_provider to have TokenProviders provider federated tokens
- Azure tries to create token provider from oidc if available else client secret
- 

#445 

Looks like a big PR but most +- are from apigen.


aabchoo:
Following up to this, there should be a similar doc on how to connect via OIDC

https://github.com/envoyproxy/ai-gateway/issues/531

api/v1alpha1/api.go
```diff
@@ -548,7 +548,24 @@ type BackendSecurityPolicyAzureCredentials struct {
...
 	// ai-gateway must be given the permission to read this secret.
 	// The key of secret should be "client-secret".
-	ClientSecretRef *gwapiv1.SecretObjectReference `json:"clientSecretRef"`
+	//
+	// +optional
+	ClientSecretRef *gwapiv1.SecretObjectReference `json:"clientSecretRef,omitempty"`
+
+	// OIDCExchangeToken specifies the oidc configurations used to obtain an oidc token. The oidc token will be
+	// used to obtain temporary credentials to access Azure.
+	// In the case both OIDC and clientSecret is defined, OIDC exchange tokens will be used to obtain Azure access token.
```
mathetake, line 557: so what happens when neither clientSecretRef nor oidcExchangeToken is defined?

mathetake (commented):
cool... left some comment but will defer to Dan for the rest

this should land before 0.2 cut right?

mathetake:
@xiaolin593 do you want to take a look as you worked on the azure stuff before

aabchoo:
> cool... left some comment but will defer to Dan for the rest
> 
> this should land before 0.2 cut right?

yeah, that would be ideal for Azure

api/v1alpha1/api.go
aabchoo: A secret will not be generated, and reconcile will be returned an error. Updated to be clearer: https://github.com/envoyproxy/ai-gateway/pull/658/commits/08354cae0aa757b159d4960b646ad86b6768aace

aabchoo: ah just needed to move it outside of the struct

mathetake (commented):
API looks good to me

internal/controller/tokenprovider/azure_token_provider.go
```diff
@@ -18,14 +18,20 @@ import (
...
 }
 
-// NewAzureTokenProvider creates a new TokenProvider with the given tenant ID, client ID, client secret, and token request options.
-func NewAzureTokenProvider(tenantID, clientID, clientSecret string, tokenOption policy.TokenRequestOptions) (TokenProvider, error) {
-	clientOptions := GetClientSecretCredentialOptions()
-	credential, err := azidentity.NewClientSecretCredential(tenantID, clientID, clientSecret, clientOptions)
+// NewAzureTokenProvider creates a new TokenProvider with the given tenant ID, client ID, tokenProvider, and token request options.
+func NewAzureTokenProvider(ctx context.Context, tenantID, clientID string, tokenProvider TokenProvider, tokenOption policy.TokenRequestOptions) (TokenProvider, error) {
+	clientOptions := GetClientAssertionCredentialOptions()
+	credential, err := azidentity.NewClientAssertionCredential(tenantID, clientID, func(_ context.Context) (string, error) {
```
yuzisun: What's the difference between `NewClientSecretCredential` and `NewClientAssertionCredential`, is here to use the oidc token to exchange the azure credential ?

aabchoo: `NewClientSecretCredential` is used to obtain azure's access token via ClientID, ClientSecret, and TenantID.

`NewClientAssertionCredential` obtains azure's access token via ClientID, TenantID, and an assertion claim (in the form of a JWT token w/ claims of auth). Could be issued by Microsoft Entra ID, or another federated identity provider ([src](https://learn.microsoft.com/en-us/entra/msal/dotnet/acquiring-tokens/web-apps-apis/confidential-client-assertions#client-assertions)).

internal/controller/tokenprovider/azure_client_secret_token_provider_test.go
```diff
@@ -0,0 +1,73 @@
...
+
+		_, err = provider.GetToken(context.Background())
+		require.Error(t, err)
+		require.Contains(t, err.Error(), "Tenant 'invalidtenantid' not found. Check to make sure you have the correct tenant ID and are signing into the correct cloud.")
+	})
+
+	t.Run("azure proxy url", func(t *testing.T) {
+		// Set environment variable for the test
+		mockProxyURL := "http://localhost:8888"
+		os.Setenv("AI_GATEWAY_AZURE_PROXY_URL", mockProxyURL)
```
mathetake: let's use t.Setenv instead

mathetake: then os.Unsetenv won't be necessary below

internal/controller/tokenprovider/azure_token_provider_test.go
```diff
@@ -49,7 +52,7 @@ func TestAzureTokenProvider_GetToken(t *testing.T) {
 		os.Setenv("AI_GATEWAY_AZURE_PROXY_URL", mockProxyURL)
```
mathetake: yeah here as well while you are at it

internal/controller/tokenprovider/azure_token_provider.go
```diff
@@ -18,14 +18,20 @@ import (
...
+		token, err := tokenProvider.GetToken(ctx)
```
mathetake, line 29: i think you should local `ctx` (the one you are ignoring as `_`)  instead of leaking the outer `ctx` here.

api/v1alpha1/api.go
```diff
@@ -532,6 +532,10 @@ type BackendSecurityPolicyOIDC struct {
 }
 
 // BackendSecurityPolicyAzureCredentials contains the supported authentication mechanisms to access Azure.
+// Only one of ClientSecretRef or OIDCExchangeToken must be specified. Credentials will not be generated if
+// neither are set.
+//
+// +kubebuilder:validation:XValidation:rule="(has(self.clientSecretRef) && !has(self.oidcExchangeToken)) || (!has(self.clientSecretRef) && has(self.oidcExchangeToken))",message="Exactly one of clientSecretRef or oidcExchangeToken must be specified"
```
mathetake, line 538: nice

PR #660, opened 2025-05-31
test: deflake e2e by forwarding to svc, not pod

### LINKED ISSUE #647: flake: Test_Examples_ProviderFallback
https://github.com/envoyproxy/ai-gateway/actions/runs/15306450239/job/43060306635

mathetake:
**Commit Message**

Previously, the pod was the target for port-forwarding. However, that can be fragile since the pod may be rolled out even after the gateway itself becomes healthy after the side car injection. This will deflake the tests by using the service as a target which allows us to avoid choosing the pod that is in the middle of termination/rollout.

Closes #647

### DIFF
--- a/examples/monitoring/monitoring.yaml
+++ b/examples/monitoring/monitoring.yaml
@@ -136,6 +136,8 @@ kind: Service
 metadata:
   name: prometheus
   namespace: monitoring
+  labels:
+    app: prometheus
 spec:
   selector:
     app: prometheus
--- a/tests/e2e/provider_fallback_test.go
+++ b/tests/e2e/provider_fallback_test.go
@@ -54,7 +54,7 @@ func Test_Examples_ProviderFallback(t *testing.T) {
 	// So, no matter how many times we try, we should always get a 503 error.
 	for i := range 5 {
 		t.Run("no-fallback/"+strconv.Itoa(i), func(t *testing.T) {
-			fwd := requireNewHTTPPortForwarder(t, egNamespace, egSelector, egDefaultPort)
+			fwd := requireNewHTTPPortForwarder(t, egNamespace, egSelector, egDefaultServicePort)
 			defer fwd.kill()
 
 			ctx, cancel := context.WithTimeout(t.Context(), 10*time.Second)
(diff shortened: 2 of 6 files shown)


mathetake:
🤞

mathetake:
ok i will re-run 10 times at least

mathetake:
ok at least 5 times passing

mathetake:
ok i believe it's fixed!

mathetake:
ping

examples/monitoring/monitoring.yaml
```diff
@@ -136,6 +136,8 @@ kind: Service
 metadata:
   name: prometheus
   namespace: monitoring
+  labels:
```
aabchoo, line 139: Is this related to the fix?

tests/e2e/provider_fallback_test.go
```diff
@@ -54,7 +54,7 @@ func Test_Examples_ProviderFallback(t *testing.T) {
 	// So, no matter how many times we try, we should always get a 503 error.
 	for i := range 5 {
 		t.Run("no-fallback/"+strconv.Itoa(i), func(t *testing.T) {
-			fwd := requireNewHTTPPortForwarder(t, egNamespace, egSelector, egDefaultPort)
+			fwd := requireNewHTTPPortForwarder(t, egNamespace, egSelector, egDefaultServicePort)
```
aabchoo, line 57: Is this the correct port? https://github.com/envoyproxy/ai-gateway/blob/main/examples/provider_fallback/base.yaml#L188

examples/monitoring/monitoring.yaml
mathetake: related yes, without this service cannot be port-forward target

tests/e2e/provider_fallback_test.go
mathetake: this 443 is fake port that is supposed to fail and not relevant here. (knid: Service != Gateway which we are talking about, so the resource you pointed is nothing to do here except for prometheus)

aabchoo (approved):

tests/e2e/provider_fallback_test.go
aabchoo: ah I see because that's example and not part of e2e 🤦

mathetake:
@aabchoo thank you for the look!!!

PR #672, opened 2025-06-02
ci: removes unnecessary line in release.yaml

mathetake:
**Commit Message**

This removes the redundant upload in release.yaml which resulted in uploading the same artifacts twice. 
The CI failed in https://github.com/envoyproxy/ai-gateway/actions/runs/15404045295/job/43343173369
while we can see the artifacts already uploaded correctly at `gh release create`.

Follow up on #513

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -46,6 +46,5 @@ jobs:
         if: ${{ !contains(github.ref, '-rc') }}
         run: |
           gh release create $TAG --draft --title $TAG --notes "To be written by the release manager" ./out/ai-gateway-crds-helm-${TAG}.tgz ./out/ai-gateway-helm-${TAG}.tgz
-          gh release upload $TAG ./out/ai-gateway-crds-helm-${TAG}.tgz ./out/ai-gateway-helm-${TAG}.tgz
         env:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -46,6 +46,5 @@ jobs:
         if: ${{ !contains(github.ref, '-rc') }}
```
aabchoo, line 45: The title says "for RC" but this GitHub action step is for stable release

mathetake: oh my good catch, this is not limited to rc and i need to fix ti

mathetake: https://github.com/envoyproxy/ai-gateway/pull/672/commits/01d156fe735e962de7a8c2e901bfdc155c02a8a0 done

aabchoo:
I'm surprised we didn't run into this problem the first time we had a release

mathetake:
because this is added in https://github.com/envoyproxy/ai-gateway/issues/513 as in the PR description 😉

mathetake:
or more precisely this https://github.com/envoyproxy/ai-gateway/pull/618

mathetake:
Or more precisely my concern https://github.com/envoyproxy/ai-gateway/pull/618#discussion_r2093598789 was corect

PR #673, opened 2025-06-03
feat: add priority to backendRef

cmaddalozzo:
**Commit Message**

Add priority field to AIGatewayRouteRuleBackendRef. This will set priority on the underlying LocalityLB endpoint in Envoy.

N/A

In order to get the Envoy Gateway to reconcile the underlying `HTTPRoute` when the priorities in the `AIGatewayRoute` spec change we need to add an annotation. This annotation just store the priorities for each rule. This ensure we only trigger reconciliation when the priorities actually change.


api/v1alpha1/api.go
```diff
@@ -277,6 +277,15 @@ type AIGatewayRouteRuleBackendRef struct {
 	// +kubebuilder:validation:Minimum=0
 	// +kubebuilder:default=1
 	Weight *int32 `json:"weight,omitempty"`
+	// Priority is the priority of the AIServiceBackend. This sets the priority on the underlying endpoints.
```
aabchoo, line 280: Can you add a cel validation test?

internal/controller/ai_gateway_route.go
```diff
@@ -276,3 +284,15 @@ func (c *AIGatewayRouteController) updateAIGatewayRouteStatus(ctx context.Contex
...
 	}
 }
+
+// Build an annotation that contains the priority of each backend ref. This is used to ensure Envoy Gateway reconciles the
+// HTTP route when the priorities change.
+func buildPriorityAnnotation(rules []aigv1a1.AIGatewayRouteRule) string {
+	priorities := make([]string, 0, len(rules))
+	for i, rule := range rules {
+		for _, br := range rule.BackendRefs {
+			priorities = append(priorities, fmt.Sprintf("%d:%s:%d", i, br.Name, *br.Priority))
```
aabchoo: > *br.Priority

this may cause problems for the CLI if the priority is not set (unlike how Kubebuilder sets it by default if omitted). Consider adding a check here if priority is defined?

internal/controller/ai_gateway_route.go
```diff
@@ -28,12 +29,13 @@ import (
...
-	selectedRouteHeaderKey        = "x-ai-eg-selected-route"
-	hostRewriteHTTPFilterName     = "ai-eg-host-rewrite"
-	aigatewayUUIDAnnotationKey    = "aigateway.envoyproxy.io/uuid"
-	egOwningGatewayNameLabel      = "gateway.envoyproxy.io/owning-gateway-name"
-	egOwningGatewayNamespaceLabel = "gateway.envoyproxy.io/owning-gateway-namespace"
+	managedByLabel                           = "app.kubernetes.io/managed-by"
+	selectedRouteHeaderKey                   = "x-ai-eg-selected-route"
+	hostRewriteHTTPFilterName                = "ai-eg-host-rewrite"
+	aigatewayUUIDAnnotationKey               = "aigateway.envoyproxy.io/uuid"
+	httpRouteBackendRefPriorityAnnotationKey = "gateway.envoyproxy.io/backend-ref-priority"
```
mathetake, line 39: nit ;could you use aigateway instead of gateway so that it's clear that it is AIGW that sets this one, not EG. This matches the uuid stuff above

api/v1alpha1/api.go
mathetake:
```suggestion
	// Priority is the priority of the AIServiceBackend for this specific rule. This sets the priority on the underlying endpoints.
```

mathetake (commented):
maybe you might want to update the provider_fallback e2e ?

internal/extensionserver/extensionserver.go
```diff
@@ -187,6 +187,9 @@ func (s *Server) maybeModifyCluster(cluster *clusterv3.Cluster) {
 		backendRef := httpRouteRule.BackendRefs[i]
 		name := backendRef.Name
 		namespace := aigwRoute.Namespace
+		if backendRef.Priority != nil {
+			endpoints.Priority = *backendRef.Priority
```
mathetake, line 191: IIUC this will override `fallback: true/false` option of EG's Backend resource right? Maybe we should document that in the API.

mathetake:
i am glad that the core change is much smaller than i thought! I think the rest is documentation + e2e tests

internal/controller/ai_gateway_route.go
cmaddalozzo: I originally tried to use `aigateway.envoyproxy.io/backend-ref-priority` but that didn't result in the `HTTPRoute` being reconciled. Setting `gateway.envoyproxy.io/backend-ref-priority` actually adds metadata to the underlying Envoy route.

https://gateway.envoyproxy.io/contributions/design/metadata/#translation

mathetake: oh ok then could you put that code comment here so that others can understand the rationale

api/v1alpha1/api.go
```diff
@@ -277,6 +277,16 @@ type AIGatewayRouteRuleBackendRef struct {
...
+	// See: https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/load_balancing/priority
+	// Note: This will override the `faillback` property of the underlying Envoy Gateway Backend
+	//
+	// Default is 0.
+	//
+	// +optional
+	// +kubebuilder:validation:Minimum=0
```
mathetake, line 287: it seems trivial but let's add the CEL validation test as @aabchoo commented above - you can take a look at tests/crdcel

cmaddalozzo: > it seems trivial but let's add the CEL validation test as @aabchoo commented above - you can take a look at tests/crdcel

I tried adding a CEL validation where `-1` was passed but this resulted in a parsing error as it cannot be parsed as an unsigned int. Since any valid unsigned int is a valid priority I am not sure we need a validation test. WDYT

mathetake: fair enough! yeah we can skip it for this one

cmaddalozzo:
> maybe you might want to update the provider_fallback e2e ?

I don't think we will be able to add this until the previous priority retry predicate PR makes it into Envoy Gateway.

https://github.com/envoyproxy/gateway/pull/6204

mathetake:
hmm i was under the impression that this could be used instead of fallback: true option in the e2e test. not sure what i am missing

cmaddalozzo:
> hmm i was under the impression that this could be used instead of fallback: true option in the e2e test. not sure what i am missing

ah, yes that will work. The part that can't be tested yet is retrying across priorities in a dynamic way.

mathetake:
cool, could you go ahead and make the change like the below;

```diff
diff --git a/examples/provider_fallback/base.yaml b/examples/provider_fallback/base.yaml
index b695ae4..ea3c5bc 100644
--- a/examples/provider_fallback/base.yaml
+++ b/examples/provider_fallback/base.yaml
@@ -41,8 +41,11 @@ spec:
               name: x-ai-eg-model
               value: us.meta.llama3-2-1b-instruct-v1:0
       backendRefs:
-        - name: provider-fallback-always-failing-upstream  # This is the primary backend and trying to speak TLS, which always fails.
+        - name: provider-fallback-always-failing-upstream
+          # This is the primary backend and trying to speak TLS, which always fails.
+          priority: 0
         - name: provider-fallback-aws
+          priority: 1
 ---
 apiVersion: aigateway.envoyproxy.io/v1alpha1
 kind: AIServiceBackend
@@ -80,8 +83,6 @@ metadata:
   name: provider-fallback-aws
   namespace: default
 spec:
-  # Indicate that this backend is a fallback backend, meaning that it will only be used if the primary backend fails.
-  fallback: true
   endpoints:
     - fqdn:
         hostname: bedrock-runtime.us-east-1.amazonaws.com
```

i believe this should work and then we are good to go (i think we can make this part of v0.2)

yuzisun:
> cool, could you go ahead and make the change like the below;
> 
> ```diff
> diff --git a/examples/provider_fallback/base.yaml b/examples/provider_fallback/base.yaml
> index b695ae4..ea3c5bc 100644
> --- a/examples/provider_fallback/base.yaml
> +++ b/examples/provider_fallback/base.yaml
> @@ -41,8 +41,11 @@ spec:
>                name: x-ai-eg-model
>                value: us.meta.llama3-2-1b-instruct-v1:0
>        backendRefs:
> -        - name: provider-fallback-always-failing-upstream  # This is the primary backend and trying to speak TLS, which always fails.
> +        - name: provider-fallback-always-failing-upstream
> +          # This is the primary backend and trying to speak TLS, which always fails.
> +          priority: 0
>          - name: provider-fallback-aws
> +          priority: 1
>  ---
>  apiVersion: aigateway.envoyproxy.io/v1alpha1
>  kind: AIServiceBackend
> @@ -80,8 +83,6 @@ metadata:
>    name: provider-fallback-aws
>    namespace: default
>  spec:
> -  # Indicate that this backend is a fallback backend, meaning that it will only be used if the primary backend fails.
> -  fallback: true
>    endpoints:
>      - fqdn:
>          hostname: bedrock-runtime.us-east-1.amazonaws.com
> ```
> 
> i believe this should work and then we are good to go (i think we can make this part of v0.2)

@mathetake I updated, going to update my doc with the same change.

examples/provider_fallback/base.yaml
```diff
@@ -42,7 +42,9 @@ spec:
               value: us.meta.llama3-2-1b-instruct-v1:0
       backendRefs:
         - name: provider-fallback-always-failing-upstream  # This is the primary backend and trying to speak TLS, which always fails.
+          priority: 0
         - name: provider-fallback-aws
+          priority: 1
```
mathetake, line 47: Can you also delete the `fallback: true` below as it's overridden by this

mathetake (approved):
Awesome

examples/provider_fallback/base.yaml
yuzisun: Yes I deleted the fallback flag below

PR #685, opened 2025-06-05
extproc: clear content-encoding only when body is mutated

mathetake:
**Commit Message**

Clearing content-type header is only necessary when the response mutation is actually happening. Otherwise, say when the original response is in gzip and not mutation is happening, the client gets gzip and no content-encoding header.

Follow up on #683


mathetake:
i think i should add a test case that doesn't require the credential enabled test

mathetake:
ok good to go

missBerg (approved):

PR #687, opened 2025-06-06
feat: add model name field to backend ref

cmaddalozzo:
**Commit Message**

Add model name field to backend ref. This allows users pass a different model name to the backend from the one that was provided in the original request. This is particularly useful for cross-provided failover and load balancing.

For example: If a user has provisioned Claude 3.5 Sonnet on both AWS and GCP they can setup a rule that will load balance across them:

```yaml
kind: AIGatewayRoute
metadata:
  name: test-route
  namespace: ai-gateway
spec:
  targetRefs:
    kind: Gateway
    name: eg
  schema:
    name: OpenAI
  rules:
  - matches:
      - headers:
        - type: Exact
          name: x-ai-eg-model
          value: claude-3-5-sonnet-v2
    backendRefs:
    - name: aws-backend
    - name: gcp-backend
```

Since the both providers will have a different name/model identifier for the provisioned endpoint the above configuration will not work. This PR makes it possible to add a `modelName` on the backend ref which specifies the providers name for the model.

With the `modelName` property we can support this use-case as follows:

```yaml
kind: AIGatewayRoute
metadata:
  name: test-route
  namespace: ai-gateway
spec:
  targetRefs:
    kind: Gateway
    name: eg
  schema:
    name: OpenAI
  rules:
  - matches:
      - headers:
        - type: Exact
          name: x-ai-eg-model
          value: claude-3-5-sonnet-v2
    backendRefs:
    - name: aws-backend
      modelName: arn:aws:bedrock:us-west-2:123456789012:provisioned-model/abc123xyz
    - name: gcp-backend
      modelName: claude-3-5-sonnet-v2@20241022
```


api/v1alpha1/api.go
```diff
@@ -267,6 +267,10 @@ type AIGatewayRouteRuleBackendRef struct {
 	// +kubebuilder:validation:MinLength=1
 	Name string `json:"name"`
 
+	// Name of the model in the backend. If provided this will override the name provided in the request.
+	// +kubebuilder:validation:MinLength=1
```
aabchoo: nit: The json is omit empty but kubebuilder validation min length is set to 1

filterapi/filterconfig.go
```diff
@@ -186,6 +186,8 @@ type Backend struct {
 	// Name of the backend, which is the value in the final routing decision
 	// matching the header key specified in the [filterapi.Config.BackendRoutingHeaderKey].
 	Name string `json:"name"`
+	// Name of the model in the backend. If provided this will override the name provided in the request.
```
aabchoo, line 189: Might want to update the test to include new ModelName field

https://github.com/envoyproxy/ai-gateway/blob/177c4147f0c3e50dc9f70df7c0accdc7bd3c9734/tests/extproc/extproc_test.go#L71

aabchoo:
The reason the extproc tests fail is because of the `modelId` field you removed from the [aws bedrock spec](https://github.com/envoyproxy/ai-gateway/pull/687/files#diff-afbaa50c55faff1e80da25dd04c952f76070b4825d88f7a36725ac1ef4c01e27L170). The test body expects the field, so remove the text `modelId: null` from [here](https://github.com/envoyproxy/ai-gateway/blob/4b5ebbd930a7de3a43bd8f8213cfcb7a611a174e/tests/extproc/testupstream_test.go#L152) + [here](https://github.com/envoyproxy/ai-gateway/blob/4b5ebbd930a7de3a43bd8f8213cfcb7a611a174e/tests/extproc/testupstream_test.go#L197) and it will pass

```
unexpected request body: got {"inferenceConfig":{},"messages":[],"system":[{"text":"You are a chatbot."}]} expected {"inferenceConfig":{},"messages":[],"modelId":null,"system":[{"text":"You are a chatbot."}]}
```

api/v1alpha1/api.go
```diff
@@ -267,6 +267,9 @@ type AIGatewayRouteRuleBackendRef struct {
 	// +kubebuilder:validation:MinLength=1
 	Name string `json:"name"`
 
+	// Name of the model in the backend. If provided this will override the name provided in the request.
+	ModelName string `json:"modelName,omitempty"`
```
mathetake: do we want to have "override" portion in the field name? e.g. https://portkey.ai/docs/product/ai-gateway/fallbacks portkey has this exact feature and they name it under override_params.

I think another layer override_params won't be needed but at least it seems to make sense to name this "modelNameOverride" or "overrideModelName"?

filterapi/filterconfig.go
```diff
@@ -186,6 +186,8 @@ type Backend struct {
...
+	ModelNameOverride string `json:"modelName"`
```
mathetake:
```suggestion
	ModelNameOverride string `json:"modelNameOverride,omittempty"`
```

mathetake (commented):
Looking good... I wonder what's holding you from making this ready for review PR?

internal/apischema/openai/openai.go
mathetake, line 1: let me confirm my understanding. Is the change necessary because there was a bug in marshaling of openai req which became necessary because of model name replace?

internal/apischema/openai/openai.go
```diff
@@ -172,6 +172,21 @@ func (s *StringOrAssistantRoleContentUnion) UnmarshalJSON(data []byte) error {
 	return errors.New("cannot unmarshal JSON data as string or assistant content parts")
 }
 
+// Marshal StringOrAssistantRoleContentUnion to JSON
+func (s *StringOrAssistantRoleContentUnion) MarshalJSON() ([]byte, error) {
```
mathetake: nit; the comment is recommended to begin with its function name:

image

internal/apischema/openai/openai.go
```diff
@@ -216,6 +246,21 @@ func (s *StringOrUserRoleContentUnion) UnmarshalJSON(data []byte) error {
...
 
+// Marshal StringOrUserRoleContentUnion to JSON
+func (s *StringOrUserRoleContentUnion) MarshalJSON() ([]byte, error) {
+	if s.Value == nil {
+		return nil, fmt.Errorf("cannot marshal StringOrUserRoleContentUnion with nil value")
+	}
+	if _, ok := s.Value.(string); ok {
+		return json.Marshal(s.Value)
+	}
+	if arr, ok := s.Value.([]ChatCompletionContentPartUserUnionParam); ok {
```
mathetake: I think ChatCompletionContentPartUserUnionParam is not properly handled in marshaling (as opposed to unmarshaling). In fact, it lacks a json tag etc. in other words, i think we need

```
func (c *ChatCompletionContentPartUserUnionParam) MarshalJSON() ([]byte, error) {
	if c.TextContent != nil {
		return json.Marshal(c.TextContent)
	}
	if c.InputAudioContent != nil {
		return json.Marshal(c.InputAudioContent)
	}
	if c.ImageContent != nil {
		return json.Marshal(c.ImageContent)
	}
	return nil, fmt.Errorf("cannot marshal ChatCompletionContentPartUserUnionParam with nil value")
}

```

mathetake: hmm, i think i found another bug around this roundtrip marshaling- 

```
`{"messages":[{"content": "Say this is a test!", "role": "user",}],"model":"llama3","temperature": 0.7}`
```

got marshaled down to
```
{"messages":[{"content":{"Value":"Say this is a test!"},"role":"user"}],"model":"llama3","temperature":0.7}
```

which seems invalid. 

I think we need a through roundtrip test where we ensure unmarshal->marshal produces the identical original json string for many test cases (note previously we didn't need marshaling before this, hence bugs are somewhat expected.)


Here's the repro for this specific case:

```
func TestRepro(t *testing.T) {
	const b = `{"messages":[{"content":"Say this is a test!","role":"user"}],"model":"llama3","temperature":0.7}`
	var openAIReq openai2.ChatCompletionRequest
	err := json.Unmarshal([]byte(b), &openAIReq)
	require.NoError(t, err, "Failed to unmarshal OpenAI request: %v", err)

	unmarshaled, err := json.Marshal(openAIReq)
	require.NoError(t, err, "Failed to marshal OpenAI request: %v", err)
	require.Equal(t, b, string(unmarshaled), "Unmarshaled OpenAI request does not match original")
}
```

internal/apischema/openai/openai.go
```diff
@@ -276,6 +321,14 @@ func (c *ChatCompletionMessageParamUnion) UnmarshalJSON(data []byte) error {
 	return nil
 }
 
+// This function marshals the ChatCompletionMessageParamUnion into JSON.
+func (c *ChatCompletionMessageParamUnion) MarshalJSON() ([]byte, error) {
+	if c.Value == nil {
+		return nil, fmt.Errorf("cannot marshal ChatCompletionMessageParamUnion with nil value")
+	}
+	return json.Marshal(c.Value)
```
mathetake: yep, at least the repro i shared ^^ will be resolved with 
```suggestion
	switch c.Type {
	case ChatMessageRoleUser:
		if userMessage, ok := c.Value.(ChatCompletionUserMessageParam); ok {
			return json.Marshal(&userMessage)
		}
	case ChatMessageRoleAssistant:
		if assistantMessage, ok := c.Value.(ChatCompletionAssistantMessageParam); ok {
			return json.Marshal(&assistantMessage)
		}
	case ChatMessageRoleSystem:
		if systemMessage, ok := c.Value.(ChatCompletionSystemMessageParam); ok {
			return json.Marshal(&systemMessage)
		}
	case ChatMessageRoleDeveloper:
		if developerMessage, ok := c.Value.(ChatCompletionDeveloperMessageParam); ok {
			return json.Marshal(&developerMessage)
		}
	case ChatMessageRoleTool:
		if toolMessage, ok := c.Value.(ChatCompletionToolMessageParam); ok {
			return json.Marshal(&toolMessage)
		}
	}
	return nil, fmt.Errorf("cannot marshal ChatCompletionMessageParamUnion with unknown type: %s", c.Type)
```

mathetake:
@cmaddalozzo i think the marshaling back to modify model field feels like a rabbit hole... I think instead we could use sjson like this 

```diff
git diff tetratetigs/main HEAD
diff --git a/go.mod b/go.mod
index 3b71d3c..6d1b36a 100644
--- a/go.mod
+++ b/go.mod
@@ -20,6 +20,7 @@ require (
        github.com/openai/openai-go v1.3.0
        github.com/prometheus/client_golang v1.22.0
        github.com/stretchr/testify v1.10.0
+       github.com/tidwall/sjson v1.2.5
        go.opentelemetry.io/otel v1.36.0
        go.opentelemetry.io/otel/exporters/prometheus v0.58.0
        go.opentelemetry.io/otel/metric v1.36.0
@@ -360,7 +361,6 @@ require (
        github.com/tidwall/gjson v1.18.0 // indirect
        github.com/tidwall/match v1.1.1 // indirect
        github.com/tidwall/pretty v1.2.1 // indirect
-       github.com/tidwall/sjson v1.2.5 // indirect
        github.com/timakin/bodyclose v0.0.0-20241222091800-1db5c5ca4d67 // indirect
        github.com/timonwong/loggercheck v0.10.1 // indirect
        github.com/tklauser/go-sysconf v0.3.12 // indirect
diff --git a/internal/extproc/translator/openai_openai.go b/internal/extproc/translator/openai_openai.go
index 93eefaf..580bf65 100644
--- a/internal/extproc/translator/openai_openai.go
+++ b/internal/extproc/translator/openai_openai.go
@@ -14,6 +14,7 @@ import (
 
        corev3 "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
        extprocv3 "github.com/envoyproxy/go-control-plane/envoy/service/ext_proc/v3"
+       "github.com/tidwall/sjson"
 
        "github.com/envoyproxy/ai-gateway/internal/apischema/openai"
 )
@@ -42,11 +43,11 @@ func (o *openAIToOpenAITranslatorV1ChatCompletion) RequestBody(raw []byte, req *
        if o.modelNameOverride != "" {
                // If modelName is set we override the model to be used for the request.
-               req.Model = o.modelNameOverride
-               out, err := json.Marshal(req)
-               if err != nil {
-                       return nil, nil, fmt.Errorf("failed to marshal request body: %w", err)
-               }
-               newBody = &out
+               raw, _ = sjson.SetBytesOptions(raw, "model", o.modelNameOverride, &sjson.Options{
+                       Optimistic:     true,
+                       ReplaceInPlace: true,
+               })
+               newBody = &raw
        }
```


I think I would prefer this over fixing all the marshaling bugs (which need lots of additional test cases )

cmaddalozzo:
> @cmaddalozzo i think the marshaling back to modify model field feels like a rabbit hole... I think instead we could use sjson like this
> 
> ```diff
> git diff tetratetigs/main HEAD
> diff --git a/go.mod b/go.mod
> index 3b71d3c..6d1b36a 100644
> --- a/go.mod
> +++ b/go.mod
> @@ -20,6 +20,7 @@ require (
>         github.com/openai/openai-go v1.3.0
>         github.com/prometheus/client_golang v1.22.0
>         github.com/stretchr/testify v1.10.0
> +       github.com/tidwall/sjson v1.2.5
>         go.opentelemetry.io/otel v1.36.0
>         go.opentelemetry.io/otel/exporters/prometheus v0.58.0
>         go.opentelemetry.io/otel/metric v1.36.0
> @@ -360,7 +361,6 @@ require (
>         github.com/tidwall/gjson v1.18.0 // indirect
>         github.com/tidwall/match v1.1.1 // indirect
>         github.com/tidwall/pretty v1.2.1 // indirect
> -       github.com/tidwall/sjson v1.2.5 // indirect
>         github.com/timakin/bodyclose v0.0.0-20241222091800-1db5c5ca4d67 // indirect
>         github.com/timonwong/loggercheck v0.10.1 // indirect
>         github.com/tklauser/go-sysconf v0.3.12 // indirect
> diff --git a/internal/extproc/translator/openai_openai.go b/internal/extproc/translator/openai_openai.go
> index 93eefaf..580bf65 100644
> --- a/internal/extproc/translator/openai_openai.go
> +++ b/internal/extproc/translator/openai_openai.go
> @@ -14,6 +14,7 @@ import (
>  
>         corev3 "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
>         extprocv3 "github.com/envoyproxy/go-control-plane/envoy/service/ext_proc/v3"
> +       "github.com/tidwall/sjson"
>  
>         "github.com/envoyproxy/ai-gateway/internal/apischema/openai"
>  )
> @@ -42,11 +43,11 @@ func (o *openAIToOpenAITranslatorV1ChatCompletion) RequestBody(raw []byte, req *
>         if o.modelNameOverride != "" {
>                 // If modelName is set we override the model to be used for the request.
> -               req.Model = o.modelNameOverride
> -               out, err := json.Marshal(req)
> -               if err != nil {
> -                       return nil, nil, fmt.Errorf("failed to marshal request body: %w", err)
> -               }
> -               newBody = &out
> +               raw, _ = sjson.SetBytesOptions(raw, "model", o.modelNameOverride, &sjson.Options{
> +                       Optimistic:     true,
> +                       ReplaceInPlace: true,
> +               })
> +               newBody = &raw
>         }
> ```
> 
> I think I would prefer this over fixing all the marshaling bugs (which need lots of additional test cases )

@mathetake 100% agree. Dan and I just discussed this and came to the same conclusion. I'll update the code to use sjson 👍

internal/extproc/translator/openai_openai.go
```diff
@@ -37,6 +39,18 @@ func (o *openAIToOpenAITranslatorV1ChatCompletion) RequestBody(raw []byte, req *
...
+		// If modelName is set we override the model to be used for the request.
+		out, err := sjson.SetBytesOptions(raw, "model", o.modelNameOverride, &sjson.Options{
+			Optimistic:     true,
+			ReplaceInPlace: true,
+		})
+		if err != nil {
+			return nil, nil, fmt.Errorf("failed to set model name: %w", err)
+		}
+		newBody = &out
+	}
```
mathetake: i think a pointer to the byte slice is unnecessary here. We can simply use `[]byte` and check len(newBody) > 0 below. That would save us one allocation per this function call. It's small but the less allocation the better (maybe escape analysis will work well to eliminate the allocation but we can't be certain about the compiler optimization).

```suggestion
	var newBody []byte
	if o.modelNameOverride != "" {
		// If modelName is set we override the model to be used for the request.
		out, err := sjson.SetBytesOptions(raw, "model", o.modelNameOverride, &sjson.Options{
			Optimistic:     true,
			ReplaceInPlace: true,
		})
		if err != nil {
			return nil, nil, fmt.Errorf("failed to set model name: %w", err)
		}
		newBody = out
	}
```

mathetake (commented):
small nit but overall LGTM! nice work!

api/v1alpha1/api.go
```diff
@@ -267,6 +267,9 @@ type AIGatewayRouteRuleBackendRef struct {
 	// +kubebuilder:validation:MinLength=1
 	Name string `json:"name"`
 
+	// Name of the model in the backend. If provided this will override the name provided in the request.
+	ModelNameOverride string `json:"modelNameOverride,omitempty"`
```
kerthcet, line 271: A quick question: will this impact the output of `/models`?

mathetake: it doesn't afaik

kerthcet: 🍺🍺

Issue #650, opened 2025-05-29
TestExtProcCustomMetrics is failing with Envoy 1.33 or older (on mac)


mathetake:
This can be closed once 1.34 on homebrew is released https://github.com/Homebrew/homebrew-core/pull/223051 and related to the special casing here:

https://github.com/envoyproxy/ai-gateway/blob/d6797ad74e564abf56e5b731003457ba6145d145/internal/extproc/server.go#L308-L309

PR #702, opened 2025-06-10
cli: use fixed prometheus port with aigw run

mathetake:
**Commit Message**

Previously, the prometheus metrics port was chosen randomly, hence it was impossible to scrape metrics when using `aigw run` command. This fixes it by using the default port 1064 used by the extproc. Note that there won't be a conflict as `aigw run` doesn't support multiple gateways.

### DIFF
--- a/site/docs/cli/run.md
+++ b/site/docs/cli/run.md
@@ -12,6 +12,10 @@ This command runs the Envoy AI Gateway locally as a standalone proxy with a give
 Since the project is primarily focused on the Kubernetes environment, this command is useful for testing the configuration locally before deploying it to a Kubernetes cluster.
 Not only does it help in testing the configuration, but it is also useful in a local development environment of the provider-agnostic AI applications.
 
+:::warning
+Currently, `aigw run` only properly works on Linux. macOS support is pending due to the Envoy v1.34+ binary not available for macOS yet. That will be resolved with https://github.com/Homebrew/homebrew-core/pull/223051.
+:::
+
 ## Default Proxy
 
 By default, `aigw run` runs the AI Gateway with a default configuration that includes a proxy that listens on port `1975`.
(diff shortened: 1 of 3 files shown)


codefromthecrypt (approved):
excellent!

site/docs/cli/run.md
```diff
@@ -12,6 +12,10 @@ This command runs the Envoy AI Gateway locally as a standalone proxy with a give
 Since the project is primarily focused on the Kubernetes environment, this command is useful for testing the configuration locally before deploying it to a Kubernetes cluster.
 Not only does it help in testing the configuration, but it is also useful in a local development environment of the provider-agnostic AI applications.
 
+:::warning
+Currently, `aigw run` only properly works on Linux. macOS support is pending due to the Envoy v1.34+ binary not available for macOS yet. That will be resolved with https://github.com/Homebrew/homebrew-core/pull/223051.
```
codefromthecrypt, line 16: technically this is use at your own risk, right? as metrics won't work until that? or there is something bigger. The workaround is to use docker when on macOS because aigw doesn't use docker internally. That said, I didn't see a guide or doc saying where the docker images are published to. Might be nice to put that somewhere. https://aigateway.envoyproxy.io/docs/cli/aigwinstall

mathetake: yeah technically at your own risk but i guess there's something bigger like https://github.com/envoyproxy/ai-gateway/issues/650

Meanwhile, as we haven't published the docker image yet, it makes sense to do so as a workaround

codefromthecrypt:
verified like this since macos is busted, and until there's a docker image for aigw

```bash
$ docker run --rm -it golang:1.24.3-alpine
/go # go install github.com/envoyproxy/ai-gateway/cmd/aigw@main
--snip--
/go # nohup aigw run > aigw.log 2>&1 &
/go # cat aigw.log
looking up the latest Envoy version
downloading https://archive.tetratelabs.io/envoy/download/v1.34.1/envoy-v1.34.1-linux-arm64.tar.xz
starting: /tmp/envoy-gateway/versions/1.34.1/bin/envoy with --admin-address-path /tmp/envoy-gateway/runs/1749885347779737913/admin-address.txt
/go # wget -qO - localhost:1064/metrics
# HELP target_info Target metadata
# TYPE target_info gauge
target_info{service_name="unknown_service:aigw",telemetry_sdk_language="go",telemetry_sdk_name="opentelemetry",telemetry_sdk_version="1.36.0"} 1
```

Issue #696, opened 2025-06-10
Support for embedding models.


pksukumar:
*Description*:
Currently we only have support for chat completion models and we have a processor to handle /v1/chat/completions.

We need a dedicated processor to support embedding models at /v1/embeddings.

mathetake:
+1

Xunzhuo:
+1

PR #726, opened 2025-06-18
feat: add support for embeddings

Xunzhuo:

Fixes: https://github.com/envoyproxy/ai-gateway/issues/696

### DIFF
--- a/internal/metrics/embedding_metrics.go
+++ b/internal/metrics/embedding_metrics.go
@@ -1,0 +1,112 @@
+// Copyright Envoy AI Gateway Authors
+// SPDX-License-Identifier: Apache-2.0
+// The full text of the Apache license is available in the LICENSE file at
+// the root of the repo.
+
+package metrics
+
+import (
+	"context"
+	"time"
+
+	"go.opentelemetry.io/otel/attribute"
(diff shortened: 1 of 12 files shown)


internal/metrics/embedding_metrics.go
```diff
@@ -0,0 +1,112 @@
...
+	"github.com/envoyproxy/ai-gateway/filterapi"
+	"github.com/envoyproxy/ai-gateway/filterapi/x"
+)
+
+const (
+	genaiOperationEmbedding = "embedding"
+)
+
+// embedding is the implementation for the embedding AI Gateway metrics.
+type embedding struct {
```
mathetake: Would it be possible generalize chatCompletion metric code? I think we can remove the entire copy-and-paste here

mathetake (commented):
cool, mainly i wonder how much we can remove the duplicate copy-and-paste code here.

After you finish implementing the code in the extproc, the next step is add tests to  https://github.com/envoyproxy/ai-gateway/blob/main/tests/extproc/testupstream_test.go

You can leave Azure and AWS Bedrock as TODO, that can be taken cared by other people.

Xunzhuo:
@mathetake sure, I will try to simplify the code with some abstractions

PR #727, opened 2025-06-18
ci: run unit tests on macos

mathetake:
**Commit Message**

Previously, we haven't run macOS CI since it was a de-facto platform for some maintainers hence it was virtually manually continuously tested. However, we are looking to provide more seamless experience for aigw CLI, it makes sense for us to do the automation of tests on macOS. 

Note that this only enables unit tests and extproc integration test job. The e2e as well as CEL validation tests are not feasible as the nested virtualization is not supported on GHA.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -41,7 +41,7 @@ jobs:
         os:
         - ubuntu-latest
         - macos-latest
-    runs-on: ${{ matrix.platform.os }}
+    runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-go@v5
@@ -54,7 +54,7 @@ jobs:
             ~/.cache/go-build
             ~/go/pkg/mod
             ~/go/bin
-          key: unittest-${{ hashFiles('**/go.mod', '**/go.sum', '**/Makefile') }}-${{ matrix.platform.arch }}
+          key: unittest-${{ hashFiles('**/go.mod', '**/go.sum', '**/Makefile') }}-${{ matrix.os }}
       - env:
           TEST_AWS_ACCESS_KEY_ID: ${{ secrets.AWS_BEDROCK_USER_AWS_ACCESS_KEY_ID }}
           TEST_AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_BEDROCK_USER_AWS_SECRET_ACCESS_KEY }}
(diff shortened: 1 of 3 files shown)


.github/workflows/tests.yaml
```diff
@@ -149,12 +158,6 @@ jobs:
             envoy_gateway_version: 1.4.0
     steps:
       - uses: actions/checkout@v4
-        if: github.event_name != 'pull_request_target'
-      - uses: actions/checkout@v4
-        with:
-          ref: ${{ github.event.pull_request.head.ref }}
-          repository: ${{ github.event.pull_request.head.repo.full_name }}
-        if: github.event_name == 'pull_request_target'
```
mathetake, line 157: cleaned up unused garbage

.github/workflows/tests.yaml
```diff
@@ -121,7 +123,13 @@ jobs:
...
             ~/go/bin
           key: extproc-tests-${{ hashFiles('**/go.mod', '**/go.sum', '**/Makefile') }}
-      - name: Install Envoy
+      - name: Install stable Envoy via func-e
+        if: matrix.version != 'latest'
+        run: |
+          go tool func-e use ${{ matrix.version }}
+          echo $HOME/.func-e/versions/${{ matrix.version }}/bin >> $GITHUB_PATH
+      - name: Install latest Envoy
+        if: matrix.version == 'latest'
```
mathetake, line 135: until the nightly support in func-e/archive-envoy https://github.com/tetratelabs/archive-envoy/issues/12 we cannot just use "go tool func-e" inside the extproc go test code... cc @codefromthecrypt not priority though

codefromthecrypt: ack

mathetake: https://github.com/envoyproxy/ai-gateway/pull/727/commits/e4ef1b4e35c4041de3d45db1f25159fa42ab468f left TODO comment

.github/workflows/tests.yaml
```diff
@@ -92,24 +98,23 @@ jobs:
...
+            os: ubuntu-latest
+    runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v4
-        if: github.event_name != 'pull_request_target'
-      - uses: actions/checkout@v4
-        with:
-          ref: ${{ github.event.pull_request.head.ref }}
-          repository: ${{ github.event.pull_request.head.repo.full_name }}
-        if: github.event_name == 'pull_request_target'
```
mathetake, line 112: same garbage

codefromthecrypt (approved):
I get it. we are externally installing envoy rather that having some lazy install inside the tests (e.g. via code called inside `ENVOY_VERSION=foo aigw`)

missBerg (approved):

PR #730, opened 2025-06-20
extproc: reverts the use of REPLACE_AND_CONTINUE

### LINKED ISSUE #721: Content-length header being removed, force-switching the request to chunked transfer endcoding
*Description*:
returned, etc.

ProcessRequestHeaders() in UpstreamFilter extproc explicitly sets the return status as `CONTINUE_AND_REPLACE` [here](https://github.com/envoyproxy/ai-gateway/blob/ad35294f40ec2375c192c82ac9aeedc9180fa44b/internal/extproc/chatcompletion_processor.go#L228). This results in envoy dropping the content-length header causing the request to switch to [chunked-transfer-encoding](https://en.wikipedia.org/wiki/Chunked_transfer_encoding). Refer [envoy-code](https://github.com/envoyproxy/envoy/blob/e9864c7706770574381bf3185de3eb2c4399fe8c/source/extensions/filters/http/ext_proc/processor_state.cc#L183)

Anthropic models hosted on GCP do not support chunked-transfer-encoding and result in below error

```json
{
  "error": {
    "code": 400,
    "message": "Prediction on deployed model (endpoint_id: <redacted-endpoint-id>, deployed_model_id: <redacted-model-id>) failed with error: \"Bad Request\".",
    "status": "INVALID_ARGUMENT"
  }
}
```

Expected Behavior:
content-length header should not be unilaterally removed by envoy + envoy-ai-gateway

*Repro steps*:

gcp-anthropic model throwing `INVALID_ARGUMENT` when chunked-transfer-encoding header is set
**Note:** `transfer-encoding: chunked` header is set
```bash
curl --request POST \
  --url https://us-east5-aiplatform.googleapis.com/v1/projects/<PROJECT-NAME>/locations/us-east5/publishers/anthropic/models/claude-3-5-haiku@20241022:rawPredict \
  --header 'authorization: Bearer <TOKEN>' \
  --header 'content-type: application/json' \
  --header 'transfer-encoding: chunked' \
  --data '{"anthropic_version": "vertex-2023-10-16","messages": [{"role": "user","content": [{"type": "text","text": "What are you doing?"}]}],"max_tokens": 256,"stream": false}'
```
Output
```json
{
  "error": {
    "code": 400,
    "message": "Prediction on deployed model (endpoint_id: <redacted-endpoint-id>, deployed_model_id: <redacted-model-id>) failed with error: \"Bad Request\".",
    "status": "INVALID_ARGUMENT"
  }
}
```

Valid response when content-length header is set
**Note:** transfer-encoding header is **NOT** set

```bash
curl --request POST \
  --url https://us-east5-aiplatform.googleapis.com/v1/projects/<PROJECT-NAME>/locations/us-east5/publishers/anthropic/models/claude-3-5-haiku@20241022:rawPredict \
  --header 'authorization: Bearer <TOKEN>' \
  --header 'content-type: application/json' \
  --header 'user-agent: vscode-restclient' \
  --data '{"anthropic_version": "vertex-2023-10-16","messages": [{"role": "user","content": [{"type": "text","text": "What are you doing?"}]}],"max_tokens": 256,"stream": false}'
```

Output

```json
{
  "id": "<redacted-response-id>",
  "type": "message",
  "role": "assistant",
  "model": "claude-3-5-haiku-20241022",
  "content": [
    {
      "type": "text",
      "text": "I want to be direct with you. I aim to help you by listening and responding to whatever task or conversation you would like to have. How can I assist you today?"
    }
  ],
  "stop_reason": "end_turn",
  "stop_sequence": null,
  "usage": {
    "input_tokens": 12,
    "cache_creation_input_tokens": 0,
    "cache_read_input_tokens": 0,
    "output_tokens": 38
  }
}
```

>**Note**: If there are privacy concerns, sanitize the data prior to
sharing.

*Environment*:
>Include the environment like gateway version, envoy version and so on.

*Logs*:

Thanks @cmaddalozzo for helping investigate

mathetake:
**Commit Message**

This tries to revert 65ca02a4d9568ec588a0f7be38424aa20e702350.  That was introduced to avoid sending request bodies twice to the extrpco. However, the current implementation of extproc will ALWAYS remove content-length header hence result in forcing the chunked transfer-encoding. That causes some issue with some AI providers as described in #721 for example.

Reverts https://github.com/envoyproxy/ai-gateway/pull/636
Fixes #721

### DIFF
--- a/internal/extensionserver/extensionserver.go
+++ b/internal/extensionserver/extensionserver.go
@@ -242,9 +242,7 @@ func (s *Server) maybeModifyCluster(cluster *clusterv3.Cluster) {
 	extProcConfig.RequestAttributes = []string{"xds.upstream_host_metadata"}
 	extProcConfig.ProcessingMode = &extprocv3http.ProcessingMode{
 		RequestHeaderMode: extprocv3http.ProcessingMode_SEND,
-		// At the upstream filter, it can access the original body in its memory, so it can perform the translation
-		// as well as the authentication at the request headers. Hence, there's no need to send the request body to the extproc.
-		RequestBodyMode: extprocv3http.ProcessingMode_NONE,
+		RequestBodyMode:   extprocv3http.ProcessingMode_BUFFERED,
 		// Response will be handled at the router filter level so that we could avoid the shenanigans around the retry+the upstream filter.
 		ResponseHeaderMode: extprocv3http.ProcessingMode_SKIP,
 		ResponseBodyMode:   extprocv3http.ProcessingMode_NONE,
--- a/internal/extproc/chatcompletion_processor.go
+++ b/internal/extproc/chatcompletion_processor.go
@@ -187,22 +187,31 @@ func (c *chatCompletionProcessorUpstreamFilter) selectTranslator(out filterapi.V
 }
 
 // ProcessRequestHeaders implements [Processor.ProcessRequestHeaders].
-//
-// At the upstream filter, we already have the original request body at request headers phase.
-// So, we simply do the translation and upstream auth at this stage, and send them back to Envoy
-// with the status CONTINUE_AND_REPLACE. This will allows Envoy to not send the request body again
-// to the extproc.
-func (c *chatCompletionProcessorUpstreamFilter) ProcessRequestHeaders(ctx context.Context, _ *corev3.HeaderMap) (res *extprocv3.ProcessingResponse, err error) {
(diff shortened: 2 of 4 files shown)


mathetake:
oh yeah wierd...

mathetake:
ok passing

wengyao04:
Do you think we can revert this change back ? TO allow aws sign with content-lengh ?

https://github.com/envoyproxy/ai-gateway/blob/main/internal/extproc/backendauth/aws.go#L116C2-L116C24

Issue #739, opened 2025-06-21
invalid bsp spec will cause panic


Xunzhuo:
*Description*:
>What issue is being seen? Describe what should be happening instead of
the bug, for example: Envoy should not crash, the expected value isn't
returned, etc.

```
8s.io/apimachinery/pkg/util/runtime.logPanic
        /home/runner/go/pkg/mod/k8s.io/apimachinery@v0.34.0-alpha.0/pkg/util/runtime/runtime.go:142
k8s.io/apimachinery/pkg/util/runtime.handleCrash
        /home/runner/go/pkg/mod/k8s.io/apimachinery@v0.34.0-alpha.0/pkg/util/runtime/runtime.go:107
k8s.io/apimachinery/pkg/util/runtime.HandleCrashWithContext
        /home/runner/go/pkg/mod/k8s.io/apimachinery@v0.34.0-alpha.0/pkg/util/runtime/runtime.go:78
runtime.gopanic
        /opt/hostedtoolcache/go/1.24.4/x64/src/runtime/panic.go:792
runtime.panicmem
        /opt/hostedtoolcache/go/1.24.4/x64/src/runtime/panic.go:262
runtime.sigpanic
        /opt/hostedtoolcache/go/1.24.4/x64/src/runtime/signal_unix.go:925
github.com/envoyproxy/ai-gateway/internal/controller.backendSecurityPolicyIndexFunc
        /home/runner/work/ai-gateway/ai-gateway/internal/controller/controller.go:254
sigs.k8s.io/controller-runtime/pkg/cache.indexByField.func1
        /home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.21.0/pkg/cache/informer_cache.go:236
k8s.io/client-go/tools/cache.(*storeIndex).updateSingleIndex
        /home/runner/go/pkg/mod/k8s.io/client-go@v0.33.1/tools/cache/thread_safe_store.go:162
k8s.io/client-go/tools/cache.(*storeIndex).updateIndices
```

*Repro steps*:
> Include sample requests, environment, etc. All data and inputs
required to reproduce the bug.

Create bsp like:

```
  type: AWSCredentials
  apiKey:
    secretRef:
      name: envoy-ai-gateway-basic-deepseek-apikey
      namespace: default
```

>**Note**: If there are privacy concerns, sanitize the data prior to
sharing.

*Environment*:
>Include the environment like gateway version, envoy version and so on.

*Logs*:
[...]

PR #740, opened 2025-06-21
fix: invalid bsp spec will cause panic

Xunzhuo:
**Commit Message**

fix: invalid bsp spec will cause panic

Fixes: #739

### DIFF
--- a/tests/crdcel/testdata/backendsecuritypolicies/apikey_with_nil_configuration.yaml
+++ b/tests/crdcel/testdata/backendsecuritypolicies/apikey_with_nil_configuration.yaml
@@ -6,7 +6,7 @@
 apiVersion: aigateway.envoyproxy.io/v1alpha1
 kind: BackendSecurityPolicy
 metadata:
-  name: apikey-with-gcp-policy
+  name: apikey-with-nil-policy
   namespace: default
 spec:
   type: APIKey
--- a/tests/crdcel/testdata/backendsecuritypolicies/apikey_with_nil_configuration.yaml
+++ b/tests/crdcel/testdata/backendsecuritypolicies/apikey_with_nil_configuration.yaml
@@ -6,12 +6,7 @@
 apiVersion: aigateway.envoyproxy.io/v1alpha1
 kind: BackendSecurityPolicy
 metadata:
-  name: azure-valid-credentials-policy
+  name: apikey-with-gcp-policy
   namespace: default
 spec:
-  type: AzureCredentials
-  azureCredentials:
-    clientID: dummy_azure_client_id
-    tenantID: dummy_azure_tenant_id
-    clientSecretRef:
-      name: dummy_azure_secret_ref_name
+  type: APIKey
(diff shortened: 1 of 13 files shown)


internal/controller/controller.go
```diff
@@ -248,13 +248,17 @@ func backendSecurityPolicyIndexFunc(o client.Object) []string {
...
+		if apiKey != nil {
+			key = getSecretNameAndNamespace(apiKey.SecretRef, backendSecurityPolicy.Namespace)
+		}
 	case aigv1a1.BackendSecurityPolicyTypeAWSCredentials:
 		awsCreds := backendSecurityPolicy.Spec.AWSCredentials
-		if awsCreds.CredentialsFile != nil {
-			key = getSecretNameAndNamespace(awsCreds.CredentialsFile.SecretRef, backendSecurityPolicy.Namespace)
-		} else if awsCreds.OIDCExchangeToken != nil {
-			key = backendSecurityPolicyKey(backendSecurityPolicy.Namespace, backendSecurityPolicy.Name)
+		if awsCreds != nil {
```
mathetake: can you fix this at the k8s layer level rather than adding unnecessary if block? when the type is AWSBedrock then this should not exist in the first place.

Xunzhuo: Sounds good to me as well, adding CEL now.

mathetake: i mean: add a new CEL validation and corresponding test cases to tests/crdcel and revert this if block

internal/controller/controller.go
```diff
@@ -248,13 +248,17 @@ func backendSecurityPolicyIndexFunc(o client.Object) []string {
 	switch backendSecurityPolicy.Spec.Type {
 	case aigv1a1.BackendSecurityPolicyTypeAPIKey:
 		apiKey := backendSecurityPolicy.Spec.APIKey
-		key = getSecretNameAndNamespace(apiKey.SecretRef, backendSecurityPolicy.Namespace)
+		if apiKey != nil {
```
mathetake: this as well. let's revert and instead reject at k8s layer with CEL validation

Xunzhuo:
PTAL @mathetake

mathetake:
can you add new test cases in https://github.com/envoyproxy/ai-gateway/blob/073d714cafbfd950e70e0b42b8e165510907e3ca/tests/crdcel/main_test.go#L104

Xunzhuo:
CEL is failed, and I wonder if we allow when setting to APIKey but APIKey is nil?

mathetake:
no can you just fix the existing test case as well.

mathetake:
add new test cases & fix the existing test case.

Xunzhuo:
Working on it

tests/crdcel/testdata/backendsecuritypolicies/apikey_with_nil_configuration.yaml
```diff
@@ -0,0 +1,12 @@
+# Copyright Envoy AI Gateway Authors
+# SPDX-License-Identifier: Apache-2.0
+# The full text of the Apache license is available in the LICENSE file at
+# the root of the repo.
+
+apiVersion: aigateway.envoyproxy.io/v1alpha1
+kind: BackendSecurityPolicy
+metadata:
+  name: apikey-with-gcp-policy
```
yuzisun:
```suggestion
  name: apikey-with-nil-policy
```

PR #744, opened 2025-06-21
fix: invalid secret will cause ep cannot start-up

### LINKED ISSUE #742: invalid secret content will cause envoyproxy cannot start-up
*Description*:
returned, etc.

Image

```
  Warning  FailedMount  19s (x7 over 51s)  kubelet            MountVolume.SetUp failed for volume "ai-gateway-envoy-ai-gateway-basic-default" : secret "envoy-ai-gateway-basic-default" not found
```

*Repro steps*:

```
apiVersion: v1
kind: Secret
metadata:
  name: envoy-ai-gateway-basic-aws-credentials
  namespace: default
type: Opaque
stringData:
  # Replace with your AWS credentials.
  apiKey: OPENAI_API_KEY  # Replace with your OpenAI API key.
```

*Logs*:

```
2025-06-21T09:46:05Z    ERROR   Reconciler error        {"controller": "gateway", "controllerGroup": "gateway.networking.k8s.io", "controllerKind": "Gateway", "Gateway": {"name":"envoy-ai-gateway-basic","namespace":"default"}, "namespace": "default", "name": "envoy-ai-gateway-basic", "reconcileID": "100f9975-0a7d-493c-a454-686fad557b06", "error": "failed to create backend auth: failed to get secret envoy-ai-gateway-basic-hunyuan-credentials: secret envoy-ai-gateway-basic-hunyuan-credentials does not contain key credentials"}
```

Xunzhuo:
**Description**

fix: invalid secret content will cause envoyproxy cannot start-up

Fixes #742

### DIFF
--- a/examples/basic/basic.yaml
+++ b/examples/basic/basic.yaml
@@ -314,10 +314,7 @@ metadata:
 type: Opaque
 stringData:
   # Replace with your AWS credentials.
-  credentials: |
-    [default]
-    aws_access_key_id = AWS_ACCESS_KEY_ID
-    aws_secret_access_key = AWS_SECRET_ACCESS_KEY
+  apiKey: OPENAI_API_KEY  # Replace with your OpenAI API key.
 ---
 apiVersion: v1
 kind: Secret
--- a/go.mod
+++ b/go.mod
@@ -17,6 +17,7 @@ require (
 	github.com/google/cel-go v0.25.0
 	github.com/google/go-cmp v0.7.0
 	github.com/google/uuid v1.6.0
+	github.com/hashicorp/go-multierror v1.1.1
 	github.com/openai/openai-go v1.5.0
 	github.com/prometheus/client_golang v1.22.0
 	github.com/stretchr/testify v1.10.0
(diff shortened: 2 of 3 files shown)


yuzisun:
@Xunzhuo The change looks good to me, can you help check the test failures?

examples/basic/basic.yaml
```diff
@@ -314,10 +314,7 @@ metadata:
 type: Opaque
 stringData:
   # Replace with your AWS credentials.
-  credentials: |
-    [default]
-    aws_access_key_id = AWS_ACCESS_KEY_ID
-    aws_secret_access_key = AWS_SECRET_ACCESS_KEY
+  apiKey: OPENAI_API_KEY  # Replace with your OpenAI API key.
```
yuzisun, line 317: AWS does not support API Key

Xunzhuo: This mis-brought by the example to reproduce the issue, will revert.

mathetake (commented):
i am not sure what semantics we want here. i wonder we should proactively check the validity of the secret at backendsecuritypolicy controller and do the early feedback instead of just logging the errors at the later stage.

Xunzhuo:
@mathetake other point is, we shouldn't let envoyproxy cannot start, for one bsp secret invalid. So I make this PR as a draft, to discuss the direction we solve this issue

mathetake:
I will take this over after #988

PR #752, opened 2025-06-23
feat: add GCP Authentication Support

sukumargaonkar:
**Description**

- Implement GCP integration via OIDC token rotator.
  This PR add support for authenticating against GCP via external OIDC tokens. It uses google's Workload Identity Federation [1] to exchange an OIDC token for GCP's access token. It also support optional service account impersonation.
- Add `ProjectName` and `Region` field in BackendSecurityPolicyGCPCredentials

1: https://cloud.google.com/iam/docs/workload-identity-federation

Previous PR: https://github.com/envoyproxy/ai-gateway/pull/635
Related Issue: https://github.com/envoyproxy/ai-gateway/issues/609

### DIFF
--- a/internal/apischema/gcp/gcp.go
+++ b/internal/apischema/gcp/gcp.go
@@ -8,9 +8,27 @@ package gcp
 import "google.golang.org/genai"
 
 type GenerateContentRequest struct {
-	Contents          []genai.Content         `json:"contents"`
-	Tools             []genai.Tool            `json:"tools"`
-	ToolConfig        *genai.ToolConfig       `json:"tool_config,omitempty"`
-	GenerationConfig  *genai.GenerationConfig `json:"generation_config,omitempty"`
-	SystemInstruction *genai.Content          `json:"system_instruction,omitempty"`
+	// Contains the multipart content of a message.
+	//
+	// https://github.com/googleapis/go-genai/blob/6a8184fcaf8bf15f0c566616a7b356560309be9b/types.go#L858
+	Contents []genai.Content `json:"contents"`
+	// Tool details of a tool that the model may use to generate a response.
+	//
+	// https://github.com/googleapis/go-genai/blob/6a8184fcaf8bf15f0c566616a7b356560309be9b/types.go#L1406
+	Tools []genai.Tool `json:"tools"`
+	// Optional. Tool config.
+	// This config is shared for all tools provided in the request.
+	//
+	// https://github.com/googleapis/go-genai/blob/6a8184fcaf8bf15f0c566616a7b356560309be9b/types.go#L1466
+	ToolConfig *genai.ToolConfig `json:"tool_config,omitempty"`
+	// Optional. Generation config.
+	// You can find API default values and more details at https://cloud.google.com/vertex-ai/generative-ai/docs/model-reference/inference#generationconfig
+	// and https://cloud.google.com/vertex-ai/generative-ai/docs/multimodal/content-generation-parameters.
+	GenerationConfig *genai.GenerationConfig `json:"generation_config,omitempty"`
+	// Optional. Instructions for the model to steer it toward better performance.
+	// For example, "Answer as concisely as possible" or "Don't use technical
+	// terms in your response".
+	//
+	// https://github.com/googleapis/go-genai/blob/6a8184fcaf8bf15f0c566616a7b356560309be9b/types.go#L858
+	SystemInstruction *genai.Content `json:"system_instruction,omitempty"`
 }


filterapi/filterconfig.go
```diff
@@ -137,9 +137,18 @@ type VersionedAPISchema struct {
...
-	APISchemaAWSBedrock  APISchemaName = "AWSBedrock"
+	// APISchemaOpenAI represents the standard OpenAI API schema.
+	APISchemaOpenAI APISchemaName = "OpenAI"
+	// APISchemaAWSBedrock represents the AWS Bedrock API schema.
+	APISchemaAWSBedrock APISchemaName = "AWSBedrock"
+	// APISchemaAzureOpenAI represents the Azure OpenAI API schema.
 	APISchemaAzureOpenAI APISchemaName = "AzureOpenAI"
+	// APISchemaGCPGemini represents the Google Cloud Gemini API schema.
+	// Used for Gemini models hosted on Google Cloud Vertex AI.
+	APISchemaGCPGemini APISchemaName = "GCPGemini"
```
yuzisun: this should be GCPVertexAI as defined here https://github.com/envoyproxy/ai-gateway/blob/main/api/v1alpha1/api.go#L458

internal/extproc/chatcompletion_processor.go
```diff
@@ -180,6 +180,10 @@ func (c *chatCompletionProcessorUpstreamFilter) selectTranslator(out filterapi.V
 		c.translator = translator.NewChatCompletionOpenAIToAWSBedrockTranslator(c.modelNameOverride)
 	case filterapi.APISchemaAzureOpenAI:
 		c.translator = translator.NewChatCompletionOpenAIToAzureOpenAITranslator(out.Version, c.modelNameOverride)
+	case filterapi.APISchemaGCPGemini:
```
yuzisun: GCPGemini -> GCPVertexAI

internal/extproc/translator/openai_gcpgemini.go
yuzisun, line 1: rename the file to openai_gcpvertexai.go

internal/extproc/translator/openai_gcpgemini.go
```diff
@@ -0,0 +1,67 @@
...
+import (
+	"io"
+
+	extprocv3 "github.com/envoyproxy/go-control-plane/envoy/service/ext_proc/v3"
+	"google.golang.org/genai"
+
+	"github.com/envoyproxy/ai-gateway/internal/apischema/openai"
+)
+
+type GenerateContentRequest struct {
```
yuzisun: GCP type should be moved under apischema
https://github.com/envoyproxy/ai-gateway/tree/main/internal/apischema

sukumargaonkar:
The current CI failure is related to test-coverage in `internal/testing` package, which can be ignored for now
```
Package coverage threshold (81%) satisfied:	FAIL
  below threshold:				coverage:	threshold:
  internal/testing				74.5% (41/55)	81%
  ```

File: internal/controller/backend_security_policy.go [hunk 4]
```diff
@@ -207,6 +225,10 @@ func getBackendSecurityPolicyAuthOIDC(spec aigv1a1.BackendSecurityPolicySpec) *e
 			return &spec.AzureCredentials.OIDCExchangeToken.OIDC
 		}
 		return nil
+	case aigv1a1.BackendSecurityPolicyTypeGCPCredentials:
```
alexagriffith, line 230: what if it is nil? we dont return or log anything - is that what we want?

alexagriffith: do we want to validate this like we do earlier in this file on line 158?

internal/controller/gateway.go
```diff
@@ -366,6 +366,23 @@ func (c *GatewayController) bspToFilterAPIBackendAuth(ctx context.Context, names
 		return &filterapi.BackendAuth{
 			AzureAuth: &filterapi.AzureAuth{AccessToken: azureAccessToken},
 		}, nil
+	case aigv1a1.BackendSecurityPolicyTypeGCPCredentials:
+		gcpCreds := backendSecurityPolicy.Spec.GCPCredentials
+		if gcpCreds == nil {
```
alexagriffith, line 372: could u reuse validateGCPCredentialsParams here? or if its already called would this be needed?

internal/extproc/translator/gemini_helper.go
```diff
@@ -0,0 +1,62 @@
+// Copyright Envoy AI Gateway Authors
```
alexagriffith, line 1: it would be easier to review if the auth was one PR and the gemini code was another. 
if this would make things difficult though I am ok with it being in one. its just a lot at once.

internal/extproc/translator/gemini_helper.go
```diff
@@ -0,0 +1,62 @@
...
+			},
+		},
+	}
+
+	// If the request body is not empty, we set the content-length header and create a body mutation
+	if len(reqBody) != 0 {
+		// Set the "content-length" header
+		headerMutation.SetHeaders = append(headerMutation.SetHeaders, &corev3.HeaderValueOption{
+			Header: &corev3.HeaderValue{
+				Key:      "content-length",
```
alexagriffith: can we make content-length a constant or use one if it already exists? I use it a few times for anthropic code also

internal/extproc/translator/openai_gcpanthropic_test.go
```diff
@@ -0,0 +1,279 @@
+// Copyright Envoy AI Gateway Authors
```
alexagriffith, line 1: remove this? anthropic tests. updated ones are on my branch

internal/extproc/translator/gemini_helper.go
alexagriffith: i see gemini request/response code is empty - maybe you meant to leave this file out?

internal/extproc/translator/openai_gcpanthropic_test.go
sukumargaonkar: need them in this PR to satisfy test-coverage requirements
feel free to replace them with actual tests when you add implementation

internal/controller/backend_security_policy.go
sukumargaonkar:
```
if spec.GCPCredentials != nil {
	return &spec.GCPCredentials.WorkLoadIdentityFederationConfig.WorkloadIdentityProvider.OIDCProvider.OIDC
}
```

you mean what if gcp credential is nil?
This function would just return nil
the caller will be expected to handle nil since this func returns a pointer

internal/controller/gateway.go
sukumargaonkar: we don't need to validate params inside GCPCredentials here as the struct is directly coming from a CR. The CEL validations on the CR itself would have ensured the gcpCreds are valid. cel-validation are defined here (https://github.com/envoyproxy/ai-gateway/blob/58a01871ba07fd61001c314616251197edf720f5/api/v1alpha1/api.go#L625-L640)

internal/extproc/translator/gemini_helper.go
sukumargaonkar: makes sense,
This PR only has auth related code. It has place holder functions for gemini req/response conversion but not the actual transformation. will make follow up PR for them

https://github.com/envoyproxy/ai-gateway/pull/752/files/920fb332d8b6fd177c22eff169e172edd7a7ef58#diff-74714447912b5ab289f502eef7696896ee87b843e39b0b228c34c8fb6639a34eR30-R42

internal/apischema/gcp/gcp.go
```diff
@@ -0,0 +1,16 @@
+// Copyright Envoy AI Gateway Authors
+// SPDX-License-Identifier: Apache-2.0
+// The full text of the Apache license is available in the LICENSE file at
+// the root of the repo.
+
+package gcp
+
+import "google.golang.org/genai"
+
+type GenerateContentRequest struct {
```
yuzisun, line 10: A you help add the reference links where these types are from and comments to the types.

internal/controller/backend_security_policy.go
```diff
@@ -154,6 +154,24 @@ func (c *BackendSecurityPolicyController) rotateCredential(ctx context.Context,
 		if err != nil {
 			return ctrl.Result{}, err
 		}
+	case aigv1a1.BackendSecurityPolicyTypeGCPCredentials:
+		if err = validateGCPCredentialsParams(bsp.Spec.GCPCredentials); err != nil {
+			return ctrl.Result{}, fmt.Errorf("invalid GCP credentials configuration: %w", err)
+		}
+
+		oidc := getBackendSecurityPolicyAuthOIDC(bsp.Spec)
```
aabchoo, line 164: Make a comment re: oidc being the only auth for GCP? Otherwise you have to validate this auth method is oidc

internal/controller/backend_security_policy.go
```diff
@@ -154,6 +154,24 @@ func (c *BackendSecurityPolicyController) rotateCredential(ctx context.Context,
...
+
+		// Create the OIDC token provider that will be used to get tokens from the OIDC provider
```
aabchoo:
```suggestion
		// Create the OIDC token provider that will be used to get tokens from the OIDC provider.
```

internal/controller/rotators/gcp_oidc_token_rotator.go
```diff
@@ -0,0 +1,331 @@
...
+) (*tokenprovider.TokenExpiry, error)
+
+// gcpOIDCTokenRotator implements Rotator interface for GCP access token exchange.
+// It handles the complete authentication flow for GCP Workload Identity Federation:
+// 1. Obtaining an OIDC token from the configured provider
+// 2. Exchanging the OIDC token for a GCP STS token
+// 3. Using the STS token to impersonate a GCP service account
+// 4. Storing the resulting access token in a Kubernetes secret
+type gcpOIDCTokenRotator struct {
+	client client.Client // Kubernetes client for interacting with the cluster
```
aabchoo:
```suggestion
	client client.Client // Kubernetes client for interacting with the cluster.
```

aabchoo: similar with below for consistency

internal/controller/rotators/gcp_oidc_token_rotator.go
```diff
@@ -0,0 +1,331 @@
...
+
+	saTokenFunc  serviceAccountTokenGenerator
+	stsTokenFunc stsTokenGenerator
+}
+
+// NewGCPOIDCTokenRotator creates a new gcpOIDCTokenRotator with the given parameters.
+func NewGCPOIDCTokenRotator(
+	client client.Client,
+	logger logr.Logger,
+	bsp aigv1a1.BackendSecurityPolicy,
```
aabchoo, line 94: Why not pass the gcp credentials directly instead of bsp?

internal/controller/rotators/gcp_oidc_token_rotator.go
```diff
@@ -0,0 +1,331 @@
...
+		oidcProvider:                   tokenProvider,
+		saTokenFunc:                    impersonateServiceAccount,
+		stsTokenFunc:                   exchangeJWTForSTSToken,
+	}, nil
+}
+
+// IsExpired implements [Rotator.IsExpired].
+// IsExpired checks if the preRotation time is before the current time.
+func (r *gcpOIDCTokenRotator) IsExpired(preRotationExpirationTime time.Time) bool {
+	// Use the common IsBufferedTimeExpired helper to determine if the token has expired
```
aabchoo, line 120:
```suggestion
	// Use the common IsBufferedTimeExpired helper to determine if the token has expired.
```

Similar with the below

internal/controller/rotators/gcp_oidc_token_rotator.go
```diff
@@ -0,0 +1,331 @@
...
+	tokenProvider tokenprovider.TokenProvider,
+) (Rotator, error) {
+	logger = logger.WithName("gcp-token-rotator")
+
+	if bsp.Spec.GCPCredentials == nil {
+		return nil, fmt.Errorf("GCP credentials are not configured in BackendSecurityPolicy %s/%s", bsp.Namespace, bsp.Name)
+	}
+
+	return &gcpOIDCTokenRotator{
+		client:                         client,
```
aabchoo, line 105: Does this client have access to a proxy if need be similar to how we have one for azure and bedrock?

internal/extproc/backendauth/gcp.go
```diff
@@ -0,0 +1,85 @@
...
+	"fmt"
+
+	corev3 "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
+	extprocv3 "github.com/envoyproxy/go-control-plane/envoy/service/ext_proc/v3"
+
+	"github.com/envoyproxy/ai-gateway/filterapi"
+)
+
+type gcpHandler struct {
+	gcpAccessToken string // The GCP access token used for authentication
```
aabchoo:
```suggestion
	gcpAccessToken string // The GCP access token used for authentication.
```

Same with below

internal/extproc/backendauth/gcp.go
```diff
@@ -0,0 +1,85 @@
...
+	region         string // The GCP region to use for requests
+	projectName    string // The GCP project to use for requests
+}
+
+func newGCPHandler(gcpAuth *filterapi.GCPAuth) (Handler, error) {
+	if gcpAuth == nil {
+		return nil, fmt.Errorf("GCP auth configuration cannot be nil")
+	}
+
+	return &gcpHandler{
```
aabchoo, line 33: thoughts on checking empty string?

internal/extproc/translator/gemini_helper.go
```diff
@@ -0,0 +1,63 @@
...
+
+	"github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
+	"github.com/envoyproxy/go-control-plane/envoy/service/ext_proc/v3"
+)
+
+const (
+	GCPModelPublisherGoogle    = "google"
+	GCPModelPublisherAnthropic = "anthropic"
+	GCPMethodGenerateContent   = "generateContent"
+	HTTPHeaderKeyContentLength = "content-length"
```
aabchoo: Just confirming, this is upstream format? Weird seeing camel case style and also dash style

internal/controller/rotators/gcp_oidc_token_rotator.go
sukumargaonkar: we need to access bsp.Name and bsp.Namespace
rather than passing them as individual params, made the entire bsp accessible inside the function

internal/extproc/translator/gemini_helper.go
sukumargaonkar: following this
https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Length

internal/extproc/translator/openai_gcpanthropic_test.go
alexagriffith: ah ok

internal/extproc/translator/gemini_helper.go
alexagriffith: my bad

internal/controller/rotators/gcp_oidc_token_rotator.go
sukumargaonkar: added proxy support

internal/extproc/backendauth/gcp.go
sukumargaonkar: added

yuzisun (approved):

internal/extproc/translator/gemini_helper.go
```diff
@@ -0,0 +1,63 @@
...
+
+	"github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
+	"github.com/envoyproxy/go-control-plane/envoy/service/ext_proc/v3"
+)
+
+const (
+	GCPModelPublisherGoogle    = "google"
+	GCPModelPublisherAnthropic = "anthropic"
+	GCPMethodGenerateContent   = "generateContent"
+	HTTPHeaderKeyContentLength = "Content-Length"
```
mathetake, line 20: can you unexport them

internal/controller/backend_security_policy.go
```diff
@@ -238,3 +262,28 @@ func (c *BackendSecurityPolicyController) updateBackendSecurityPolicyStatus(ctx
 		c.logger.Error(err, "failed to update BackendSecurityPolicy status")
 	}
 }
+
+func validateGCPCredentialsParams(gcpCreds *aigv1a1.BackendSecurityPolicyGCPCredentials) error {
```
mathetake, line 266: shouldn't all of this condition be part of k8s CEL validation? (If so, it's redundant)

internal/controller/rotators/gcp_oidc_token_rotator.go
```diff
@@ -0,0 +1,370 @@
...
+	tokenTypeAccessToken = "urn:ietf:params:oauth:token-type:access_token" //nolint:gosec
+	// tokenTypeJWT indicates the subject token type is a JWT.
+	tokenTypeJWT = "urn:ietf:params:oauth:token-type:jwt" //nolint:gosec
+	// stsTokenScope is the OAuth scope for GCP cloud platform operations.
+	stsTokenScope = "https://www.googleapis.com/auth/cloud-platform" //nolint:gosec
+)
+
+// serviceAccountTokenGenerator defines a function type for generating a GCP service account access token
+// using an STS token and impersonation configuration.
+type serviceAccountTokenGenerator func(
```
mathetake, line 50: this needs a gcp prefix in the struct name

internal/controller/rotators/gcp_oidc_token_rotator.go
```diff
@@ -0,0 +1,370 @@
...
+	// grantTypeTokenExchange is the OAuth 2.0 grant type for token exchange.
+	grantTypeTokenExchange = "urn:ietf:params:oauth:grant-type:token-exchange" //nolint:gosec
+	// gcpIAMScope is the OAuth scope for IAM operations in GCP.
+	gcpIAMScope = "https://www.googleapis.com/auth/iam" //nolint:gosec
+	// tokenTypeAccessToken indicates the requested token type is an access token.
+	tokenTypeAccessToken = "urn:ietf:params:oauth:token-type:access_token" //nolint:gosec
+	// tokenTypeJWT indicates the subject token type is a JWT.
+	tokenTypeJWT = "urn:ietf:params:oauth:token-type:jwt" //nolint:gosec
+	// stsTokenScope is the OAuth scope for GCP cloud platform operations.
+	stsTokenScope = "https://www.googleapis.com/auth/cloud-platform" //nolint:gosec
```
mathetake, line 45: `gcp` prefix is needed here as well

internal/controller/rotators/gcp_oidc_token_rotator.go
```diff
@@ -0,0 +1,370 @@
...
+type serviceAccountTokenGenerator func(
+	ctx context.Context,
+	stsToken string,
+	saConfig aigv1a1.GCPServiceAccountImpersonationConfig,
+	opts ...option.ClientOption,
+) (*tokenprovider.TokenExpiry, error)
+
+// stsTokenGenerator defines a function type for exchanging a JWT token for a GCP STS token
+// using Workload Identity Federation configuration.
+type stsTokenGenerator func(
```
mathetake, line 59: can you also add gcp prefix here

internal/extproc/backendauth/gcp.go
```diff
@@ -0,0 +1,89 @@
...
+
+	// Find and update the ":path" header by prepending the prefix.
+	for _, hdr := range headerMut.SetHeaders {
+		if hdr.Header != nil && hdr.Header.Key == ":path" {
+			pathHeaderFound = true
+			// Update the string value if present.
+			if len(hdr.Header.Value) > 0 {
+				suffixPath := hdr.Header.Value
+				hdr.Header.Value = fmt.Sprintf("%s/%s", prefixPath, suffixPath)
+			}
```
mathetake, line 62: i don't think you need this in practice. (Envoy's API naming is really bad here, but) we should stick to .RawValue for setting headers. (.Value is for formatter) https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/base.proto#config-core-v3-headervalue

yuzisun:
@sukumargaonkar can you take a look at @mathetake’s comments and address in the follow up PR?

mathetake:
@sukumargaonkar also could you add more unit test around here?
image

Issue #705, opened 2025-06-13
doc: add "supported endpoints" page


mathetake:
*Description*:

Since we only have support for a subset of OpenAI API, it would be beneficial to add a page we explicitly state which endpoints are supported etc.

Xunzhuo:
we would better to add this after landing `/v1/embeddings`

PR #755, opened 2025-06-24
feat: add support for embeddings

### LINKED ISSUE #696: Support for embedding models.

Xunzhuo:
**Description**

This commit adds an initial implementation /embeddings endpoint. This only implements an OpenAI translator for it, so AWS Bedrock, and Azure are not implemented. 

Fixes #696 

Clipboard_Screenshot_1750738098

### DIFF
--- a/internal/extproc/embeddings_processor.go
+++ b/internal/extproc/embeddings_processor.go
@@ -18,7 +18,6 @@ import (
 	corev3 "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
 	extprocv3 "github.com/envoyproxy/go-control-plane/envoy/service/ext_proc/v3"
 	typev3 "github.com/envoyproxy/go-control-plane/envoy/type/v3"
-	"google.golang.org/protobuf/types/known/structpb"
 
 	"github.com/envoyproxy/ai-gateway/filterapi"
 	"github.com/envoyproxy/ai-gateway/filterapi/x"
@@ -306,7 +305,7 @@ func (e *embeddingsProcessorUpstreamFilter) ProcessResponseBody(ctx context.Cont
 	e.metrics.RecordTokenUsage(ctx, tokenUsage.InputTokens, tokenUsage.TotalTokens)
 
 	if body.EndOfStream && len(e.config.requestCosts) > 0 {
-		resp.DynamicMetadata, err = e.maybeBuildDynamicMetadata()
+		resp.DynamicMetadata, err = buildDynamicMetadata(e.config, &e.costs, e.requestHeaders)
 		if err != nil {
 			return nil, fmt.Errorf("failed to build dynamic metadata: %w", err)
 		}
@@ -336,10 +335,6 @@ func (e *embeddingsProcessorUpstreamFilter) SetBackend(ctx context.Context, b *f
 	e.onRetry = rp.upstreamFilterCount > 1
 	rp.upstreamFilter = e
 	return
-}
-
-func (e *embeddingsProcessorUpstreamFilter) maybeBuildDynamicMetadata() (*structpb.Struct, error) {
-	return buildDynamicMetadata(e.config, &e.costs, e.requestHeaders)
 }
 
 func parseOpenAIEmbeddingBody(body *extprocv3.HttpBody) (modelName string, rb *openai.EmbeddingRequest, err error) {
--- a/internal/metrics/embeddings_metrics_test.go
+++ b/internal/metrics/embeddings_metrics_test.go
@@ -7,7 +7,6 @@ package metrics
 
 import (
 	"testing"
-	"time"
 
 	"github.com/stretchr/testify/assert"
 	"go.opentelemetry.io/otel/attribute"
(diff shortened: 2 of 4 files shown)


examples/embeddings/embeddings.yaml
mathetake, line 1: why do you need another example for embeddings? could you simply add additional code to tests/e2e/basic_test.go?

examples/README.md
mathetake, line 1: can you do this in another PR ?

internal/metrics/embeddings_metrics.go
```diff
@@ -0,0 +1,108 @@
+// Copyright Envoy AI Gateway Authors
```
mathetake, line 1: Looks like you ignored my previous comment, but the entire code here should not be copied and pasted but instead you could reuse almost all the code from the existing chat_completion_metrics.go by passing "operation". 

https://opentelemetry.io/docs/specs/semconv/registry/attributes/gen-ai/#gen-ai-operation-name

site/docs/capabilities/embeddings.md
mathetake, line 1: I appreciate your doc here but could you do this in another PR together with chat completion (per #705)? this creates a weird diff with this existing /chat/completion end point.

Also, i don't think the embedding endpoint has anything new compared to chat/completion so this doc seems a duplicate.

site/docs/getting-started/connect-providers/aws-bedrock.md
```diff
@@ -8,6 +8,12 @@ sidebar_position: 3
 
 This guide will help you configure Envoy AI Gateway to work with AWS Bedrock's foundation models.
 
+## Supported Features
```
mathetake: let's do this "support matrix" in another PR for #705, so can you revert this for now.

site/docs/getting-started/connect-providers/azure-openai.md
```diff
@@ -9,6 +9,14 @@ sidebar_position: 3
 
 This guide will help you configure Envoy AI Gateway to work with Azure OpenAI's foundation models.
 
+## Supported Features
```
mathetake: let's do this "support matrix" in another PR for #705, so can you revert this for now.

site/docs/getting-started/connect-providers/openai.md
```diff
@@ -123,6 +140,40 @@ spec:
         - name: envoy-ai-gateway-basic-openai
 ```
 
+### Embedding Models
```
mathetake: can you delete this and simply add a new matching to AIGatewayRoute above

mathetake (commented):
left some comments but this is exciting!

site/docs/getting-started/connect-providers/openai.md
Xunzhuo: Done

site/docs/getting-started/connect-providers/azure-openai.md
Xunzhuo: Done

site/docs/getting-started/connect-providers/aws-bedrock.md
Xunzhuo: Done

site/docs/capabilities/embeddings.md
Xunzhuo: Removed

examples/README.md
Xunzhuo: Reverted

examples/embeddings/embeddings.yaml
Xunzhuo: Make sense, removed

internal/metrics/embeddings_metrics.go
Xunzhuo: done, see: https://github.com/envoyproxy/ai-gateway/pull/755/commits/6dd5d61d126c3f148a66db298bd21ff43c7da06b

filterapi/x/x.go
```diff
@@ -68,3 +68,26 @@ type ChatCompletionMetrics interface {
...
+	SetModel(model string)
+	// SetBackend sets the selected backend when the routing decision has been made. This is usually called
+	// after parsing the request body to determine the model and invoke the routing logic.
+	SetBackend(backend *filterapi.Backend)
+
+	// RecordTokenUsage records token usage metrics for embeddings (only input and total tokens are relevant).
+	RecordTokenUsage(ctx context.Context, inputTokens, totalTokens uint32, extraAttrs ...attribute.KeyValue)
+	// RecordRequestCompletion records latency metrics for the entire request
+	RecordRequestCompletion(ctx context.Context, success bool, extraAttrs ...attribute.KeyValue)
+}
```
mathetake: can you remove this for now? let's not add x component unless someone requests. I think i am even looking to remove the chat completion one as well

internal/apischema/openai/openai.go
```diff
@@ -184,6 +184,15 @@ func (s *StringOrArray) UnmarshalJSON(data []byte) error {
 		return nil
 	}
 
+	// Try to unmarshal as array of strings (for embeddings)
```
mathetake: super nit, but could you add period for all code comments not only here?

```suggestion
	// Try to unmarshal as array of strings (for embeddings).
```

internal/extproc/embeddings_processor.go
```diff
@@ -0,0 +1,393 @@
...
+	}
+	e.handler = backendHandler
+	e.originalRequestBody = rp.originalRequestBody
+	e.originalRequestBodyRaw = rp.originalRequestBodyRaw
+	e.onRetry = rp.upstreamFilterCount > 1
+	rp.upstreamFilter = e
+	return
+}
+
+func (e *embeddingsProcessorUpstreamFilter) maybeBuildDynamicMetadata() (*structpb.Struct, error) {
```
mathetake: can you generalize this and share it with chat completion?

internal/extproc/translator/openai_embeddings.go
```diff
@@ -0,0 +1,140 @@
...
+	}
+
+	var resp openai.EmbeddingResponse
+	if err := json.NewDecoder(body).Decode(&resp); err != nil {
+		return nil, nil, tokenUsage, fmt.Errorf("failed to unmarshal body: %w", err)
+	}
+	tokenUsage = LLMTokenUsage{
+		InputTokens: uint32(resp.Usage.PromptTokens), //nolint:gosec
+		TotalTokens: uint32(resp.Usage.TotalTokens),  //nolint:gosec
+		// Embeddings don't have output tokens, only input and total
```
mathetake:
```suggestion
		// Embeddings don't have output tokens, only input and total.
```

internal/metrics/chat_completion_metrics.go
```diff
@@ -12,18 +12,14 @@ import (
...
 	"go.opentelemetry.io/otel/metric"
 
-	"github.com/envoyproxy/ai-gateway/filterapi"
 	"github.com/envoyproxy/ai-gateway/filterapi/x"
 )
 
 // chatCompletion is the implementation for the chat completion AI Gateway metrics.
 type chatCompletion struct {
-	metrics        *genAI
+	*baseMetrics
```
mathetake: i don't think this needs to be a pointer (less allocation:)
```suggestion
	baseMetrics
```

internal/metrics/embeddings_metrics.go
```diff
@@ -0,0 +1,49 @@
...
+
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/metric"
+
+	"github.com/envoyproxy/ai-gateway/filterapi/x"
+)
+
+// embeddings is the implementation for the embeddings AI Gateway metrics.
+type embeddings struct {
+	*baseMetrics
```
mathetake: same comment
```suggestion
	baseMetrics
```

mathetake (commented):
almost there!!!

filterapi/x/x.go
Xunzhuo: Can you elaborate what I need to remove here and what the x component means?

mathetake: could you remove all the change in this x/x.go as these are not needed at this point

Xunzhuo: When I am going to remove the EmbeddingsMetrics interface, it will cause the mock logics in internal/extproc/embeddings_processor_test.go becomes weird(embeddingsProcessorUpstreamFilter needs a metrics interface to run its mock), thoughts on it? Rmove the mock test or keep the interface?

mathetake: ahhh sorry i misunderstood the situation... i meant was let's remove `CustomEmbeddings`, but let's keep `EmbeddingsMetrics`.

```
// NewCustomEmbeddingsMetrics is the function to create a custom embeddings AI Gateway metrics over
// the default metrics. This is nil by default and can be set by the custom build of external processor.
var NewCustomEmbeddingsMetrics NewCustomEmbeddingsMetricsFn

// NewCustomEmbeddingsMetricsFn is the function to create a custom embeddings AI Gateway metrics.
type NewCustomEmbeddingsMetricsFn func(meter metric.Meter) EmbeddingsMetrics
```

internal/extproc/embeddings_processor.go
Xunzhuo: Working on it

internal/metrics/embeddings_metrics_test.go
```diff
@@ -0,0 +1,230 @@
+// Copyright Envoy AI Gateway Authors
```
mathetake, line 1: i think the test here can be deleted as it's duplicate of the baseMetrics method which are already tested in chat_completion_metrics_test.go.

Instead, shall we have one test for RecordTokenUsage which is the only diff from the base.

site/docs/getting-started/connect-providers/openai.md
```diff
@@ -98,7 +98,9 @@ If you encounter issues:
 
 ## Configuring More Models
 
-To use more models, add more [AIGatewayRouteRule]s to the `basic.yaml` file with the [model alias] in the `value` field. For example, to use [o1]
+To use more models, add more [AIGatewayRouteRule]s to the `basic.yaml` file with the [model alias] in the `value` field.
+
+For example, for chat completion models, to use [o1], for embedding models, to use [text-embedding-ada-002](https://platform.openai.com/docs/models/text-embedding-ada-002):
```
mathetake:
```suggestion
For example, let's add [o1] as a chat completion model, and [text-embedding-ada-002](https://platform.openai.com/docs/models/text-embedding-ada-002) as an embedding model:
```

internal/extproc/embeddings_processor.go
```diff
@@ -0,0 +1,351 @@
...
```
mathetake: i think this is redundant, so you can just call buildDynamicMetadata directly at the callsite

internal/metrics/embeddings_metrics_test.go
Xunzhuo: Removed and kept only RecordTokenUsage related testcases

site/docs/getting-started/connect-providers/openai.md
Xunzhuo: Done

internal/extproc/embeddings_processor.go
Xunzhuo: Done

mathetake (approved):
Thank you! I want to note that we should be able to share more code with chat_completions processor but I agree that's a bit difficult due to the way it's written. I will go ahead and merge for now to unblock you and others to work on the other provider support.

The follow up PR I would like to see the resulution of #705 where we have a table to show which endpoints are supported by each backend. Maybe editing the existing "supported providers page" to include it or adding another page.

anyways thank you for working on this one!

PR #756, opened 2025-06-24
ci: add codeql workflows

Xunzhuo:
**Description**

This adds the codeql action, against a repository's source code to find security vulnerabilities. It automatically uploads the results to GitHub so they can be displayed on pull requests and in the repository's security tab.

### DIFF
--- a/.github/workflows/codeql.yaml
+++ b/.github/workflows/codeql.yaml
@@ -26,9 +26,6 @@ jobs:
 
     strategy:
       fail-fast: false
-      matrix:
-        language:
-        - go
 
     steps:
     - name: Checkout repository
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql.yaml
```diff
@@ -0,0 +1,54 @@
...
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
+
+    strategy:
+      fail-fast: false
+      matrix:
+        language:
+        - go
```
mathetake: unnecessary matrix (at least for now), so let's remove

Xunzhuo: Updated

PR #758, opened 2025-06-24
ci: add osv-scanner workflows

Xunzhuo:
**Commit Message**

ci: add osv-scanner workflows

### DIFF
--- a/.github/workflows/osv-scanner.yaml
+++ b/.github/workflows/osv-scanner.yaml
@@ -1,0 +1,44 @@
+name: OSV-Scanner
+
+on:
+  pull_request:
+    branches:
+    - "main"
+  merge_group:
+    branches:
+    - "main"
+  push:
+    branches:
+    - "main"
+  schedule:
+  - cron: '44 15 * * 5'
+
+permissions:
+  contents: read
+
+jobs:
+  scan-scheduled:
+    if: ${{ github.event_name == 'push' || github.event_name == 'schedule' }}
+    uses: "google/osv-scanner-action/.github/workflows/osv-scanner-reusable.yml@e69cc6c86b31f1e7e23935bbe7031b50e51082de"  # v2.0.2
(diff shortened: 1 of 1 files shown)


mathetake:
@Xunzhuo super nit but could you write a commit message in English sentences just like any other commit and PRs?

mathetake:
no need to prefix "ci: " in your commit message

.github/workflows/osv-scanner.yaml
```diff
@@ -0,0 +1,44 @@
+name: OSV-Scanner
+
+on:
+  pull_request:
+    branches:
+    - "main"
+  merge_group:
+    branches:
+    - "main"
```
mathetake: we are not using merge queue

Xunzhuo: Updated

Xunzhuo:
> @Xunzhuo super nit but could you write a commit message in English sentences just like any other commit and PRs?

Clipboard_Screenshot_1750820083

Sorry I dont get it, it is in English and also based on the https://www.conventionalcommits.org/en/v1.0.0/

Clipboard_Screenshot_1750820154

mathetake:
no, we are only checking the semantics for the title, and that's what the CI checks it. Please check the existing commits and follow the style.

mathetake:
in other words, the "<prefix>: adds foo for bar" is only checked for the PR title which is used as a first line of commit message. The content in the description follows after that. What you referred to is for the first line of the commit and that's the only thing the CI checks.

mathetake:
sorry i will revert this as this overlaps with trivy as well as OSV scanner is not aware of the state of the repository's security data (e.g. i ignored some npm packages) which will cause a double maintenance.

Xunzhuo:
I was going to close this, I realized the revert reason same as you, so nice revert! Haha

PR #759, opened 2025-06-24
ci: add the codecov check

Xunzhuo:
**Description**

This commit adds an additional coverage check by codecov. Codecov can check the patch level coverage and will add useful annotations on the PR that enable reviewers to notice uncovered lines. This is complementary to the existing go-test-coverage which can be run locally vs codecov cannot be run locally.

### DIFF
--- a/.github/workflows/build_and_test.yaml
+++ b/.github/workflows/build_and_test.yaml
@@ -31,6 +31,7 @@ concurrency:
 permissions:
   contents: read
   packages: write
+  id-token: write
 
 jobs:
   unittest:
@@ -78,6 +79,7 @@ jobs:
           files: ./out/go-test-coverage.out
           name: codecov-envoy-ai-gateway
           verbose: true
+          # https://github.com/codecov/codecov-action/issues/1594#issuecomment-2394913029
           use_oidc: ${{ !(github.event_name == 'pull_request' && github.event.pull_request.head.repo.fork) }}
 
   test_crdcel:


mathetake:
could you explore how this will play with the existing coverage setup with go-testcoverage? is it possible to migrate `make test-coverage` to use the setting here?  https://github.com/envoyproxy/ai-gateway/blob/main/.testcoverage.yml

mathetake:
so the idea is to give all contributors the ability to reproduce the CI failure and alerts by codecov bot. Otherwise, it's useless as we see in many projects that the comments by bod get ignored and just become a noise.

Makefile
```diff
@@ -129,9 +129,7 @@ test: ## Run the unit tests for the codebase.
 # This runs the unit tests for the codebase with coverage check.
 .PHONY: test-coverage
 test-coverage: ## Run the unit tests for the codebase with coverage check.
-	@mkdir -p $(OUTPUT_DIR)
-	@$(MAKE) test GO_TEST_ARGS="-coverprofile=$(OUTPUT_DIR)/go-test-coverage.out -covermode=atomic -coverpkg=github.com/envoyproxy/ai-gateway/... $(GO_TEST_ARGS)"
-	@go tool go-test-coverage --config=.testcoverage.yml
```
mathetake: cool, last request is to could you remove `go-test-coverage` from go.mod tool section?

Makefile
```diff
@@ -129,9 +129,7 @@ test: ## Run the unit tests for the codebase.
...
+	@$(MAKE) test GO_TEST_ARGS="-coverprofile=coverage.xml -covermode=atomic -coverpkg=github.com/envoyproxy/ai-gateway/... $(GO_TEST_ARGS)"
```
mathetake: help me understand this correctly but this itself is not failing the CI right? How can a contributor know if the current change passes all the coverage checks without pushing it? That's why I used go-test-coverage in the first place, and i wonder if it's possible to do so with codecov setting.

yuzisun: codecov does not show full project coverage

image

mathetake (approved):

PR #761, opened 2025-06-24
ci: add stale workflow

Xunzhuo:
**Commit Message**

This adds the stale workflow to detect the issues/PRs which are inactivated.

We usually use it as a reminder for both maintainer and contributors, not a action to directly close the PR/issues. So we only enable the `days-before-stale` and disable `days-before-close`.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -21,8 +21,6 @@ jobs:
       uses: actions/stale@5bef64f19d7facfb25b37b414482c7164d639639  # v9.1.0
       with:
         repo-token: ${{ secrets.GITHUB_TOKEN }}
-        # Different amounts of days for issues/PRs are not currently supported but there is a PR
-        # open for it: https://github.com/actions/stale/issues/214
         days-before-stale: 30
         days-before-close: -1
         stale-issue-message: >
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,41 @@
+on:
+  workflow_dispatch:
+  schedule:
+  - cron: '0 */4 * * *'
+
+permissions:
+  contents: read
+
+jobs:
+  prune_stale:
+    permissions:
+      issues: write  # for actions/stale to close stale issues
+      pull-requests: write  # for actions/stale to close stale PRs
+    name: Prune Stale
+    runs-on: ubuntu-22.04
+    # do not run it in forked repos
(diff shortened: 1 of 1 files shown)


.github/workflows/stale.yml
```diff
@@ -0,0 +1,41 @@
...
+
+    steps:
+    - name: Prune Stale
+      uses: actions/stale@5bef64f19d7facfb25b37b414482c7164d639639  # v9.1.0
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        # Different amounts of days for issues/PRs are not currently supported but there is a PR
+        # open for it: https://github.com/actions/stale/issues/214
+        days-before-stale: 30
+        days-before-close: -1
```
mathetake, line 25: so we won't close the issue/prs? could you explain the rationale

.github/workflows/stale.yml
```diff
@@ -0,0 +1,41 @@
...
+    # do not run it in forked repos
+    if: github.repository == 'envoyproxy/ai-gateway'
+
+    steps:
+    - name: Prune Stale
+      uses: actions/stale@5bef64f19d7facfb25b37b414482c7164d639639  # v9.1.0
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        # Different amounts of days for issues/PRs are not currently supported but there is a PR
+        # open for it: https://github.com/actions/stale/issues/214
```
mathetake: it looks it's supported?

.github/workflows/stale.yml
Xunzhuo: Yes, we added the stale labels for now, do not close it.

.github/workflows/stale.yml
Xunzhuo: Yes will remove this line, we dont have the requests to diff the days for issue and PR.

.github/workflows/stale.yml
mathetake: i am genuinely curious why not close?

Xunzhuo: It is a maintenance choice for how to deal with the stale PR and issue, we usually use it as a reminder for both maintainer and contributors, not a action to directly close the PR/issues

mathetake: cool, could you put that rationale here in this yaml so that others can see what we were thinking at this point.

Xunzhuo: Updated in description

Issue #549, opened 2025-04-02
BackendSecurityPolicy should target AIServiceBackend


arkodg:
*Description*:
>Describe the issue.

Noticed that the AIServiceBackend targets a BackendSecurityPolicy
https://aigateway.envoyproxy.io/docs/api/#aiservicebackendspec

The idiomatic design pattern is for the Policy to target the Backend using `targetRefs` field
https://gateway-api.sigs.k8s.io/geps/gep-2648/

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

Issue #675, opened 2025-06-03
Consider making AIServiceBackend an abstraction for Backends for AI Services


arkodg:
*Description*:
>Describe the desired behavior, what scenario it enables and how it
would be used.

Today https://aigateway.envoyproxy.io/docs/api/#aiservicebackendspec references two resources

* A BackendRef resource
* A BackendSecurityPolicyRef

For the egress case where clients are connecting to external model services, it will help if these references are converted into absolute values/fields are embedded instead of references, since the persona responsible for managing the AIServiceBackend is the same.

This would make it more inline with AIGatewayRoute which is an abstraction on top of HTTPRoute for routing AI traffic

Relates to https://github.com/envoyproxy/ai-gateway/issues/603

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

Xunzhuo:
+1 for this, would love to make API simplify.

the ideal approach I think:

1. GatewayClass -> Gateway -> AIGatewayRoute -> AIServiceBackend

the current approach is:

1. GatewayClass -> Gateway -> AIGatewayRoute -> AIServiceBackend + n * Backend + n * BackendSecurityPolicy + n * BackendTLSPolicy

the user-facing API grows large, and this will bring some issues:
1. the UX is bad and people need to understand a lot of resources we define.
2. the system logics becomes more complex.
3. the frontend engineer feels bad about it, need a lof of works in console building

yuzisun:
The proposal here is to embed the `Backend` and `BackendSecurityPolicy` type in the `AIServiceBackend` right?

arkodg:
yeah @yuzisun, also `Service` prefix can be removed, since its redundant (a `Service` is a Backend), and can be reduced to `AIBackend` (even `AIGatewayRoute` could be simplified to `AIRoute`, since a `Route` always attaches to a `Gateway/Listener` parentRef)

Xunzhuo:
Agreed with @arkodg 

GatewayClass -> Gateway -> AIGatewayRoute -> AIServiceBackend

To:

GatewayClass -> Gateway -> AIRoute -> AIBackend

yuzisun:
Let’s create a poll with the community ?

Xunzhuo:
sure @yuzisun 

### the current approach is

```
GatewayClass -> Gateway -> AIGatewayRoute -> AIServiceBackend + n * Backend + n * BackendSecurityPolicy + n * BackendTLSPolicy
```

[...]

PR #762, opened 2025-06-24
feat: support targetRef in backend security policy

Xunzhuo:
**Commit Message**

feat: support targetRef in backend security policy

Fixes: https://github.com/envoyproxy/ai-gateway/issues/549


mathetake:
thank for the work here... but i think this involves the API refactoring discussion we have in like #675 , so let's wait for it to reach the consensus  cc @missBerg

Xunzhuo:
Yes, I did this earlier than the discussion, so let us hold.

examples/basic/bsp-target-ref.yaml
```diff
@@ -0,0 +1,92 @@
+# Copyright Envoy AI Gateway Authors
```
mathetake: could you delete this file, and instead fix all existing examples to use targetRefs?

mathetake:
i will take over this pr

mathetake:
i pushed some changes. PTAL @Xunzhuo

Xunzhuo:
thanks @mathetake!

examples/basic/basic.yaml
```diff
@@ -122,6 +122,8 @@ spec:
     name: envoy-ai-gateway-basic-azure
     kind: Backend
     group: gateway.envoyproxy.io
+  # Deprecated: Use targetRefs in BackendSecurityPolicy instead. See "envoy-ai-gateway-basic-openai-apikey" below.
+  # This is mainly to ensure backward compatibility in tests.
   backendSecurityPolicyRef:
```
Xunzhuo: Should we update all the references?

mathetake: done

PR #706, opened 2025-06-13
feat: adds support for almost-OpenAI-compatible providers

mathetake:
**Commit Message**

This adds support for almost-OpenAI-compatible providers where the only difference with OpenAI is that it has different "prefix". For example, Gemini API uses "/v1beta/openai" prefix. Another example is that Cohere AI [2]  uses "/compatibility/v1" prefix.

[1] https://ai.google.dev/gemini-api/docs/openai
[2] https://docs.cohere.com/v2/docs/compatibility-api

PR #765, opened 2025-06-24
controller: defaults to /v1 for OpenAI when empty version

mathetake:
**Commit Message**

Previously, the controller didn't fallback to /v1 version for OpenAI schema since #706 which starts utilizing schema.version field. The field was previously non nil hence in many cases, it was set to empty, and that caused the backward incompatibility. This fixes it by defaulting to v1 when the field is set but empty string.

Follow up on #706


api/v1alpha1/api.go
```diff
@@ -419,9 +419,9 @@ type VersionedAPISchema struct {
 
 	// Version is the version of the API schema.
 	//
-	// When the name is set to "OpenAI", this equals to the prefix of the OpenAI API endpoints, and
-	// this defaults to "v1" if not set. For example, "chat completions" API endpoint will be
-	// "/v1/chat/completions" if the version is set to "v1".
```
codefromthecrypt, line 424: is this really semantically meaningful? "/v1" is not a part of the api, it is just a convention for the base URL and isn't /v1 or even a different version indicator if azure openai

https://github.com/openai/openai-openapi/blob/manual_spec/openapi.yaml#L15

I would suggest we don't attempt to normalize a part of the base URL and treat the openai api as it is now, which is rooted after the /v1 part.

mathetake: correct, semantically meaningless. This is more about how to support almost-compatible openaai providers like e.g. groq (https://console.groq.com/docs/openai) which has `openai/v1` base and Gemini (https://ai.google.dev/gemini-api/docs/openai) which has `v1beta/openai/` base. 

In that sense, i feel like we should rename this to `base` or similar maybe rather than version.

codefromthecrypt: yeah base url is only thing with meaning and we can align/blame docs and naming convention on openai ;) https://github.com/openai/openai-python/blob/1a85f249ecdf13278d16185a1f9ae908e96e03d2/README.md?plain=1#L759

Issue #648, opened 2025-05-28
Support HTTPRoute -> InferencePool Directly


robscott:
*Description*:
All of the OSS Inference Extension docs focus on a common use case where a HTTPRoute points directly to an InferencePool. This expectation is now being translated into [conformance tests](https://github.com/kubernetes-sigs/gateway-api/tree/main/conformance). To ensure Envoy AI Gateway can pass those conformance tests, it would be useful if it supported this direct attachment model from HTTPRoute -> InferencePool.

yuzisun:
@robscott thanks for opening this issue! As we discussed before HTTPRoute is primarily designed for ingress use case while envoy ai gateway was originally proposed to handle egress use case, so we created AIGatewayRoute resource to handle LLM traffic. Later on when we wanted to support inference extension API and to unify the ingress/egress use cases, we attached the InferencePool to AIGatewayRoute. 

There is no reason we can not attach InferencePool to HTTPRoute,  however currently HTTPRoute is not directly user facing as we do not want to introduce the confusion that for different use case user has to go for different API. 

I am happy to discuss with you with a better option, possibly if we can support egress use case as part of Gateway API.

arkodg:
from a capability perspective, since this project is already using the [Extension Server](https://github.com/envoyproxy/ai-gateway/tree/main/internal/extensionserver) mechanism, one way to additionally support this extension (on top of attaching to AIGatewayRoute), is to treat InferencePool as an ExtensionResource https://gateway.envoyproxy.io/docs/tasks/extensibility/extension-server/ and edit the xDS to support this functionality

arkodg:
hey @robscott closing this one as a dup of https://github.com/envoyproxy/ai-gateway/issues/423 where @mathetake +1s [directly HTTPRoute->Inference pool support](https://github.com/envoyproxy/ai-gateway/issues/423#issuecomment-2957560133)

PR #771, opened 2025-06-25
docs: add epp integration proposal

Xunzhuo:
**Description**

This PR adds the proposal for supporting Integration with Endpoint Picker(GIE)

Related to https://github.com/envoyproxy/ai-gateway/issues/423


missBerg:
@yanavlasov you may appreciate this document 😊

mathetake:
will take a look tomorrow (in the us timezone)

docs/proposals/003-epp-integration-proposal/proposal.md
```diff
@@ -0,0 +1,610 @@
...
+    group: aigateway.envoyproxy.io
+```
+
+To integrate with the GIE, there are two options:
+
+#### Option 1: Add InferencePool as an backendRef on AIGatewayRoute Level
+
+This requires to expand the `AIGatewayRouteRuleBackendRef` with `BackendObjectReference`
+
+##### Current
```
mathetake: nit: can you use ` ```diff ` block to merge "Current" and "Target" to make it visible

docs/proposals/003-epp-integration-proposal/proposal.md
```diff
@@ -0,0 +1,610 @@
...
+    group: gateway.envoyproxy.io
+  backendSecurityPolicyRef:
+    name: envoy-ai-gateway-basic-openai-apikey
+    kind: BackendSecurityPolicy
+    group: aigateway.envoyproxy.io
+```
+
+To integrate with the GIE, there are two options:
+
+#### Option 1: Add InferencePool as an backendRef on AIGatewayRoute Level
```
mathetake, line 199: I am +1 to this since this is the same as the initial proposal.

docs/proposals/003-epp-integration-proposal/proposal.md
```diff
@@ -0,0 +1,610 @@
...
+
+<!-- /toc -->
+
+## Summary
+
+This propopal aims to land integration with other endpoint picker in Envoy AI Gateway, expand EAGW abilities with other EPP implementations, like Gateway API Inference Extension, AIBrix Plugin, semantic router etc.
+
+This is a core functionality in EAGW`s vision, make the routing more intelligent.
+
+![](http://liuxunzhuo.oss-cn-chengdu.aliyuncs.com/2025-06-25-090714.png)
```
yuzisun, line 30: Could we add this image to the envoy ai gateway repo?

docs/proposals/003-epp-integration-proposal/proposal.md
```diff
@@ -0,0 +1,610 @@
...
+
+Then everything the EPP can do is done, envoyproxy is going to work, the logics is the next section.
+
+### How Envoy works with GIE?
+
+EnvoyProxy is the data plane who actually forwards the request, unlike the ways we forward the traffic to the kubernetes service/endpoints, the destination is unknown for the control plane, so it cannot be pre-generated to EnvoyProxy.
+
+Above section tells, the destination is chosen by EPP and the information is located in header and metadata, so the way envoy determines is to read the header or the metadata to pick the target endpoint.
+
+There are two approaches envoy can work in this scenario:
```
yuzisun, line 109: Is there any pros and cons between the two approaches ?

docs/proposals/003-epp-integration-proposal/proposal.md
```diff
@@ -0,0 +1,610 @@
...
+            - type: Exact
+              name: x-ai-eg-model
+              value: vllm-llama3-8b-instruct
+      backendRefs:
+        - name: vllm-llama3-8b-instruct
+        	group: inference.networking.x-k8s.io
+          kind: InferencePool
+```
+
+#### Option 2: Add InferencePool as an backendRef on AIServiceBackend Level
```
yuzisun, line 253: I like this option as `InferencePool` should be at the AIServiceBackend level.

mathetake (commented):
Generally looks good (I like your break down of how it works) with some comments:

* The reason why I prefer the AIGatewyRouteBackendRef level is that InferencePool will never need BackendSecurityPolicy as well as the schema configuration. We should be able to assume they are using OpenAI format as GAIE's implementation itself is based on that assumption.
* We have to document that cross provider fallback won't work at the InferencePool due to the way Envoy cluster is configured on both EG and this extension server level insertion. This is I believe cannot be resolved.
* We also have to document that users are not allowed to define multiple InferencePool and/or with normal AIServiceBackends in a single route rule. This can be enforced at k8s CEL validation layer. The reason is similar as to why the fallback cannot work above. The multiple backends in a route rule thingy assumes that an Envoy cluster contains multiple backends as LocalityLbEndpoints for each Backend whose metadata can be used to distinguish which  AIServiceBackend it belongs to, and with that, the extproc can determine translator etc. Also, EG level fallback/priority works on that level.



One last question though is that wha do we do about https://github.com/envoyproxy/ai-gateway/issues/648. It's just for conformance test but at the end of the day, almost the entire logic lives within the extproc, so i don't think that matters in reality. Having said that though, I would like to see our impl pass the conformance test as-is as well. wdyt?

yuzisun:
> * The reason why I prefer the AIGatewyRouteBackendRef level is that InferencePool will never need BackendSecurityPolicy as well as the schema configuration. We should be able to assume they are using OpenAI format as GAIE's implementation itself is based on that assumption.

This might not be true, `InferencePool` is designed for self hosted model endpoints which itself can be protected with authentication and authorization, so you still need a `BackendSecurityPolicy` in that case.

yuzisun:
> * We have to document that cross provider fallback won't work at the InferencePool due to the way Envoy cluster is configured on both EG and this extension server level insertion. This is I believe cannot be resolved unless EG allows us to use aggregate cluster. (Or maybe it's possible to use it at our extension server level?)

This is a good point, the use case does exist though as we want to fallback to an InferencePool if the cloud model endpoint is unhealthy. For the time being we can add the validation.

docs/proposals/003-epp-integration-proposal/proposal.md
```diff
@@ -0,0 +1,610 @@
...
+## Summary
+
+This propopal aims to land integration with other endpoint picker in Envoy AI Gateway, expand EAGW abilities with other EPP implementations, like Gateway API Inference Extension, AIBrix Plugin, semantic router etc.
+
+This is a core functionality in EAGW`s vision, make the routing more intelligent.
+
+![](http://liuxunzhuo.oss-cn-chengdu.aliyuncs.com/2025-06-25-090714.png)
+
+## Goals
++ Integrate with EPP to expand the Envoy AI Gateway abilities
```
yuzisun, line 33: What are the EPP algorithms we plan to support initially ? would like to see more details on the EPP implementation

mathetake: i believe the EEP is selected via InferencePool.spec.extensionRef (https://gateway-api-inference-extension.sigs.k8s.io/reference/spec/#extensionreference) and that's agnostic of the implementation where users can freely specify their own EEP's deployment.

mathetake: for example, they can deploy the GAIE reference impl or their own custom extproc cc @kerthcet

mathetake:
yeah, so if we somehow will be able to resolve the fallback issue (== cluster definition and priority issue) with aggregated cluster, maybe it's better to have it on AIServiceBackend level. My concern is that, as I commented above, that might end up making it difficult for the direct HTTPRoute support, though i am not sure whether we want to do that in the first place. The direct support portion can be deferred later i guess as i don't think that really matters in reality.

docs/proposals/003-epp-integration-proposal/proposal.md
```diff
@@ -0,0 +1,610 @@
...
+
++ generate ext-proc cluster based on the InferencePool extensionRef
++ generate ext-proc http filter based on the InferencePool extensionRef
+
+##### route level config
+
++ patch the ext-proc filter into the route configuration based on which route the InferencePool is linked with
++ add the cluster with loadbalancing policy or ORIGINAL_DST to understand the header and route  `x-gateway-destination-endpoint`
+
+#### Resource Generation
```
Xunzhuo, line 340: Still need to decide for two points, the resource generation and works with EG.

docs/proposals/003-epp-integration-proposal/proposal.md
Xunzhuo: Yes, and for how llm-d doing is GW + GIE as a core and the scheduler is a higher level for that. So basically if we support this proposal, we can be an option for llm-d scheduler GW implementation.

yuzisun: I understand user can plugin their own epp implementation. I guess my question is that what envoy ai gateway offers as out of the box AI aware routing like kv-aware, prefix-aware and disaggregated prefilling routing. LLM-d is implementation details of envoy ai gateway.

Xunzhuo: @yuzisun the first goal for envoy ai gw is to take GIE as a default EPP implementation, it provides the routing algorithm you raised above.

docs/proposals/003-epp-integration-proposal/proposal.md
mathetake: i don't think we have currently any bandwidth to maintain this dynamic one, so let's start small with the static configuration by end user.

docs/proposals/003-epp-integration-proposal/proposal.md
```diff
@@ -0,0 +1,610 @@
...
+
+This section is about how to manage the GIE kubernetes resource, in short, there are two approaches, static or dynamic.
+
++ static: end-user manage the GIE deployment, service etc resource.
++ dynamic: end-user do not care about the life cycle of GIE resources, Envoy AI Gateway take care of that.
+
+#### Work with Envoy Gateway
+
+There are two work-in-process PRs in upstream:
```
Xunzhuo, line 355: Thoughts on the two approaches? I prefer the first one, we created the Backend + EEP for LLMInferencePool Backend Type.

Xunzhuo: cc @mathetake

mathetake: I would prefer the extension server approach. We can later revisit how to utilize the EG resources but after i saw the discussion here https://github.com/envoyproxy/gateway/issues/6234, i am not sure if it's ready for us to tie our implementation with EG level API. Not only that, personally i feel it's easier to understand what's going on if we consolidate all the logic in the extension server.

Xunzhuo: Yep, if we take the extension server approach, maybe we can reuse the EEP for generating the ext proc server filter and cluster and we modify the route and add original dst cluster for it?

mathetake: plus i think with this approach, we can pass the conformance test without modifying it (https://github.com/envoyproxy/ai-gateway/issues/648)?

Xunzhuo: @mathetake yes, if we create a httproute by the aigwroute and sycned the InferencePool configuration with it, for the conformance view, it is matched with its requirements.

Xunzhuo:
Hey forks, a left decision to make is the place we add the InferencePool.
I am +1 for adding that in AIGWRoute level, pretty straightforward, and also we can starting the implementation when https://github.com/envoyproxy/gateway/pull/6342 is landing.

Envoy Gateway will support custom backendRef via extension server.

The workflow will be like:
1. Enable Envoy Gateway extension server with InferencePool backendResources
2. Install GIE related resources first (CRDs , deployment)
3. Create InferencePool refers to the ext-proc service
4. Use InferencePool as the AIGatewayRoute Backend (limitation is one InferencePool per Rule)
5. Envoy AI GW sync it to the managed HTTPRoute with BackendRef of InferencePool
6. Envoy AI GW also creates an EnvoyExtensionPolicy with ext-proc info of InferencePool, and targetRef is the HTTPRoute.
7. Envoy GW carries the InferencePool resource with the Cluster by calling `PostClusterModify`
8. Envoy AI GW modifies the Cluster with Original Dst Cluster and enable the header match with `x-gateway-destination-endpoint`
9. Client calls to EnvoyProxy, EnvoyProxy talks to the GIE, GIE adds the header, EnvoyProxy forwards with the header.

mathetake:
yep, i am +1 for Route Rule level too which matches the underlying HTTPRoute extension ref style.

As for the backend security attachment pointed out by @yuzisun, i think with this https://github.com/envoyproxy/ai-gateway/issues/549 refactoring, we can allow BSP to be attached to the pool. That way we can also allow InferencePool to use the BSP. i guess this seems to be the way.

mathetake:
@Xunzhuo could you refactor the doc based on the discussion/agreement we had, and then i can stamp

docs/proposals/003-epp-integration-proposal/proposal.md
kerthcet: Second this, I think it's unaccessible now.

docs/proposals/003-epp-integration-proposal/proposal.md
```diff
@@ -0,0 +1,610 @@
...
+    -   [Configuration Generation](#configuration-generation)
+    -   [Work with Envoy Gateway](#aigatewayroute)
+
+<!-- /toc -->
+
+## Summary
+
+This propopal aims to land integration with other endpoint picker in Envoy AI Gateway, expand EAGW abilities with other EPP implementations, like Gateway API Inference Extension, AIBrix Plugin, semantic router etc.
+
+This is a core functionality in EAGW`s vision, make the routing more intelligent.
```
kerthcet, line 28: I guess the main goal here is more flexible.

docs/proposals/003-epp-integration-proposal/proposal.md
kerthcet: What we want here is just a flexible way to embed our own logic but let the ai gateway help us handle the traffic forwarding.

I think even in the long term, maybe ai gw should not touch this part as well, llm-d, aibrix, our own platform llmaz all follow the same way just because the implementation is complex and domain specific, for example, it needs to collect metrics for decision make. It would be surprised if aigw wants to do this.

I'm not saying this pattern is always right, because all the things goes too fast. And if there're any very general algorithms there, also happy to see here.

docs/proposals/003-epp-integration-proposal/proposal.md
```diff
@@ -0,0 +1,610 @@
...
+
+![](http://liuxunzhuo.oss-cn-chengdu.aliyuncs.com/2025-06-25-090714.png)
+
+## Goals
++ Integrate with EPP to expand the Envoy AI Gateway abilities
++ Integrate with the existing CRD and features well
```
kerthcet, line 34: Maybe we should add Non-Goals as we discussed above:

No EPP implementations would be supported as the out-of-the-box routing algorithms, would consider this in the future if needed.

docs/proposals/003-epp-integration-proposal/proposal.md
```diff
@@ -0,0 +1,610 @@
...
+  targetPortNumber: 8000
+  selector:
+    app: vllm-llama3-8b-instruct
+  extensionRef:
+    name: vllm-llama3-8b-instruct-epp
+    port: 9002
+    failureMode: FailClose
+```
+
+The control plane will generate the corresponding ext proc config (filter + cluster) to envoy, Take the inferencePool above as an example, the destination would be `vllm-llama3-8b-instruct-epp:9002` in the same namespace with the InferencePool.
```
kerthcet, line 65: What about non-GAIE, no where to store the information.

docs/proposals/003-epp-integration-proposal/proposal.md
kerthcet: +1

docs/proposals/003-epp-integration-proposal/proposal.md
```diff
@@ -0,0 +1,610 @@
...
+}
+```
+
+##### Target
+
+```
+// AIGatewayRouteRuleBackendRef is a reference to a backend with a weight.
+type AIGatewayRouteRuleBackendRef struct {
+
+	gwapiv1.BackendObjectReference
```
kerthcet: One confusing thing is the default value of backendObjectReference.kind is Service, but we're AIServiceBackend, I think they're not the same. And seems the fields are nested too deep ...

docs/proposals/003-epp-integration-proposal/proposal.md
kerthcet, line 253: Why I feel this one is more reasonable?

docs/proposals/003-epp-integration-proposal/proposal.md
```diff
@@ -0,0 +1,610 @@
...
+    kind: InferencePool
+```
+
+#### Configuration Generation
+
+no matter which one we decide for the above two approaches, we need to figure out how/what configuration we need to generate.
+
+based on the background, we need to generate such configurations:
+
+##### ext-proc config
```
kerthcet, line 330: Maybe more on the integration with non-GIE projects, I think it's reasonable since we want to work out a general solution.

docs/proposals/003-epp-integration-proposal/proposal.md
kerthcet: Sorry, I didn't get the point here, I think we can not create the whole GIE resources in ai gw, then how can we say manage the lifecycle? 

+1 to start with the base line.

kerthcet (commented):
Thanks @Xunzhuo for all these efforts, left some comments, generally would like to see more integrations with non-GIE implementations. I think also one of the goals of this proposal? Maybe could elaborate more once we made our technical decisions.

docs/proposals/003-epp-integration-proposal/proposal.md
mathetake: how about i think we can simply expand this implementation to support a generic k8s service as a backendref target, and that service can be a custom eep beyond inference pool.

docs/proposals/003-epp-integration-proposal/proposal.md
Xunzhuo: @kerthcet Actually this is not binded to the GIE, just focus on the InferencePool API, with this, we can support any endpoint picker.

docs/proposals/003-epp-integration-proposal/proposal.md
Xunzhuo: Yes currently, we need user to install the GIE deployment themselves. This is considering that we want to just focus on the InferencePool API, we just manages the ext-proc config and route level config, help the EPP connects to EnvoyProxy in the right place. This can allow end-user to implement custom EPP and integrate with Envoy AI Gateway.

If we manage to GIE resource, maybe we are binded to the EPP implementation

docs/proposals/003-epp-integration-proposal/proposal.md
Xunzhuo: @kerthcet as I commented below, the Envoy AI GW only cares about the InferencePool API, so actually, after this landing, we can support any endpoint picker.

Xunzhuo:
For the non-GIE EPP implementation, there are some requirements for it.
1. it should be an ext-proc server
2. it should be picking the endpoint by adding `x-gateway-destination-endpoint` header in the request header
3. it should be awared of the InferencePool API (it tells Envoy AI Gateway how to connect to the ext-proc server, and which routes should be attached to the ext-proc route level filter, and modifies the cluster with original dst cluster with header `x-gateway-destination-endpoint`)

After meeting the requirements above, non-GIE EPP implementation just needs to deploy and create InferencePool to link it to the AIGatewayRoute, then when request comes to the envoy, it will talk to the non-GIE EPP ext-proc server, ask for the x-gateway-destination-endpoint in the header, then forward to the picked endpoint.

cc @kerthcet

docs/proposals/003-epp-integration-proposal/proposal.md
yuzisun: I do not want to exclude this possibility, there are use cases for picking a cloud model endpoint like QoS which is not the focus of llm-d, aibrix.

docs/proposals/003-epp-integration-proposal/proposal.md
yuzisun: overrideHostSources supports a list which is more flexible. The request is forwarded to the first address and subsequent addresses are used for request retries or hedging.

docs/proposals/003-epp-integration-proposal/proposal.md
yuzisun: btw KServe is going with the llm-d EPP implementation and KServe's `LLMInferenceService` API handles the EPP deployment, so I do not think this needs to be done in Envoy AI Gateway.  see https://github.com/kserve/kserve/blob/master/pkg/apis/serving/v1alpha1/llm_inference_service_types.go#L171

docs/proposals/003-epp-integration-proposal/proposal.md
yuzisun: agree with the extension server approach for now.

Xunzhuo:
Some points left to discuss:
1. LB policy: use host override policy or use original dst?
2. Fallback support: since we support cluster level hook in eg, can we support fallback between the InferencePool and AIBackendService? (one InferencePool and n*AIServiceBackend at one rule, we just need to patch the cluster carried with the inferencePool resource)
3. Token ratelimit: can we support token based ratelimit to inferencePool?

mathetake:
> 1. LB policy: use host override policy or use original dst?

you can start with original_dst as it's currently [used by the reference implementation repo](https://github.com/kubernetes-sigs/gateway-api-inference-extension/blob/8d01161ec48d6b49cd371f179551b35da46e6fd6/test/testdata/envoy.yaml#L152).

> 2. Fallback support: 

[As I said above](https://github.com/envoyproxy/ai-gateway/pull/771#pullrequestreview-2967931573), i don't think it will work without a huge refactoring of EG, not here. EG's fallback works **WITHIN** one cluster where multiple backends in a single route rule exists as a separate localityLBEndpoints. On the other hand, this case, InfrencePool cannot coexist with other normal backends in one single cluster as the cluster will be configured to use original_dst (or override_host) stuff. Maybe there's a solution but atm i have no concrete idea. We can ignore this and document this limitation at the initial impl.

> 3. Token ratelimit: can we support token based ratelimit to inferencePool?

As long as the upstream filter extproc is configured properly, i should work. In other words, don't forget to insert upstream extproc filter just like we do in the current extension server code;)

yuzisun:
> > 1. LB policy: use host override policy or use original dst?
> 
> you can start with original_dst as it's currently [used by the reference implementation

My only concern is that original_dst does not seem like support fallback, what if the epp service is not available and the request will fail?

mathetake:
yeah we can use the override_host then

Xunzhuo:
For the host override lbpolicy, we need to create a service with the same selector of the inference pool, and the endpoint selection should be in the service endpoints. Istio creates a service for the inference pool to meet host override policy requirements.

mathetake:
up to you but I would start with the simpler impl

Xunzhuo:
Sure I think this cannot be too hard to implement the host override(generate a service and pin it to one more backendRef with InferencePool), but we can do it later. Will document it in design doc.

Xunzhuo:
I will start to summarize to design doc today. Since the extension server is going to merge now

mathetake:
yep, this is exciting!

mathetake (approved):
Well done!

docs/proposals/003-epp-integration-proposal/proposal.md
kerthcet: Yeah, it's good to have general ones. But can you elaborate more about QoS endpoints, what do it means?

kerthcet:
I know GIE is definitely the in-tree support for our gateway, but I'm not quite understand why we need to have inferencePool for non-GIE implementations as well, to me, I hope not, also would like to hear advices from @yuzisun wearing the hat of kserve.

> how about i think we can simply expand this implementation to support a generic k8s service as a backendref target, and    that service can be a custom eep beyond inference pool.

I think @mathetake advice is what we hope to have, relying on kubernetes native objects only.

kerthcet:
Also only me can not see the arch diagram?

Xunzhuo:
The reason we support InferencePool is just using it to determine which routes should use the Original DST or Host Override Policy, and tell envoy AI GW how to connect to the EPP ext-proc.

So I don't think we support it for binding integrations to GIE.

Xunzhuo:
> Also only me can not see the arch diagram?

Removed that for now, it is a graph we use in KubeCon

PR #776, opened 2025-06-26
feat: add generated apis/v1alpha1 clientset

isyangban:
**Description**

- Generate typed clientset for apis/v1alpha1
- Add make clientgen target.
- Updates CI formatting/linters to skip the generated directories.

- Not sure if you are happy with maintaining the code in-tree
- Should I generate listers and informers as well?


mathetake:
to be honest, i am not sure we want this until v1.0.

mathetake:
Due to the same reason as #782, I would like to avoid the maintenance burden until v1.0 as this will add not only the additional step but also noise in the API change related API. I personally do not envision anyone using EAIG programmatically at the moment, so i am going to close this one. Thanks for giving it a shot anyways!

PR #777, opened 2025-06-26
extproc: pass model name override as part of dynamic metadata

aabchoo:
The goal here is to give user access to the model override name used to query the inference provider.

### DIFF
--- a/internal/extproc/chatcompletion_processor.go
+++ b/internal/extproc/chatcompletion_processor.go
@@ -29,6 +29,10 @@ import (
 	"github.com/envoyproxy/ai-gateway/internal/llmcostcel"
 )
 
+const (
+	ModelNameOverride = "model_name_override"
+)
+
 // ChatCompletionProcessorFactory returns a factory method to instantiate the chat completion processor.
 func ChatCompletionProcessorFactory(ccm x.ChatCompletionMetrics) ProcessorFactory {
 	return func(config *processorConfig, requestHeaders map[string]string, logger *slog.Logger, isUpstreamFilter bool) (Processor, error) {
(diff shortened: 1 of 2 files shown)


internal/extproc/chatcompletion_processor.go
```diff
@@ -29,6 +29,10 @@ import (
 	"github.com/envoyproxy/ai-gateway/internal/llmcostcel"
 )
 
+const (
+	ModelNameOverride = "model_name_override"
```
mathetake: i see this looks unnecessary package scope constant.  could you either just inline it inside the function or prefix with a meaningful context + unexport it

internal/extproc/chatcompletion_processor.go
```diff
@@ -397,6 +401,11 @@ func (c *chatCompletionProcessorUpstreamFilter) maybeBuildDynamicMetadata() (*st
 		c.logger.Info("Setting request cost metadata", "type", rc.Type, "cost", cost, "metadataKey", rc.MetadataKey)
 		metadata[rc.MetadataKey] = &structpb.Value{Kind: &structpb.Value_NumberValue{NumberValue: float64(cost)}}
 	}
+
+	if c.modelNameOverride != "" {
+		metadata[ModelNameOverride] = &structpb.Value{Kind: &structpb.Value_StringValue{StringValue: c.modelNameOverride}}
```
mathetake: model override is per (backend, route rule) information, so this changes on retry. do we want to add additional information like the backend name here or as another metadata?

aabchoo: Hmm that's a good point - we currently only need the override but I can see scenarios where the route and backend may come in handy. Will add it! Thanks for the suggestion

aabchoo: What do you think about only setting the backend + override because route rule can be determined based on the AIGatewayRouteRuleMatch headers passed?

mathetake: setting backend name + override sounds good

mathetake:
On the naming consistency with #781, I will do some refactoring after this and #781 lands, so don't worry about it too much. Mainly I would like to avoid blocking one PR over another.

mathetake:
could you merge the main branch? The metadata construction routine has changed a bit since #696

internal/extproc/chatcompletion_processor.go
aabchoo: updated it for both embedded and chat complete

Issue #723, opened 2025-06-18
Add performance metrics (TTFT and ITL) in dynamic metadata or response header


wengyao04:
*Description*:
>Describe the desired behavior, what scenario it enables and how it
would be used.

TTFT and ITL are important performance metrics and AI-Gateway publish TTFT and ITL as prometheus metrics. But users may needs these number for internal analysis or reporting.

Could we send TTFT and ITL in the response header or in the dynamic metadata ?

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

mathetake:
could you clarify exactly what TTFT and ITL means and stand for for someone like me ?

> But users may needs these number for internal analysis or reporting.

am i understanding correctly that the user here means a downstream client (the user of openai SDK?)

mathetake:
if the user here means the platform team  (who deploys/maintains EAIG), then i think emitting these metrics (including the all the existing ones) as metadata makes sense. The metadata data can be logged as part of access logging which can be used to analyze metrics per individual request

wengyao04:
> could you clarify exactly what TTFT and ITL means and stand for for someone like me ?
> 
> > But users may needs these number for internal analysis or reporting.
> 
> am i understanding correctly that the user here means a downstream client (the user of openai SDK?)

TTFT is time-to-first-token and ITL is inter-token-latency (a nice definition https://docs.nvidia.com/nim/benchmarking/llm/latest/metrics.html). In ai-gateway, we collect both TTFT and ITL https://github.com/envoyproxy/ai-gateway/blob/main/internal/metrics/chat_completion_metrics.go#L127-L134.

The user here means downstream client and they may use these metrics for some analysis

mathetake:
oh ok, sending to the downstream client means that the metadata is not enough but we need to add them as response headers yeah. 

I think it makes sense and we can have the implementation

mathetake:
assigning you @wengyao04 👍

wengyao04:
The TTFT and ITL metrics are only available for stream mode and we calculate it when receive stream body. However, envoy ext_proc does not support header mutation after data processing for streaming https://github.com/envoyproxy/envoy/issues/25410

Could we add these in dynamic meta data that users can extract it in their ext_proc ?

mathetake:
ah i see the problem... it's possible to expose them via extproc but not sure if we/you really want to allow downstream clients to access extproc (== Envoy proxy container).

mathetake:
[...]

PR #781, opened 2025-06-26
extproc: add TTFT and ITL in dynamic metadata

yduwcui:
**Description**

This PR is to add TTFT(time to first token) and ITL(inter token latency) in dynamic metadata, so that downstream clients can access these data.

issues #723


internal/metrics/chat_completion_metrics.go
```diff
@@ -126,11 +128,20 @@ func (c *chatCompletion) RecordTokenLatency(ctx context.Context, tokens uint32,
...
 		// Calculate time between tokens.
-		itl := time.Since(c.lastTokenTime).Seconds() / float64(tokens)
-		c.metrics.outputTokenLatency.Record(ctx, itl, metric.WithAttributes(attrs...))
+		c.interTokenLatency = (time.Since(c.lastTokenTime).Seconds() * 1000) / float64(tokens)
+		c.metrics.outputTokenLatency.Record(ctx, c.interTokenLatency, metric.WithAttributes(attrs...))
 	}
 	c.lastTokenTime = time.Now()
 }
+
+func (c *chatCompletion) GetTimeToFirstToken() float64 {
```
mathetake: could you add the same style comment for interface implementation elsewhere:
```suggestion
// GetTimeToFirstToken implements [ChatCompletion.GetTimeToFirstToken].
func (c *chatCompletion) GetTimeToFirstToken() float64 {
```

internal/metrics/chat_completion_metrics.go
```diff
@@ -126,11 +128,20 @@ func (c *chatCompletion) RecordTokenLatency(ctx context.Context, tokens uint32,
 
 	if !c.firstTokenSent {
 		c.firstTokenSent = true
-		c.metrics.firstTokenLatency.Record(ctx, time.Since(c.requestStart).Seconds(), metric.WithAttributes(attrs...))
+		c.timeToFirstToken = time.Since(c.requestStart).Seconds() * 1000
```
mathetake: comment on why *1000 and can you add unit test for this change? any functional change essentially needs to be tested with unit tests.

internal/metrics/chat_completion_metrics.go
```diff
@@ -126,11 +128,20 @@ func (c *chatCompletion) RecordTokenLatency(ctx context.Context, tokens uint32,
...
+		c.metrics.firstTokenLatency.Record(ctx, c.timeToFirstToken, metric.WithAttributes(attrs...))
 	} else if tokens > 0 {
 		// Calculate time between tokens.
-		itl := time.Since(c.lastTokenTime).Seconds() / float64(tokens)
-		c.metrics.outputTokenLatency.Record(ctx, itl, metric.WithAttributes(attrs...))
+		c.interTokenLatency = (time.Since(c.lastTokenTime).Seconds() * 1000) / float64(tokens)
```
mathetake: same comment

internal/metrics/chat_completion_metrics.go
mathetake: [ChatCompletion.GetTimeToFirstToken] maybe needs to be updated with [x.ChatCompletionMetrics.GetTimeToFirstToken] though ... the comment hasn't been updated after we refactored

yduwcui: sure, will update this!

internal/metrics/chat_completion_metrics.go
yduwcui: it's because the `timeToFirstToken` and `interTokenLatency` are very small numbers in seconds, so want to covert them into milliseconds, and stay in float64 type

mathetake: i meant could you put these rationale in the code comment? It's not obvious to new people to know what's going on without the context. This is for future contributors..

yduwcui: sure will do that, thanks for the feedback!

internal/extproc/chatcompletion_processor.go
```diff
@@ -397,6 +397,12 @@ func (c *chatCompletionProcessorUpstreamFilter) maybeBuildDynamicMetadata() (*st
 		c.logger.Info("Setting request cost metadata", "type", rc.Type, "cost", cost, "metadataKey", rc.MetadataKey)
 		metadata[rc.MetadataKey] = &structpb.Value{Kind: &structpb.Value_NumberValue{NumberValue: float64(cost)}}
 	}
+	if c.stream {
+		timeToFirstToken := c.metrics.GetTimeToFirstToken()
+		metadata["tokenlatency-ttft"] = &structpb.Value{Kind: &structpb.Value_NumberValue{NumberValue: float64(timeToFirstToken)}}
```
yuzisun: let's make sure we have the same naming convention for the metadata key, already seeing discrepancies as the other PR from @aabchoo 
https://github.com/envoyproxy/ai-gateway/pull/777

internal/metrics/chat_completion_metrics.go
```diff
@@ -126,11 +128,20 @@ func (c *chatCompletion) RecordTokenLatency(ctx context.Context, tokens uint32,
 
 	if !c.firstTokenSent {
 		c.firstTokenSent = true
-		c.metrics.firstTokenLatency.Record(ctx, time.Since(c.requestStart).Seconds(), metric.WithAttributes(attrs...))
+		c.timeToFirstToken = time.Since(c.requestStart).Seconds() * 1000
+		c.metrics.firstTokenLatency.Record(ctx, c.timeToFirstToken, metric.WithAttributes(attrs...))
```
nacx, line 73: Changing the units for this will be a breaking change for anyone already monitoring this value.

It was initially set to seconds according to [the semantic conventions](https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-metrics/#metric-gen_aiservertime_to_first_token) for this metric. If the value is changed, probably the histogram buckets should be changed accordingly as well, which I'm not sure I would recommend.

What about keeping the metrics unchanged, and if you want the metadata value to be in `ms`, use the `ms` jsut there? The method name could be changed to `GetTimeToFirstTokenMs()` to be explicit about it.

internal/extproc/chatcompletion_processor.go
yduwcui: will change it to camel case

internal/metrics/chat_completion_metrics.go
yduwcui: thanks for the suggestion! will keep the metrics unchanged and return milliseconds in `GetTimeToFirstTokenMs()`

internal/extproc/chatcompletion_processor.go
aabchoo: The metadata key in the tests uses snake case so I think we should keep consistent with that -- or change it

yduwcui: will follow the test cases to use snake case

mathetake:
On the naming consistency with #777, I will do some refactoring after this and #777 lands, so don't worry about it too much. Mainly I would like to avoid blocking one PR over another.

internal/extproc/translator/translator.go
```diff
@@ -120,4 +120,8 @@ type LLMTokenUsage struct {
 	OutputTokens uint32
 	// TotalTokens is the total number of tokens consumed.
 	TotalTokens uint32
+	// TimeToFirstTokenMs is the time to first token in milliseconds.
+	TimeToFirstTokenMs float64
+	// InterTokenLatencyMs is the inter token latency in milliseconds.
+	InterTokenLatencyMs float64
```
mathetake: hmm, these do not belong here as this is for usage numbers not stats.

yduwcui: added `TimeToFirstTokenMs` and `InterTokenLatencyMs` to `LLMTokenUsage` because in [PR755 ](https://github.com/envoyproxy/ai-gateway/pull/755), `buildDynamicMetadata` is no longer a member function of type `chatCompletionProcessorUpstreamFilter`, and inside `buildDynamicMetadata` function we need to access TTFT and ITL, so added them to `LLMTokenUsage` struct

yduwcui: added `TimeToFirstTokenMs` and `InterTokenLatencyMs` to `LLMTokenUsage` because in [PR755](https://github.com/envoyproxy/ai-gateway/pull/755), `buildDynamicMetadata` is no longer a member function of type `chatCompletionProcessorUpstreamFilter`, and inside `buildDynamicMetadata` function we need to access TTFT and ITL, so added them to `LLMTokenUsage` struct

mathetake: yes no longer a member of the function. As you see this struct is LLMCosts, not stats or whatsoever and that's the reason why it is inside translator package in the first place (created through understanding the body). These two number has completely separate semantics compared to the existing ones. So this is definitely not right place to put these stats. 

Instead you can add these metadata outside buildDynamicMetadata in chat completion processor's method as this is only for chat completions.

yduwcui: I can rename `LLMTokenUsage` to `LLMTokenStats`, let me know your thoughts

mathetake: > I can rename LLMTokenUsage to LLMTokenStats, let me know your thoughts

please do not do that. see the comment above `// LLMTokenUsage represents the token usage reported usually by the backend API in the response body.` as well as my previous comment. Their semantics are completely different.

yduwcui: will remove from `LLMTokenUsage` and add a function in chatcompletion_processor to add these two values

mathetake (commented):
almost there!

internal/extproc/chatcompletion_processor.go
```diff
@@ -361,6 +366,19 @@ func (c *chatCompletionProcessorUpstreamFilter) SetBackend(ctx context.Context,
 	return
 }
 
+func (c *chatCompletionProcessorUpstreamFilter) mergeWithTokenLatencyMetadata(metadata *structpb.Struct) {
```
mathetake, line 371: could you write a unit test for this?

yduwcui: sure will do

internal/extproc/chatcompletion_processor.go
```diff
@@ -329,10 +329,15 @@ func (c *chatCompletionProcessorUpstreamFilter) ProcessResponseBody(ctx context.
 	}
 
 	if body.EndOfStream && len(c.config.requestCosts) > 0 {
-		resp.DynamicMetadata, err = buildDynamicMetadata(c.config, &c.costs, c.requestHeaders, c.modelNameOverride, c.backendName)
+		metadata, err := buildDynamicMetadata(c.config, &c.costs, c.requestHeaders, c.modelNameOverride, c.backendName)
 		if err != nil {
 			return nil, fmt.Errorf("failed to build dynamic metadata: %w", err)
 		}
+		if c.stream {
+			// Adding token latency information to metadata
```
mathetake: super nit but I would like any code comment to end with period.

```suggestion
			// Adding token latency information to metadata.
```

internal/metrics/chat_completion_metrics_test.go
```diff
@@ -157,6 +157,33 @@ func TestRecordRequestCompletion(t *testing.T) {
 	assert.Greater(t, sum, 0.0)
 }
 
+func TestGetTimeToFirstTokenMsAndGetInterTokenLatencyMs(t *testing.T) {
```
mathetake, line 160: does this test really need to be this complex? it feels like a duplicate of the test above. Something like the following seems to be enough:

```
func TestGetTimeToFirstTokenMsAndGetInterTokenLatencyMs {
    c := chatCompletion{timeToFirstToken: 1, interTokenLatency: 2}
    require.Equal(t, c. GetTimeToFirstTokenMs(), ....
}
```

yuzisun (approved):

internal/extproc/chatcompletion_processor.go
yuzisun: Can we add a lint rule for this ?

PR #782, opened 2025-06-26
feat: ai-gateway-provisioner helm chart

ion-elgreco:
**Description**
AI gateway provisioner helm chart to make it easier to deploy an AI gateway service. Let me know your thoughts on this :)

### DIFF
--- a/manifests/charts/ai-gateway-provisioner/Chart.yaml
+++ b/manifests/charts/ai-gateway-provisioner/Chart.yaml
@@ -1,0 +1,37 @@
+# Copyright Envoy AI Gateway Authors
+# SPDX-License-Identifier: Apache-2.0
+# The full text of the Apache license is available in the LICENSE file at
+# the root of the repo.
+
+apiVersion: v2
+name: ai-gateway-provisioner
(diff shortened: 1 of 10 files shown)


mathetake:
hmm, API will likely change including breaking changes until v1.0, so I definitely would like to avoid having this as it will add additional maintenance when making changes to API.

ion-elgreco:
@mathetake alright, let's keep the PR just as an example then

mathetake:
open PRs are mainly for waiting for review stage, so I would rather not keep this open as v1.0 won't happen at least for one or two years from now. Whether being closed or kept open doesn't matter for demonstration purpose as you can still see the change from GitHub UI. Thanks for your work here anyways, this might be helpful for some people.

ion-elgreco:
@mathetake I could publish it in a repo maintained by me, and then we could mention it as a community maintained chart to help provision AI-gw

ion-elgreco:
@mathetake I've published it under here: https://github.com/ion-elgreco/envoy-aigw-provisioner

Once you guys feel AIGW is in a good state, we can bring it back this repo :)

PR #789, opened 2025-06-27
chore: adds internal/testing for coverage ignore

mathetake:
**Description**

This will add `internal/testing` directly to the ignore target of test coverage. That package's code path varies depending on the credentials so it is flaky, not only it is a test only package.


mathetake:
thank you @sukumargaonkar for pointing this out!

Issue #773, opened 2025-06-25
Add CRDs to main chart


ion-elgreco:
*Description*:
Simplify deployment step, by adding CRDs in main chart, then you can still seperately deploy it if you disable it on the main chart.

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

mathetake:
This seems reasonable. Is this what EG does like including CRD in the main chart as well? @Xunzhuo

Xunzhuo:
Yes, we install it along with gateway-helm chart, but has a separate crd chart for its upgrade

cvrajeesh:
Created a PR https://github.com/envoyproxy/ai-gateway/pull/792 which adds the crds to the main chart

PR #792, opened 2025-06-28
chore: add crds to the ai-gateway-helm chart

cvrajeesh:
**Description**
Include crds in the main chart (`ai-gateway-helm`)

Fixes: #773


yuzisun:
@cvrajeesh this won’t work as helm doe not update crd during upgrade, see https://helm.sh/docs/chart_best_practices/custom_resource_definitions. So we decided to go with option 2 to maintain as a separate chart.

cvrajeesh:
Yeah, it won't upgrade crds but including will give a better DX for the first time installers. 

Envoy Gateway chart includes the crds in the main chart in addition to the crd helm chart

mathetake:
yep, i think it makes sense for first time installers IMO

mathetake:
@cvrajeesh can you edit the getting started doc (https://github.com/envoyproxy/ai-gateway/blob/main/site/docs/getting-started/installation.md#installing-envoy-ai-gateway) to mention 
 * For first time installation the crd helm chart is not necessary and 
 * If there's already the installation, users need to apply the crd helm chart in addition to the main char

cvrajeesh:
@mathetake updated the Helm installation instructions, also included instructions to skip the CRD installation

site/docs/getting-started/installation.md
```diff
@@ -11,20 +11,36 @@ This guide will walk you through installing Envoy AI Gateway and its required co
...
 
-kubectl wait --timeout=2m -n envoy-ai-gateway-system deployment/ai-gateway-controller --for=condition=Available
+After the CRDs are installed, you can install the AI Gateway Helm chart without re-installing the CRDs by using the `--skip-crds` flag.
+
+```shell
+helm upgrade -i aieg oci://docker.io/envoyproxy/ai-gateway-helm \
+    --version v0.0.0-latest \
+    --namespace envoy-ai-gateway-system \
+    --create-namespace \
+    --skip-crds
```
mathetake, line 43: i am no expert but is this flag necessary? I was thinking if the CRD is already there, then CRD in the chart will be ignored?

cvrajeesh: If the CRDs are already present, then there will be a warning https://helm.sh/docs/chart_best_practices/custom_resource_definitions/#:~:text=If%20the%20CRD%20already%20exists%2C%20it%20will%20be%20skipped%20with%20a%20warning.%20If%20you%20wish%20to%20skip%20the%20CRD%20installation%20step%2C%20you%20can%20pass%20the%20%2D%2Dskip%2Dcrds%20flag.

mathetake: oh ok that makes sense

mathetake:
could you fix the CI?

cvrajeesh:
@mathetake  done

mathetake (approved):

Issue #612, opened 2025-05-13
default AIGatewayRoute backend


codefromthecrypt:
*Description*:

Right now, it appears that you need to set a custom header in order to route requests to a single backend (e.g. ollama).

e.g. `curl ... -H 'aigw-backend-selector: ollama'` will pass to ollama like this.
```
    - matches:
        - headers:
            - type: Exact
              name: aigw-backend-selector
              value: ollama
      backendRefs:
        - name: ollama
          namespace: default
```

Here's the full configuration I would like to simplify

<details>

```yaml
apiVersion: gateway.networking.k8s.io/v1
kind: GatewayClass
metadata:
  name: aigw-run
spec:
  controllerName: gateway.envoyproxy.io/gatewayclass-controller
---
apiVersion: gateway.networking.k8s.io/v1
kind: Gateway
metadata:
  name: aigw-run
spec:
  gatewayClassName: aigw-run
  listeners:
    - name: http
      protocol: HTTP
      port: 1975
  infrastructure:
    parametersRef:
      group: gateway.envoyproxy.io
      kind: EnvoyProxy
      name: envoy-ai-gateway
---
apiVersion: aigateway.envoyproxy.io/v1alpha1
kind: AIGatewayRoute
metadata:
  name: aigw-run
spec:
  schema:
    name: OpenAI
  targetRefs:
    - name: aigw-run
      kind: Gateway
      group: gateway.networking.k8s.io
  rules:
    - matches:
        - headers:
            - type: Exact
              name: aigw-backend-selector
              value: ollama
      backendRefs:
        - name: ollama
          namespace: default
apiVersion: aigateway.envoyproxy.io/v1alpha1
kind: AIServiceBackend
metadata:
  name: ollama
spec:
  timeouts:
    request: 3m
  schema:
    name: OpenAI
  backendRef:
    name: ollama
    kind: Backend
    group: gateway.envoyproxy.io
---
apiVersion: gateway.envoyproxy.io/v1alpha1
kind: Backend
metadata:
  name: ollama
spec:
  endpoints:
    - ip:
        address: 0.0.0.0
        port: 11434
```
</details>

However, as far as I know, you can't specify a default backend. This makes programmatic access not portable.

For example, here's how to add the headers with openai
```python
    client = openai.Client(default_headers={"aigw-backend-selector": "ollama"})
```

[...]

Issue #73, opened 2025-01-09
discussion: add "standard routing header" that gets prioritized 


mathetake:
per https://github.com/envoyproxy/ai-gateway/pull/71#discussion_r1907155026

Currently, users have to specify all routing conditions in LLMRoute.Rules. For example, if an AI Gateway user wants to allow clients to select backend by themselves, they need to define something like

```yaml
apiVersion: aigateway.envoyproxy.io/v1alpha1
kind: LLMRoute
metadata:
  name: some-route
  namespace: default
spec:
  inputSchema:
    schema: OpenAI
  rules:
    - matches:
      - headers:
        - type: Exact
          name: some-random-header-that-can-be-sent-directly-by-clients
          value: foo
      backendRefs:
        - name: somebackend
 ....
```

and tells the clients that use `-H some-random-header-that-can-be-sent-directly-by-clients: somebackend` to select the backend. This is not that inconvenient because at the end of the day there won't be hundreds of backends but at most tens.

On the other hand, we can provide the "standard routing header", say `x-ai-gateway-backend`, that effectively ignores the `LLMRoute.Rules` and routes the requests to the backend specified in the value of the standard header. To do so, we need to know which backends can be routed via that headers at the HTTPRoute construction phase. One way is to allow routing to any backend in the same namespace as LLMRoute, or only route to backends that _appear_ in LLMRoute.Rules.

Personally, i feel this "implicit routing" might makes things more confusing, but agree this will provide better experience, in fact this is how the PoC works which is because it didn't have the Rules and routing inside extproc.

i am opening this issue as I feel at least this is worth the discussion.

cc @envoyproxy/assignable

Krishanx92:
Are we discussing keeping only one of the options? I would prefer to keep both options and prioritize the standard header if it is set. Additionally, I suggest leaving that decision to the user level by defining a flag in the LLMRoute. For example, if the gateway user enables the flag, clients can use the standard x-ai-gateway-backend header to specify which backend they want to route to.

To determine which backends can be routed via headers, the gateway should route the request to "somebackend" only if it appears in one of the LLMRoute.Rules or is in the same namespace, which seems fine to me. However, if needed in the future, we can introduce an explicit allowed backend list for the spec as well.

PR #793, opened 2025-06-28
refactor: use Envoy native router

mathetake:
**Description**

This commit removes the handwritten header matching code from the extproc, and instead starts utilizing the hardened envoy native router.

Historically, we had only one giant extproc filter where we did all logics including model name extraction, routing and then body transformation & upstream authorization. Since #599, we split into two external processor filters; one sits at the normal HTTP router and the other is configured at the per-cluster upstream HTTP filter. In theory, the one at HTTP router has only one job on request path: extracting model name from the request body. However, due to the historical reason, the handwritten router logic component remained, and that comes with not only a maintenance cost (forcing a complex extproc & control plane orchestration) but also a potential security vulnerability. In fact, writing header matching logic can be an easy attack surface, so if it's possible, we should avoid writing our own header matching (routing logic) but should rely on the battle-tested hardened envoy native router.

With this commit, now a regex matching is available as well as there's no difference between HTTPRoute's matching and AIGatewayRoute's matching implementation. This also opens up a possibility to support path matching in our rule.

Ref #612 
Ref #73


Xunzhuo (approved):
love the simplification! let us get this in

Xunzhuo:
It looks like a conflict @mathetake can resolve it?

Xunzhuo:
@mathetake feel free to merge it after CI passed

mathetake:
Ok looks like I need to add more coverage

PR #794, opened 2025-06-30
refactor: separate api to individual files

Xunzhuo:
**Description**

Currently, we kept all API defs in a single file, as the API grows, it becomes large and unmaintainable.

This PR separates API into different files grouping by CRDs.

ref: https://github.com/envoyproxy/gateway/tree/main/api/v1alpha1


api/v1alpha1/shared_types.go
```diff
@@ -0,0 +1,125 @@
...
+)
+
+const (
+	// AIModelHeaderKey is the header key whose value is extracted from the request by the ai-gateway.
+	// This can be used to describe the routing behavior in HTTPRoute referenced by AIGatewayRoute.
+	AIModelHeaderKey = "x-ai-eg-model"
+)
+
+// LLMRequestCost configures each request cost.
+type LLMRequestCost struct {
```
mathetake, line 74: it's OK but this is only used in AIGatewayRoute, so not "shared" as opposed to being put here.

PR #813, opened 2025-07-01
extproc: account for parallel tool calls

### LINKED ISSUE #736: Add parallel tooling for AWS Bedrock
*Description*:
>Describe the desired behavior, what scenario it enables and how it
would be used.

Support multiple tool calls within the same response message for bedrock.

```
    MESSAGES_WITH_PARALLEL_TOOL_RESPONSE = [
        {
            "role": "user",
            "content": "What is the weather in Dallas, Texas and Orlando, Florida in Fahrenheit?",
        },
        {
            "role": "assistant",
            "tool_calls": [
                {
                    "id": "tool-1",
                    "type": "function",
                    "function": {
                        "name": WEATHER_TOOL["function"]["name"],
                        "arguments": '{"city": "Dallas", "state": "TX", '
                        '"unit": "fahrenheit"}',
                    },
                },
                {
                    "id": "tool-2",
                    "type": "function",
                    "function": {
                        "name": WEATHER_TOOL["function"]["name"],
                        "arguments": '{"city": "Orlando", "state": "Fl", '
                        '"unit": "fahrenheit"}',
                    },
                },
            ],
        },
        {
            "role": "tool",
            "tool_call_id": "tool-1`",
            "content": "The weather in Dallas TX is 88 degrees fahrenheit with mostly "
            "cloudy skies and a chance of rain in the evening.",
        },
        {
            "role": "tool",
            "tool_call_id": "tool-2",
            "content": "The weather in Orlando FL is 77 degrees fahrenheit with clear"
            "skies.",
        },
    ]
```

openai.BadRequestError: Error code: 400 - {'type': 'error', 'error': {'type': 'ValidationException:http://internal.amazon.com/coral/com.amazon.bedrock/', 'code': '400', 'message': 'Expected toolResult blocks at messages.2.content for the following Ids: tool-2'}}

[optional *Relevant Links*:]

aabchoo:
**Description**
Resolves #736

Assistant that calls multiple tools are expected to group tool result in the same message. Adding logic for that!


aabchoo:
edit: I think I will leave it to upstream to handle that

internal/extproc/translator/openai_awsbedrock.go
```diff
@@ -405,7 +408,14 @@ func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) openAIMessageToBedrockMes
 			if err != nil {
 				return err
 			}
-			bedrockReq.Messages = append(bedrockReq.Messages, bedrockMessage)
+
+			latestMessage := bedrockReq.Messages[len(bedrockReq.Messages)-1]
```
aabchoo: only concern here is if it's possible to send the tool messages before the assistant message

aabchoo: ie https://github.com/envoyproxy/ai-gateway/blob/main/internal/extproc/translator/openai_awsbedrock_test.go#L76-L101

yuzisun: The assistant message with tool_calls comes first, indicating the intention to use a tool, and then the tool message provides the result of that tool's execution. But there could be another assistant message with the tool results.

yuzisun: I think instead of using latestMessage, we can find the message with user role which has `ToolResult`, since we are merging to a single user role with tool results then it should only has one such message in the array.

aabchoo: All Tool messages are aggregated together? It's not aggregated with the tools  it's called with?

internal/extproc/translator/openai_awsbedrock.go
```diff
@@ -405,10 +406,28 @@ func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) openAIMessageToBedrockMes
 			if err != nil {
 				return err
 			}
+
+			for i+1 < openAIReqMessageLen {
```
yuzisun, line 410:
```suggestion
                         // Coalesce consecutive tool messages following a user message
			for i+1 < openAIReqMessageLen {
```

internal/extproc/translator/openai_awsbedrock.go
```diff
@@ -405,10 +406,28 @@ func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) openAIMessageToBedrockMes
...
+				nextMessage := &openAIReq.Messages[i+1]
+				if nextMessage.Type != openai.ChatMessageRoleTool {
+					break
+				}
+
+				nextToolMessage := nextMessage.Value.(openai.ChatCompletionToolMessageParam)
```
yuzisun: there is a type assertion risk here, consider using the "comma ok" idiom.

internal/extproc/translator/openai_awsbedrock.go
```diff
@@ -405,10 +406,28 @@ func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) openAIMessageToBedrockMes
...
+				nextBedrockMessage, err := o.openAIMessageToBedrockMessageRoleTool(&nextToolMessage, awsbedrock.ConversationRoleUser)
+				if err != nil {
+					return err
+				}
```
yuzisun, line 423: check the length of the content in case it is empty

yuzisun (approved):

PR #819, opened 2025-07-02
feat: implement GCP Gemini request and response translation

sukumargaonkar:
**Description**

This PR request and response translation for gcp-gemini models

Issue: https://github.com/envoyproxy/ai-gateway/issues/609

This PR only support basic requests with text and images
Future PRs will add support for tools, streaming-requests etc.


internal/extproc/translator/util.go
```diff
@@ -0,0 +1,38 @@
...
+	"encoding/base64"
+	"fmt"
+	"regexp"
+)
+
+const (
+	MimeTypeImageJPEG = "image/jpeg"
+	MimeTypeImagePNG  = "image/png"
+	MimeTypeImageGIF  = "image/gif"
+	MimeTypeImageWEBP = "image/webp"
```
mathetake: could you please not export unnecessarily?

internal/extproc/translator/openai_gcpvertexai.go
```diff
@@ -49,10 +61,88 @@ func (o *openAIToGCPVertexAITranslatorV1ChatCompletion) ResponseHeaders(headers
...
+		}
+	}
+
+	headerMutation, bodyMutation = buildGCPRequestMutations(nil, openAIRespBytes)
+
+	return headerMutation, bodyMutation, usage, nil
+}
+
+func (o *openAIToGCPVertexAITranslatorV1ChatCompletion) openAIMessageToGeminiMessage(openAIReq *openai.ChatCompletionRequest) (gcp.GenerateContentRequest, error) {
+	// Convert OpenAI messages to Gemini Contents and SystemInstruction
```
mathetake: nit; for any code comment, make sure each comment ends with period just like a normal sentence (ack some comments already lacking it but as long as i review, i will keep commenting on this 
```suggestion
	// Convert OpenAI messages to Gemini Contents and SystemInstruction.
```

sukumargaonkar: got it. will keep in mind

internal/extproc/translator/gemini_helper.go
```diff
@@ -20,30 +28,488 @@ const (
...
+		return nil, nil, fmt.Errorf("unsupported content type in assistant message: %T", v)
+	}
+
+	return parts, knownToolCalls, nil
+}
+
+// toGeminiGenerationConfig converts OpenAI request to Gemini GenerationConfig
+func toGeminiGenerationConfig(openAIReq *openai.ChatCompletionRequest) (*genai.GenerationConfig, error) {
+	if openAIReq == nil {
+		return nil, fmt.Errorf("input request is nil")
```
mathetake: when does this happen? if not at all expected, then let's panic or simply remove this if block

internal/extproc/translator/gemini_helper.go
```diff
@@ -20,30 +28,488 @@ const (
...
+		switch msgUnion.Type {
+		case openai.ChatMessageRoleDeveloper:
+			msg := msgUnion.Value.(openai.ChatCompletionDeveloperMessageParam)
+			inst, err := fromDeveloperMsg(msg)
+			if err != nil {
+				return nil, nil, fmt.Errorf("error converting developer message: %w", err)
+			}
+			if len(inst) != 0 {
+				if systemInstruction == nil {
+					systemInstruction = &genai.Content{Role: genai.RoleUser}
```
mathetake: maybe put code comment on why this user insertion is needed? Let's make sure that any non trivial logic like this should have a code comment for future contributors.

internal/extproc/translator/gemini_helper.go
```diff
@@ -20,30 +28,488 @@ const (
...
+			gcpParts = append(gcpParts, parts...)
+		case openai.ChatMessageRoleTool:
+			msg := msgUnion.Value.(openai.ChatCompletionToolMessageParam)
+			part, err := fromToolMsg(msg, knownToolCalls)
+			if err != nil {
+				return nil, nil, fmt.Errorf("error converting tool message: %w", err)
+			}
+			gcpParts = append(gcpParts, part)
+		case openai.ChatMessageRoleAssistant:
+			// flush any accumulated user/tool parts before assistant
```
mathetake:
```suggestion
			// Flush any accumulated user/tool parts before assistant.
```

internal/extproc/translator/openai_gcpvertexai.go
sukumargaonkar: should we add [godot](https://golangci-lint.run/usage/linters/#godot)?

mathetake: oh wow i didn't know about that linter. let's add it...would you mind opening a pr?

internal/extproc/translator/gemini_helper.go
sukumargaonkar: I agree with what you said
ideally it never happens as the openAIReq is passed from the [RequestBody()](https://github.com/sukumargaonkar/ai-gateway/blob/cb12d44ece4316cdd996ae0315d7a1ee13fab0ae/internal/extproc/translator/openai_gcpvertexai.go#L33) function. added the check defensively

should i instead add the check in [RequestBody()](https://github.com/sukumargaonkar/ai-gateway/blob/cb12d44ece4316cdd996ae0315d7a1ee13fab0ae/internal/extproc/translator/openai_gcpvertexai.go#L33) and change the input param here from pointer to struct?

mathetake: you can remove the check altogether as that's not expected at all. no need to nil check.

Speaking of using struct, no, we shouldn't as the data structure openai.ChatCompletionReques is huge enough that doesn't fit in cpu registers and result in unnecessary costly function call.

internal/extproc/translator/gemini_helper.go
sukumargaonkar: fair point.
tested request without the role set. works
the api allows it to be unset. will remove it

api/v1alpha1/registry.go
```diff
@@ -19,10 +19,10 @@ func init() {
 const GroupName = "aigateway.envoyproxy.io"
 
 var (
-	// schemeGroupVersion is group version used to register these objects
+	// schemeGroupVersion is group version used to register these objects.
```
mathetake: could you not fix unrelated stuff mixed up in a single PR? I understand why you made this change but I would appreciate it if you could either revert this or make another PR. I don't want to mess up the commit history of each file. E.g. if we allow this, this `api/v1alpha1/registry.go` will have a commit with `feat: implement GCP Gemini request and response translation` message, which is nothing to do with the actual change.

mathetake: applies to any other similar change. This PR should solely focus on the translation logic. This would help reviewers focus on what's important and help others in the future looking at this PR.

internal/extproc/translator/openai_gcpvertexai.go
```diff
@@ -49,10 +59,89 @@ func (o *openAIToGCPVertexAITranslatorV1ChatCompletion) ResponseHeaders(headers
...
 	headerMutation *extprocv3.HeaderMutation, bodyMutation *extprocv3.BodyMutation, tokenUsage LLMTokenUsage, err error,
 ) {
-	// TODO: Implement response body translation from GCP Gemini to OpenAI format
-	_, _, _ = respHeaders, body, endOfStream
-	return nil, nil, LLMTokenUsage{}, nil
+	// Read the body
+	bodyBytes, err := io.ReadAll(body)
+	if err != nil {
+		return nil, nil, LLMTokenUsage{}, fmt.Errorf("error reading response body: %w", err)
+	}
```
mathetake: this unnecessarily copies the body. You can directly unmarshal into the struct via `json.NewDecoder(body).Decode`.

internal/extproc/translator/gemini_helper.go
```diff
@@ -20,30 +28,485 @@ const (
...
+			if len(inst) != 0 {
+				if systemInstruction == nil {
+					systemInstruction = &genai.Content{}
+				}
+				systemInstruction.Parts = append(systemInstruction.Parts, inst...)
+			}
+		case openai.ChatMessageRoleSystem:
+			msg := msgUnion.Value.(openai.ChatCompletionSystemMessageParam)
+			devMsg := systemMsgToDeveloperMsg(msg)
+			inst, err := fromDeveloperMsg(devMsg)
```
yuzisun: Understand you are trying to use `fromDeveloperMsg`, but looks a bit unnecessary that it goes from system message -> developer message ->  system instruction.

internal/extproc/translator/openai_gcpvertexai.go
mathetake: i will take care of this

mathetake: https://github.com/envoyproxy/ai-gateway/pull/820

internal/extproc/translator/gemini_helper.go
sukumargaonkar: true, the alternative is to have two functions `fromDeveloperMsg` and `fromSystemMsg` which have pretty much identical function body.

prefer that approach?

api/v1alpha1/registry.go
sukumargaonkar: makes sense. will keep in mind
for now, will remove changes from non-relevant files

internal/extproc/translator/openai_gcpvertexai.go
sukumargaonkar: Thanks for making the PR

mathetake:
https://github.com/envoyproxy/ai-gateway/pull/752 @sukumargaonkar could you fix the remaining comments in this PR as well before this PR? it shouldn't take much cycles

sukumargaonkar:
> #752 @sukumargaonkar could you fix the remaining comments in this PR as well before this PR? it shouldn't take much cycles

yes, working on addressing https://github.com/envoyproxy/ai-gateway/pull/819#pullrequestreview-2980692814

was having issues setting it up locally

mathetake:
I am not going through the detail but one generic question: how are you going to translate "reasoning_effort" portion ? It is either low, medium or high (https://platform.openai.com/docs/api-reference/responses-streaming/response/incomplete) and I think it should be translated to the corresponding `thinking_budget` parameter in Vertex AI Gemini.

FWIW, Gemini on AI Studio's openai compatible endpoint translates 

> "low", "medium", and "high", which map to 1,024, 8,192, and 24,576 tokens, respectively.

according to their documentation here (https://ai.google.dev/gemini-api/docs/openai).


Can we do exactly the same thing or do you have different idea? We would love to see the parameter supported

sukumargaonkar:
> I am not going through the detail but one generic question: how are you going to translate "reasoning_effort" portion ? It is either low, medium or high (https://platform.openai.com/docs/api-reference/responses-streaming/response/incomplete) and I think it should be translated to the corresponding `thinking_budget` parameter in Vertex AI Gemini.
> 
> FWIW, Gemini on AI Studio's openai compatible endpoint translates
> 
> > "low", "medium", and "high", which map to 1,024, 8,192, and 24,576 tokens, respectively.
> 
> according to their documentation here (https://ai.google.dev/gemini-api/docs/openai).
> 
> Can we do exactly the same thing or do you have different idea? We would love to see the parameter supported

This PR only handles basic text and image input
future PRs will address tools and thinking/reasoning requests

broke it down to make reviewing easier

But good point, will keep your comment in mind for future PRs

internal/extproc/translator/openai_gcpvertexai.go
```diff
@@ -46,10 +57,94 @@ func (o *openAIToGCPVertexAITranslatorV1ChatCompletion) ResponseHeaders(headers
...
+
+	// Parse the GCP response.
+	var gcpResp genai.GenerateContentResponse
+	if err = json.NewDecoder(body).Decode(&gcpResp); err != nil {
+		return nil, nil, LLMTokenUsage{}, fmt.Errorf("error decoding GCP response: %w", err)
+	}
+
+	var openAIRespBytes []byte
+	if len(gcpResp.Candidates) > 0 {
```
mathetake: I _guess_ if we have the if block here, `openAIRespBytes` has the zero-length, hence the resulting the body mutation is also nil, which results in the raw GCP response will be returned to the downstream client who is likely to be using OpenAI SDK? I think that seems like problematic. So the question would be like

* When `len(gcpResp.Candidates)` == 0 happens?
* If the case can happen or we cannot be certain, should be make sure that the empty openai response will be constructed in the else block here so that the downstream OpenAI SDK client won't receive the GCP raw response?

sukumargaonkar: good point.
removed the if condition
updated the corresponding test-case

internal/extproc/translator/gemini_helper.go
```diff
@@ -20,30 +28,485 @@ const (
...
+
 func buildGCPModelPathSuffix(publisher, model, gcpMethod string) string {
 	pathSuffix := fmt.Sprintf("publishers/%s/models/%s:%s", publisher, model, gcpMethod)
 	return pathSuffix
 }
 
 // buildGCPRequestMutations creates header and body mutations for GCP requests
 // It sets the ":path" header, the "content-length" header and the request body.
-func buildGCPRequestMutations(path string, reqBody []byte) (*ext_procv3.HeaderMutation, *ext_procv3.BodyMutation) {
+func buildGCPRequestMutations(path *string, reqBody []byte) (*ext_procv3.HeaderMutation, *ext_procv3.BodyMutation) {
```
mathetake: people usually do not use a pointer to the string. This unnecessarily results in escaping the string header (a pair of the length of and pointer to the buffer) to heap. You can just use `len(path) != 0` to check if the string empty or not since I believe the empty path is invalid anyways.

internal/extproc/translator/openai_gcpvertexai.go
```diff
@@ -46,10 +57,92 @@ func (o *openAIToGCPVertexAITranslatorV1ChatCompletion) ResponseHeaders(headers
...
 ) {
-	// TODO: Implement response body translation from GCP Gemini to OpenAI format.
-	_, _, _ = respHeaders, body, endOfStream
-	return nil, nil, LLMTokenUsage{}, nil
+	if statusStr, ok := respHeaders[statusHeaderName]; ok {
+		var status int
+		if status, err = strconv.Atoi(statusStr); err == nil {
+			if !isGoodStatusCode(status) {
+				// TODO: Parse GCP error response and convert to OpenAI error format.
+				// For now, just return error response as-is.
```
yuzisun, line 68: Can we prioritize this TODO, i think it is important translation to deliver user error response.

sukumargaonkar: plan to do this in the next PR

internal/controller/rotators/gcp_oidc_token_rotator.go
```diff
@@ -65,10 +65,10 @@ type stsTokenGenerator func(
...
 // gcpOIDCTokenRotator implements Rotator interface for GCP access token exchange.
 // It handles the complete authentication flow for GCP Workload Identity Federation:
-//  1. Obtaining an OIDC token from the configured provider
-//  2. Exchanging the OIDC token for a GCP STS token
-//  3. Using the STS token to impersonate a GCP service account
-//  4. Storing the resulting access token in a Kubernetes secret
+//  1. Obtaining an OIDC token from the configured provider.
+//  2. Exchanging the OIDC token for a GCP STS token.
+//  3. Using the STS token to impersonate a GCP service account.
+//  4. Storing the resulting access token in a Kubernetes secret.
```
mathetake: can you revert this ?

internal/extproc/translator/gemini_helper.go
mathetake: yes, can you go ahead and have two different functions instead of obfuscate the actual logic by going through unnecessary code path

internal/extproc/translator/gemini_helper.go
```diff
@@ -6,11 +6,19 @@
 package translator
```
mathetake, line 6: All functions here must be prefixed with gemini or gcp. Any function is visible inside the package, simply using generic names would be distracting to others when working on totally different context.

mathetake: and also can we be consistent about the use of the word gemini/gcp here?

mathetake:
almost there!

internal/extproc/translator/gemini_helper.go
sukumargaonkar: should i create a sub package?

sukumargaonkar: renamed functions 

```
toGeminiContents --> openAIMessagesToGeminiContents
fromDeveloperMsg --> developerMsgToGeminiParts
fromUserMsg --> userMsgToGeminiParts
fromToolMsg --> toolMsgToGeminiParts
toGeminiGenerationConfig --> openAIReqToGeminiGenerationConfig
toOpenAIChoices --> geminiCandidatesToOpenAIChoices
toOpenAIFinishReason --> geminiFinishReasonToOpenAI
extractTextParts --> extractTextFromGeminiParts
extractToolCalls --> extractToolCallsFromGeminiParts
toOpenAIUsage --> geminiUsageToOpenAIUsage
toLogprobs --> geminiLogprobsToOpenAILogprobs
```

sukumargaonkar: @mathetake this looks good?

mathetake: yep, looks much better!

mathetake (approved):
i am not going through the actual translation logic but LGTM on the other parts generally! Thanks! I will defer to @yuzisun for the final stamping

mathetake:
on second thought it seems like there would be a conflict with @alexagriffith's PR #838, so I am going ahead and merging to unlock you guys to work on subsequent stuff

Issue #604, opened 2025-05-08
Support DynamicLoadBalancing beyond AIE(API inference extension)


kerthcet:
*Description*:

DynamicLoadBalancing extends the gateway routing system by picking the endpoints(Pod IPs etc.)directly, it's great. But after I checked the code, I found two limitations here:
- From the API perspective, the `AIGatewayRoute` particularly, only AIE is supported, see the code link below. For users don't employ AIE like us, we hope to have this support as well. I know we can update the configmap to enable this, but it's a bit complex. https://github.com/envoyproxy/ai-gateway/blob/be2b479b04bc7a219b0c8239143bfbabebdcd615/internal/controller/ai_gateway_route.go#L295-L302
- The Pod IPs are attached to the Backend configurations, which means once Pods updates, we have to update the configurations. Have no idea whether this is an efficient way, maybe still useful to some users with static Pod IPs, but to us, we hope to manage the Pods ourselves. So we may implement a new DynamicLoadBalancing to replace the default one.

This is a blocking issue for us, what we're doing right now is we're building the `metrics-based router`, we have all the Pods informations by collecting necessary metrics and want to bake it as a plugin into the Envoy AI Gateway. So within the dynamicLoadBalaner, we can make wise routing decisions by picking Pod endpoint directly.

Correct me if I misunderstood the system design here. Thanks!

mathetake:
yeah let's come back to this after #599 - almost all the ip address resolution, etc code will be irrelevant after the refactoring

kerthcet:
I will revisit this these days, it would be great if @mathetake could give me some advices about how to extend the gateway on Pod IP level.

mathetake:
so the refactoring is going on in https://github.com/envoyproxy/ai-gateway/pull/614 which i will resume after #629 lands. We are basically trying to make Envoy AI Gateway agnostic of the ip-based endpoint picker implementation and let the users deploy their own extproc or whatever Envoy filter to select the endpoint. In other words, these dynlib or ip resolver stuff will be all removed

kerthcet:
Thanks for the response, we hope to help accelerate the process because it's a blocking issue for us right now, anything I can help here. 

Besides, if we can't accomplish this recently, can we(I mean our team) look for other solutions to bypass this temporary? Thanks.

PR #823, opened 2025-07-04
feat: add support for endpoint picker

Xunzhuo:
**Description**

This PR addes support for inferencePool, which allows Envoy AI Gateway to integrate with ANY endpoint picker who is supported the inferencePool.

By integrating with the Endpoint Picker like Gateway API Inference Extenstion or the non-GIE EPP, it can expand Envoy AI Gateway`s abilities to advanced scheduleing algorithm to optimize inference.

Fixes: https://github.com/envoyproxy/ai-gateway/issues/423
Fixes https://github.com/envoyproxy/ai-gateway/issues/604
Fixes: https://github.com/envoyproxy/ai-gateway/issues/648

Some follow-up: https://github.com/envoyproxy/ai-gateway/issues/911

### DIFF
--- a/examples/inference-pool/base.yaml
+++ b/examples/inference-pool/base.yaml
@@ -127,8 +127,8 @@ spec:
           ports:
             - containerPort: 9002
             - containerPort: 9003
-        - name: metrics
-          containerPort: 9090
+            - name: metrics
+              containerPort: 9090
           livenessProbe:
             grpc:
               port: 9003


internal/controller/gateway.go
```diff
@@ -189,11 +196,150 @@ func (c *GatewayController) ensureExtensionPolicy(ctx context.Context, gw *gwapi
...
+
+	eppServiceName := string(inferencePool.Spec.ExtensionRef.Name)
+	var eppServicePort int32 // Default port.
+	if inferencePool.Spec.ExtensionRef.PortNumber != nil {
+		eppServicePort = int32(*inferencePool.Spec.ExtensionRef.PortNumber)
+	} else {
+		// The port number on the service running the extension. When unspecified,
+		// implementations SHOULD infer a default value of 9002 when the Kind is
+		// Service.
+		eppServicePort = 9002
```
yuzisun: make this a constant as other places like tests will need to use it

internal/controller/gateway.go
```diff
@@ -189,11 +196,150 @@ func (c *GatewayController) ensureExtensionPolicy(ctx context.Context, gw *gwapi
...
+					Hostname: fmt.Sprintf("%s.%s.svc.cluster.local", eppServiceName, route.Namespace),
+					Port:     eppServicePort,
+				},
+			}},
+		},
+	}
+
+	if err := c.client.Create(ctx, &backend); err != nil && !apierrors.IsAlreadyExists(err) {
+		return fmt.Errorf("failed to create EPP backend: %w", err)
+	}
```
yuzisun: We need to handle the update case as well as InferencePool resource can be updated like FQDN or port.

internal/controller/gateway.go
```diff
@@ -189,11 +196,150 @@ func (c *GatewayController) ensureExtensionPolicy(ctx context.Context, gw *gwapi
...
+							Group:     ptr.To(gwapiv1.Group("gateway.envoyproxy.io")),
+							Namespace: ptr.To(gwapiv1.Namespace(gw.Namespace)),
+						},
+					}},
+				},
+			}},
+		},
+	}
+
+	if err := c.client.Create(ctx, eppExtPolicy); err != nil && !apierrors.IsAlreadyExists(err) {
```
yuzisun: Same here need to handle the update case if the resource exists already, we ran into issue before if the extension policy already exists then it won't update when ai gateway controller code changes here like updating processing mode etc.

internal/controller/gateway.go
mathetake: instead of creating eep, how about adding extproc into the specific route which routes to the inference pool lb policy cluster? that why other normal routes won't need to talk to eep unnecessarily. This could be done in extension server i guess?

internal/controller/gateway.go
mathetake: instead of creating eep, how about adding extproc into the specific route which routes to the inference pool lb policy cluster? that why other normal routes won't need to talk to eep unnecessarily. This could be done in extension server i guess?

mathetake:
instead of creating eep, how about adding extproc into the specific route which routes to the inference pool lb policy cluster? that way other normal routes won't need to talk to eep unnecessarily. This could be done in extension server i guess?

yuzisun:
> instead of creating eep, how about adding extproc into the specific route which routes to the inference pool lb policy cluster? that way other normal routes won't need to talk to eep unnecessarily. This could be done in extension server i guess?

That’s a very good point. We need to make sure normal routes do not go through epp.

Xunzhuo:
Yep, that is reasonable :)

mathetake:
having said that, one concern about the per route is that it might not work well with ClearRouteCache: true which is set by our AI Gateway extproc. The EPP extproc must come after the ai gateway extproc since until then envoy doesn't know the destination. However, per route filter config might not work well with the deferred route calculation. Maybe it's not the case but something i am worried about it now...

internal/extproc/chatcompletion_processor.go
```diff
@@ -341,13 +345,22 @@ func (c *chatCompletionProcessorUpstreamFilter) SetBackend(ctx context.Context,
 		panic("BUG: expected routeProcessor to be of type *chatCompletionProcessorRouterFilter")
 	}
 	rp.upstreamFilterCount++
+	if c.isEndpointPicker() {
+		b = &filterapi.Backend{
+			Name:              "endpoint-picker",
+			Schema:            filterapi.VersionedAPISchema{Name: filterapi.APISchemaOpenAI},
+			ModelNameOverride: c.requestHeaders["x-ai-eg-model"],
```
mathetake: i believe this must be constructed from the same source as non inference pool backend, e.g. AIGatewayRouteRuleBackendRef. Since we cannot access the endpoint level metadata in this case, I wonder if it might be a good idea to insert the header mutation filter BEFORE the upstream AI Gateway extproc and insert the "backend name" as well as "model name override" from the corresponding AIGatewayRouteRuleBackendRef as headers. Then retrieve these headers in here. [You can find the header mutation filter example here](https://github.com/envoyproxy/ai-gateway/blob/2004c864453c7a0594a1393a11b1b5edabfa3361/tests/extproc/envoy.yaml#L975-L977).

For schema, yes you can put the constant OpenAI schema, but I think you forget `Version: "v1"` here like:
```go
filterapi.VersionedAPISchema{Name: filterapi.APISchemaOpenAI, Version: "v1"}
```

tests/e2e/e2e_test.go
```diff
@@ -169,10 +169,46 @@ func cleanupKindCluster(testsFailed bool) {
...
 	return kubectlApplyManifest(ctx, infExtURL)
 }
 
+func installVLLMDeployment(ctx context.Context) (err error) {
+	const vllmURL = "https://github.com/kubernetes-sigs/gateway-api-inference-extension/raw/main/config/manifests/vllm/sim-deployment.yaml"
+	return kubectlApplyManifest(ctx, vllmURL)
+}
+
+func installInferenceModel(ctx context.Context) (err error) {
+	const inferenceModelURL = "https://github.com/kubernetes-sigs/gateway-api-inference-extension/raw/main/config/manifests/inferencemodel.yaml"
```
mathetake: this should be pinned to a specific version v0.4.0 maybe

tests/e2e/e2e_test.go
```diff
@@ -169,10 +169,46 @@ func cleanupKindCluster(testsFailed bool) {
...
+	return kubectlApplyManifest(ctx, inferenceModelURL)
+}
+
+func installInferencePoolResources(ctx context.Context) (err error) {
+	const inferencePoolURL = "https://github.com/kubernetes-sigs/gateway-api-inference-extension/raw/main/config/manifests/inferencepool-resources.yaml"
```
mathetake: same

tests/e2e/e2e_test.go
```diff
@@ -169,10 +169,46 @@ func cleanupKindCluster(testsFailed bool) {
 }
 
 func installInferenceExtensionCRD(ctx context.Context) (err error) {
-	const infExtURL = "https://github.com/kubernetes-sigs/gateway-api-inference-extension/releases/download/v0.2.0/manifests.yaml"
+	const infExtURL = "https://github.com/kubernetes-sigs/gateway-api-inference-extension/releases/latest/download/manifests.yaml"
```
mathetake: let's pin it to 0.4.0

mathetake (commented):
nits

Xunzhuo:
blocked by  https://github.com/envoyproxy/ai-gateway/pull/860

internal/extensionserver/post_cluster_modify.go
```diff
@@ -0,0 +1,385 @@
...
+			result = append(result, buildExtProcClusterForInferencePool(pool))
+		}
+	}
+
+	return result
+}
+
+// buildExtProcClusterForInferencePool builds and returns a "STRICT_DNS" cluster from the given pool.
+func buildExtProcClusterForInferencePool(pool *gwaiev1a2.InferencePool) *clusterv3.Cluster {
+	if pool == nil || pool.Spec.EndpointPickerConfig.ExtensionRef == nil {
```
mathetake: i think you should either remove this if block of panic instead of silencing the error.

internal/internalapi/internalapi.go
```diff
@@ -14,6 +14,8 @@ const (
 	InternalEndpointMetadataNamespace = "aigateway.envoy.io"
 	// InternalMetadataBackendNameKey is the key used to store the backend name
 	InternalMetadataBackendNameKey = "per_route_rule_backend_name"
+	// InternalMetadataInferencePoolKey is the key used to store the inference pool
+	InternalMetadataInferencePoolKey = "per_route_rule_inference_pool"
```
mathetake: can you not put this in here? As the documentation says this package for sharing things among multiple places (controller<>extproc). VS InternalMetadataInferencePoolKey is only used in the extension server.

internal/extensionserver/post_cluster_modify.go
```diff
@@ -0,0 +1,385 @@
...
+		if pool := getInferencePoolByMetadata(cluster); pool != nil {
+			result = append(result, buildExtProcClusterForInferencePool(pool))
+		}
+	}
+
+	return result
+}
+
+// buildExtProcClusterForInferencePool builds and returns a "STRICT_DNS" cluster from the given pool.
+func buildExtProcClusterForInferencePool(pool *gwaiev1a2.InferencePool) *clusterv3.Cluster {
```
mathetake: let's make it more explicit about what this does since we have two kinds of clusters dealing with in this package regarding InferencePool. (one is for route and one is for endpoint picker)  

```suggestion
func buildExtProcClusterForInferencePoolEndpointPicker(pool *gwaiev1a2.InferencePool) *clusterv3.Cluster {
```

internal/extensionserver/post_cluster_modify.go
```diff
@@ -0,0 +1,385 @@
...
+						PortNumber: ptr.To(gwaiev1a2.PortNumber(port)),
+					},
+				},
+			},
+		},
+	}
+}
+
+// buildClustersForInferencePool builds and returns a "STRICT_DNS" cluster for each InferencePool cluster.
+func buildClustersForInferencePool(clusters []*clusterv3.Cluster) []*clusterv3.Cluster {
```
mathetake:
```suggestion
func buildClustersForInferencePoolEndpointPickers(clusters []*clusterv3.Cluster) []*clusterv3.Cluster {
```

internal/extensionserver/post_listener_modify.go
```diff
@@ -0,0 +1,108 @@
...
+		// If a inference dummy ext proc filter already exists, update it. Otherwise, create it.
+		_, baIndex, err := findInferencePoolExtProc(httpConManager.HttpFilters)
+		if err != nil {
+			s.log.Error(err, "failed to unmarshal the existing inference pool ext proc filter")
+			continue
+		}
+		dummyExtProc := &httpconnectionmanagerv3.HttpFilter{}
+		if baIndex == -1 {
+			s.log.Info("Creating a dummy inference pool ext proc filter")
+			// Create a dummy filter.
```
mathetake: this definitely needs a comment on why a dummy filter is needed in the first place. Imagine someone without context coming in and taking a look at the code.

internal/extensionserver/post_cluster_modify.go
```diff
@@ -0,0 +1,385 @@
...
+func dnsNameForInferencePool(pool *gwaiev1a2.InferencePool) string {
+	ns := pool.GetNamespace()
+	svc := pool.Spec.EndpointPickerConfig.ExtensionRef.Name
+	return fmt.Sprintf("%s.%s.svc", svc, ns)
+}
+
+// portForInferencePool returns the port number for the given InferencePool.
+func portForInferencePool(pool *gwaiev1a2.InferencePool) uint32 {
+	if p := pool.Spec.ExtensionRef.ExtensionReference.PortNumber; p == nil {
+		return uint32(9002)
```
mathetake: where 9002 comes from? I see many places 9002 is used. Shall we define a constant and put a documentation or reference on where 9002 comes from.

internal/extensionserver/post_listener_modify.go
```diff
@@ -0,0 +1,108 @@
...
+			dummyExtProc = dummyHTTPFilterForInferencePool()
+		}
+		if baIndex == -1 {
+			length := len(httpConManager.HttpFilters)
+			httpConManager.HttpFilters = append(httpConManager.HttpFilters, httpConManager.HttpFilters[length-1])
+			httpConManager.HttpFilters[length-1] = dummyExtProc
```
mathetake: let's put comment on why it's before the last router filter like we discussed

internal/extensionserver/post_translate_modify.go
```diff
@@ -0,0 +1,342 @@
...
+	"google.golang.org/protobuf/types/known/wrapperspb"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	aigv1a1 "github.com/envoyproxy/ai-gateway/api/v1alpha1"
+	"github.com/envoyproxy/ai-gateway/internal/internalapi"
+)
+
+const (
+	ExtProcUDSClusterName    = "ai-gateway-extproc-uds"
+	ExtProcNameInferencePool = "envoy.filters.http.ext_proc/inferencepool"
```
mathetake: can you un-export these constants?

internal/extensionserver/post_virtualhost_modify.go
```diff
@@ -0,0 +1,45 @@
...
+	for _, route := range req.VirtualHost.Routes {
+		if dr := route.GetDirectResponse(); dr != nil {
+			if strings.Contains(dr.Body.GetInlineString(), "No matching route found") {
+				s.log.Info("found route not found response, modify it")
+			}
+			if inferencePool != nil {
+				if route.TypedPerFilterConfig == nil {
+					route.TypedPerFilterConfig = make(map[string]*anypb.Any)
+				}
+				route.TypedPerFilterConfig[ExtProcNameInferencePool] = inferencePool
```
mathetake: I see this seems problematic. When users make a request, say /v1/chat/completions, this route will be hit 100% at the very beginning phase of HCM at the filter chain construction. Hence this per_route extproc will be activated regardless of the route after the clearCache by the AI Gateway Extrpco. That means our objective to avoid sending requests to EPPs when the target backends are not inference pool is not satisfied. (I think on the other hand, that's why you are passing the tests.)
I still do not have a workaround for this, but like would it be possible to run the EPP extproc at the cluster level instead of router level by like setting a dummy header for the match? I don't know exactly when the original_dst/overide_host picks an ip address and port, but if it is possible to override the ip:port from the upstream extproc at the cluster level, then we can workaround this (i am not saying it is posibble....)

mathetake: I think I came up with a workaround (not verified it works): 

* Insert/Enable the EPP extproc filters at the HCM/listener level without disabling it. This will make the EPP extproc filters available on any route. However, make sure that if there's no route referencing InferencePool in a listener, then do not insert this filter.
* Disabling the EPP extproc filters on non EPP routes via [ExtProcPerRoute config level](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/ext_proc/v3/ext_proc.proto#extensions-filters-http-ext-proc-v3-extprocperroute). This per route config will be dynamically fetched by the Envoy's extproc filter implementation, so it should be available even after the clear route cache. This will add a slight overhead on non-InferencePool route but should be almost negligible (just a boolean check).
* Enabling the EPP extproc filters on EPP routes via ExtProcPerRoute config level like the opposite of above.

so with this you won't need to "No matching route found" route like this. Instead, we can focus on the listener level extproc filter insertions and route level disabling of extproc filters.


In other words, this is the opposite of the current code: enabling by default and disabling it on unnecessary routes.

Xunzhuo: Not sure about if the ExtProcPerRoute will work, but will give it a try. I am afraid of all routes will call epp ext-proc if disabled dont work.

So if it is true, after this try, I prefer we fallback to this approach and add limitations on this feature:
1.  adding support for inference on route level, not rule level, only one inference pool per aigatewayroute
2.  inference pool cannot mix use with aiservicebackend, afaik there is no clear need for mix reuse both of them in a single aigwroute, if do, end-user can sparate them to different aigwroute I think.

Xunzhuo: Had a test on config, new approach tested passed, they dont affect each other.

```
kind: AIGatewayRoute
metadata:
  name: inference-pool
  namespace: default
spec:
  schema:
    name: OpenAI
  targetRefs:
    - name: inference-pool
      kind: Gateway
  rules:
    - matches:
        - headers:
            - type: Exact
              name: x-ai-eg-model
              value: meta-llama/Llama-3.1-8B-Instruct
      backendRefs:
        - group: inference.networking.x-k8s.io
          kind: InferencePool
          name: vllm-llama3-8b-instruct
          modelNameOverride: meta-llama/Llama-3.1-8B-Instruct
    - matches:
        - headers:
            - type: Exact
              name: x-ai-eg-model
              value: mistral:latest
      backendRefs:
        - group: inference.networking.x-k8s.io
          kind: InferencePool
          name: mistral
          modelNameOverride: mistral:latest
    - matches:
        - headers:
            - type: Exact
              name: x-ai-eg-model
              value: some-cool-self-hosted-model
      backendRefs:
        - name: envoy-ai-gateway-basic-testupstream
```

Xunzhuo: Route config is like:
```
						"routes": [{
								"match": {
									"prefix": "/",
									"headers": [{
										"name": "x-ai-eg-model",
										"string_match": {
											"exact": "meta-llama/Llama-3.1-8B-Instruct"
										}
									}]
								},
								"route": {
									"cluster": "httproute/default/inference-pool/rule/0",
									"auto_host_rewrite": false,
									"timeout": "60s",
									"upgrade_configs": [{
										"upgrade_type": "websocket"
									}],
									"append_x_forwarded_host": true
								},
								"metadata": {
									"filter_metadata": {
										"envoy-gateway": {
											"resources": [{
												"kind": "HTTPRoute",
												"name": "inference-pool",
												"annotations": {
													"backend-ref-priority": "0:vllm-llama3-8b-instruct:0,1:mistral:0,2:envoy-ai-gateway-basic-testupstream:0"
												},
												"namespace": "default"
											}]
										},
										"aigateway.envoy.io": {
											"per_route_rule_inference_pool": "default/vllm-llama3-8b-instruct/vllm-llama3-8b-instruct-epp/9002"
										}
									}
								},
								"typed_per_filter_config": {
									"envoy.filters.http.ext_proc/endpointpicker_mistral_default_ext_proc": {
										"@type": "type.googleapis.com/envoy.extensions.filters.http.ext_proc.v3.ExtProcPerRoute",
										"disabled": true
									},
									"envoy.filters.http.ext_proc/envoyextensionpolicy/default/ai-eg-eep-inference-pool/extproc/0": {
										"@type": "type.googleapis.com/envoy.config.route.v3.FilterConfig",
										"config": {}
									}
								},
								"name": "httproute/default/inference-pool/rule/0/match/0/*"
							},
							{
								"match": {
									"prefix": "/",
									"headers": [{
										"name": "x-ai-eg-model",
										"string_match": {
											"exact": "mistral:latest"
										}
									}]
								},
								"route": {
									"cluster": "httproute/default/inference-pool/rule/1",
									"auto_host_rewrite": false,
									"timeout": "60s",
									"upgrade_configs": [{
										"upgrade_type": "websocket"
									}],
									"append_x_forwarded_host": true
								},
								"metadata": {
									"filter_metadata": {
										"aigateway.envoy.io": {
											"per_route_rule_inference_pool": "default/mistral/mistral-epp/9002"
										},
										"envoy-gateway": {
											"resources": [{
												"namespace": "default",
												"kind": "HTTPRoute",
												"name": "inference-pool",
												"annotations": {
													"backend-ref-priority": "0:vllm-llama3-8b-instruct:0,1:mistral:0,2:envoy-ai-gateway-basic-testupstream:0"
												}
											}]
										}
									}
								},
								"typed_per_filter_config": {
									"envoy.filters.http.ext_proc/envoyextensionpolicy/default/ai-eg-eep-inference-pool/extproc/0": {
										"@type": "type.googleapis.com/envoy.config.route.v3.FilterConfig",
										"config": {}
									},
									"envoy.filters.http.ext_proc/endpointpicker_vllm-llama3-8b-instruct_default_ext_proc": {
										"@type": "type.googleapis.com/envoy.extensions.filters.http.ext_proc.v3.ExtProcPerRoute",
										"disabled": true
									}
								},
								"name": "httproute/default/inference-pool/rule/1/match/0/*"
							},
							{
								"match": {
									"prefix": "/",
									"headers": [{
										"name": "x-ai-eg-model",
										"string_match": {
											"exact": "some-cool-self-hosted-model"
										}
									}]
								},
								"route": {
									"cluster": "httproute/default/inference-pool/rule/2",
									"auto_host_rewrite": true,
									"timeout": "60s",
									"upgrade_configs": [{
										"upgrade_type": "websocket"
									}],
									"append_x_forwarded_host": true
								},
								"metadata": {
									"filter_metadata": {
										"envoy-gateway": {
											"resources": [{
												"name": "inference-pool",
												"namespace": "default",
												"kind": "HTTPRoute",
												"annotations": {
													"backend-ref-priority": "0:vllm-llama3-8b-instruct:0,1:mistral:0,2:envoy-ai-gateway-basic-testupstream:0"
												}
											}]
										}
									}
								},
								"typed_per_filter_config": {
									"envoy.filters.http.ext_proc/endpointpicker_mistral_default_ext_proc": {
										"@type": "type.googleapis.com/envoy.extensions.filters.http.ext_proc.v3.ExtProcPerRoute",
										"disabled": true
									},
									"envoy.filters.http.ext_proc/envoyextensionpolicy/default/ai-eg-eep-inference-pool/extproc/0": {
										"@type": "type.googleapis.com/envoy.config.route.v3.FilterConfig",
										"config": {}
									},
									"envoy.filters.http.ext_proc/endpointpicker_vllm-llama3-8b-instruct_default_ext_proc": {
										"@type": "type.googleapis.com/envoy.extensions.filters.http.ext_proc.v3.ExtProcPerRoute",
										"disabled": true
									}
								},
								"name": "httproute/default/inference-pool/rule/2/match/0/*"
							},
```

Xunzhuo: listener config is like:
```
											{
												"name": "envoy.filters.http.ext_proc/endpointpicker_mistral_default_ext_proc",
												"typed_config": {
													"@type": "type.googleapis.com/envoy.extensions.filters.http.ext_proc.v3.ExternalProcessor",
													"grpc_service": {
														"envoy_grpc": {
															"cluster_name": "endpointpicker_mistral_default_ext_proc",
															"authority": "mistral-epp.default.svc:9002"
														}
													},
													"processing_mode": {
														"request_header_mode": "SEND",
														"response_header_mode": "SEND",
														"request_body_mode": "FULL_DUPLEX_STREAMED",
														"response_body_mode": "FULL_DUPLEX_STREAMED",
														"request_trailer_mode": "SEND",
														"response_trailer_mode": "SEND"
													},
													"message_timeout": "5s"
												}
											},
											{
												"name": "envoy.filters.http.ext_proc/endpointpicker_vllm-llama3-8b-instruct_default_ext_proc",
												"typed_config": {
													"@type": "type.googleapis.com/envoy.extensions.filters.http.ext_proc.v3.ExternalProcessor",
													"grpc_service": {
														"envoy_grpc": {
															"cluster_name": "endpointpicker_vllm-llama3-8b-instruct_default_ext_proc",
															"authority": "vllm-llama3-8b-instruct-epp.default.svc:9002"
														}
													},
													"processing_mode": {
														"request_header_mode": "SEND",
														"response_header_mode": "SEND",
														"request_body_mode": "FULL_DUPLEX_STREAMED",
														"response_body_mode": "FULL_DUPLEX_STREAMED",
														"request_trailer_mode": "SEND",
														"response_trailer_mode": "SEND"
													},
													"message_timeout": "5s"
												}
											},
											{
												"name": "envoy.filters.http.router",
												"typed_config": {
													"@type": "type.googleapis.com/envoy.extensions.filters.http.router.v3.Router",
													"suppress_envoy_headers": true
												}
											}
```

data.json
mathetake, line 1: could you remove this lol

internal/extensionserver/post_virtualhost_modify.go
mathetake: looks exactly what i wanted to see. really nice

data.json
Xunzhuo: yes, will remove in code polish

Xunzhuo:
After https://github.com/envoyproxy/gateway/pull/6524 lands, new algorithm would be:
1. find inferencepool relevant listener
2. insert epp extproc config into listener
3. find unrelated routes under relevant listener
4. insert extproc perroute to disable these routes

mathetake:
looks good and it would be much simpler

Xunzhuo:
new approach based on https://github.com/envoyproxy/gateway/pull/6524 landed it in https://github.com/envoyproxy/ai-gateway/pull/823/commits/03573fdcb44d6680be2e7970874458416202412f, and e2e test passed locally

Clipboard_Screenshot_1752669028

Xunzhuo:
Clipboard_Screenshot_1752832117

e2e passed

api/v1alpha1/ai_gateway_route_helper.go
```diff
@@ -13,14 +13,20 @@ const (
...
+	inferencePoolKind = "InferencePool"
 )
 
 // GetTimeoutsWithDefaults returns the timeouts with default values applied when not specified.
 // This ensures that AI Gateway routes have appropriate timeout defaults for AI workloads.
 func (r *AIGatewayRouteRule) GetTimeoutsOrDefault() *gwapiv1.HTTPRouteTimeouts {
 	defaultTimeout := defaultRequestTimeout
 
-	if r.Timeouts == nil {
+	// Handle nil receiver
```
mathetake:
```suggestion
```

internal/internalapi/internalapi.go
```diff
@@ -16,9 +16,19 @@ const (
...
 // i.e., the unique identifier for a backend that is associated with a specific
 // route rule in a specific AIGatewayRoute.
 func PerRouteRuleRefBackendName(namespace, name, routeName string, routeRuleIndex, refIndex int) string {
 	return fmt.Sprintf("%s/%s/route/%s/rule/%d/ref/%d", namespace, name, routeName, routeRuleIndex, refIndex)
 }
+
+// ClusterRefInferencePool generates a unique reference for an InferencePool cluster.
+func ClusterRefInferencePool(namespace, name, serviceName string, servicePort uint32) string {
+	return fmt.Sprintf("%s/%s/%s/%d", namespace, name, serviceName, servicePort)
+}
```
mathetake: can you move this to extensionserver package? no reason to exist here as it's not shared across multiple context.

internal/internalapi/internalapi.go
```diff
@@ -16,9 +16,19 @@ const (
 	InternalMetadataBackendNameKey = "per_route_rule_backend_name"
 )
 
+const (
+	// EndpointPickerHeEndpointPickerHeaderKeyader is the header key used to specify the target backend endpoint.
+	EndpointPickerHeaderKey = "x-gateway-destination-endpoint"
```
mathetake:
```suggestion
	// EndpointPickerHeaderKey is the header key used to specify the target backend endpoint.
    // This is the default header name in the reference implementation:
    // https://github.com/kubernetes-sigs/gateway-api-inference-extension/blob/2b5b337b45c3289e5f9367b2c19deef021722fcd/pkg/epp/server/runserver.go#L63
	EndpointPickerHeaderKey = "x-gateway-destination-endpoint"
```

internal/extproc/server_test.go
```diff
@@ -316,8 +316,9 @@ func TestServer_setBackend(t *testing.T) {
 			str, err := prototext.Marshal(tc.md)
 			require.NoError(t, err)
 			s, _ := requireNewServerWithMockProcessor(t)
-			s.config.backends = map[string]*processorConfigBackend{"openai": {}}
-			err = s.setBackend(t.Context(), nil, "aaaaaaaaaaaa", &extprocv3.ProcessingRequest{
+			s.config.backends = map[string]*processorConfigBackend{"openai": {b: &filterapi.Backend{Name: "openai"}}}
+			mockProc := &mockProcessor{}
+			err = s.setBackend(t.Context(), mockProc, "aaaaaaaaaaaa", false, &extprocv3.ProcessingRequest{
```
mathetake: can you modify the test here so that we can both test false and true?

internal/extproc/chatcompletion_processor.go
```diff
@@ -336,13 +337,17 @@ func (c *chatCompletionProcessorUpstreamFilter) SetBackend(ctx context.Context,
...
 	rp, ok := routeProcessor.(*chatCompletionProcessorRouterFilter)
 	if !ok {
 		panic("BUG: expected routeProcessor to be of type *chatCompletionProcessorRouterFilter")
 	}
 	rp.upstreamFilterCount++
 	c.metrics.SetBackend(b)
 	c.modelNameOverride = b.ModelNameOverride
+	if c.modelNameOverride == "" && isEndpointPicker {
+		c.modelNameOverride = c.requestHeaders[c.config.modelNameHeaderKey]
+	}
```
mathetake: i think you don't need this one. modelNameOverride is optional

internal/extensionserver/post_cluster_modify.go
```diff
@@ -0,0 +1,87 @@
...
+		// No backend extension resources, skip modification and return cluster as-is.
+		return &egextension.PostClusterModifyResponse{Cluster: req.Cluster}, nil
+	}
+
+	// Parse InferencePools resources from BackendExtensionResources.
+	// BackendExtensionResources contains unstructured Kubernetes resources that were
+	// referenced in the AIGatewayRoute's BackendRefs with non-empty Group and Kind fields.
+	// If we found an InferencePool, configure the cluster for ORIGINAL_DST.
+	if inferencePools := s.constructInferencePoolsFrom(req.PostClusterContext.BackendExtensionResources); inferencePools != nil {
+		s.handleInferencePoolCluster(req.Cluster, inferencePools[0])
```
mathetake, line 52: ~why are we handling only the first pool? I think cluster metadata should be different among them. Otherwise, they are all sharing the same "backend name" in the extproc and the metrics are conflated. I think there's no harm to create original_dst for each of them.~


edited: i see this cluster is already per-route-rule so already one inference pool hah.

internal/extensionserver/post_route_modify.go
```diff
@@ -0,0 +1,44 @@
...
+		// No backend extension resources, skip.
+		return &egextension.PostRouteModifyResponse{Route: req.Route}, nil
+	}
+
+	// Parse InferencePool resources from BackendExtensionResources.
+	inferencePools := s.constructInferencePoolsFrom(req.PostRouteContext.ExtensionResources)
+
+	// If we found an InferencePool, configure the route with the ext_proc per-route config.
+	if inferencePools != nil {
+		req.Route.GetRoute().HostRewriteSpecifier = &routev3.RouteAction_AutoHostRewrite{
```
mathetake, line 41: can you put a comment on why the host rewrite disabled? I remember you encountered a weird issue but putting comment here would be helpful as a future reference.

internal/extensionserver/post_route_modify.go
```diff
@@ -0,0 +1,44 @@
...
+			AutoHostRewrite: wrapperspb.Bool(false),
+		}
+		if req.Route.TypedPerFilterConfig == nil {
+			req.Route.TypedPerFilterConfig = make(map[string]*anypb.Any)
+		}
+		buildEPPMetadataForRoute(req.Route, inferencePools[0])
```
mathetake, line 47: is this because we limit a inferencepool per route rule at AIGatewayRoute level right?

internal/extensionserver/post_translate_modify.go
```diff
@@ -0,0 +1,491 @@
...
+		upstreamCodec.Name = "envoy.filters.http.upstream_codec"
+		upstreamCodec.ConfigType = &httpconnectionmanagerv3.HttpFilter_TypedConfig{
+			TypedConfig: mustToAny(&upstream_codecv3.UpstreamCodec{}),
+		}
+		po.HttpFilters = append(po.HttpFilters, upstreamCodec)
+	}
+	cluster.TypedExtensionProtocolOptions[httpProtocolOptions] = mustToAny(po)
+}
+
+func (s *Server) maybeModifyListenerAndRoutes(listeners []*listenerv3.Listener, routes []*routev3.RouteConfiguration) {
```
mathetake, line 350: this function is a little bit hard to follow without the general idea of what it's doing. Could you add the function comment here on what it does?

internal/controller/ai_gateway_route.go
```diff
@@ -233,16 +233,33 @@ func (c *AIGatewayRouteController) newHTTPRoute(ctx context.Context, dst *gwapiv
...
+							Group:     (*gwapiv1.Group)(br.Group),
+							Kind:      (*gwapiv1.Kind)(br.Kind),
+							Name:      gwapiv1.ObjectName(br.Name),
+							Namespace: (*gwapiv1.Namespace)(&aiGatewayRoute.Namespace),
+						},
+						Weight: br.Weight,
+					}},
+				)
+			} else {
+				// Handle AIServiceBackend reference (existing logic).
```
mathetake:
```suggestion
				// Handle AIServiceBackend reference.
```

internal/controller/gateway.go
```diff
@@ -243,18 +243,24 @@ func (c *GatewayController) reconcileFilterConfigSecret(ctx context.Context, gw
...
-				backendObj, err = c.backend(ctx, aiGatewayRoute.Namespace, backendRef.Name)
-				if err != nil {
-					return fmt.Errorf("failed to get AIServiceBackend %s: %w", b.Name, err)
-				}
-				b.Schema = schemaToFilterAPI(backendObj.Spec.APISchema)
-				if bspRef := backendObj.Spec.BackendSecurityPolicyRef; bspRef != nil {
-					b.Auth, err = c.bspToFilterAPIBackendAuth(ctx, aiGatewayRoute.Namespace, string(bspRef.Name))
+				if backendRef.IsInferencePool() {
+					schema := aiGatewayRoute.Spec.APISchema
+					b.Schema = filterapi.VersionedAPISchema{Name: filterapi.APISchemaName(schema.Name), Version: ptr.Deref(schema.Version, "v1")}
```
mathetake, line 249:
```suggestion
					// We assume that InferencePools are all OpenAI schema.
					b.Schema = filterapi.VersionedAPISchema{Name: filterapi.APISchemaName(schema.Name), Version: ptr.Deref(schema.Version, "v1")}
```

internal/extensionserver/post_cluster_modify.go
mathetake: can you put an assertion on the length of inference poole?
```suggestion
		if len(inferencePools) != 1 {
			panic("BUG: at most one inferencepool can be referenced per route rule")
		} 
		s.handleInferencePoolCluster(req.Cluster, inferencePools[0])
```

internal/extensionserver/post_route_modify.go
mathetake: Let's put an assertion here
```suggestion
		if len(inferencePools) != 1 {
			panic("BUG: at most one inferencepool can be referenced per route rule")
		}
		buildEPPMetadataForRoute(req.Route, inferencePools[0])
```

internal/extensionserver/post_translate_modify.go
```diff
@@ -0,0 +1,491 @@
...
+			continue
+		}
+		routeConfigName := findListenerRouteConfig(listener)
+		listenerRouteMatrix[listener.Name] = routeConfigName
+		listenerMatrix[listener.Name] = listener
+	}
+
+	// inferencePoolRoutes builds a matrix of route configs and the inference pools they use.
+	routeCfgMatrix := make(map[string]*routev3.RouteConfiguration)
+	inferencePoolRoutes := make(map[string]map[string]*gwaiev1a2.InferencePool)
```
mathetake:
```suggestion
	routeNameToVHRouteNameToInferencePool := make(map[string]map[string]*gwaiev1a2.InferencePool)
```

internal/extensionserver/post_translate_modify.go
```diff
@@ -0,0 +1,491 @@
...
+		upstreamCodec.ConfigType = &httpconnectionmanagerv3.HttpFilter_TypedConfig{
+			TypedConfig: mustToAny(&upstream_codecv3.UpstreamCodec{}),
+		}
+		po.HttpFilters = append(po.HttpFilters, upstreamCodec)
+	}
+	cluster.TypedExtensionProtocolOptions[httpProtocolOptions] = mustToAny(po)
+}
+
+func (s *Server) maybeModifyListenerAndRoutes(listeners []*listenerv3.Listener, routes []*routev3.RouteConfiguration) {
+	listenerRouteMatrix := make(map[string]string)
```
mathetake: can you avoid using "matrix" in the variable name but instead be explicit like 
```suggestion
	listenerNameToRouteName := make(map[string]string)
```

internal/extensionserver/post_translate_modify.go
```diff
@@ -0,0 +1,491 @@
...
+	listenerMatrix := make(map[string]*listenerv3.Listener)
```
mathetake:
```suggestion
	listenerNameToListener := make(map[string]*listenerv3.Listener)
```

internal/extensionserver/post_translate_modify.go
```diff
@@ -0,0 +1,491 @@
...
+		if strings.HasPrefix(listener.Name, "envoy-gateway") {
+			continue
+		}
+		routeConfigName := findListenerRouteConfig(listener)
+		listenerRouteMatrix[listener.Name] = routeConfigName
+		listenerMatrix[listener.Name] = listener
+	}
+
+	// inferencePoolRoutes builds a matrix of route configs and the inference pools they use.
+	routeCfgMatrix := make(map[string]*routev3.RouteConfiguration)
```
mathetake:
```suggestion
	routeNameToRoute := make(map[string]*routev3.RouteConfiguration)
```

internal/extensionserver/post_translate_modify.go
```diff
@@ -0,0 +1,491 @@
...
+						inferencePoolRoutes[routeCfg.Name] = make(map[string]*gwaiev1a2.InferencePool)
+					}
+					inferencePoolRoutes[routeCfg.Name][route.Name] = pool
+				}
+			}
+		}
+	}
+
+	// listenerPoolMatrix builds a matrix of listeners and the inference pools they use.
+	listenerPoolMatrix := make(map[string][]*gwaiev1a2.InferencePool)
```
mathetake:
```suggestion
	listenerToInferencePools := make(map[string][]*gwaiev1a2.InferencePool)
```

internal/extensionserver/post_translate_modify.go
```diff
@@ -0,0 +1,491 @@
...
+
+// patchListenerWithInferencePoolFilters adds the necessary HTTP filters to the listener to support InferencePool backends.
+func (s *Server) patchListenerWithInferencePoolFilters(listener *listenerv3.Listener, inferencePools []*gwaiev1a2.InferencePool) {
+	// First, get the filter chains from the listener.
+	filterChains := listener.GetFilterChains()
+	defaultFC := listener.DefaultFilterChain
+	if defaultFC != nil {
+		filterChains = append(filterChains, defaultFC)
+	}
+	// Go over all of the chains, and add the dummy inference pool http filter.
```
mathetake: this comment seems old and obsolete? could you update?

internal/extensionserver/post_translate_modify.go
Xunzhuo: done

internal/extensionserver/post_route_modify.go
Xunzhuo: done

internal/extensionserver/post_cluster_modify.go
Xunzhuo: done

api/v1alpha1/ai_gateway_route_helper.go
Xunzhuo: done

internal/internalapi/internalapi.go
Xunzhuo: done

internal/internalapi/internalapi.go
Xunzhuo: done

internal/extproc/server_test.go
Xunzhuo: done

Xunzhuo: done

internal/extensionserver/post_route_modify.go
Xunzhuo: done

internal/controller/ai_gateway_route.go
Xunzhuo: done

internal/controller/gateway.go
Xunzhuo: done

internal/extensionserver/post_translate_modify.go
Xunzhuo: done

internal/extensionserver/post_translate_modify.go
Xunzhuo: done

internal/extensionserver/post_translate_modify.go
Xunzhuo: done

internal/extensionserver/post_translate_modify.go
Xunzhuo: done

internal/extensionserver/post_translate_modify.go
Xunzhuo: done

internal/extensionserver/post_translate_modify.go
Xunzhuo: done

internal/extproc/chatcompletion_processor.go
Xunzhuo: this is only left one review i dont make a change. the reason is when i have tested inferencepool with no model override, i got a 400 `no Content-Length header, using request body length` from the testupstream. the reason i think is the route level epp extproc removed the content-length header, so set modelNameOverride explicitly when it is empty can trigger the content-length header addition in upstream level filter processing header. which prevents the 400 from the testupstream

api/v1alpha1/ai_gateway_route.go
```diff
@@ -277,17 +286,42 @@ type AIGatewayRouteRule struct {
...
+	// +kubebuilder:validation:MaxLength=63
+	// +kubebuilder:validation:Pattern=`^$|^[a-zA-Z]([-a-zA-Z0-9]*[a-zA-Z0-9])?$`
+	Kind *string `json:"kind,omitempty"`
+
 	// Name of the model in the backend. If provided this will override the name provided in the request.
+	// This field is ignored when referencing InferencePool resources.
 	ModelNameOverride string `json:"modelNameOverride,omitempty"`
 
-	// Weight is the weight of the AIServiceBackend. This is exactly the same as the weight in
+	// Weight is the weight of the backend. This is exactly the same as the weight in
```
yuzisun, line 324: I think we need to document that priority and weight does not apply to InferencePool, might need some validations

Xunzhuo: Yes, that is already documented in design doc, and also we don't allow multiple backendRef when having inferencePool in an rule, so priority wouldn't work.

internal/extproc/chatcompletion_processor.go
yuzisun: Why does epp extproc remove the content length header if it is set? that does not sound right.

internal/extensionserver/post_cluster_modify.go
```diff
@@ -0,0 +1,96 @@
...
+func (s *Server) handleInferencePoolCluster(cluster *clusterv3.Cluster, inferencePool *gwaiev1a2.InferencePool) {
+	// Configure cluster for ORIGINAL_DST with header-based load balancing.
+	// ORIGINAL_DST type allows Envoy to route to destinations specified in HTTP headers.
+	cluster.ClusterDiscoveryType = &clusterv3.Cluster_Type{Type: clusterv3.Cluster_ORIGINAL_DST}
+
+	// CLUSTER_PROVIDED load balancing policy is required for ORIGINAL_DST clusters.
+	cluster.LbPolicy = clusterv3.Cluster_CLUSTER_PROVIDED
+
+	// Set a reasonable connection timeout. This is quite long to accommodate AI workloads.
+	cluster.ConnectTimeout = durationpb.New(1000 * time.Second)
```
yuzisun: 1000 seconds is too long for a connection to be established .

Xunzhuo: Good catch, updated with 10s

yuzisun:
@Xunzhuo @mathetake could you hold off the merging, we will run our integration testing suite today to make sure nothing is breaking.

Xunzhuo:
@yuzisun sure, how long will it take? Can you send a ping when it's done : )

internal/extproc/chatcompletion_processor.go
mathetake: yeah that doesn't seem right, and having this solely for content-length header workaround is not the right way to at least. The reason it's missing content-length header is that the upstream AI Gateawy Filter uses REPLACE_AND_CONTINUE option at header mutation phase. That's why we are having header mutation filter after the ai gateway upstream filter (see https://github.com/envoyproxy/ai-gateway/pull/818). so could you remove this line and add the header mutating filter like this https://github.com/envoyproxy/ai-gateway/blob/45c18d3dda0167a09690b142344eedc0e7a4b784/internal/extensionserver/extensionserver.go#L276-L297

internal/extensionserver/post_cluster_modify.go
```diff
@@ -0,0 +1,96 @@
...
+	egextension "github.com/envoyproxy/gateway/proto/extension"
+	clusterv3 "github.com/envoyproxy/go-control-plane/envoy/config/cluster/v3"
+	"google.golang.org/protobuf/types/known/durationpb"
+	gwaiev1a2 "sigs.k8s.io/gateway-api-inference-extension/api/v1alpha2"
+
+	"github.com/envoyproxy/ai-gateway/internal/internalapi"
+)
+
+// ClusterRefInferencePool generates a unique reference for an InferencePool cluster.
+func ClusterRefInferencePool(namespace, name, serviceName string, servicePort uint32) string {
```
mathetake:
```suggestion
// clusterRefInferencePool generates a unique reference for an InferencePool cluster.
func clusterRefInferencePool(namespace, name, serviceName string, servicePort uint32) string {
```

internal/extproc/chatcompletion_processor.go
```diff
@@ -352,6 +356,9 @@ func (c *chatCompletionProcessorUpstreamFilter) SetBackend(ctx context.Context,
 	c.originalRequestBodyRaw = rp.originalRequestBodyRaw
 	c.onRetry = rp.upstreamFilterCount > 1
 	c.stream = c.originalRequestBody.Stream
+	if isEndpointPicker {
+		c.logger.Debug("selected backend", slog.String("picked_endpoint", pickedEndpoint), slog.String("backendName", b.Name), slog.String("modelNameOverride", c.modelNameOverride), slog.String("originalRequestBodyRaw", string(c.originalRequestBodyRaw)))
```
mathetake: this line is crazy costly implicitly. the expression evaluation happens before the logging level check inside logger.Debug. so making string out of `string(c.originalRequestBodyRaw)` happens regardless of the log level. maybe you might want to delete this line or use `Enabled` to avoid the evaluation. Either way i don't think you would need the body here as it will be logged if the log level is set to debug somwhere in server.go
```suggestion
		if c.logger.Enabled(ctx, slog.LevelDebug) { 
				c.logger.Debug("selected backend", slog.String("picked_endpoint", pickedEndpoint), slog.String("backendName", b.Name), slog.String("modelNameOverride", c.modelNameOverride))
		}
```

internal/extensionserver/post_cluster_modify.go
Xunzhuo: done

internal/extproc/chatcompletion_processor.go
Xunzhuo: done

internal/extproc/chatcompletion_processor.go
Xunzhuo: done, rebased with main

mathetake:
LGTM and let's wait for the extensive tests by Dan's team before landing... exciting!!

examples/inference-pool/base.yaml
```diff
@@ -0,0 +1,212 @@
...
+            - "json"
+            - -grpcPort
+            - "9002"
+            - -grpcHealthPort
+            - "9003"
+          ports:
+            - containerPort: 9002
+            - containerPort: 9003
+        - name: metrics
+          containerPort: 9090
```
yuzisun: containerPort needs to be under ports

examples/inference-pool/base.yaml
```diff
@@ -0,0 +1,212 @@
...
+            - --zap-encoder
+            - "json"
+            - -grpcPort
+            - "9002"
+            - -grpcHealthPort
+            - "9003"
+          ports:
+            - containerPort: 9002
+            - containerPort: 9003
+        - name: metrics
```
yuzisun: metrics container is missing image?

Xunzhuo: there is no metrics container here, this is under epp container

examples/inference-pool/base.yaml
Xunzhuo: it is under ports just with a name here

Xunzhuo: https://pkg.go.dev/k8s.io/api/core/v1#ContainerPort

examples/inference-pool/base.yaml
Xunzhuo: the epp deployment is from https://raw.githubusercontent.com/kubernetes-sigs/gateway-api-inference-extension/main/config/manifests/inferencepool-resources.yaml

yuzisun: I think the metric is another port? the indentation is wrong then

examples/inference-pool/base.yaml
yuzisun:
```suggestion
          ports:
            - containerPort: 9002
            - containerPort: 9003
            - name: metrics
               containerPort: 9090
```

Xunzhuo: done

yuzisun (approved):

PR #831, opened 2025-07-07
controller: ensure eg rollout when deployed as daemonset

yuzisun:
**Description**
This PR handles the rollout for envoy gateway during ai gateway extproc upgrade when deployed as daemonset.

Related Issues/PRs (if applicable)
Related PR: https://github.com/envoyproxy/ai-gateway/pull/699

### DIFF
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -31,18 +31,17 @@ import (
 )
 
 type flags struct {
-	extProcLogLevel                 string
-	extProcImage                    string
-	extProcImagePullPolicy          corev1.PullPolicy
-	enableLeaderElection            bool
-	logLevel                        zapcore.Level
-	extensionServerPort             string
-	tlsCertDir                      string
-	tlsCertName                     string
-	tlsKeyName                      string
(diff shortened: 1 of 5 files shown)


cmd/controller/main.go
```diff
@@ -114,6 +115,11 @@ func parseAndValidateFlags(args []string) (flags, error) {
 		"envoy-gateway-system",
 		"The namespace where the Envoy Gateway system components are installed.",
 	)
+	enableEnvoyGatewayDaemonSetMode := fs.Bool(
+		"enableEnvoyGatewayDaemonSetMode",
```
mathetake: what do you think about instead of explicitly specifying the daemonset, simply adding annotation to both deoloyment and daemonset with the matching `gateway.envoyproxy.io/owning-gateway-name` label? I think it's simpler

yuzisun: actually I think it might be already there, so you are suggesting removing the flag all together and check both deployment and daemonset by this label right?

mathetake: yeah

PR #837, opened 2025-07-07
controller: return 404 instead of 500 for no matching

mathetake:
**Description**

Before #793, the case where no matching route found was handled in the extproc and the 404 immediate response was returned from there, but after that, it naturally results in the "unreachable" default route and swallowed the indication of no matching and it made it impossible to reason about the 500 error on that case. In other words, this fixes the regression in #793 to return the proper 404 response.


internal/controller/ai_gateway_route.go
```diff
@@ -206,20 +229,18 @@ func (c *AIGatewayRouteController) newHTTPRoute(ctx context.Context, dst *gwapiv
 		})
 	}
 
-	// Adds the default route rule with "/" path. This is necessary because Envoy's router selects the backend
-	// before entering the filters. So, all requests would result in a 404 if there is no default route. In practice,
-	// this default route is not used because our AI Gateway filters is the one who actually calculates the route based
-	// on the given Rules. If it doesn't match any backend, 404 will be returned from the AI Gateway filter as an immediate
-	// response.
```
mathetake, line 213: note: this was no longer valid after #793 hence this fixes it

missBerg (approved):

PR #838, opened 2025-07-08
feat: add gcp anthropic basic translation

alexagriffith:
**Description**

add basic gcp anthropic translation support 

still need to do:
More robust tooling/edge case/non-basic parameter translations
Streaming suport


internal/apischema/openai/openai.go
```diff
@@ -463,6 +479,11 @@ type ChatCompletionRequest struct {
 	// refs: https://platform.openai.com/docs/api-reference/chat/create#chat-create-max_tokens
 	MaxTokens *int64 `json:"max_tokens,omitempty"` //nolint:tagliatelle //follow openai api
```
xiaolin593: `MaxTokens` is marked as deprecated in OpenAI doc. Can you update the comment to indicate that?

internal/apischema/openai/openai.go
```diff
@@ -519,7 +546,7 @@ type ChatCompletionRequest struct {
 
 	// ParallelToolCalls enables multiple tools to be returned by the model.
 	// Docs: https://platform.openai.com/docs/guides/function-calling/parallel-function-calling
-	ParallelToolCalls bool `json:"parallel_tool_calls,omitempty"` //nolint:tagliatelle //follow openai api
+	ParallelToolCalls *bool `json:"parallel_tool_calls,omitempty"` //nolint:tagliatelle //follow openai api
```
xiaolin593: why not just scalar type?

internal/extproc/translator/openai_gcpanthropic.go
```diff
@@ -6,59 +6,692 @@
...
 )
 
-// NewChatCompletionOpenAIToGCPAnthropicTranslator implements [Factory] for OpenAI to GCP Anthropic translation.
-// This translator converts OpenAI ChatCompletion API requests to GCP Anthropic API format.
-func NewChatCompletionOpenAIToGCPAnthropicTranslator() OpenAIChatCompletionTranslator {
-	return &openAIToGCPAnthropicTranslatorV1ChatCompletion{}
+// currently a requirement for GCP Vertex / Anthropic API https://docs.anthropic.com/en/api/claude-on-vertex-ai
+const (
+	anthropicVersionKey   = "anthropic_version"
+	anthropicVersionValue = "vertex-2023-10-16"
```
xiaolin593: Make the value configurable?

internal/extproc/translator/openai_gcpanthropic.go
```diff
@@ -6,59 +6,692 @@
...
+	gcpBackendError       = "GCPBackendError"
+	defaultMaxTokens      = int64(100)
+	tempNotSupportedError = "temperature %.2f is not supported by Anthropic (must be between 0.0 and 1.0)"
```
xiaolin593, line 33: Drop the variable and make error in-place.

internal/extproc/translator/openai_gcpanthropic.go
```diff
@@ -6,59 +6,692 @@
...
+
+// AnthropicContent Anthropic request/response structs.
+type AnthropicContent struct {
+	Type   string                            `json:"type"`
+	Text   string                            `json:"text"`
+	Source *anthropic.Base64ImageSourceParam `json:"source,omitempty"`
+}
+
+// NewChatCompletionOpenAIToAnthropicTranslator creates a new translator.
+func NewChatCompletionOpenAIToAnthropicTranslator() OpenAIChatCompletionTranslator {
```
xiaolin593: Is this conversion generic and apply to both GCP and AWS Anthropic? File name `openai_gcpanthropic.go` indicates it only applies to GCP?

internal/extproc/translator/openai_gcpanthropic.go
```diff
@@ -6,59 +6,692 @@
...
+// currently a requirement for GCP Vertex / Anthropic API https://docs.anthropic.com/en/api/claude-on-vertex-ai
+const (
+	anthropicVersionKey   = "anthropic_version"
+	anthropicVersionValue = "vertex-2023-10-16"
+	gcpBackendError       = "GCPBackendError"
+	defaultMaxTokens      = int64(100)
+	tempNotSupportedError = "temperature %.2f is not supported by Anthropic (must be between 0.0 and 1.0)"
+)
+
+var errStreamingNotSupported = errors.New("streaming is not yet supported for GCP Anthropic translation")
```
xiaolin593, line 36: move to in-place as it only references once.

mathetake:
@alexagriffith could you resolve the conflict you had with @sukumargaonkar's PR?

internal/extproc/translator/openai_gcpanthropic.go
```diff
@@ -6,59 +6,692 @@
...
+// NewChatCompletionOpenAIToAnthropicTranslator creates a new translator.
+func NewChatCompletionOpenAIToAnthropicTranslator() OpenAIChatCompletionTranslator {
+	return &openAIToAnthropicTranslatorV1ChatCompletion{}
+}
+
+func anthropicToOpenAIFinishReason(stopReason anthropic.StopReason) (openai.ChatCompletionChoicesFinishReason, error) {
+	switch stopReason {
+	// The most common stop reason. Indicates Claude finished its response naturally.
+	// or Claude encountered one of your custom stop sequences.
+	// TODO: A better way to return pause_turning
```
xiaolin593: nit: pause_turn

internal/extproc/translator/openai_gcpanthropic.go
```diff
@@ -6,59 +6,692 @@
...
+		return openai.ChatCompletionChoicesFinishReasonContentFilter, nil
+	default:
+		return "", fmt.Errorf("received invalid stop reason %v", stopReason)
+	}
+}
+
+// validateTemperatureForAnthropic checks if the temperature is within Anthropic's supported range (0.0 to 1.0).
+// Returns an error if the value is greater than 1.0.
+func validateTemperatureForAnthropic(temp *float64) error {
+	if temp != nil && *temp > 1.0 {
```
xiaolin593:
```suggestion
	if temp != nil && (*temp < 0.0 || *temp > 1.0) {
```

internal/extproc/translator/openai_gcpanthropic.go
```diff
@@ -6,59 +6,692 @@
...
+			var inputSchema map[string]interface{}
+			if openAITool.Function.Parameters != nil {
+				// Directly assert the 'any' type to the expected map structure.
+				schema, ok := openAITool.Function.Parameters.(map[string]interface{})
+				if !ok {
+					err = fmt.Errorf("tool parameters for '%s' are not a valid JSON object", openAITool.Function.Name)
+					return
+				}
+				inputSchema = schema
+			}
```
xiaolin593: We should perform a JSON Schema validation check for `input_schema` block. 

-  https://docs.anthropic.com/en/docs/agents-and-tools/tool-use/implement-tool-use#example-simple-tool-definition

- https://json-schema.org/

internal/extproc/translator/openai_gcpanthropic.go
```diff
@@ -6,59 +6,692 @@
...
+				imageURL := contentPart.ImageContent.ImageURL.URL
+				switch {
+				case isDataURI(imageURL):
+					contentType, data, err := parseDataURI(imageURL)
+					if err != nil {
+						return nil, fmt.Errorf("failed to parse image URL: %w", err)
+					}
+					base64Data := base64.StdEncoding.EncodeToString(data)
+					appPDF := string(constant.ValueOf[constant.ApplicationPDF]())
+					switch contentType {
```
xiaolin593: There are 5 nested loops(`switch` and `for`) up to this point and it reduces code readability. 
Can you wrap some of them into separate function to keep conversion clean?

i.e. from line 217 to 236, give a function name `convertImageContentToAnthropic`, and I think it will make each of them unit testable.

internal/extproc/translator/openai_gcpanthropic.go
```diff
@@ -6,59 +6,692 @@
...
+		// Concatenate all text parts for completeness.
+		var sb strings.Builder
+		for _, part := range v {
+			if part.TextContent != nil {
+				sb.WriteString(part.TextContent.Text)
+			}
+		}
+		return sb.String()
+	case nil:
+		return ""
```
xiaolin593: nit: move `nil` check earlier in switch.

internal/extproc/translator/openai_gcpanthropic.go
```diff
@@ -6,59 +6,692 @@
...
+			if param, ok := msg.Value.(openai.ChatCompletionDeveloperMessageParam); ok {
+				systemBlocks = append(systemBlocks, anthropic.TextBlockParam{Text: extractSystemPromptFromDeveloperMsg(param)})
+			}
+		case openai.ChatMessageRoleUser:
+			message := msg.Value.(openai.ChatCompletionUserMessageParam)
+			var content []anthropic.ContentBlockParamUnion
+			content, err = openAIToAnthropicContent(message.Content.Value)
+			if err != nil {
+				return
+			}
```
xiaolin593, line 368: Are we suppressing the error? Can we do `return (nil, nil, err)`. Same for below.

internal/extproc/translator/openai_gcpanthropic.go
```diff
@@ -6,59 +6,692 @@
...
+	} else if openAIReq.MaxTokens != nil {
+		maxTokens = *openAIReq.MaxTokens
+	}
+
+	// Translate openAI contents to anthropic params.
+	// 2. Translate messages and system prompts.
+	messages, systemBlocks, err := openAIToAnthropicMessages(openAIReq.Messages)
+	if err != nil {
+		return
+	}
```
xiaolin593, line 436: bubble up the error? same as below.

internal/extproc/translator/openai_gcpanthropic_test.go
```diff
@@ -7,273 +7,1105 @@ package translator
...
+		translator := NewChatCompletionOpenAIToAnthropicTranslator()
+		_, bm, err := translator.RequestBody(nil, multiSystemReq, false)
+		require.NoError(t, err)
+		body := bm.GetBody()
+		require.Equal(t, firstMsg, gjson.GetBytes(body, "system.0.text").String())
+		require.Equal(t, secondMsg, gjson.GetBytes(body, "system.1.text").String())
+		require.Equal(t, thirdMsg, gjson.GetBytes(body, "messages.0.content.0.text").String())
+	})
+
+	t.Run("Streaming Request Error", func(t *testing.T) {
```
xiaolin593, line 144: Is this because not supported yet?

internal/apischema/openai/openai.go
alexagriffith: because openai has it as optional , it can be nil

internal/extproc/translator/openai_gcpanthropic.go
alexagriffith: its not configurable for gcp

internal/extproc/translator/openai_gcpanthropic.go
alexagriffith: we use it in testing too

internal/extproc/translator/openai_gcpanthropic.go
alexagriffith: if its 0.0 its okay actually

alexagriffith: technically a valid value

internal/apischema/openai/openai.go
alexagriffith: see the line above, it is already noted

internal/extproc/translator/openai_gcpanthropic.go
alexagriffith: https://docs.anthropic.com/en/api/claude-on-vertex-ai

alexagriffith: https://github.com/anthropics/anthropic-sdk-go/blob/main/vertex/vertex.go#L20

internal/extproc/translator/openai_gcpanthropic.go
alexagriffith: id rather keep it since we use it in the test, that ok?

internal/extproc/translator/openai_gcpanthropic.go
alexagriffith: the err is defined in the output https://go.dev/tour/basics/7

internal/extproc/translator/openai_gcpanthropic.go
xiaolin593: We allow user to configure model version. 

https://github.com/envoyproxy/ai-gateway/blob/7d293fa14a414d048de11f2599442361716137c3/filterapi/filterconfig.go#L98

https://github.com/envoyproxy/ai-gateway/blob/7d293fa14a414d048de11f2599442361716137c3/examples/basic/basic.yaml#L120

Looks like bind AIGW to `anthropic-sdk-go` SDK, in that case, can we directly reference the on in SDK instead of creating a new one?

internal/extproc/translator/openai_gcpanthropic.go
xiaolin593: not sure if this is standard, in the past the error msg was suggested to put directly in where it references. I'll leave it up to @envoyproxy/ai-gateway-maintainers to check code.

internal/extproc/translator/openai_gcpanthropic.go
xiaolin593: how about negative value -2.5 ?

internal/extproc/translator/openai_gcpanthropic.go
alexagriffith: this is different than model version, for gcp this is the only allowed value (see first link I sent) but sure I can use the anthropic sdk const

internal/extproc/translator/openai_gcpanthropic.go
alexagriffith: we wanted to make it generic but then we kinda reverted to not over engineer it yet, i can make naming specific again

xiaolin593: or drop the `gcp` from file name if that's the intention to make it generic?

alexagriffith: well right now it is specific so I kinda agree with you we can keep gcp for now. its a small lift to change it if we need

alexagriffith: done

internal/extproc/translator/openai_gcpanthropic.go
alexagriffith: done!

internal/extproc/translator/openai_gcpanthropic.go
alexagriffith: added! good catch btw didn't notice anthropic sdk exported the const

internal/extproc/translator/openai_gcpanthropic.go
alexagriffith: good idea! made two new helper methods - this logic was already covered by unit tests though, lmk what you think

alexagriffith:
> @alexagriffith could you resolve the conflict you had with @sukumargaonkar's PR?

done! thank you

internal/extproc/translator/openai_gcpanthropic.go
alexagriffith: thanks done

internal/extproc/translator/openai_gcpanthropic.go
alexagriffith: ah my bad, I read it wrong. You are right

alexagriffith: good catch

alexagriffith: updated and added a test

internal/extproc/translator/openai_gcpanthropic.go
alexagriffith: done, can you check it again? made a validate helper

.gitignore
```diff
@@ -36,3 +36,7 @@ site/yarn-debug.log*
 site/yarn-error.log*
 site/static/.DS_Store
 site/temp
+
+/hack/
```
sukumargaonkar: probably added by mistake

alexagriffith: need to remove before merge

internal/extproc/translator/openai_gcpanthropic.go
```diff
@@ -6,59 +6,728 @@
...
+	return params, nil
+}
+
+// RequestBody implements [Translator.RequestBody] for GCP.
+func (o *openAIToGCPAnthropicTranslatorV1ChatCompletion) RequestBody(_ []byte, openAIReq *openai.ChatCompletionRequest, _ bool) (
+	*extprocv3.HeaderMutation, *extprocv3.BodyMutation, error,
+) {
+	params, err := buildAnthropicParams(openAIReq)
+	if err != nil {
+		return nil, nil, err
```
xiaolin593: just return instead of `nil, nil, err`?

internal/extproc/translator/openai_gcpanthropic.go
```diff
@@ -6,59 +6,728 @@
...
+	}
+
+	body, err := json.Marshal(params)
+	if err != nil {
+		return nil, nil, err
```
xiaolin593: same above.

xiaolin593 (commented):
Looks good to me now except a few minor comments. @mathetake please take a look.

internal/extproc/translator/util.go
```diff
@@ -36,3 +43,56 @@ func parseDataURI(uri string) (string, []byte, error) {
 	}
 	return contentType, bin, nil
 }
+
+// buildGCPRequestMutations creates header and body mutations for GCP requests
+// It sets the ":path" header, the "content-length" header and the request body.
+func buildGCPRequestMutations(path string, reqBody []byte) (*ext_procv3.HeaderMutation, *ext_procv3.BodyMutation) {
```
yuzisun: this mutation does not look like GCP specific, can we rename it? Also this needs a test

alexagriffith: can add - we do test the mutations but i can add separate util test for this if we want also

internal/extproc/translator/openai_gcpanthropic.go
alexagriffith: done

internal/extproc/translator/util.go
alexagriffith: util file has 95% test coverage - do you want me to make util_test file to test these? may not add coverage but it will be dedicated to these so makes sense

alexagriffith: added

internal/apischema/openai/openai.go
```diff
@@ -424,6 +424,22 @@ type ChatCompletionResponseFormatJSONSchema struct {
 	Strict      bool   `json:"strict"`
 }
 
+// Reasoning represents the reasoning options for o-series models.
+// Docs: https://platform.openai.com/docs/api-reference/responses/create#responses-create-reasoning
+type Reasoning struct {
```
yuzisun: let's move this change to a separate PR as it is not handling reasoning in the scope of this PR.

internal/extproc/translator/openai_gcpanthropic.go
```diff
@@ -6,59 +6,729 @@
...
 	"github.com/envoyproxy/ai-gateway/internal/apischema/openai"
 )
 
-// NewChatCompletionOpenAIToGCPAnthropicTranslator implements [Factory] for OpenAI to GCP Anthropic translation.
-// This translator converts OpenAI ChatCompletion API requests to GCP Anthropic API format.
+// currently a requirement for GCP Vertex / Anthropic API https://docs.anthropic.com/en/api/claude-on-vertex-ai
+const (
+	anthropicVersionKey   = "anthropic_version"
+	gcpBackendError       = "GCPBackendError"
+	defaultMaxTokens      = int64(100)
```
alexagriffith: thoughts on having a default?? @yuzisun

internal/apischema/openai/openai.go
alexagriffith: https://github.com/envoyproxy/ai-gateway/pull/868

internal/extproc/translator/openai_gcpanthropic.go
```diff
@@ -520,12 +500,18 @@ func buildAnthropicParams(openAIReq *openai.ChatCompletionRequest) (params *anth
...
 
 	// Handle stop sequences.
-	stopSequences, err := extractStopSequencesFromPtrSlice(openAIReq.Stop)
+	stopSequences, err := processStop(openAIReq.Stop)
 	if err != nil {
 		return &anthropic.MessageNewParams{}, err
 	}
 	if len(stopSequences) > 0 {
-		params.StopSequences = stopSequences
+		var stops []string
```
alexagriffith, line 469: maybe want to change this, i didn't realize gcp gemini & anthropic want non ptr slice while bedrock wants ptr slice

internal/extproc/translator/util.go
```diff
@@ -117,3 +119,14 @@ func processStop(data interface{}) ([]*string, error) {
 		return nil, fmt.Errorf("invalid type for stop parameter: expected string, []string, []*string, or nil, got %T", v)
 	}
 }
+
+// coalesce returns the value of the first non-nil pointer,
+// or the default value if all pointers are nil.
+func coalesce[T any](defaultValue T, ptrs ...*T) T {
```
alexagriffith: if we dont want default max tokens I can just use cmp.Or, but maxtokens is required value for anthropic

internal/extproc/translator/openai_gcpanthropic.go
```diff
@@ -6,59 +6,715 @@
...
+	tempNotSupportedError = "temperature %.2f is not supported by Anthropic (must be between 0.0 and 1.0)"
+)
+
+var errStreamingNotSupported = errors.New("streaming is not yet supported for GCP Anthropic translation")
+
+// openAIToGCPAnthropicTranslatorV1ChatCompletion where we can store information for streaming requests.
+type openAIToGCPAnthropicTranslatorV1ChatCompletion struct{}
+
+// Option defines a function that configures the translator.
+type Option func(*openAIToGCPAnthropicTranslatorV1ChatCompletion)
```
mathetake: let's remove this as it's not used. You can bring it back when it's needed

internal/extproc/translator/openai_gcpanthropic.go
```diff
@@ -6,59 +6,715 @@
...
+
+// AnthropicContent Anthropic request/response structs.
+type AnthropicContent struct {
+	Type   string                            `json:"type"`
+	Text   string                            `json:"text"`
+	Source *anthropic.Base64ImageSourceParam `json:"source,omitempty"`
+}
+
```
mathetake: also unused

internal/extproc/translator/openai_gcpanthropic.go
mathetake: i would prefer inline if it's only used in one place in main code + tests but not strong. you can keep it

internal/extproc/translator/openai_gcpanthropic.go
```diff
@@ -6,59 +6,715 @@
...
+// Returns an error if the value is greater than 1.0.
+func validateTemperatureForAnthropic(temp *float64) error {
+	if temp != nil && (*temp < 0.0 || *temp > 1.0) {
+		return fmt.Errorf(tempNotSupportedError, *temp)
+	}
+	return nil
+}
+
+func isDataURI(uri string) bool {
+	return strings.HasPrefix(uri, "data:")
```
mathetake: nit; can you move this inline as it's only used in once place + not for unit-testability where you also have `strings.HasSuffix(strings.ToLower(imageURL), ".pdf"):` so let's make it consistent

internal/extproc/translator/openai_gcpanthropic.go
```diff
@@ -6,59 +6,715 @@
...
+}
+
+func isSupportedImageMediaType(mediaType string) bool {
```
mathetake:
```suggestion
func isAnthropicSupportedImageMediaType(mediaType string) bool {
```

internal/extproc/translator/openai_gcpanthropic.go
```diff
@@ -6,59 +6,715 @@
...
+		anthropic.Base64ImageSourceMediaTypeImageGIF,
+		anthropic.Base64ImageSourceMediaTypeImageWebP:
+		return true
+	default:
+		return false
+	}
+}
+
+// validateFunctionParameters checks if the given parameters object is a valid JSON schema.
+func validateFunctionParameters(params any) error {
```
mathetake: This feels like doing something in a different layer that the to-Anthropic translation. This basically does only the validation of a pure OpenAI request input, and nothing Anthropic specific. I think my question would bel like are we supposed to do this kind of pure validation of OpenAI request input in the first place. I feel it should be fine without doing it as it will be validated by the provider anyways. On the other hand, if we want, I think we should consolidate these validation in a central place before entering translator subroutine outside this package.

This comment assumes that FunctionDefinition.Parameters must be JsonSchema for all providers. At least, it seems true for OpenAI and AWS Bedrock? 

This is non blocking comment but something we should discuss!

internal/extproc/translator/openai_gcpanthropic.go
```diff
@@ -6,59 +6,715 @@
...
+	}
+	if _, err = compiler.Compile(schemaJSON); err != nil {
+		return fmt.Errorf("invalid JSON schema for tool parameters: %w", err)
+	}
+
+	return nil
+}
+
+// translateToolChoice converts the OpenAI tool_choice parameter to the Anthropic format.
+func translateToolChoice(openAIToolChoice any, disableParallelToolUse anthropicParam.Opt[bool]) (anthropic.ToolChoiceUnionParam, error) {
```
mathetake:
```suggestion
func translateToolChoiceForAnthropic(openAIToolChoice any, disableParallelToolUse anthropicParam.Opt[bool]) (anthropic.ToolChoiceUnionParam, error) {
```

internal/extproc/translator/openai_gcpanthropic.go
```diff
@@ -6,59 +6,715 @@
...
+	return &anthropic.MessageParam{
+		Role:    anthropic.MessageParamRoleAssistant,
+		Content: contentBlocks,
+	}, nil
+}
+
+// openAIToAnthropicMessages converts OpenAI messages to Anthropic message params type, handling all roles and system/developer logic.
+func openAIToAnthropicMessages(openAIMsgs []openai.ChatCompletionMessageParamUnion) (anthropicMessages []anthropic.MessageParam, systemBlocks []anthropic.TextBlockParam, err error) {
+	for i := range openAIMsgs {
+		msg := openAIMsgs[i]
```
mathetake: This will be (really) slightly efficient since it won't copy the value

```suggestion
		msg := &openAIMsgs[i]
```

internal/extproc/translator/openai_gcpanthropic.go
```diff
@@ -6,59 +6,715 @@
...
+			anthropicMessages = append(anthropicMessages, anthropicMsg)
+		case openai.ChatMessageRoleAssistant:
+			assistantMessage := msg.Value.(openai.ChatCompletionAssistantMessageParam)
+
+			var messages *anthropic.MessageParam
+			messages, err = openAIMessageToAnthropicMessageRoleAssistant(&assistantMessage)
+			if err != nil {
+				return
+			}
+			anthropicMessages = append(anthropicMessages, *messages)
```
mathetake: if you make openAIMessageToAnthropicMessageRoleAssistant return a value instead of pointer, then we can avoid an allocation here. The change would look like:

```diff
diff --git a/internal/extproc/translator/openai_gcpanthropic.go b/internal/extproc/translator/openai_gcpanthropic.go
index 47eeb10..9f6ae50 100644
--- a/internal/extproc/translator/openai_gcpanthropic.go
+++ b/internal/extproc/translator/openai_gcpanthropic.go
@@ -344,7 +344,7 @@ func anthropicRoleToOpenAIRole(role anthropic.MessageParamRole) (string, error)
 
 // openAIMessageToAnthropicMessageRoleAssistant converts an OpenAI assistant message to Anthropic content blocks.
 // The tool_use content is appended to the Anthropic message content list if tool_calls are present.
-func openAIMessageToAnthropicMessageRoleAssistant(openAiMessage *openai.ChatCompletionAssistantMessageParam) (anthropicMsg *anthropic.MessageParam, err error) {
+func openAIMessageToAnthropicMessageRoleAssistant(openAiMessage *openai.ChatCompletionAssistantMessageParam) (anthropicMsg anthropic.MessageParam, err error) {
        contentBlocks := make([]anthropic.ContentBlockParamUnion, 0)
        if v, ok := openAiMessage.Content.Value.(string); ok && len(v) > 0 {
                contentBlocks = append(contentBlocks, anthropic.NewTextBlock(v))
@@ -381,7 +381,7 @@ func openAIMessageToAnthropicMessageRoleAssistant(openAiMessage *openai.ChatComp
                contentBlocks = append(contentBlocks, anthropic.ContentBlockParamUnion{OfToolUse: &toolUse})
        }
 
-       return &anthropic.MessageParam{
+       return anthropic.MessageParam{
                Role:    anthropic.MessageParamRoleAssistant,
                Content: contentBlocks,
        }, nil
@@ -416,12 +416,12 @@ func openAIToAnthropicMessages(openAIMsgs []openai.ChatCompletionMessageParamUni
                case openai.ChatMessageRoleAssistant:
                        assistantMessage := msg.Value.(openai.ChatCompletionAssistantMessageParam)
 
-                       var messages *anthropic.MessageParam
+                       var messages anthropic.MessageParam
                        messages, err = openAIMessageToAnthropicMessageRoleAssistant(&assistantMessage)
                        if err != nil {
                                return
                        }
-                       anthropicMessages = append(anthropicMessages, *messages)
+                       anthropicMessages = append(anthropicMessages, messages)
                case openai.ChatMessageRoleTool:
                        toolMsg := msg.Value.(openai.ChatCompletionToolMessageParam)
                        var content []anthropic.ContentBlockParamUnion
```

mathetake (commented):
left a few comments, but on the right track and almost there!

internal/extproc/translator/openai_gcpanthropic.go
alexagriffith: yeah will remove think it was left over from a refactor I did

internal/extproc/translator/openai_gcpanthropic.go
alexagriffith: I added this in per @xiaolin593 comment asking me to add it, @xiaolin593 would you like to add anything here?

internal/extproc/translator/openai_gcpanthropic.go
yuzisun: I think an explicit configuration might be preferred here if we want envoy ai gateway to manage the defaults due to the varying nature of LLM costs and usage patterns.

yuzisun: Let's remove the default here and open an issue for discussion.

internal/extproc/translator/openai_gcpanthropic.go
yuzisun: minimally we can move this to a common util file so later we can port to other translations.

mathetake: yeah I would prefer this to be removed if it seems a bit hard to extract and refactor later where the shared logic is, but not strong opinion. But the fact that this validation will be performed by the upstream makes this code path seem unnecessary to be honest. anyways at least, we can move this to a common util file as Dan suggested. it's you call

internal/extproc/translator/openai_gcpanthropic.go
alexagriffith: ok so we can remove it and instead we will return an error if its not set because its required in anthropic struct

alexagriffith: updated

internal/extproc/translator/openai_gcpanthropic.go
alexagriffith: yes, for some reason I thought we could use it across aws , gemini and this but it seems we all do it slightly differently, we may want to revisit this in future and see how we can simplify / be consistent

internal/extproc/translator/openai_gcpanthropic.go
alexagriffith: so it seems like @mathetake you are saying we probably shouldnt be handling this validation at all and @yuzisun you mention to move it to util so others can use it - is there a use case for other code now? if not I can remove right now if its doubling the validation

internal/extproc/translator/openai_gcpanthropic.go
alexagriffith: done

internal/extproc/translator/openai_gcpanthropic.go
alexagriffith: done

internal/extproc/translator/openai_gcpanthropic.go
alexagriffith: done

internal/extproc/translator/openai_gcpanthropic.go
alexagriffith: done

internal/extproc/translator/openai_gcpanthropic.go
alexagriffith: removed

internal/extproc/translator/openai_gcpanthropic.go
alexagriffith: done

internal/extproc/translator/openai_gcpanthropic.go
alexagriffith: done

internal/extproc/translator/openai_gcpanthropic.go
mathetake: yeah let's remove it for now +1  (if we do the pure validation, then we should do that on all fields, not limited to function call schema... that's my feeling right now)

alexagriffith: updated

.gitignore
alexagriffith: removed

internal/extproc/translator/openai_gcpanthropic.go
```diff
@@ -23,48 +45,641 @@ type openAIToGCPAnthropicTranslatorV1ChatCompletion struct {
...
+
 	modelName := openAIReq.Model
 	if o.modelNameOverride != "" {
 		// Use modelName override if set.
 		modelName = o.modelNameOverride
 	}
-	pathSuffix := buildGCPModelPathSuffix(GCPModelPublisherAnthropic, modelName, GCPMethodGenerateContent)
+	pathSuffix := buildGCPModelPathSuffix(GCPModelPublisherAnthropic, modelName, specifier)
+	// b. Set the "anthropic_version" key in the JSON body
+	// Using same logic as anthropic go SDK: https://github.com/anthropics/anthropic-sdk-go/blob/main/vertex/vertex.go#L78
```
mathetake: can you make this reference link to be pinned to a specific commit in the repo (permalink) instead of a branch? otherwise this will be soon invalid

internal/extproc/translator/openai_gcpanthropic.go
```diff
@@ -23,48 +45,641 @@ type openAIToGCPAnthropicTranslatorV1ChatCompletion struct {
...
+	body, _ = sjson.SetBytes(body, anthropicVersionKey, anthropicVertex.DefaultVersion)
```
mathetake, line 513: let me confirm if i understand this correctly: is this set via sjons because it's an anthropic specific field and not part of GCP/Vertex schema in apischema package? if so this makes sense

mathetake (commented):
almost LGTM with one more request: would you mind adding a few end to end test case in tests/extproc? especially around ResponseError error handling code path would be nice

internal/extproc/translator/openai_gcpanthropic.go
alexagriffith: yes, used the same logic as https://github.com/anthropics/anthropic-sdk-go/blob/main/vertex/vertex.go#L74

alexagriffith:
> almost LGTM with one more request: would you mind adding a few end to end test case in tests/extproc? especially around ResponseError error handling code path would be nice

yes, can that be in a separate PR and I will do it next? this one is getting large

internal/extproc/translator/openai_gcpanthropic.go
alexagriffith: this ok? https://github.com/anthropics/anthropic-sdk-go/blob/e252e284244755b2b2f6eef292b09d6d1e6cd989/bedrock/bedrock.go#L167

alexagriffith: good point!

mathetake (approved):
sure!

mathetake:
thanks for all the work and collaboration here!

Issue #701, opened 2025-06-10
test flake: TestStartConfigWatcher


mathetake:
https://github.com/envoyproxy/ai-gateway/actions/runs/15565072541/job/43827016738

```
--- FAIL: TestStartConfigWatcher (0.91s)
    watcher_test.go:172: 
        	Error Trace:	/home/runner/work/ai-gateway/ai-gateway/internal/extproc/watcher_test.go:172
        	Error:      	Not equal: 
        	            	expected: 3
        	            	actual  : 2
        	Test:       	TestStartConfigWatcher
FAIL
```

yuluo-yx:
pls assign to me. i can try

googs1025:
/assign 

I will try this 😄

googs1025:
@mathetake Sorry to bother you, if you can please give me some advice or guidance, thank you! 😄 

Before submitting PR, I want to try to understand this problem first. 🤔 

I am not sure whether the test tick is too short, which leads to the inaccurate judgment of `stat.ModTime().Sub(cw.lastMod) <= 0`?

codebase: 

```go
// loadConfig loads a new config from the given path and updates the Receiver by
// calling the [Receiver.Load].
func (cw *configWatcher) loadConfig(ctx context.Context) error {
	var cfg *filterapi.Config
	stat, err := os.Stat(cw.path)
	switch {
	case err != nil && os.IsNotExist(err):
		// If the file does not exist, do not fail (which could lead to the extproc process to terminate).
		// Instead, load the default configuration and keep running unconfigured.
		cfg = filterapi.MustLoadDefaultConfig()
	case err != nil:
		return err
	}

	if cfg != nil {
		if cw.usingDefaultCfg { // Do not re-reload the same thing on every tick.
			return nil
		}
		cw.l.Info("config file does not exist; loading default config", slog.String("path", cw.path))
		cw.lastMod = time.Now()
		cw.usingDefaultCfg = true
	} else {
		cw.usingDefaultCfg = false
		if stat.ModTime().Sub(cw.lastMod) <= 0 {
			return nil
		}
		cw.l.Info("loading a new config", slog.String("path", cw.path))
		cw.lastMod = stat.ModTime()
		cfg, err = filterapi.UnmarshalConfigYaml(cw.path)
		if err != nil {
			return err
		}
	}

	if err = cw.rcv.LoadConfig(ctx, cfg); err != nil {
		return fmt.Errorf("failed to load config: %w", err)
	}
	return nil
}
[...]

PR #841, opened 2025-07-08
test: adds real provider embeddings test & update doc

mathetake:
**Description**

This adds embeddings endpoint tests with the providers that support the endpoint. This only added the providers for which we have credentials. According to the testing situation we have right now, this also clarifies in the "Supported Endpoints" page that which provider is tested and which is not for each endpoint.


site/docs/supported-endpoints.md
```diff
@@ -143,3 +132,5 @@ To learn more about configuring and using the Envoy AI Gateway with these endpoi
 - **[Usage-Based Rate Limiting](./capabilities/usage-based-ratelimiting.md)** - Configure token-based rate limiting and cost controls
 - **[Provider Fallback](./capabilities/fallback.md)** - Set up automatic failover between providers for high availability
 - **[Metrics and Monitoring](./capabilities/metrics.md)** - Monitor usage, costs, and performance metrics
+
+[issue#609]: https://github.com/envoyproxy/ai-gateway/issues/609
```
aabchoo, line 137: Could you explain why you added this?

tests/extproc/real_providers_test.go
```diff
@@ -83,21 +83,36 @@ func TestWithRealProviders(t *testing.T) {
...
-		}
+		t.Run("chat/completions", func(t *testing.T) {
+			for _, tc := range []realProvidersTestCase{
+				{name: "openai", modelName: "gpt-4o-mini", required: internaltesting.RequiredCredentialOpenAI},
+				{name: "aws-bedrock", modelName: "us.meta.llama3-2-1b-instruct-v1:0", required: internaltesting.RequiredCredentialAWS},
+				{name: "azure-openai", modelName: "o1", required: internaltesting.RequiredCredentialAzure},
+				{name: "gemini", modelName: "gemini-2.0-flash-lite", required: internaltesting.RequiredCredentialGemini},
+				{name: "groq", modelName: "llama-3.1-8b-instant", required: internaltesting.RequiredCredentialGroq},
+				{name: "grok", modelName: "grok-3", required: internaltesting.RequiredCredentialGrok},
+				{name: "sambanova", modelName: "Meta-Llama-3.1-8B-Instruct", required: internaltesting.RequiredCredentialSambaNova},
```
aabchoo, line 94: sambanova isn't part of the supported endpoints above?

tests/extproc/real_providers_test.go
```diff
@@ -83,21 +83,36 @@ func TestWithRealProviders(t *testing.T) {
...
+					requireEventuallyChatCompletionNonStreamingRequestOK(t, tc.modelName, "Say this is a test")
+				})
+			}
+		})
+		t.Run("embeddings", func(t *testing.T) {
+			for _, tc := range []realProvidersTestCase{
+				{name: "openai", modelName: "text-embedding-3-small", required: internaltesting.RequiredCredentialOpenAI},
+				{name: "gemini", modelName: "gemini-embedding-exp-03-07", required: internaltesting.RequiredCredentialGemini},
+				{name: "sambanova", modelName: "E5-Mistral-7B-Instruct", required: internaltesting.RequiredCredentialSambaNova},
+				{name: "deepinfra", modelName: "BAAI/bge-base-en-v1.5", required: internaltesting.RequiredCredentialDeepInfra},
```
aabchoo, line 108: Is there a reason why the model for deep infra is different between embedding and chat/completions?

mathetake: because they (or any other provider i guess) have a different model for embeddings vs chat completion.

site/docs/supported-endpoints.md
mathetake: you can see the note section above in the table

tests/extproc/real_providers_test.go
mathetake: oh wow good catch

site/docs/supported-endpoints.md
aabchoo: oh it doesn't show up on the enriched view

tests/extproc/real_providers_test.go
mathetake: you see that is not only for deepinfra but for all four providers

aabchoo (approved):

mathetake:
@aabchoo thanks for the look!!!

mathetake:
the unit test job is failing due to the unrelated flaky: #701

PR #842, opened 2025-07-08
cli: adds default route test

### LINKED ISSUE #612: default AIGatewayRoute backend

mathetake:
**Description**

This adds an additional test to aigw run command so that we can verify that setting the default route is possible. 

Closes #612

### DIFF
--- a/cmd/aigw/run_test.go
+++ b/cmd/aigw/run_test.go
@@ -93,6 +93,12 @@ func TestRun(t *testing.T) {
 			modelName: "us.meta.llama3-2-1b-instruct-v1:0",
 			required:  internaltesting.RequiredCredentialAWS,
 		},
+		{
+			testName: "openai with fallback route",
+			// gpt-4o is not explicitly listed in the route, but it should still work by matching the fallback route.
+			modelName: "gpt-4o",
+			required:  internaltesting.RequiredCredentialOpenAI,
+		},
 	} {
 		t.Run(tc.testName, func(t *testing.T) {
 			client := openai.NewClient(option.WithBaseURL("http://localhost:1975" + "/v1/"))
--- a/cmd/aigw/run_test.go
+++ b/cmd/aigw/run_test.go
@@ -79,10 +79,11 @@ func TestRun(t *testing.T) {
 		return true
 	}, 120*time.Second, 1*time.Second)
 
-	for _, tc := range []struct {
+	type testCase struct {
 		testName, modelName string
 		required            internaltesting.RequiredCredential
-	}{
+	}
+	cases := []testCase{
 		{
 			testName:  "openai",
 			modelName: "gpt-4o-mini",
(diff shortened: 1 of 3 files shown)


cmd/aigw/run_test.go
```diff
@@ -93,6 +93,12 @@ func TestRun(t *testing.T) {
 			modelName: "us.meta.llama3-2-1b-instruct-v1:0",
 			required:  internaltesting.RequiredCredentialAWS,
 		},
+		{
```
aabchoo, line 97: nit: add `mistral:latest` case defaulting to Ollama?

aabchoo (approved):
lgtm

Issue #851, opened 2025-07-10
Add Weekly/Daily publish for images/helm charts


aabchoo:
*Description*:
>Describe the desired behavior, what scenario it enables and how it
would be used.

Currently, we publish `latest` images and `v0.0.0-latest` helm charts per merge to main. This is great when needing to get the latest changes, but there are instances where we'd want to only pull up to a specific PR. One scenario could be a fix was added a week ago with more changes were added after but we only want the fix.

Proposal:
- Have a daily or weekly automated GitHub action that will publish a unique
- Have a manual Github action to generate a unique image+helm tag

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

mathetake:
i like this and we should use the commit hash instead of :latest tag for the main branch char images as well!

mathetake:
or i think we can simply publish the hash-tagged helm charts on every main branch commit instead of manual or scheduled push

aabchoo:
> i like this and we should use the commit hash instead of :latest tag for the main branch char images as well!


Preferences on `latest-hash` vs `hash` as the tag?

> or i think we can simply publish the hash-tagged helm charts on every main branch commit instead of manual or scheduled push

I would prefer this!

mathetake:
i think simple let's use a commit hash as-is without anything else

PR #852, opened 2025-07-10
ci: commits to main pushes image/helm with sha suffix

aabchoo:
**Description**

Replacing `latest` suffix for helm charts and docker image. Replacing suffix with commit message.

#851

### DIFF
--- a/.github/workflows/build_and_test.yaml
+++ b/.github/workflows/build_and_test.yaml
@@ -264,5 +264,5 @@ jobs:
           username: ${{ vars.DOCKERHUB_USERNAME }}
           password: ${{ secrets.DOCKERHUB_PASSWORD }}
       - run: |
-          make helm-push HELM_CHART_VERSION=v0.0.0-latest
-          make helm-push HELM_CHART_VERSION=0.0.0-latest
+          make helm-push HELM_CHART_VERSION=v0.0.0-${{ github.sha }}
+          make helm-push HELM_CHART_VERSION=0.0.0-${{ github.sha }}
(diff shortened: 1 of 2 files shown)


.github/workflows/build_and_test.yaml
```diff
@@ -264,5 +264,5 @@ jobs:
           username: ${{ vars.DOCKERHUB_USERNAME }}
           password: ${{ secrets.DOCKERHUB_PASSWORD }}
       - run: |
-          make helm-push HELM_CHART_VERSION=v0.0.0-latest
-          make helm-push HELM_CHART_VERSION=0.0.0-latest
```
mathetake: i think we need to keep this to make the latest version documentation work. so can you instead of delete the existing one but add these commit hash based one?

.github/workflows/build_and_test.yaml
```diff
@@ -264,5 +264,5 @@ jobs:
...
+          make helm-push HELM_CHART_VERSION=v0.0.0-${{ github.sha }}
+          make helm-push HELM_CHART_VERSION=0.0.0-${{ github.sha }}
```
mathetake: Also i think to make the helm chart to point to the commit hash tagged containers, you will need to specify TAG=${{ github.sha }}

```suggestion
          make helm-push HELM_CHART_VERSION=0.0.0-${{ github.sha }} TAG=${{ github.sha }}
```

mathetake: you can see if this work correctly by passing them to helm-package locally

aabchoo:
Added back latest for both images + helm charts. This means we will be building twice for latest + sha images

.github/workflows/build_and_test.yaml
aabchoo: Ran 

` helm package manifests/charts/ai-gateway-crds-helm --app-version 7d293fa14a414d048de11f2599442361716137c3  --version 0.0.0-7d293fa14a414d048de11f2599442361716137c3
`

I see that the appVersion has been updated `7d293fa14a414d048de11f2599442361716137c3`

PR #854, opened 2025-07-10
test: adds weighted load balancing case

mathetake:
**Description**

This adds an additional e2e test case to ensure the weights and priority works correctly and matches our intuition. That is, when weights are assigned at host/backend level not ip address level. For example, let's say we have backends Foo and Bar has the same priority as well as the same weights. Even when Foo's FQDN resolves 1 ip address vs Bar resolves 10000 ip addresses, we expect the traffic split evenly distributed between Foo and Bar, not 1: 10000 distribution.

### DIFF
--- a/tests/e2e/testdata/testupstream.yaml
+++ b/tests/e2e/testdata/testupstream.yaml
@@ -166,7 +166,7 @@ metadata:
   name: testupstream-canary
   namespace: default
 spec:
-  replicas: 10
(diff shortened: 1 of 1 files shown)


mathetake:
no idea why "no healthy upstream" happening in irrelevant part of the test

mathetake:
marking as draft until e2e stabilizes but at least this all passes locally

PR #875, opened 2025-07-12
examples: use numAttemptsPerPriority in fallback

mathetake:
**Description**

numAttemptsPerPriority API was recently introduced in Envoy Gateway which allows users to specify how many times Envoy should send the retry requests to an endpoint in the same priority. For example, let's say when we have a backend p0 and p1 where p0 and p1 has the priority 0 and 1 respectively. Previously, when we want to send a request to p1 on retry, the passive health check must be configured and need to wait for envoy to eject the backend. With this numAttemptsPerPriority API, we no longer need to rely on it and allows us to do the fine-control on the retry backend selection.

Since we dropped v1.4 EG support on the main branch and we are planning to do our next version release based on v1.5 (not released yet), we can showcase this API in our examples now.

Implemented in EG: https://github.com/envoyproxy/gateway/pull/6204


tests/e2e/e2e_test.go
```diff
@@ -23,7 +23,7 @@ import (
 )
 
 const (
-	egDefaultVersion     = "v1.4.0"
+	egDefaultVersion     = "v0.0.0-latest"
```
mathetake, line 26: this was forgotten in #859

mathetake:
hoping e2e passes 🤞

Issue #877, opened 2025-07-13
log panic in backend security controller with credential rotation


yuzisun:
*Description*:
observed a log panic during credential rotation.
```
2025-07-13T03:28:13Z	ERROR	Observed a panic	{"controller": "backendsecuritypolicy", "controllerGroup": 
"aigateway.envoyproxy.io", "controllerKind": "BackendSecurityPolicy", "BackendSecurityPolicy": {"name":"envoy-
ai-gateway-basic-gcp-credentials","namespace":"gateway"}, "namespace": "gateway", "name": "envoy-ai-
gateway-basic-gcp-credentials", "reconcileID": "7c2a154d-28ca-4ba1-8598-755ac72cc92f", "panic": "odd 
number of arguments passed as key-value pairs for logging", "stacktrace": "goroutine 271 
[running]:\nk8s.io/apimachinery/pkg/util/runtime.logPanic({0x28f16c0, 0x4000decf00}, {0x1e5ae20, 
0x4000896830})\n\t/home/runner/go/pkg/mod/k8s.io/apimachinery@v0.34.0-
alpha.0/pkg/util/runtime/runtime.go:132 +0x98\nsigs.k8s.io/controller-runtime/pkg/internal/controller.
(*Controller[...]).Reconcile.func1()\n\t/home/runner/go/pkg/mod/sigs.k8s.io/controller-
runtime@v0.21.0/pkg/internal/controller/controller.go:108 +0xf0\npanic({0x1e5ae20?, 
0x4000896830?})\n\t/opt/hostedtoolcache/go/1.24.4/x64/src/runtime/panic.go:792 
+0x124\ngo.uber.org/zap/zapcore.CheckWriteAction.OnWrite(0x0?, 0x0?, {0x4001766240?, 0x0?, 
0x0?})\n\t/home/runner/go/pkg/mod/go.uber.org/zap@v1.27.0/zapcore/entry.go:196 
+0x78\ngo.uber.org/zap/zapcore.(*CheckedEntry).Write(0x4000d39110, {0x4000a2ae00, 0x1, 
0x1})\n\t/home/runner/go/pkg/mod/go.uber.org/zap@v1.27.0/zapcore/entry.go:262 +0x1b8\ngo.uber.org/zap.
(*Logger).DPanic(0x24fa072?, {0x257dc62?, 0x1e5ae20?}, {0x4000a2ae00, 0x1, 
[...]

PR #878, opened 2025-07-13
fix: panic in  BackendSecurityPolicyController

googs1025:
**Description**

This commit fixes log panic in  BackendSecurityPolicyController

Fixes https://github.com/envoyproxy/ai-gateway/issues/877

None

### DIFF
--- a/internal/controller/backend_security_policy.go
+++ b/internal/controller/backend_security_policy.go
@@ -205,11 +205,8 @@ func (c *BackendSecurityPolicyController) executeRotation(ctx context.Context, r
 						fmt.Sprintf("successfully rotated credential for %s in namespace %s of auth type %s, renewing in %f minutes",
 							bsp.Name, bsp.Namespace, bsp.Spec.Type, requeue.Minutes()))
 				} else {
-					c.logger.Error(fmt.Errorf("newly rotated credential is already expired %s", rotationTime),
-						"credential is already expired after rotation",
-						"namespace", bsp.Namespace,
-						"name", bsp.Name,
-					)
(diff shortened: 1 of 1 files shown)


internal/controller/backend_security_policy.go
```diff
@@ -205,8 +205,11 @@ func (c *BackendSecurityPolicyController) executeRotation(ctx context.Context, r
 						fmt.Sprintf("successfully rotated credential for %s in namespace %s of auth type %s, renewing in %f minutes",
 							bsp.Name, bsp.Namespace, bsp.Spec.Type, requeue.Minutes()))
 				} else {
-					c.logger.Error(fmt.Errorf("newly rotated credential is already expired %s",
-						rotationTime), "namespace", bsp.Namespace, "name", bsp.Name)
+					c.logger.Error(fmt.Errorf("newly rotated credential is already expired %s", rotationTime),
```
googs1025: `func (l Logger) Error(err error, msg string, keysAndValues ...any)`

Previously, msg info was missed, which caused panic, but I am not sure whether it is appropriate to fill the duplicate msg

yuzisun: how about following ? 
```go
err = fmt.Errorf("newly rotated credential is already expired %s", rotationTime)
c.logger.Error(err, err.Error(), "namespace", bsp.Namespace, "name", bsp.Name)
```

yuzisun: previously err is also not assigned

googs1025: done

yuzisun (approved):

Issue #873, opened 2025-07-12
create default filters per AIGatewayRoute with ownerReference to it


mathetake:
*Description*:

Currently, AIGateayRoute will be translated with two static HTTPRouteFilters as you can see here https://github.com/envoyproxy/ai-gateway/blob/f037bd2e9559f446f6a38e505fe7a570b73ba1ea/internal/controller/ai_gateway_route.go#L98-L120

However, they are statically generated, not per AIGatewayRoute, so there is no garbage collecting mechanism to it. For example, deleting AIGatewayRoute, etc won't clean them up.

We can create these two resources per AIGatewayRoute, and add the ownerreference to them to make sure that these two filter resources will be cleaned up. The required change should look similar to the one you see in the issue description of #872

googs1025:
I have seen that we need to change from the original HTTPRouteFilter (each namespace has only two specific defaultHTTPRouteFilters) to create two corresponding defaultHTTPRouteFilters for each AIGatewayRoute and set the ownerreference. 🤔 

I am willing to try this issue

PR #888, opened 2025-07-14
fix: add OwnerReference for HTTPRouteFilter

googs1025:
**Description**
This commit add OwnerReference for HTTPRouteFilter when creating

Fixes: https://github.com/envoyproxy/ai-gateway/issues/873

None

### DIFF
--- a/internal/controller/ai_gateway_route.go
+++ b/internal/controller/ai_gateway_route.go
@@ -158,7 +158,7 @@ func (c *AIGatewayRouteController) syncAIGatewayRoute(ctx context.Context, aiGat
 		if err := c.client.Get(ctx, client.ObjectKey{Name: base.Name, Namespace: base.Namespace}, &f); err != nil {
 			if apierrors.IsNotFound(err) {
 				if err = ctrlutil.SetControllerReference(aiGatewayRoute, base, c.client.Scheme()); err != nil {
-					return fmt.Errorf("failed to set owner reference for HTTPRouteFilter %s: %w", base.Name, err)
+					panic(fmt.Errorf("BUG: failed to set controller reference for HTTPRouteFilter: %w", err))
 				}
 				// Create the filter if it does not exist.
 				if err = c.client.Create(ctx, base); err != nil {
--- a/tests/controller/controller_test.go
+++ b/tests/controller/controller_test.go
@@ -23,7 +23,6 @@ import (
 	"github.com/stretchr/testify/require"
 	"go.uber.org/goleak"
 	corev1 "k8s.io/api/core/v1"
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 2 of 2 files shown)


internal/controller/ai_gateway_route.go
```diff
@@ -131,12 +151,15 @@ func (c *AIGatewayRouteController) syncAIGatewayRoute(ctx context.Context, aiGat
...
-	// Check if the static default HTTPRouteFilters exist in the namespace.
-	filters := defaultHTTPRouteFilters(aiGatewayRoute.Namespace)
+	// Check if the static default HTTPRouteFilters exist per AIGatewayRoute.
+	filters := generateHTTPRouteFilters(aiGatewayRoute)
 	for _, base := range filters {
 		var f egv1a1.HTTPRouteFilter
 		if err := c.client.Get(ctx, client.ObjectKey{Name: base.Name, Namespace: base.Namespace}, &f); err != nil {
 			if apierrors.IsNotFound(err) {
+				if err = ctrlutil.SetControllerReference(aiGatewayRoute, base, c.client.Scheme()); err != nil {
+					return fmt.Errorf("failed to set owner reference for HTTPRouteFilter %s: %w", base.Name, err)
```
mathetake: this should be a panic just like my comment in another pr

tests/controller/controller_test.go
```diff
@@ -315,6 +303,42 @@ func TestAIGatewayRouteController(t *testing.T) {
...
+			if err == nil {
+				return true
+			}
+			if !apierrors.IsNotFound(err) {
+				t.Logf("unexpected error when fetching filter %s: %v", hostRewriteKey, err)
+				return false
+			}
+			ok, _ := ctrlutil.HasOwnerReference(f.OwnerReferences, origin, c.Scheme())
+			require.True(t, ok, "expected hostRewriteFilter to have owner reference to AIGatewayRoute")
+			return false
```
mathetake: this is very weird. you are checking if err == nil at the beginning and at the bottom you are checking ownerreference.  you should check the owner reference on ok case
```suggestion
	var f egv1a1.HTTPRouteFilter
	err = c.Get(t.Context(), hostRewriteKey, &f)
	if err != nil {
		t.Logf("expected to get hostRewriteFilter %s, but got error: %v", hostRewriteHTTPFilterName, err)
		return false
	}
	ok, _ := ctrlutil.HasOwnerReference(f.OwnerReferences, origin, c.Scheme())
	require.True(t, ok, "expected hostRewriteFilter to have owner reference to AIGatewayRoute")
	return true
```

tests/controller/controller_test.go
mathetake, line 1: can you add the deletion check in `t.Run("delete route", func(t *testing.T) {` below?

tests/controller/controller_test.go
googs1025: Thanks for pointing out that. this was a logic error...

internal/controller/ai_gateway_route.go
googs1025: done

tests/controller/controller_test.go
googs1025: The envtest.Environment does not seem to have gc deletion. I actually tested it and it seems that does not work.

some issue: https://github.com/kubernetes-sigs/kubebuilder/pull/1175

docs: https://github.com/kubernetes-sigs/kubebuilder/blob/master/docs/book/src/reference/envtest.md#testing-considerations

googs1025: > The envtest.Environment does not seem to have gc-related deletion. I actually tested it and it seems that the function does not work.

mathetake (approved):
Thank you for your contributing!

PR #892, opened 2025-07-15
feat: add configurable anthropic version

alexagriffith:
**Description**

right now there is only one option allowed, but it should be configurable and use the default if nothing set


mathetake (approved):
💯

PR #895, opened 2025-07-16
docs: update epp design logics

Xunzhuo:
**Description**

1. This PR updates the EPP integration design logics.
2. This PR addes the request flow diagram for EPP integration.

### DIFF
--- a/docs/proposals/003-epp-integration-proposal/proposal.md
+++ b/docs/proposals/003-epp-integration-proposal/proposal.md
@@ -760,7 +760,7 @@ Current implementation is able to support InferencePool using AIGatewayRoute as 
 
 But there are some advantages for AIGatewayRoute + InferencePool:
 
-+ Native OpenAPI Support: we can use native Open AI Spec to call different inference pool in a same listener, because we support to parse the mode from the body, and refresh the cache with route level ai-gw extproc. But using HTTPRoute directly cannot do that, it has to add more routeMatch to do it in a same listener like headerMatch, or separate it into different listeners.
++ Native OpenAI Support: we can use native Open AI Spec to call different inference pool in a same listener, because we support to parse the mode from the body, and refresh the cache with route level ai-gw extproc. But using HTTPRoute directly cannot do that, it has to add more routeMatch to do it in a same listener like headerMatch, or separate it into different listeners.
 + Token Ratelimit: we can support token ratelimit, for the self hosted models, we can extract the token usage from the response and reduce the rate limit budget.
 + Advanced Observability: we can expose more metrics in our ai-gw ext-proc.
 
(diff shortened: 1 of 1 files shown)


mathetake (commented):
thanks, let's merge this after https://github.com/envoyproxy/ai-gateway/pull/823 lands

Xunzhuo:
@mathetake sure!

docs/proposals/003-epp-integration-proposal/proposal.md
```diff
@@ -488,7 +488,9 @@ spec:
       messageTimeout: 5s
 ```
 
-This direction is to reuse the abilities of Envoy Gateway, and generate the Backend and EnvoyExtensionPolicy to deal with the InferencePool
+This direction is to reuse the abilities of Envoy Gateway, and generate the Backend and EnvoyExtensionPolicy to deal with the InferencePool.
```
missBerg:
```suggestion
This direction is to reuse the capabilities of Envoy Gateway and generate the Backend and EnvoyExtensionPolicy to manage the InferencePool.
```

docs/proposals/003-epp-integration-proposal/proposal.md
```diff
@@ -488,7 +488,9 @@ spec:
...
+
+But it cannot support rule level InferencePool support. EnvoyExtensionPolicy can only target on HTTPRoute/Gateway Level, so it cannot support multiple InferencePool or mixed use AIServiceBackend in one AIGatewayRoute.
```
missBerg:
```suggestion
But it cannot provide rule-level InferencePool support. EnvoyExtensionPolicy can only target `HTTPRoute`/`Gateway`, therefore it cannot support multiple `InferencePool` resources or mixed use `AIServiceBackend` in one AIGatewayRoute.
```

docs/proposals/003-epp-integration-proposal/proposal.md
```diff
@@ -540,13 +640,130 @@ The complete integration workflow follows these steps:
...
+    Envoy->>Client: Response
+```
 
 ## Implementation Considerations and Limitations
 
+Current implementation is able to support InferencePool using AIGatewayRoute as well as HTTPRoute.
+
+But there are some advantages for AIGatewayRoute + InferencePool:
+
++ Native OpenAPI Support: we can use native Open AI Spec to call different inference pool in a same listener, because we support to parse the mode from the body, and refresh the cache with route level ai-gw extproc. But using HTTPRoute directly cannot do that, it has to add more routeMatch to do it in a same listener like headerMatch, or separate it into different listeners.
```
yuzisun:
```suggestion
+ Native OpenAI Support: we can use native Open AI Spec to call different inference pool in a same listener, because we support to parse the mode from the body, and refresh the cache with route level ai-gw extproc. But using HTTPRoute directly cannot do that, it has to add more routeMatch to do it in a same listener like headerMatch, or separate it into different listeners.
```

yuzisun: I think this essentially means `AIGatewayRoute` can support body based routing while `HTTPRoute` can only do header based routing right?

Xunzhuo: yes, we can smartly pick it from the body

mathetake (approved):

Issue #864, opened 2025-07-11
GCP Integration: Gemini Streaming Request


missBerg:
*Description*:
>Describe the desired behavior, what scenario it enables and how it
would be used.

[optional *Relevant Links*:]
Sub-issue of #609

PR #900, opened 2025-07-16
feat: add streaming support for GCP Vertex AI

sukumargaonkar:
**Description**
Add streaming support for Gemini API.
This PR implements transformations from gemini-streaming-chunks to openai-chunks

Issues:
- https://github.com/envoyproxy/ai-gateway/issues/609
- https://github.com/envoyproxy/ai-gateway/issues/864

### DIFF
--- a/internal/extproc/translator/gemini_helper.go
+++ b/internal/extproc/translator/gemini_helper.go
@@ -15,7 +15,6 @@ import (
 
 	"github.com/google/uuid"
 	"google.golang.org/genai"
-	"k8s.io/utils/strings/slices"
 
 	"github.com/envoyproxy/ai-gateway/internal/apischema/openai"
 )
@@ -633,9 +632,6 @@ func geminiLogprobsToOpenAILogprobs(logprobsResult genai.LogprobsResult) openai.
 // buildGCPModelPathSuffix constructs a path Suffix with an optional queryParams where each string is in the form of "%s=%s".
 func buildGCPModelPathSuffix(publisher, model, gcpMethod string, queryParams ...string) string {
 	pathSuffix := fmt.Sprintf("publishers/%s/models/%s:%s", publisher, model, gcpMethod)
-
-	// If there are no query parameters, return the path suffix as is.
-	queryParams = slices.Filter(queryParams[:0], queryParams, func(q string) bool { return q != "" })
 
 	if len(queryParams) > 0 {
 		pathSuffix += "?" + strings.Join(queryParams, "&")
--- a/internal/extproc/translator/gemini_helper.go
+++ b/internal/extproc/translator/gemini_helper.go
@@ -11,6 +11,7 @@ import (
 	"mime"
 	"net/url"
 	"path"
+	"strings"
 
 	"github.com/google/uuid"
 	"google.golang.org/genai"
@@ -19,11 +20,12 @@ import (
 )
 
 const (
-	GCPModelPublisherGoogle    = "google"
-	GCPModelPublisherAnthropic = "anthropic"
-	GCPMethodGenerateContent   = "generateContent"
-	GCPMethodRawPredict        = "rawPredict"
-	HTTPHeaderKeyContentLength = "Content-Length"
+	GCPModelPublisherGoogle        = "google"
+	GCPModelPublisherAnthropic     = "anthropic"
+	GCPMethodGenerateContent       = "generateContent"
+	GCPMethodStreamGenerateContent = "streamGenerateContent"
+	GCPMethodRawPredict            = "rawPredict"
+	HTTPHeaderKeyContentLength     = "Content-Length"
 )
 
 // -------------------------------------------------------------
@@ -514,6 +516,10 @@ func geminiFinishReasonToOpenAI(reason genai.FinishReason) openai.ChatCompletion
 		return openai.ChatCompletionChoicesFinishReasonStop
 	case genai.FinishReasonMaxTokens:
 		return openai.ChatCompletionChoicesFinishReasonLength
+	case "":
+		// For intermediate chunks in a streaming response, the finish reason is an empty string.
+		// This is normal behavior and should not be treated as an error.
+		return ""
 	default:
 		return openai.ChatCompletionChoicesFinishReasonContentFilter
 	}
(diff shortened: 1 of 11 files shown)


internal/extproc/translator/gemini_helper.go
```diff
@@ -19,11 +20,12 @@ import (
...
-	GCPModelPublisherAnthropic = "anthropic"
-	GCPMethodGenerateContent   = "generateContent"
-	GCPMethodRawPredict        = "rawPredict"
-	HTTPHeaderKeyContentLength = "Content-Length"
+	GCPModelPublisherGoogle        = "google"
+	GCPModelPublisherAnthropic     = "anthropic"
+	GCPMethodGenerateContent       = "generateContent"
+	GCPMethodStreamGenerateContent = "streamGenerateContent"
+	GCPMethodRawPredict            = "rawPredict"
+	HTTPHeaderKeyContentLength     = "Content-Length"
```
mathetake: could you unexport everything here

internal/extproc/translator/openai_gcpvertexai_test.go
```diff
@@ -427,12 +479,41 @@ func TestOpenAIToGCPVertexAITranslatorV1ChatCompletion_ResponseBody(t *testing.T
 			},
 			wantTokenUsage: LLMTokenUsage{},
 		},
+		{
+			name: "single stream chunk response",
```
mathetake, line 483: i can't find multiple chunks test case. take a look at TestOpenAIToAWSBedrockTranslatorV1ChatCompletion_Streaming_ResponseBody and let's have a similar one

sukumargaonkar: covered them in https://github.com/envoyproxy/ai-gateway/blob/7ef05590e41d0d1afa538bb992c90af1ddde8fb2/internal/extproc/translator/openai_gcpvertexai_test.go#L652

mathetake: ah nice. then the rest is can you add an e2e extproc test case similar to https://github.com/envoyproxy/ai-gateway/blob/18379e87ed56e3fd9ad9124df61154a20e46d7dc/tests/extproc/testupstream_test.go#L246

internal/extproc/translator/gemini_helper.go
```diff
@@ -514,6 +516,10 @@ func geminiFinishReasonToOpenAI(reason genai.FinishReason) openai.ChatCompletion
 		return openai.ChatCompletionChoicesFinishReasonStop
 	case genai.FinishReasonMaxTokens:
 		return openai.ChatCompletionChoicesFinishReasonLength
+	case "":
+		// For intermediate chunks in a streaming response, the finish reason is an empty string.
```
mathetake, line 520: is this documented or it's just a finding? if it is documented, let's put a link

internal/extproc/translator/gemini_helper.go
```diff
@@ -623,7 +629,55 @@ func geminiLogprobsToOpenAILogprobs(logprobsResult genai.LogprobsResult) openai.
 	}
 }
 
-func buildGCPModelPathSuffix(publisher, model, gcpMethod string) string {
+func buildGCPModelPathSuffix(publisher, model, gcpMethod string, pathParams map[string]string) string {
```
mathetake: why do you need to use map for this? only possible content is `map[string]string{"alt": "sse"}` and i don't see the necessity to use map (causing allocation?). Also, it's converted to a list immediately here and what's the reason for map in the first place. can you just accept variadic string?
```suggestion
// buildGCPModelPathSuffix constructs a path Suffix with an optional queryParams where each string is in the form of "%s=%s".
func buildGCPModelPathSuffix(publisher, model, gcpMethod string, queryParams ...string) string {
```

internal/extproc/translator/gemini_helper.go
sukumargaonkar: not documented in gemini api explicitly, but the finish reason is marked as optional in [gemini docs](https://cloud.google.com/vertex-ai/generative-ai/docs/reference/rest/v1/GenerateContentResponse#Candidate) and [open-ai docs](https://platform.openai.com/docs/api-reference/chat-streaming/streaming)

this change ensure the`geminiFinishReasonToOpenAI` handles the case when finish reason is empty

internal/extproc/translator/gemini_helper.go
```diff
@@ -623,7 +630,56 @@ func geminiLogprobsToOpenAILogprobs(logprobsResult genai.LogprobsResult) openai.
 	}
 }
 
-func buildGCPModelPathSuffix(publisher, model, gcpMethod string) string {
+// buildGCPModelPathSuffix constructs a path Suffix with an optional queryParams where each string is in the form of "%s=%s".
+func buildGCPModelPathSuffix(publisher, model, gcpMethod string, queryParams ...string) string {
 	pathSuffix := fmt.Sprintf("publishers/%s/models/%s:%s", publisher, model, gcpMethod)
+
+	// If there are no query parameters, return the path suffix as is.
+	queryParams = slices.Filter(queryParams[:0], queryParams, func(q string) bool { return q != "" })
```
mathetake: i don't think we want the code like this -- empty string is not a valid query param so it should be handled by a caller.

internal/extproc/translator/openai_gcpvertexai.go
```diff
@@ -38,7 +43,20 @@ func (o *openAIToGCPVertexAITranslatorV1ChatCompletion) RequestBody(_ []byte, op
...
+	// Choose the correct endpoint based on streaming.
+	var pathParams string
+	method := gcpMethodGenerateContent
+	if o.stream {
+		// For streaming requests, use the streamGenerateContent endpoint with SSE format.
+		method = gcpMethodStreamGenerateContent
+		pathParams = "alt=sse"
+	}
+
+	pathSuffix := buildGCPModelPathSuffix(gcpModelPublisherGoogle, modelName, method, pathParams)
```
mathetake: With this you don't need that checked i mentioned above.
```suggestion
	// Choose the correct endpoint based on streaming.
	var pathSuffix string
	if o.stream {
		// For streaming requests, use the streamGenerateContent endpoint with SSE format.
		pathSuffix = buildGCPModelPathSuffix(gcpModelPublisherGoogle, modelName, gcpMethodStreamGenerateContent, "alt=sse"
	} else {
		pathSuffix = buildGCPModelPathSuffix(gcpModelPublisherGoogle, modelName, gcpMethodGenerateContent)
	}
```

mathetake (approved):

mathetake:
i am going ahead an merge while ignoring the test coverage flake!

PR #910, opened 2025-07-18
test: add anthropic extproc tests

alexagriffith:
**Description**

adding extproc anthropic basic & error case tests


mathetake (commented):
Thank you!!

PR #912, opened 2025-07-20
feat: support aigw install/uninstall

Xunzhuo:
**Description**

This PR added support for aigw install/uninstall which improves UX.


mathetake:
let's put this on hold for now since i am not sure if we want to prioritize this at the moment and this certainly needs some discussion

Issue #791, opened 2025-06-27
Integrate OpenTelemetry to provide more observability


xiaolin593:
*Description*:

Integrate OpenTelemetry to add distributed trace(span & attribute) support to
1.  provide visibility into request lifecycle
2. analyze ai-gateway performance 
3. allow end-to-end debugging

mathetake:
is this relevant to https://github.com/envoyproxy/ai-gateway/issues/747 ?

ion-elgreco:
> is this relevant to [#747](https://github.com/envoyproxy/ai-gateway/issues/747) ?

Pretty much yeah, @xiaolin593 can you join the discussion there? @codefromthecrypt suggested to use OpenInference

mikeldking:
> > is this relevant to [#747](https://github.com/envoyproxy/ai-gateway/issues/747) ?
> 
> Pretty much yeah, [@xiaolin593](https://github.com/xiaolin593) can you join the discussion there? [@codefromthecrypt](https://github.com/codefromthecrypt) suggested to use OpenInference

We don't currently have any c++ instrumentation, but we'd be happy to help review a change that allows envoy AI gateway to report traces in openinference format! the use cases you mentioned are already supported in OSS phoenix

ion-elgreco:
@mikeldking to my understanding it needs to be a wasm plugin, I see there is JS support there: https://github.com/Arize-ai/openinference#javascript, so that might work

xiaolin593:
I'm currently instrumenting traces & spans with [OpenTelemetry Go SDK](https://github.com/open-telemetry/opentelemetry-go) in our internal extproc and having opentelemetry collector in same cluster to receive them. I hope we could use same SDK in ai-gateway but open t other way. 

Briefly took a look at [openinference](https://github.com/Arize-ai/openinference?tab=readme-ov-file), it doesn't have Go SDK?

codefromthecrypt:
@xiaolin593 otel SDK is the base layer for any instrumentation. So, for example in python there are additional libraries in openinference for corresponding libraries like openai. In the case of ai-gateway, it isn't necessary to have the instrumentation hosted inside the openinference repo, as we can just follow the conventions for the span data. So, basically, we use the specification, not a shared library, as ai-gateway is a service not a library IIUC. Does this make sense?

https://github.com/Arize-ai/openinference/blob/main/spec/llm_spans.md

[...]

PR #914, opened 2025-07-21
test: new fake OpenAI server for shared, deterministic tests

codefromthecrypt:
**Description**

This PR introduces a fake OpenAI API server using `go-vcr` to enable fast, credential-free, and reproducible integration tests for the AI Gateway external processor (`extproc`). The changes enhance testing reliability, eliminate dependency on live API calls, and reduce CI flakiness and costs. Key changes include:

- **Fake Shared OpenAI Server**: Added `internal/testing/fakeopenai` package with pre-recorded YAML cassettes for common OpenAI endpoints (e.g., chat completions, streaming, tools, multimodal). Supports automatic recording of new interactions when `OPENAI_API_KEY` is set.
- **ExtProc Testing Enhancements**:
  - On-demand `extproc` binary build in tests (via `EXTPROC_BIN` env or auto-compile).
  - New `tests/extproc` suite using the fake server, covering various request scenarios with assertions on responses and status codes.
  - Emits startup message ("AI Gateway External Processor is ready") to stderr for reliable readiness detection.
- **Config & Tooling Updates**: Added `.env.ollama` and a working docker-compose configuration for ad-hoc testing of extproc
- **Dependencies**: Added `gopkg.in/dnaeon/go-vcr.v4` for cassette recording/replay.

- Related to upcoming OpenTelemetry integration work #791
- Inspired by @anuraaga’s contributions to OpenTelemetry OpenAI instrumentation both in contrib and OpenInference.

Cassettes were recorded based on analysis of OpenInference instrumentation and edge cases that often break traces.

**Future work**

OpenAI is not the only inference API supported, but it is special as it is
the most common frontend and backend for AI Gateway. This is why we expose the
requests, as we will often proxy these even if the backend is not OpenAI
compatible.

The recording process would remain consistent for other cloud services, such as
Anthropic or Bedrock, though there could be variations in how requests are
scrubbed for secrets or handled for request signing. In a future refactoring,
we could extract the core recording infrastructure into a separate package,
reducing this one to just cassette constants and OpenAI-specific request
recording and handling details. Most of the code could be reused for other
backends.

For additional insights, refer to OpenTelemetry instrumentation, which often
employs VCR for LLM frameworks as well.

Here are key parts of the OpenTelemetry Botocore Bedrock instrumentation that
deals with request signing and recording:
- [test_botocore_bedrock.py#](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/77f3171bd4d0ca8eb5501c8c493364f7b6c8859a/instrumentation/opentelemetry-instrumentation-botocore/tests/test_botocore_bedrock.py#L403)
- [conftest.py](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/77f3171bd4d0ca8eb5501c8c493364f7b6c8859a/instrumentation/opentelemetry-instrumentation-botocore/tests/conftest.py#L77)

Here are key parts of OpenInference Anthropic instrumentation, which handles
their endpoint.
- [test_instrumentor.py](https://github.com/Arize-ai/openinference/blob/main/python/instrumentation/openinference-instrumentation-anthropic/tests/openinference/anthropic/test_instrumentor.py)


tests/extproc/README.md
```diff
@@ -0,0 +1,55 @@
+# AI Gateway ExtProc Tests
```
codefromthecrypt: FYI once this is merged, I will progress to do similar for aigw (standalone), so that we can easily test ad-hoc with the same config used in unit tests, ideally port collision free ***


** (there is a little bit of dancing in cmd/aigw we'll need to do because of duplicate prom registrations, but we'll get through)

internal/testing/fakeopenai/cassettes/chat-parallel-tools.yaml
```diff
@@ -0,0 +1,135 @@
...
+        proto: HTTP/1.1
+        proto_major: 1
+        proto_minor: 1
+        content_length: -1
+        host: api.openai.com
+        body: |-
+            {
+              "messages": [
+                {
+                  "content": "Answer in up to 3 words: Which ocean contains Bouvet Island?",
```
codefromthecrypt: oops I will fix this as my prompt didn't trigger a tool call. I will re-record before merge time

internal/testing/fakeopenai/handler.go
```diff
@@ -0,0 +1,256 @@
...
+		return fmt.Errorf("failed to create recorder: %w", err)
+	}
+	defer func() {
+		if stopErr := rec.Stop(); stopErr != nil && err == nil {
+			err = fmt.Errorf("failed to stop recorder: %w", stopErr)
+		}
+	}()
+
+	// Create a new request to the real API using the configured base URL..
+	// Strip /v1 prefix if present since apiBase already includes it..
```
codefromthecrypt: TODO I will also scrub for extra '..' ( was trying to satisfy lint gods and went to far )

internal/testing/fakeopenai/requests.go
```diff
@@ -0,0 +1,240 @@
...
+	// Set headers..
+	req.Header.Set("Content-Type", "application/json")
+	req.Header.Set("X-Cassette-Name", cassetteName)
+
+	return req, nil
+}
+
+// requestBodies contains the actual request body for each cassette..
+// These are kept private as they should only be accessed via NewRequest..
+var requestBodies = map[string]string{
```
codefromthecrypt: this is a relic. I'll update to grab it from the cassette directly.

internal/testing/fakeopenai/sse.go
```diff
@@ -0,0 +1,152 @@
...
+	// Check if we have a partial event at EOF..
+	if event.Data != "" || event.Event != "" {
+		return event, io.EOF
+	}
+
+	return nil, io.EOF
+}
+
+// ReadChatCompletionStream reads and parses OpenAI chat completion chunks from an SSE stream..
+func ReadChatCompletionStream(r io.Reader) ([]ChatCompletionChunk, string, error) {
```
codefromthecrypt: this is a utility to verify OpenInference formatted otel span data, which isn't in this change. I can remove it or leave it.

cmd/extproc/mainlib/main.go
```diff
@@ -160,6 +160,10 @@ func Main(ctx context.Context, args []string, stderr io.Writer) (err error) {
 			l.Error("Failed to shutdown health check server gracefully", "error", err)
 		}
 	}()
+
+	// Emit startup message to stderr when all listeners are ready.
+	fmt.Fprintln(stderr, "AI Gateway External Processor is ready")
```
mathetake: would it be possible to use the logger `l` for consistency or any reason not to do it? otherwise, only this line would be non structured text but other logs are all structured json

cmd/extproc/mainlib/main_test.go
```diff
@@ -205,3 +208,54 @@ func TestStartHealthCheckServer(t *testing.T) {
 		})
 	}
 }
+
+// TestExtProcStartupMessage ensures other programs can rely on the startup message to STDERR.
+func TestExtProcStartupMessage(t *testing.T) {
```
mathetake, line 213: +1

internal/testing/fakeopenai/README.md
mathetake, line 1: can you move to under `tests/internal/fakeopenai` instead of internal/testing which I meant for being used in unit tests in the main packages vs this is for integration tests. Or i think maybe I would like to consolidate them together, as there's not much reason except for my slight preference on separation, so i am neutral

tests/extproc/main_test.go
```diff
@@ -0,0 +1,195 @@
...
+	cmd.Stdout = io.Discard
+	cmd.Stderr = stderrWriter // Only write to our pipe, not to os.Stderr.
+	require.NoError(t, cmd.Start())
+
+	// Wait for Envoy to emit "starting main dispatch loop".
+	waitForEnvoyReady(stderrReader)
+}
+
+// TestEnvironment holds all the services needed for OTEL tests.
+type TestEnvironment struct {
```
mathetake: unexport ?
```suggestion
type testEnvironment struct {
```

tests/extproc/main_test.go
```diff
@@ -0,0 +1,195 @@
...
+
+// Close cleans up all resources in reverse order.
+func (t *TestEnvironment) Close() {
+	if t.openAIServer != nil {
+		t.openAIServer.Close()
+	}
+}
+
+// SetupTestEnvironment starts all required services and returns ports and a closer.
+func SetupTestEnvironment(t *testing.T) *TestEnvironment {
```
mathetake:
```suggestion
// setupTestEnvironment starts all required services and returns ports and a closer.
func setupTestEnvironment(t *testing.T) *TestEnvironment {
```

tests/extproc/extproc_test.go
```diff
@@ -32,6 +33,16 @@ const (
...
+	var err error
+	// Build extproc binary once for all tests.
+	if extprocBin, err = buildExtProcOnDemand(); err != nil {
+		log.Fatalf("failed to start tests due to build error: %v", err)
+	}
+	if _, err = exec.LookPath("envoy"); err != nil {
+		log.Fatalf("envoy binary not found in PATH: %v", err)
+	}
+
+	// continue existing code.
```
mathetake: 😄 
```suggestion
```

tests/extproc/envoy.yaml
```diff
@@ -209,6 +209,8 @@ static_resources:
                     typed_config:
                       "@type": type.googleapis.com/envoy.extensions.filters.http.ext_proc.v3.ExternalProcessor
                       allow_mode_override: true
+                      request_attributes:
+                        - xds.upstream_host_metadata
```
mathetake: i don't think we need this at the router level extproc, so could you try removing it? otherwise there seems a bug

tests/extproc/envoy_aigw_local.yaml
```diff
@@ -0,0 +1,110 @@
...
+                direct_response:
+                  status: 404
+                  body:
+                    inline_string: 'not forwarding paths except in cmd/extproc/mainlib/main.go'
+                typed_per_filter_config:
+                  envoy.filters.http.ext_proc:
+                    "@type": type.googleapis.com/envoy.config.route.v3.FilterConfig
+                    disabled: true
+          http_filters:
+          # Simulate real config injected via EnvoyExtensionPolicy
```
mathetake: so unfortunately this is totally different from the actual envoy config generated by both aigw and the k8s controller mode. I think this is accidentally working and i have a slight idea of why it's working if it's already passing tests, but this differs from the actual how it works. In that sense `aigw translate` is useless since the translation result does not reflect the reality. That is because (unfortunately in my opinion, due to the lack of many API in EG) the final envoy resources are fine tuned via the extension server mechanism in internal/extensionserver package, which makes the translation result different from this yaml.

If you take a look at the original envoy.yaml in this directory, you would see two extprocs filters (!= not two processes, just filters) are configured, and that's intentional to enable the fallback across multiple providers like falling back from local model to AWS/OpenAI etc. https://aigateway.envoyproxy.io/docs/latest/concepts/architecture/data-plane. The existing envoy.yaml is aligned with the actual config generated by aigw run & the control plane in other words

I am ok to go with this as-is since i see the main purpose of this config is to have a minimal working stuff, not for the alignment with exactly how it works. On the other hand, this might cause a trouble like; while things working at the extproc layer with this tests, not working at k8s & "aigw run" command since the configuration setups are different from here and there. wdyt?

mathetake (commented):
left some comments but before going forward i would love to understand if it's possible to deprecate testupstream_test.go tests where we already use a "fake" homemade server. I this these test and this fakeopenai tests are somewhat overlapped in the sense of using a fake server (not the replay portion which i like about this one), so i would like to migrate to this recording style. 

Meanwhile, I would slightly prefer keep one envoy config in `tests/extproc` test rather than two different envoy configs in the same directly where each will be used in different tests `testupstream_test.go,realproviders_test.go` vs `chat_completions_test.go`.

tests/extproc/README.md
mathetake: could you also mention the other existing tests using testupstream and how they differ from this https://github.com/envoyproxy/ai-gateway/blob/main/tests/extproc/testupstream_test.go ?  Or maybe cutting a subdirectory for all the new files in this PRs better for separation since the existing tests are not using the ones added here?

codefromthecrypt: actually I wanted to remove the openai -> openai tests from testupstream_test.go once the approach was ok as we don't want redundant I think.

internal/testing/fakeopenai/requests.go
codefromthecrypt: not a relic. I will update the comment. This is needed for re-recording ;) (you shouldn't depend on the cassette itself to know what request you should record!

tests/extproc/envoy.yaml
codefromthecrypt: I can try removing it, in my Otel tests there were errors in the log about missing attributes and finally hunted down to this. It might now be the case anymore or triggered with these files.

tests/extproc/envoy_aigw_local.yaml
codefromthecrypt: where I started was Otel and I couldn't find any yaml that worked without errors. So, I pared this down to the minimum that would work with extproc (ack I have a partial understanding still). I do prefer parity with intent more than what I was able to get working, so happy to try to match esp as I have tests now!

internal/testing/fakeopenai/README.md
codefromthecrypt: oops moved to the wrong internal, yeah will do

codefromthecrypt:
I cleaned up the fakeopenai server, but I still have some work to do to try to use the old envoy yaml, also remove redundant tests, and try to use internal/apischema/openai to describe the recordings. should be done in an hour or two

tests/extproc/envoy_aigw_local.yaml
codefromthecrypt: The existing envoy.yaml is tightly coupled with many clouds and slow testupstream which makes it difficult to progress before "stop the world" refactor of everything. I will make the openai one as similar as possible, yet still runnable with docker against ollama which is a very important goal of this PR. We need to be able to quickly ad-hoc especially when adding tracing to this, and there will be overlap with the "testupstream" thing vs VCR until other things migrate. Doing that can happen in later PRs and aren't urgent.

cmd/extproc/mainlib/main.go
codefromthecrypt: yeah this was intentionally using non-buffered stderr (like envoy does) which allows processes to reliably block on it. Should I put comments to that effect or should we just hope the buffering doesn't impact barriers? You can push your preference if I'm not around.

tests/extproc/envoy_aigw_local.yaml
codefromthecrypt: OK I think I get what you are saying more. Our destination is parity with real envoy config aigw would make, and a simplified envoy.yaml might pass for the wrong reasons. Yet, challenges remain..

It's a chicken-egg issue at this low layer of extproc to even know the config it might beed to run, because the owner of that generation is aigw -> eg, etc. We have envoy.yaml which while hand edited, is likely very close, and we have good reason to be cautious about deviations.

*Future idea*: Run aigw as a library with a startup hook to func-e, capture the real envoy config and have Go tests compare that with test config in this directory. For example, it could have a working aigw yaml for ollama and simply splat out the envoy. There is chicken egg on this part, too, but something like this would reduce drift and misconfiguration anxiety. Also, for newcomers like me, it makes it a lot more transparent what we actually must have for testing vs all the stuff that happens to be in a big file ;)

*Tactically in this PR*: Align new config closely to existing (e.g., dual filters), even if the new tests don't use those features, because it is closer to prod.

Hope this works, but you can feel free to revise if not!

codefromthecrypt:
ok I was unable to get the tests/extproc/envoy.yaml working with my test fixtures and also have it pass with ollama locally etc. It is a "kitchen sink" config, containing many different things used by all tests, and I don't think that is sustainable for reasons including deleting several of my hours trying to re-use it despite being mostly not about local openai connections.


Solving the entire surface of all tests is too much to do in this PR, and I'm not the right one to do that (yet at least). However, with this in, I can progress tracing which is something I'm a specialist at. Then, as I gain experience maybe I can fail less at this configuration.

TL;DR; I put the new tests into a "vcr" directory so that we can at least begin the process of less monolithic test execution, free of port conflicts and without risk of accidentally copy/pasting wrong responses etc. Feel free to revise this branch without my permission. Thanks for your attention to this matter ;)

codefromthecrypt:
added a footnote into the description about future work, taken from the fakeopenai README. signing off, and enjoy!

cmd/extproc/mainlib/main.go
mathetake: the logger is using the same stderr io.Writer variable `l := slog.New(slog.NewTextHandler(stderr, &slog.HandlerOptions{Level: flags.logLevel})` and i think it works exactly the same in terms of blocking?  I will push the change

mathetake:
TestChatCompletions race during the test... let me take alook ..

```
Read at 0x00c00019b4e8 by goroutine 13:
  strings.(*Builder).Write()
      /Users/runner/hostedtoolcache/go/1.24.4/arm64/src/strings/builder.go:83 +0xc0
  io.copyBuffer()
      /Users/runner/hostedtoolcache/go/1.24.4/arm64/src/io/io.go:431 +0x258
  io.copyBuffer()
      /Users/runner/hostedtoolcache/go/1.24.4/arm64/src/io/io.go:429 +0x218
  io.Copy()
      /Users/runner/hostedtoolcache/go/1.24.4/arm64/src/io/io.go:388 +0x64
  os.genericWriteTo()
      /Users/runner/hostedtoolcache/go/1.24.4/arm64/src/os/file.go:275 +0x18
  os.(*File).WriteTo()
      /Users/runner/hostedtoolcache/go/1.24.4/arm64/src/os/file.go:253 +0x90
  io.copyBuffer()
      /Users/runner/hostedtoolcache/go/1.24.4/arm64/src/io/io.go:411 +0xa8
  io.Copy()
      /Users/runner/hostedtoolcache/go/1.24.4/arm64/src/io/io.go:388 +0x5c
  os/exec.(*Cmd).writerDescriptor.func1()
      /Users/runner/hostedtoolcache/go/1.24.4/arm64/src/os/exec/exec.go:596 +0x38
  os/exec.(*Cmd).Start.func2()
      /Users/runner/hostedtoolcache/go/1.24.4/arm64/src/os/exec/exec.go:749 +0x3c
  os/exec.(*Cmd).Start.gowrap1()
      /Users/runner/hostedtoolcache/go/1.24.4/arm64/src/os/exec/exec.go:761 +0x44

Previous write at 0x00c00019b4e8 by goroutine 14:
  strings.(*Builder).Write()
      /Users/runner/hostedtoolcache/go/1.24.4/arm64/src/strings/builder.go:83 +0x13c
  io.(*multiWriter).Write()
      /Users/runner/hostedtoolcache/go/1.24.4/arm64/src/io/multi.go:85 +0xa8
  io.copyBuffer()
      /Users/runner/hostedtoolcache/go/1.24.4/arm64/src/io/io.go:431 +0x258
  io.copyBuffer()
      /Users/runner/hostedtoolcache/go/1.24.4/arm64/src/io/io.go:429 +0x218
  io.copyBuffer()
      /Users/runner/hostedtoolcache/go/1.24.4/arm64/src/io/io.go:429 +0x218
  io.copyBuffer()
      /Users/runner/hostedtoolcache/go/1.24.4/arm64/src/io/io.go:429 +0x218
  io.copyBuffer()
      /Users/runner/hostedtoolcache/go/1.24.4/arm64/src/io/io.go:429 +0x218
  io.copyBuffer()
      /Users/runner/hostedtoolcache/go/1.24.4/arm64/src/io/io.go:429 +0x218
  io.copyBuffer()
      /Users/runner/hostedtoolcache/go/1.24.4/arm64/src/io/io.go:429 +0x218
  io.copyBuffer()
      /Users/runner/hostedtoolcache/go/1.24.4/arm64/src/io/io.go:429 +0x218
  io.copyBuffer()
      /Users/runner/hostedtoolcache/go/1.24.4/arm64/src/io/io.go:429 +0x218
  io.copyBuffer()
      /Users/runner/hostedtoolcache/go/1.24.4/arm64/src/io/io.go:429 +0x218
  io.copyBuffer()
      /Users/runner/hostedtoolcache/go/1.24.4/arm64/src/io/io.go:429 +0x218
  io.copyBuffer()
      /Users/runner/hostedtoolcache/go/1.24.4/arm64/src/io/io.go:429 +0x218
  io.copyBuffer()
      /Users/runner/hostedtoolcache/go/1.24.4/arm64/src/io/io.go:429 +0x218
  io.copyBuffer()
      /Users/runner/hostedtoolcache/go/1.24.4/arm64/src/io/io.go:429 +0x218
  io.Copy()
      /Users/runner/hostedtoolcache/go/1.24.4/arm64/src/io/io.go:388 +0x64
  os.genericWriteTo()
      /Users/runner/hostedtoolcache/go/1.24.4/arm64/src/os/file.go:275 +0x18
  os.(*File).WriteTo()
      /Users/runner/hostedtoolcache/go/1.24.4/arm64/src/os/file.go:253 +0x90
  io.copyBuffer()
      /Users/runner/hostedtoolcache/go/1.24.4/arm64/src/io/io.go:411 +0xa8
  io.Copy()
      /Users/runner/hostedtoolcache/go/1.24.4/arm64/src/io/io.go:388 +0x5c
  os/exec.(*Cmd).writerDescriptor.func1()
      /Users/runner/hostedtoolcache/go/1.24.4/arm64/src/os/exec/exec.go:596 +0x38
  os/exec.(*Cmd).Start.func2()
      /Users/runner/hostedtoolcache/go/1.24.4/arm64/src/os/exec/exec.go:749 +0x3c
  os/exec.(*Cmd).Start.gowrap1()
      /Users/runner/hostedtoolcache/go/1.24.4/arm64/src/os/exec/exec.go:761 +0x44

Goroutine 13 (running) created at:
  os/exec.(*Cmd).Start()
      /Users/runner/hostedtoolcache/go/1.24.4/arm64/src/os/exec/exec.go:748 +0xbe4
  github.com/envoyproxy/ai-gateway/tests/extproc/vcr.requireEnvoyWithOutput()
      /Users/runner/work/ai-gateway/ai-gateway/tests/extproc/vcr/main_test.go:155 +0x814
  github.com/envoyproxy/ai-gateway/tests/extproc/vcr.setupTestEnvironment()
make: *** [test-extproc] Error 1
      /Users/runner/work/ai-gateway/ai-gateway/tests/extproc/vcr/main_test.go:201 +0x260
  github.com/envoyproxy/ai-gateway/tests/extproc/vcr.TestChatCompletions()
      /Users/runner/work/ai-gateway/ai-gateway/tests/extproc/vcr/chat_completions_test.go:21 +0x30
  testing.tRunner()
      /Users/runner/hostedtoolcache/go/1.24.4/arm64/src/testing/testing.go:1792 +0x180
  testing.(*T).Run.gowrap1()
      /Users/runner/hostedtoolcache/go/1.24.4/arm64/src/testing/testing.go:1851 +0x40

Goroutine 14 (running) created at:
  os/exec.(*Cmd).Start()
      /Users/runner/hostedtoolcache/go/1.24.4/arm64/src/os/exec/exec.go:748 +0xbe4
  github.com/envoyproxy/ai-gateway/tests/extproc/vcr.requireEnvoyWithOutput()
      /Users/runner/work/ai-gateway/ai-gateway/tests/extproc/vcr/main_test.go:155 +0x814
  github.com/envoyproxy/ai-gateway/tests/extproc/vcr.setupTestEnvironment()
      /Users/runner/work/ai-gateway/ai-gateway/tests/extproc/vcr/main_test.go:201 +0x260
  github.com/envoyproxy/ai-gateway/tests/extproc/vcr.TestChatCompletions()
      /Users/runner/work/ai-gateway/ai-gateway/tests/extproc/vcr/chat_completions_test.go:21 +0x30
  testing.tRunner()
      /Users/runner/hostedtoolcache/go/1.24.4/arm64/src/testing/testing.go:1792 +0x180
  testing.(*T).Run.gowrap1()
      /Users/runner/hostedtoolcache/go/1.24.4/arm64/src/testing/testing.go:1851 +0x40
==================
--- FAIL: TestChatCompletions (0.44s)
    --- FAIL: TestChatCompletions/chat-basic (0.02s)
        testing.go:1490: race detected during execution of test
```

mathetake:
why only ubuntu extproc jobs are failing...

mathetake:
can't repro on my local ubuntu but the seems not a flake on CI ..

tests/extproc/extproc_test.go
```diff
@@ -129,6 +129,8 @@ func requireRunEnvoy(t *testing.T, accessLogPath string) {
 		"-c", envoyYamlPath,
 		"--log-level", "warn",
 		"--concurrency", strconv.Itoa(max(runtime.NumCPU(), 2)),
+		// This allows multiple Envoy instances to run in parallel.
+		"--base-id", strconv.Itoa(time.Now().Nanosecond()),
```
codefromthecrypt, line 133: all your base are belong to us

codefromthecrypt:
Thanks for helping this in safely, @mathetake, like the safe buffers and also didn't know about the base-id thing. TIL!

Issue #844, opened 2025-07-08
Support Non-OpenAI API params in requests


sukumargaonkar:
*Description*:
>Describe the desired behavior, what scenario it enables and how it
would be used.

AI-gateway translates input request from OpenAI to whatever API the backend supports. Thus only fields in backend API that are represented in OpenAI API can be used. Fields in backend API that don't have any corresponding field in OpenAI cannot be used. (See examples below)
A way to express such non-openai native fields in the input will be useful as different API (OpenAI API, Gemini, Bedrock) start diverging.

Examples of Fields in backend APIs that do not have a corresponding OpenAI field:
- [RoutingConfig](https://cloud.google.com/vertex-ai/docs/reference/rest/v1/GenerationConfig#RoutingConfig) from GCP Gemini API
- [cachedContent](https://cloud.google.com/vertex-ai/docs/reference/rest/v1/projects.locations.endpoints/generateContent#request-body) from GCP Gemini API
- `googleSearch`, `codeExecution` fields in GCP Gemini's [Tool Part](https://cloud.google.com/vertex-ai/docs/reference/rest/v1/projects.locations.cachedContents#Tool)

*Note: The examples above are GCP Gemini specific but similar fields will exists in other APIs as well*

**Security Implications**  
- The json patch is part if the input request itself, i.e. its coming from the same source and not a third party.
- There might be other security implications. **Input Appreciated**

*Proposal*  
Support arbitary json-patch in OpenAI's [extra_body](https://github.com/openai/openai-python?tab=readme-ov-file#undocumented-request-params) param.
The json-patch can be read by upstream-extproc from request's extra_body param and applied as a last step after the request-translation is done.

Since the json-patch will depend on the backend API, allow specifying distince json-patches for each backend API. This will be required if retries across different backends needs to be supported.

Example OpenAI request with json-patch

```json
{
    "model": "gemini-2.0-flash-001",
    "messages": [
        {
            "role": "system",
            "content": "You are a numbers genius."
        },
        {
            "role": "user",
            "content": "how do I list prime numbers?"
        }
    ],
    "max_tokens": 250,
    "temperature": 0.1,
    "stream": false,
    "extra_body": {
        "aigateway.envoy.io": {
            "json_patches": {
                "GCPVertexAI": [
                    {
                        "op": "add",
                        "path": "/cachedContent",
                        "value": "some-cache-key"
                    },
                    {
                        "op": "add",
                        "path": "/generationConfig",
                        "value": {
[...]

PR #919, opened 2025-07-22
feat: support json-patch in request extra_body

sukumargaonkar:
**Description**

Add JSON patch support in request.extra_body to enable backend-specific field additions that are not natively supported in the OpenAI API schema. The JSON patches will be applied after the translator transforms the request body but before sending it to the backend.

The `proposal.md` below describes the process in detail

Issue: https://github.com/envoyproxy/ai-gateway/issues/844

- This will be presented in upcoming community meeting (24th July), marking as draft PR till then
- Refer to the `proposal.md` for overview
- Early feedback is appreciated

### DIFF
--- a/docs/proposals/004-json-patch-support/proposal.md
+++ b/docs/proposals/004-json-patch-support/proposal.md
@@ -70,10 +70,6 @@ The implementation extends the OpenAI ChatCompletionRequest schema with an `extr
 ```go
 type ChatCompletionRequest struct {
     // ...existing fields...
-    ExtraBody *ExtraBody `json:"extra_body,omitempty"`
-}
-
-type ExtraBody struct {
     AIGateway *AIGatewayExtensions `json:"aigateway.envoy.io,omitempty"`
 }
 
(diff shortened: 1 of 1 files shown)


sukumargaonkar:
Discussions  from weekly meeting
- openai sdk pulls everything from `extra_body` field and puts it at the root level. This will not work with current implementation. (@gavrissh)
- Instead of json_patches, should we start supporting vendor-specific fields at the root level itself? (@johnugeorge)

johnugeorge:
Extending from the community discussion, I see the following concerns with this PR

1.   This will not work with official OpenAI clients who directly want to talk to the AI gateway.  The design assumes that there is a layer in between that can add Json Patches (which is not true for all deployments)
2.   extra_body field in official OpenAI python client is flattened internally to keep the fields at the root level.  With this behaviour, the current PR might not work. Also, it is better not to override extra_body field to fit to our definition as the behaviour can change any time(eg: The same ield can be used for a different purpose tomorrow)

There are two types of OpenAI compatible backend types
1. Servers that implement OpenAI protocol as it is without any changes in official OpenAI Spec
2. Servers that extend OpenAI protocol with new fields but no breaking changes with respect to standard OpenAI spec
        eg: vllm which adds extra fields like guided choices  on top of the /chat/completions 
        
     This can be supported using official OpenAI clients using extra_body definitions. (https://docs.vllm.ai/en/latest/features/structured_outputs.html#online-serving-openai-api) 
        
     At the server side, the best bet is to keep fields at the root level. See https://github.com/sashabaranov/go-openai/blob/master/chat.go#L311-L324 where vllm specific fields are packed at the root level. This will work with all OpenAI clients.  Can we take this route?
        
     cc @yuzisun

johnugeorge:
Also if server needs to support direct openAI Clients for Gemini generationConfig, server should support extra body field at the top level

 'extra_body': {
        "google": {
          "thinking_config": {
            "thinking_budget": 800,
            "include_thoughts": True
          }
 https://ai.google.dev/gemini-api/docs/openai#python_2

sukumargaonkar:
@johnugeorge based on your suggestions and discussions with @yuzisun, made another proposal.
PTAL: https://github.com/envoyproxy/ai-gateway/pull/919

sukumargaonkar:
closing in favor of https://github.com/envoyproxy/ai-gateway/pull/961

PR #920, opened 2025-07-22
feat: anthropic update to support multi tool

alexagriffith:
**Description**

update the tool logic to support multi tool for gcp anthropic translation


internal/extproc/translator/openai_gcpanthropic.go
```diff
@@ -148,14 +150,15 @@ func translateOpenAItoAnthropicTools(openAITools []openai.Tool, openAIToolChoice
 
 			// The parameters for the function are expected to be a JSON Schema object.
 			// We can pass them through as-is.
-			toolParam.InputSchema = anthropic.ToolInputSchemaParam{
-				Properties: openAITool.Function.Parameters,
-				// TODO: support extra fields.
-				ExtraFields: nil,
+			if openAITool.Function.Parameters != nil {
```
alexagriffith, line 153: the type is any, so should have a check here

alexagriffith: also when i ran integration tests, the json schema was invalid and throwing errors, this section is the fix for that.

internal/extproc/translator/openai_gcpanthropic.go
```diff
@@ -383,35 +416,53 @@ func openAIToAnthropicMessages(openAIMsgs []openai.ChatCompletionMessageParamUni
...
-			var trb anthropic.ToolResultBlockParamContentUnion
-			for _, c := range content {
-				if c.OfText != nil {
-					trb.OfText = c.OfText
-				} else if c.OfImage != nil {
-					trb.OfImage = c.OfImage
+			// Aggregate all consecutive tool messages into a single user message
+			// to support parallel tool use.
+			var toolResultBlocks []anthropic.ContentBlockParamUnion
+			for i < len(openAIMsgs) && openAIMsgs[i].Type == openai.ChatMessageRoleTool {
```
alexagriffith, line 423: support multi tool calling

internal/extproc/translator/openai_gcpanthropic.go
```diff
@@ -148,14 +150,40 @@ func translateOpenAItoAnthropicTools(openAITools []openai.Tool, openAIToolChoice
...
+				paramsMap, ok := openAITool.Function.Parameters.(map[string]interface{})
+				if !ok {
+					err = fmt.Errorf("failed to cast tool parameters to map[string]interface{}")
+					return
+				}
+
+				inputSchema := anthropic.ToolInputSchemaParam{}
+
+				var typeVal string
+				if typeVal, ok = paramsMap["type"].(string); ok {
```
mathetake, line 163: do we have a unit test case where "type" is not present in the map? if not, let's add one.

internal/extproc/translator/openai_gcpanthropic.go
```diff
@@ -148,14 +150,40 @@ func translateOpenAItoAnthropicTools(openAITools []openai.Tool, openAIToolChoice
...
+					inputSchema.Type = constant.Object(typeVal)
+				}
+
+				var propsVal map[string]interface{}
+				if propsVal, ok = paramsMap["properties"].(map[string]interface{}); ok {
```
mathetake, line 168: the same comment for "type" above

internal/extproc/translator/openai_gcpanthropic.go
```diff
@@ -148,14 +150,40 @@ func translateOpenAItoAnthropicTools(openAITools []openai.Tool, openAIToolChoice
...
+					inputSchema.Properties = propsVal
+				}
+
+				var requiredVal []interface{}
+				if requiredVal, ok = paramsMap["required"].([]interface{}); ok {
```
mathetake, line 173: also the same comment for "type" above

internal/extproc/translator/openai_gcpanthropic.go
```diff
@@ -351,18 +380,21 @@ func openAIMessageToAnthropicMessageRoleAssistant(openAiMessage *openai.ChatComp
 
 // openAIToAnthropicMessages converts OpenAI messages to Anthropic message params type, handling all roles and system/developer logic.
 func openAIToAnthropicMessages(openAIMsgs []openai.ChatCompletionMessageParamUnion) (anthropicMessages []anthropic.MessageParam, systemBlocks []anthropic.TextBlockParam, err error) {
-	for i := range openAIMsgs {
+	i := 0
+	for i < len(openAIMsgs) {
```
mathetake: Let's make it for-loop scope so that it's obvious that `i` is only used in the inner scope
```suggestion
	for i := 0; i < len(openAIMsgs); {
```

internal/extproc/translator/openai_gcpanthropic.go
alexagriffith: done

internal/extproc/translator/openai_gcpanthropic.go
alexagriffith: done

internal/extproc/translator/openai_gcpanthropic.go
alexagriffith: done

internal/extproc/translator/openai_gcpanthropic.go
alexagriffith: done

mathetake (approved):

Issue #891, opened 2025-07-15
Cannot unmarshal JSON data as string or assistant content


ion-elgreco:
*Description*:
When running a request to AIGW, I get this error.

These are the logs:

```
time=2025-07-15T11:49:26.431Z level=DEBUG msg="handling a new stream" config_uuid=ai-gateway-provision-ai-gateway-provisioner
time=2025-07-15T11:49:26.431Z level=DEBUG msg="request headers processing" request_id=030d14b6-8312-4c6c-829c-bdbab9b4e9c0 is_upstream_filter=false request_headers.:authority=[REDACTED] request_headers.:path=/v1/chat/completions request_headers.:method=POST request_headers.:scheme=https request_headers.accept=application/json request_headers.user-agent="CustomOpenAIClient/JS 5.7.0" request_headers.x-stainless-retry-count=0 request_headers.x-stainless-lang=js request_headers.x-stainless-package-version=5.7.0 request_headers.x-stainless-os=Linux request_headers.x-stainless-arch=x64 request_headers.x-stainless-runtime=node request_headers.x-stainless-runtime-version=v20.19.3 request_headers.authorization=[REDACTED] request_headers.content-type=application/json request_headers.accept-language=* request_headers.sec-fetch-mode=cors request_headers.accept-encoding="br, gzip, deflate" request_headers.content-length=1651 request_headers.x-forwarded-for=10.244.4.127 request_headers.x-forwarded-proto=https request_headers.x-envoy-external-address=[REDACTED]  request_headers.x-request-id=030d14b6-8312-4c6c-829c-bdbab9b4e9c0
time=2025-07-15T11:49:26.431Z level=DEBUG msg="request headers processed" request_id=030d14b6-8312-4c6c-829c-bdbab9b4e9c0 is_upstream_filter=false response=request_headers:{}
[...]

PR #922, opened 2025-07-23
fix: extproc openai assistant message type

isyangban:
**Description**

According to https://platform.openai.com/docs/api-reference/chat/create
Assistant message's content should be string or array. 

https://github.com/envoyproxy/ai-gateway/issues/891


internal/apischema/openai/openai_test.go
```diff
@@ -164,7 +164,7 @@ func TestOpenAIChatCompletionMessageUnmarshal(t *testing.T) {
 			in: []byte(`{"model": "gpu-o4",
                         "messages": [
                          {"role": "assistant", "content": "you are a helpful assistant"},
-                         {"role": "assistant", "content": "{'text': 'you are a helpful assistant'}"}
```
isyangban, line 167: I think the previous code wasn’t valid JSON in the first place (maybe a hallucination?).

yuzisun: I think this gets decoded into the singular assistant content type which you have fixed with the array type.

PR #924, opened 2025-07-23
fix: incorrect content-length broken http connections

Xunzhuo:
**Description**


mathetake:
as per the offline comment, i am not sure what this is fixing so closing

PR #930, opened 2025-07-24
ci: use Envoy v1.35 for extproc test

mathetake:
**Description**

Envoy v1.35 has been released today, so this upgrades the Envoy version used in the extproc integration tests.

https://github.com/envoyproxy/envoy/releases/tag/v1.35.0


mathetake:
There's a bug in the build...

Xunzhuo:
what is that?

mathetake:
Should be good now

Xunzhuo:
what happened, the test is flaky?

mathetake:
so the macos binary is being built here https://github.com/tetratelabs/archive-envoy which is used by func-e, and func-e is used by EG's standalone mode and used in the failing test. There was a bug in macos built and i fixed it

PR #931, opened 2025-07-24
refactor: move common metadata key into internal api

Xunzhuo:
**Description**

This PR moves common metadata key into internal api.


mathetake (approved):
Makes sense

PR #945, opened 2025-07-25
cli: fixes race with EG logs for aigw run

mathetake:
**Description**

Previously, there was a flake in the use of bytes.Buffer passed to EG command as well as on our aigw run side, we are reading it concurrently. This fixes it temporarily by passing os.Stdout or io.Discard. 

The race was occasionally happening on CI which was causing the flake, so this fixes it.

### DIFF
--- a/cmd/aigw/run.go
+++ b/cmd/aigw/run.go
@@ -173,8 +173,8 @@ func run(ctx context.Context, c cmdRun, stdout, stderr io.Writer) error {
 	server := root.GetRootCommand()
 	// TODO: enable the log by default after the issue is resolved: https://github.com/envoyproxy/gateway/issues/6596
 	if c.Debug {
-		server.SetOut(os.Stdout)
-		server.SetErr(os.Stderr)
+		server.SetOut(stdout)
+		server.SetErr(stderr)
 	} else {
 		server.SetOut(io.Discard)
 		server.SetErr(io.Discard)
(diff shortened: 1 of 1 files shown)


cmd/aigw/run.go
```diff
@@ -171,14 +171,18 @@ func run(ctx context.Context, c cmdRun, stdout, stderr io.Writer) error {
 	// Then the agent will read the resources from the file pointed inside the config and start the Envoy process.
 
 	server := root.GetRootCommand()
-	egOut := &bytes.Buffer{}
-	server.SetOut(egOut)
-	server.SetErr(egOut)
+	// TODO: enable the log by default after the issue is resolved: https://github.com/envoyproxy/gateway/issues/6596
+	if c.Debug {
+		server.SetOut(os.Stdout)
+		server.SetErr(os.Stderr)
```
aabchoo: Just clarification: Why is `os.Stderr` used here but the [logger](https://github.com/envoyproxy/ai-gateway/pull/945/files#diff-258a62d2df66bb6e0c7a447ebfb9f34b7bf1d243c6b241f786950f1457f04ab7R80) uses the io.writer stderr param.

aabchoo (approved):

cmd/aigw/run.go
mathetake: good catch!

Issue #939, opened 2025-07-25
Support OpenAI service_tier field in request.


xiaolin593:
*Description*:
Include `service_tier` in API to allow user specify the processing type used for serving the request per OpenAI request_body schema https://platform.openai.com/docs/api-reference/chat/create

Current API schema doesn't include such field.

gavrissh:
I would like to pick this one, if not already done

Issue #847, opened 2025-07-09
Support Native Anthropic API


xiaolin593:
*Description*:
Support native Anthropic API so that user can enable Claude Code via Anthropic as some features are not available in OpenAI schema.

It's evident that Anthropic API is becoming an unified API for Claude model as cloud providers both GCP and AWS support Anthropic API.

 - Anthropic Native API
    - [x] Support native Anthropic API `v1/messages` for first party Anthropic  
    - [X] Support native Anthropic API `v1/messages` for [GCP Anthropic](https://github.com/envoyproxy/ai-gateway/pull/1022) 
    - [x] Support native Anthropic API `v1/messages` for AWS Anthropic

*Relevant Links*:
- [Anthropic Claude Code](https://docs.anthropic.com/en/docs/claude-code/overview)

- [AWS Anthropic Claude Message](https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-anthropic-claude-messages.html)

- [GCP Claude Model](https://cloud.google.com/vertex-ai/generative-ai/docs/partner-models/claude)

yuzisun:
[Claude code SDK ](https://docs.anthropic.com/en/docs/claude-code/sdk) supports first party Anthropic API and third party AWS bedrock, GCP Vertex. Currently AIGW only supports OpenAI API, however there is no reason to translate back and forth from Anthropic to OpenAI and then back to Anthropic to access Claude models. Their API structures and conversational paradigms for handling tools, thinking differ, reflecting their underlying design philosophies, so it is beneficial for Envoy AI Gateway to support Anthropic API natively to access Claude models.

mathetake:
Am I understanding correctly that this is about supporting Anthropic API as an input schema?

yuzisun:
> Am I understanding correctly that this is about supporting Anthropic API as an input schema?

Correct, so user can use anthropic SDK or Claude code which is Anthropic's agentic coding tool directly to access Claude family models via Envoy AI Gateway.

mathetake:
sounds good

yuzisun:
[...]

PR #947, opened 2025-07-25
Support OpenAI service_tier field

gavrissh:
**Description**
Support for the `service_tier` field in the OpenAI request and response

Fixes: #939


mathetake:
so for OpenAI backends, the transformation won't happen (bodies won't be mutated), so just adding the schema definition like the current change on the PR here won't have any impact unless it is used in the translation to, say, AWS Bedrock, and others where body mutations happen. what's your initial intent in the original issue? @xiaolin593

gavrissh:
> so for OpenAI backends, the transformation won't happen (bodies won't be mutated), so just adding the schema definition like the current change on the PR here won't have any impact unless it is used in the translation to, say, AWS Bedrock, and others where body mutations happen. what's your initial intent in the original issue? @xiaolin593

@mathetake @xiaolin593 Checked through the providers, `service_tier` field in the current definition is not supported by other providers

Only Anthropic has a field with the same name, but the accepted values are different - https://docs.anthropic.com/en/api/service-tiers#using-service-tiers 
The field is ignored when sent using chat/completions - https://docs.anthropic.com/en/api/openai-sdk#simple-fields

If the intent is to only support OpenAI, then my understanding is that no translation would be needed.

yuzisun:
> > so for OpenAI backends, the transformation won't happen (bodies won't be mutated), so just adding the schema definition like the current change on the PR here won't have any impact unless it is used in the translation to, say, AWS Bedrock, and others where body mutations happen. what's your initial intent in the original issue? @xiaolin593
> 
> @mathetake @xiaolin593 Checked through the providers, `service_tier` field in the current definition is not supported by other providers
> 
> Only Anthropic has a field with the same name, but the accepted values are different - https://docs.anthropic.com/en/api/service-tiers#using-service-tiers The field is ignored when sent using chat/completions - https://docs.anthropic.com/en/api/openai-sdk#simple-fields
> 
> If the intent is to only support OpenAI, then my understanding is that no translation would be needed.

yes the intent here is to support OpenAI service tier and it defaults to auto if not specified
https://openai.com/api-scale-tier/
https://openai.com/api-priority-processing/

internal/apischema/openai/openai.go
```diff
@@ -632,6 +632,16 @@ type ChatCompletionRequest struct {
...
 
+	// ServiceTier:string or null - Defaults to auto
+	// Specifies the processing type used for serving the request.
+	// If set to 'auto', then the request will be processed with the service tier configured in the Project settings. Unless otherwise configured, the Project will use 'default'.
+	// If set to 'default', then the request will be processed with the standard pricing and performance for the selected model.
+	// If set to 'flex' or 'priority', then the request will be processed with the corresponding service tier. Contact sales to learn more about Priority processing.
+	// When the service_tier parameter is set, the response body will include the service_tier value based on the processing mode actually used to serve the request. 
+	// This response value may be different from the value set in the parameter.
+	// Docs: https://platform.openai.com/docs/api-reference/chat/create#chat-create-service_tier
+	ServiceTier *string `json:"service_tier,omitempty"`
```
yuzisun, line 646: We can add a comment that currently this field is only intended for OpenAI processing, it is possible that in future this field can be opened for other vendors with translation. Otherwise we may need to take the [inline extension approach](https://github.com/envoyproxy/ai-gateway/pull/961) and treat OpenAI as another model vendor.

xiaolin593:
The original intent is to allow user to specify different processing tier in request to OpenAI. `service_tier` field also will be returned in response too for user to make routing decision based on processing capacity or cost.

I guess translation part would become obvious when Anthropic Native API is available.
OpenAI service_tier supported value: auto, default, priority or flex
Anthropic service_tier supported value: auto, standard_only

gavrissh:
> The original intent is to allow user to specify different processing tier in request to OpenAI. `service_tier` field also will be returned in response too for user to make routing decision based on processing capacity or cost.
> 
> I guess translation part would become obvious when Anthropic Native API is available. OpenAI service_tier supported value: auto, default, priority or flex Anthropic service_tier supported value: auto, standard_only

@xiaolin593, I have updated the PR as per your suggestion. Could you have a look

internal/apischema/openai/openai.go
```diff
@@ -632,6 +632,16 @@ type ChatCompletionRequest struct {
 	// Docs: https://platform.openai.com/docs/api-reference/chat/create#chat-create-seed
 	Seed *int `json:"seed,omitempty"`
 
+	// ServiceTier:string or null - Defaults to auto
+	// Specifies the processing type used for serving the request.
+	// If set to 'auto', then the request will be processed with the service tier configured in the Project settings. Unless otherwise configured, the Project will use 'default'.
+	// If set to 'default', then the request will be processed with the standard pricing and performance for the selected model.
+	// If set to 'flex' or 'priority', then the request will be processed with the corresponding service tier. Contact sales to learn more about Priority processing.
```
xiaolin593:
```suggestion
	// If set to 'flex' or 'priority', then the request will be processed with the corresponding service tier.
```

internal/extproc/translator/openai_gcpanthropic.go
```diff
@@ -78,6 +79,17 @@ func validateTemperatureForAnthropic(temp *float64) error {
 	return nil
 }
 
+func translateAnthropicServiceTier(serviceTier *string) (anthropic.MessageNewParamsServiceTier, error) {
```
xiaolin593: I'm not sure if GCP Vertex API  or AWS Bedrock supports service_tier in request/response for Anthropic models. I do not find GCP API reference for `service_tier`. Can you add reference in comment if you find one?

If `service_tier` isn't supported by Vertex API or AWS Bedrock, then we do not need OpenAI -> Anthropic translation.

My comment that responded to @mathetake 's question in https://github.com/envoyproxy/ai-gateway/pull/947#issuecomment-3138443313 was to suggest to have OpenAI -> Anthropic API translation if we decide that OpenAI schema is the only standard for envoy-ai-gateway, but that is not the case per discussion in https://github.com/envoyproxy/ai-gateway/issues/847. 

wdyt @yuzisun ?

gavrissh: Understood, the previous comment got me confused.
If we plan to support Anthropic native `/v1/messages/` API as per #847, then I agree that the translator changes made in this PR are redundant and the provider API is not present for this field

gavrissh:
Based on the comments above 
- Intend to support this field only for OpenAI processing
- Anthropic supports this field, but the Vertex API and AWS Bedrock API don't expose them, hence no translator changes expected

@yuzisun @mathetake @xiaolin593 PR is ready for review

mathetake:
> Intend to support this field only for OpenAI processing

Let me repeat my comment again that adding this field doesn't have any impact on OpenAI processing. When the backend is in OpenAI schema, AIGW doesn't mutate the body. Parsed request body is only used for translation (except for "model" field), meaning not used for OpenAI. In other words, "service_tier" is **already** passed to OpenAI format backends.

If this won't be used at all in the translation, then this PR doesn't do anything for us but at the same time harmless though

yuzisun:
> > Intend to support this field only for OpenAI processing
> 
> Let me repeat my comment again that adding this field doesn't have any impact on OpenAI processing. When the backend is in OpenAI schema, AIGW doesn't mutate the body. Parsed request body is only used for translation (except for "model" field), meaning not used for OpenAI. In other words, "service_tier" is **already** passed to OpenAI format backends.
> 
> If this won't be used at all in the translation, then this PR doesn't do anything for us but at the same time harmless though

totally agree, I think the purpose here is for validation and schema documentation so user know what are the supported fields on envoy ai gateway as we maintain this as a compatible OpenAI schema.

mathetake:
yep makes sense, so i will edit the title to avoid the confusion

mathetake:
edited the PR title and description to reflect the exact change 
___
Title: extproc: add service_tier field into OpenAI schema
Description: This commit adds the service_tier field in the OpenAI schema definition. Note that the field is already supported without this change since the parsed body is not used for OpenAI format backends except for the model field.

internal/apischema/openai/openai.go
```diff
@@ -635,6 +635,17 @@ type ChatCompletionRequest struct {
...
 
+	// ServiceTier:string or null - Defaults to auto
+	// Specifies the processing type used for serving the request.
+	// If set to 'auto', then the request will be processed with the service tier configured in the Project settings. Unless otherwise configured, the Project will use 'default'.
+	// If set to 'default', then the request will be processed with the standard pricing and performance for the selected model.
+	// If set to 'flex' or 'priority', then the request will be processed with the corresponding service tier.
+	// When the service_tier parameter is set, the response body will include the service_tier value based on the processing mode actually used to serve the request.
+	// This response value may be different from the value set in the parameter.
+	// Docs: https://platform.openai.com/docs/api-reference/chat/create#chat-create-service_tier
+	// Field is only intended for OpenAI processing.
```
mathetake:
```suggestion
```

PR #951, opened 2025-07-26
feat: add configurable HTTP headers as metrics labels

### LINKED ISSUE #848: Flexible metrics label customization without building a custom extproc
*Description*:

Currently, the only way to add additional metrics label is to use [the custom metrics interface](https://github.com/envoyproxy/ai-gateway/blob/f02cd42d98080e566a89ca5e1f2eb6a8b492033e/filterapi/x/x.go#L18-L20) and rebuild your own external processor binary. That works for advanced users who are in any ways willing to rebuild the ext proc container and deploy by themselves. However, there's clearly a huge hurdle for a lot people, especially people who start looking at this project or maybe are not familiar with Go, or want to use the vanilla upstream artifacts.

100% of the use case i have heard about this metrics customization so far is about "adding additional labels from headers". Hence, headers is the only context that can be used with the custom metrics interface right now.

So I propose that we add a configuration to allow users to specify header names to be added to the metrics. More precisely, I am proposing to add `--metrics-header-names` to the extproc command line flags where users specify header names to be added to the metrics.

For example, with `--metrics-header-names=foo,bar`, the extproc will add the labels corresponding to the headers whose names are either foo or bar.

surenraju-careem:
**Description**
Implements feature to allow users to specify which HTTP request headers should be added as labels to Prometheus/OpenTelemetry metrics via a new command-line flag --metricsHeaderNames.

Changes:
- Add --metricsHeaderNames flag to extproc with comma-separated header names
- Update metrics system to extract header values and add as metric labels
- Add header name sanitization for Prometheus label compatibility
- Update Helm chart to expose the new configuration option
- Update controller to propagate the flag to extproc deployments
- Add comprehensive unit and integration tests
- Add proper handling of missing headers with 'unknown' fallback value

The feature allows operators to add request headers like user-id, team-id as metric labels without rebuilding the extproc binary, enabling better observability and monitoring capabilities.

Fixes #848


surenraju-careem:
Screenshot 2025-07-26 at 4 07 27 PM

mathetake:
nice, this looks cool. i will review Monday

surenraju-careem:
Thank you @mathetake
Are these flaky test runs or something to look into?  
https://github.com/envoyproxy/ai-gateway/actions/runs/16541900689/job/46785575063?pr=951
https://github.com/envoyproxy/ai-gateway/actions/runs/16541900689/job/46785575058?pr=951
https://github.com/envoyproxy/ai-gateway/actions/runs/16541900689/job/46785575078?pr=951

mathetake:
They are all your tests added here. Maybe it looks conflict of the port etc. you can run the full tests locally and see how it goes. It looks pretty much reproducible.

surenraju-careem:
@mathetake latest pipeline passed. Cloud you please have a look?

cmd/extproc/mainlib/main.go
```diff
@@ -85,6 +91,42 @@ func parseAndValidateFlags(args []string) (extProcFlags, error) {
...
+				if (r >= 'a' && r <= 'z') || (r >= '0' && r <= '9') || r == '_' {
+					return r
+				}
+				return '_'
+			}, header)
+
+			// Ensure it doesn't start with a number (Prometheus requirement).
+			if len(sanitized) > 0 && sanitized[0] >= '0' && sanitized[0] <= '9' {
+				sanitized = "header_" + sanitized
+			}
```
mathetake: shouldn't this be happening at much earlier phase: the main routine of the controller at the startup time?

internal/metrics/base_metrics.go
```diff
@@ -59,12 +74,49 @@ func (b *baseMetrics) SetBackend(backend *filterapi.Backend) {
...
 		attribute.Key(genaiAttributeSystemName).String(b.backend),
 		attribute.Key(genaiAttributeRequestModel).String(b.model),
 	)
+
+	// Add header values as attributes if configured.
+	if b.metricsHeaderNames != nil && b.requestHeaders != nil {
+		for _, sanitizedHeaderName := range b.metricsHeaderNames {
+			// Convert the sanitized header name back to the original format for lookup.
+			// The sanitized name has underscores, but the actual header name might have dashes.
+			originalHeaderName := strings.ReplaceAll(sanitizedHeaderName, "_", "-")
```
mathetake: similar comment: i don't think we need to do this for each request?

internal/metrics/base_metrics.go
```diff
@@ -59,12 +74,49 @@ func (b *baseMetrics) SetBackend(backend *filterapi.Backend) {
...
+				}
+			}
+
+			// If header value is empty, use "unknown" to maintain consistent cardinality.
+			if headerValue == "" {
+				headerValue = "unknown"
+			}
+
+			// Create an attribute key with "header_" prefix to avoid conflicts.
+			attrKey := "header_" + sanitizedHeaderName
```
mathetake: same comment: can we do the initialization at the startup.

Also can we make sure this is a request header, not response header?
```suggestion
			// Create an attribute key with "header_" prefix to avoid conflicts.
			attrKey := "request_header_" + sanitizedHeaderName
```

cmd/extproc/mainlib/main.go
```diff
@@ -70,6 +71,11 @@ func parseAndValidateFlags(args []string) (extProcFlags, error) {
 	)
 	fs.IntVar(&flags.metricsPort, "metricsPort", 1064, "port for the metrics server.")
 	fs.IntVar(&flags.healthPort, "healthPort", 1065, "port for the health check HTTP server.")
+	fs.StringVar(&flags.metricsHeaderNames,
+		"metricsHeaderNames",
```
mathetake: let's add a prefix "request" just in case later we need response header use case
```suggestion
		"metricsRequestHeaderNames",
```

cmd/extproc/mainlib/main.go
```diff
@@ -37,11 +37,48 @@ import (
...
+		// Convert to lowercase and sanitize for Prometheus compatibility.
+		// Replace any non-alphanumeric character (except underscore) with underscore.
+		originalName := strings.ToLower(headerName)
+		sanitized := ""
+		for _, r := range originalName {
+			if (r >= 'a' && r <= 'z') || (r >= '0' && r <= '9') || r == '_' {
+				sanitized += string(r)
+			} else {
+				sanitized += "_"
+			}
```
mathetake, line 72: sorry the same comment again- can you 
* do this conversion in the controller's start up time (cmd/controller), not extproc (cmd/extproc)
* rename the extproc's flag to `extProcPrometheusMetricsRequestHeaderNames`

surenraju-careem: What do you think about accepting comma-separated key-value pairs (e.g., header1:label1,header2:label2) instead of just a comma-separated list of header names(e.g., header1,header2) ? This would give users more flexibility in how they label the metrics, right?

mathetake: i think that's a good idea

surenraju-careem:
Moving to new PR https://github.com/envoyproxy/ai-gateway/pull/981

Issue #953, opened 2025-07-28
[BUG] Basic usage standard config cannot startup


Xunzhuo:
*Description*:

This is caused by we enabled the InferencePool in default, but not installed it.

```
2025-07-28T10:55:33.603Z        ERROR   provider.controller-runtime.source.Kind source/kind.go:75        if kind is a CRD, it should be installed before calling Start    {"runner": "provider", "kind": "InferencePool.inference.networking.x-k8s.io", "error": "no matches for kind \"InferencePool\" in version \"inference.networking.x-k8s.io/v1alpha2\""}
2025-07-28T10:55:43.602Z        ERROR   provider.controller-runtime.source.Kind source/kind.go:75        if kind is a CRD, it should be installed before calling Start    {"runner": "provider", "kind": "InferencePool.inference.networking.x-k8s.io", "error": "no matches for kind \"InferencePool\" in version \"inference.networking.x-k8s.io/v1alpha2\""}
2025-07-28T10:55:53.602Z        ERROR   provider.controller-runtime.source.Kind source/kind.go:75        if kind is a CRD, it should be installed before calling Start    {"runner": "provider", "kind": "InferencePool.inference.networking.x-k8s.io", "error": "no matches for kind \"InferencePool\" in version \"inference.networking.x-k8s.io/v1alpha2\""}
2025-07-28T10:56:03.600Z        ERROR   provider.controller-runtime.source.Kind source/kind.go:75        if kind is a CRD, it should be installed before calling Start    {"runner": "provider", "kind": "InferencePool.inference.networking.x-k8s.io", "error": "no matches for kind \"InferencePool\" in version \"inference.networking.x-k8s.io/v1alpha2\""}
2025-07-28T10:56:13.602Z        ERROR   provider.controller-runtime.source.Kind source/kind.go:75        if kind is a CRD, it should be installed before calling Start    {"runner": "provider", "kind": "InferencePool.inference.networking.x-k8s.io", "error": "no matches for kind \"InferencePool\" in version \"inference.networking.x-k8s.io/v1alpha2\""}
2025-07-28T10:56:23.601Z        ERROR   provider.controller-runtime.source.Kind source/kind.go:75        if kind is a CRD, it should be installed before calling Start    {"runner": "provider", "kind": "InferencePool.inference.networking.x-k8s.io", "error": "no matches for kind \"InferencePool\" in version \"inference.networking.x-k8s.io/v1alpha2\""}
[...]

PR #954, opened 2025-07-28
fix: separate init inferencepool config

Xunzhuo:
**Description**

Fixes: https://github.com/envoyproxy/ai-gateway/issues/953

I am thinking if we should let users install inferencepool in default, so this problem can be solved as well.


mathetake:
i think i would like this direction. As long as it'c clearly documented as you already did, it should be fine

mathetake (approved):

PR #961, opened 2025-07-28
feat: support vendor-specific fields in ChatCompletion

sukumargaonkar:
**Description**
Add support for vendor-specific parameters through an `aigateway.envoy.io` field in OpenAI requests. This will enable backend-specific field additions that are not natively supported in the OpenAI API schema. The vendor-specific fields will be applied after the translator transforms the request body but before sending it to the backend.

The proposal.md below describes the process in detail

Issue: https://github.com/envoyproxy/ai-gateway/issues/844
Previous Proposal: https://github.com/envoyproxy/ai-gateway/pull/919

- Refer to the `proposal.md` for overview
- This PR builds on the comment [1] from previous proposal
- Reviews appreciated

1: https://github.com/envoyproxy/ai-gateway/pull/919#issuecomment-3116467502


docs/proposals/004-vendor-specific-fields/proposal.md
```diff
@@ -0,0 +1,90 @@
...
+  "aigateway.envoy.io": {
+    "GCPVertexAI": {
+      "generationConfig": {
+        "thinkingConfig": {
+          "includeThoughts": true,
+          "thinkingBudget": 1000
+        }
+      }
+    },
+    "GCPAnthropic": {
```
yuzisun: I think this should be from the supported APISchema we defined on AIServiceBackend. Both AWSAnthropic and GCPAnthropic are following the Anthropic API.

docs/proposals/004-vendor-specific-fields/proposal.md
```diff
@@ -0,0 +1,90 @@
...
+  "model": "gemini-1.5-pro",
+  "messages": [
+    {
+      "role": "user",
+      "content": "Explain quantum computing and show me a simple code example."
+    }
+  ],
+  "temperature": 0.7,
+  "max_tokens": 2000,
+  "aigateway.envoy.io": {
```
yuzisun: Is this extra section compatible with the OpenAI SDK?

docs/proposals/004-vendor-specific-fields/proposal.md
gavrissh: Are there any potential differences in the APISchema between GCP-supported Anthropic and AWS-supported Anthropic? If yes, is it better to separate it out?

docs/proposals/004-vendor-specific-fields/proposal.md
```diff
@@ -0,0 +1,90 @@
...
+    GCPVertexAI  *GCPVertexAIVendorFields  `json:"GCPVertexAI,omitempty"`
+    GCPAnthropic *GCPAnthropicVendorFields `json:"GCPAnthropic,omitempty"`
+}
+
+// GCP Vertex AI vendor-specific fields.
+type GCPVertexAIVendorFields struct {
+    GenerationConfig *GCPVertexAIGenerationConfig `json:"generationConfig,omitempty"`
+}
+
+// AWS Bedrock vendor-specific fields.
```
gavrissh: nit misplaced comment?

johnugeorge:
@sukumargaonkar  This will not work with native clients. We need to keep embeddable structs(json: inline tag) so as to have fields at the top level and avoid intermediate layers.

docs/proposals/004-vendor-specific-fields/proposal.md
```diff
@@ -0,0 +1,90 @@
...
+
+For example:
+- GCP Vertex AI's `thinkingConfig` for advanced reasoning models.
+- AWS Bedrock's `guardrail` parameters for content filtering.
+
+## Schema Extensions
+The `ChatCompletionRequest` struct is extended to include a new field for vendor-specific parameters
+
+```go
+type ChatCompletionRequest struct {
```
gavrissh, line 32: For a native OpenAI client this will be the request

```
completion = client.chat.completions.create(
    model="<model>",
    messages=[
        {"role": "user", "content": "Classify this sentiment: vLLM is wonderful!"}
    ],
    extra_body={
        "thinking": {
                "type": "enabled",
                "budget_tokens": 1000
          }
    }
) 
```

Prefer it being embedded, something like this
```
type ChatCompletionRequest struct {
    // ...existing fields...
    
    // Embed vendor fields directly
    GCPVertexAI  `json:",inline,omitempty"`
    GCPAnthropic `json:",inline,omitempty"`
}

// All GCP Vertex AI supported extra fields
type GCPVertexAI struct {
    GenerationConfig *GCPVertexAIGenerationConfig `json:"generationConfig,omitempty"`
}

// All GCP Anthropic supported extra fields
type GCPAnthropic struct {
    Thinking *anthropic.ThinkingConfigParamUnion `json:"thinking,omitempty"`
}
```

@johnugeorge @mathetake @yuzisun 
Just thinking out loud, is it possible to make sure only one vendor fields are used based on the supported backend? Say the request was to be routed for `GCPVertexAI`, then only the `GCPVertexAIFields` would be included in the upstream request

sukumargaonkar: the translator will be called after the routing decision. so each translator is backend-api specific 
that way the translator gets to choose which fields it wants to read
https://github.com/envoyproxy/ai-gateway/pull/961/files#diff-d51f70dad4d583c82b80ccbb57ef387ede21eb8e5ea91e37eef8d14441e8e7f7R276-R301

gavrissh: During the translator to provider, we do pass the raw request body as a param - https://github.com/envoyproxy/ai-gateway/blob/main/internal/extproc/translator/openai_gcpvertexai.go#L38

Could we pick only the vendor-specific fields from there? This way, we would not have a field name collision across vendors.

thoughts?

sukumargaonkar:
not sure i follow.

with field name collision i meant, if in future multiple vendor add the same field but with different meaning in their specific api's
eg. (hypothetically) anthropic and vllm adding `cache-control` but with different sub-fields

docs/proposals/004-vendor-specific-fields/proposal.md
sukumargaonkar: not sure i follow.

with field name collision i meant, if in future multiple vendor add the same field but with different meaning in their specific api's
eg. (hypothetically) anthropic and vllm adding cache-control but with different sub-fields

sukumargaonkar: after discussions on slack.
updated the PR as per gavrissh's code snippet above

internal/apischema/openai/openai.go
```diff
@@ -1141,3 +1147,19 @@ func (t *JSONUNIXTime) UnmarshalJSON(s []byte) error {
 func (t JSONUNIXTime) Equal(other JSONUNIXTime) bool {
 	return time.Time(t).Unix() == time.Time(other).Unix()
 }
+
+// GCPVertexAIVendorFields contains GCP Vertex AI (Gemini) vendor-specific fields.
+type GCPVertexAIVendorFields struct {
+	// GenerationConfig holds Gemini generation configuration options.
```
mathetake, line 1153: can you put a link to the GCP's documentation?

internal/apischema/openai/openai.go
```diff
@@ -1141,3 +1147,19 @@ func (t *JSONUNIXTime) UnmarshalJSON(s []byte) error {
...
+	GenerationConfig *GCPVertexAIGenerationConfig `json:"generationConfig,omitzero"`
+}
+
+// GCPVertexAIGenerationConfig represents Gemini generation configuration options.
+type GCPVertexAIGenerationConfig struct {
+	ThinkingConfig *genai.GenerationConfigThinkingConfig `json:"thinkingConfig,omitzero"`
```
mathetake, line 1165: same comment, let's put a reference for the future contributors

internal/apischema/openai/openai.go
```diff
@@ -1141,3 +1147,19 @@ func (t *JSONUNIXTime) UnmarshalJSON(s []byte) error {
...
+}
+
+// GCPAnthropicVendorFields contains GCP Anthropic-specific fields.
+type GCPAnthropicVendorFields struct {
+	Thinking *anthropic.ThinkingConfigParamUnion `json:"thinking,omitzero"`
```
mathetake, line 1173: same

internal/extproc/translator/openai_gcpvertexai.go
```diff
@@ -271,9 +271,34 @@ func (o *openAIToGCPVertexAITranslatorV1ChatCompletion) openAIMessageToGeminiMes
 		SystemInstruction: systemInstruction,
 	}
 
+	// Apply vendor-specific fields after standard OpenAI-to-Gemini translation.
+	// Vendor fields take precedence over translated fields when conflicts occur.
+	o.applyVendorSpecificFields(openAIReq, &gcr)
+
 	return gcr, nil
```
mathetake: Can you make this `openAIMessageToGeminiMessage` function return pointer? 

Now that we call applyVendorSpecificFields, we escape `gcp.GenerateContentRequest` to heap here inside this function (depending on how Go compiler works though. Escape analysis might work this case but we cannot be sure.) Since at the end of the day, at the call site of openAIMessageToGeminiMessage, we also end up escaping `gcp.GenerateContentRequest` when calling json.Marshal, returning already a pointer from here makes sense because that allows us to have only one escape vs twice

internal/apischema/openai/vendor_fields_test.go
```diff
@@ -0,0 +1,244 @@
...
+			name: "Request with multiple vendor fields",
+			jsonData: []byte(`{
+				"model": "claude-3",
+				"messages": [
+					{
+						"role": "user",
+						"content": "Multiple vendors test"
+					}
+				],
+				"GCPVertexAI": {
```
yuzisun: is the test correct here? in the json body request we no longer have this vendor name

docs/proposals/004-vendor-specific-fields/proposal.md
```diff
@@ -0,0 +1,83 @@
...
+    GenerationConfig *GCPVertexAIGenerationConfig `json:"generationConfig,omitempty"`
+}
+
+// GCPVertexAIGenerationConfig represents Gemini generation configuration options.
+type GCPVertexAIGenerationConfig struct {
+    ThinkingConfig *genai.GenerationConfigThinkingConfig `json:"thinkingConfig,omitempty"`
+}
+
+// GCPAnthropicVendorFields contains GCP Anthropic-specific fields.
+type GCPAnthropicVendorFields struct {
```
yuzisun: GCPAnthropicVendorFields -> AnthropicVendorFields, doc also need to reflect the change

yuzisun (approved):

PR #963, opened 2025-07-29
controller: eliminate EEP in favor of extension server logic

mathetake:
**Description**

WIP


Xunzhuo (commented):
let me try to understand the changes:
1. we were generated eep at gateway level, so when a aigwroute refered to an gateway, we cannot use any self created httproute
2. currently we dont generate it anymore, we patch the ai-gw extproc at extensions server, and enabled it in auto-generated httproute (there is an annotation to determine it). So we can attach self-created httproute to an aigwroute.

do i miss anything else for this change?

Xunzhuo (commented):
a nit question is that, should we add owner reference on self-created httproute?

internal/extensionserver/post_translate_modify.go
```diff
@@ -490,3 +499,188 @@ func (s *Server) patchVirtualHostWithInferencePool(vh *routev3.VirtualHost, infe
...
+}
+
+func shouldAIGatewayExtProcBeInserted(filters []*httpconnectionmanagerv3.HttpFilter) bool {
+	for _, f := range filters {
+		if f.Name == aiGatewayExtProcName ||
+			// This is a backwards compatibility check for the old filter name generated by the EnvoyExtensionPolicy.
+			// https://github.com/envoyproxy/ai-gateway/blob/v0.2.1/internal/controller/gateway.go#L133
+			//
+			// TODO: remove after v0.3 release.
+			strings.Contains(f.Name, "ai-eg-eep-") {
```
Xunzhuo, line 645: trying to understand this logic, when an old version of envoy ai gw upgraded, what will happen, will there be two aigw extproc? and if the old eep policy should be deleted directly

mathetake:
> should we add owner reference on self-created httproute?

wait, I was completely under the impression that we have one, but looks like i don't...! Definitely we should do that.

mathetake:
> a nit question is that, should we add owner reference on self-created httproute?

We already have: https://github.com/envoyproxy/ai-gateway/blob/7deecd02e23f11b0e4dd94b42873d4a4cae27899/internal/controller/ai_gateway_route.go#L188

internal/extensionserver/post_translate_modify.go
mathetake: No, there won't be two extproc. Their filter chain will remain exactly the same because of this backward compat if statement (except that typed_filter_config is added unconditionally but the key won't exist for them so no problem), and users are encouraged to delete EEP by hand after upgrade, which results in switching to this new version. That's the migration path I described in the PR description.

Xunzhuo:
@mathetake yep, i mean should we add owner reference on the httproute not generated by aigwroute

mathetake:
why would you want to add ?

Xunzhuo (approved):

Xunzhuo:
> why would you want to add ?

i mean we shouldnt add

mathetake:
I think forcing users to eliminate EEP by hand seems weird. I will do a follow up PR to auto-delete the existing eep

PR #972, opened 2025-07-29
test: openinference span recording from python

codefromthecrypt:
**Description**

This creates the tests which gate completion of tracing support (#791). These are intentionally disabled as functionality isn't yet written, but raising this as a separate part allows smaller code review in the future, and also allows others to participate without blocking (e.g. by skipping only certain cassettes, e.g. at first all but "chat-basic")

**Related issues**

#791

It is important that we know what the span data expectations are, hence the testopeninference package. This was one of the most difficult things to figure out, so I hope this helps. docker won't run unless we re-record the spans. Also, in the future, we can pin the python requirements.txt for a specific openinference version to know what we are aiming to be compatible with.

cc @mikeldking maybe something like this could eventually go upstream, granted that we need a copy here for parity with the cassettes we are recording, because the cassettes are used for reasons besides otel. food for thought.


go.mod
```diff
@@ -344,6 +360,7 @@ require (
 	github.com/securego/gosec/v2 v2.22.2 // indirect
 	github.com/sergi/go-diff v1.3.2-0.20230802210424-5b0b94c5c0d3 // indirect
 	github.com/shirou/gopsutil/v3 v3.24.5 // indirect
```
codefromthecrypt, line 362: ps I think when envoy gateway gets bumped, it will include new func-e and hopefully kill the old gopsutil dip

codefromthecrypt:
sorry I need to rewrite to increase coverage

codefromthecrypt:
ok should be good now. I also added a "why" section

---

## Why is this needed?

This package bridges two open source communities:

* Envoy AI Gateway, who are experts in GenAI routing and policy, and typically  program in Go and C++.
* OpenInference, who are experts in GenAI observability and typically program  in Python and JavaScript.

As this project grows from dozens to hundreds of request shapes, it is important to have a standardized bridge between the two communities, so that troubleshooting or getting to the same page is easy.

PR #979, opened 2025-07-31
feat: support inferencepool status

Xunzhuo:
**Description**


tests/e2e/testdata/conformance.yaml
```diff
@@ -0,0 +1,11 @@
+# Copyright Envoy AI Gateway Authors
+# SPDX-License-Identifier: Apache-2.0
+# The full text of the Apache license is available in the LICENSE file at
+# the root of the repo.
+
```
mathetake: can you rename this file to `inference-extension-conformance.yaml`

tests/e2e/conformance_test.go
```diff
@@ -0,0 +1,46 @@
...
+	options.ConformanceProfiles.Insert(gie.GatewayLayerProfileName)
+	defaultTimeoutConfig := config.DefaultTimeoutConfig()
+	defaultTimeoutConfig.HTTPRouteMustHaveCondition = 10 * time.Second
+	defaultTimeoutConfig.HTTPRouteMustNotHaveParents = 10 * time.Second
+	defaultTimeoutConfig.GatewayMustHaveCondition = 10 * time.Second
+	config.SetupTimeoutConfig(&defaultTimeoutConfig)
+	options.TimeoutConfig = defaultTimeoutConfig
+	options.GatewayClassName = "inference-pool"
+	options.SkipTests = []string{
+		"EppUnAvailableFailOpen",
```
mathetake, line 43: let me confirm my understanding; so to pass this, we have to be able to configure the failureModeAllow: false/true of inference pool extprocs? https://github.com/envoyproxy/ai-gateway/blob/a35e9adae8d7ce655b4fff0602ef0cce4a67b8dd/internal/extensionserver/inferencepool.go#L289

tests/e2e/conformance/envoy-ai-gateway-report.yaml
```diff
@@ -0,0 +1,30 @@
+# Copyright Envoy AI Gateway Authors
```
mathetake: I am not sure if we need to check in this file. Can we delete? Instead, you can add this to .gitignore i guess?

tests/e2e/conformance_test.go
```diff
@@ -0,0 +1,46 @@
...
+	v1 "sigs.k8s.io/gateway-api/conformance/apis/v1"
+	"sigs.k8s.io/gateway-api/conformance/utils/config"
+)
+
+func TestGatewayAPIInferenceExtension(t *testing.T) {
+	const manifest = "testdata/conformance.yaml"
+	require.NoError(t, kubectlApplyManifest(t.Context(), manifest))
+
+	options := gie.DefaultOptions(t)
+	options.ReportOutputPath = "./conformance/envoy-ai-gateway-report.yaml"
```
mathetake:
```suggestion
	options.ReportOutputPath = "infernece-extension-conformance-test-report.yaml"
```

Also let's add `infernece-extension-conformance-test-report.yaml` to .gitignore as per my comment

internal/controller/controller.go
```diff
@@ -135,6 +135,16 @@ func StartControllers(ctx context.Context, mgr manager.Manager, config *rest.Con
...
 
+	inferencePoolC := NewInferencePoolController(c, kubernetes.NewForConfigOrDie(config), logger.
+		WithName("inference-pool"))
+	if err = TypedControllerBuilderForCRD(mgr, &gwaiev1a2.InferencePool{}).
+		Watches(&gwapiv1.Gateway{}, inferencePoolC.gatewayEventHandler()).
+		Watches(&aigv1a1.AIGatewayRoute{}, inferencePoolC.routeEventHandler()).
+		Watches(&gwapiv1.HTTPRoute{}, inferencePoolC.httpRouteEventHandler()).
+		Complete(inferencePoolC); err != nil {
+		return fmt.Errorf("failed to create controller for InferencePool: %w", err)
+	}
```
mathetake: does this not fail for the cluster where inferencepool CRD is not installed? If this fails, then shall we make this optional by simply emitting the error log like ("failed to start the inference pool cotroller. This is most likely due to Gateway API Inference Extension CRDs are not installed on the cluster. Please referece to the Envoy AI Gateawy documentation for details."). Alternativel, we can add a flag to the cnotroller like `--inferenceExtensionEnabled` etc

internal/controller/inference_pool.go
```diff
@@ -0,0 +1,409 @@
...
+		}
+	}
+
+	return referencedGateways, nil
+}
+
+// validateExtensionReference checks if the ExtensionReference service exists.
+func (c *InferencePoolController) validateExtensionReference(ctx context.Context, inferencePool *gwaiev1a2.InferencePool) error {
+	// Check if ExtensionRef is specified.
+	if inferencePool.Spec.ExtensionRef == nil {
```
mathetake, line 125: isn't this part of CEL validation? Do we need this?

internal/controller/inference_pool.go
```diff
@@ -0,0 +1,409 @@
...
+		if string(parentRef.Name) != gatewayName {
+			continue
+		}
+
+		// Check namespace - if not specified in parentRef, it defaults to the route's namespace.
+		// But since we're checking cross-namespace references, we need to check if the namespace matches.
+		if parentRef.Namespace != nil {
+			if string(*parentRef.Namespace) == gatewayNamespace {
+				return true
+			}
```
mathetake, line 206:
```suggestion
		if parentRef.Namespace != nil && string(*parentRef.Namespace) == gatewayNamespace {
			return true
        }
```

internal/controller/inference_pool.go
```diff
@@ -0,0 +1,409 @@
...
+
+// validateExtensionReference checks if the ExtensionReference service exists.
+func (c *InferencePoolController) validateExtensionReference(ctx context.Context, inferencePool *gwaiev1a2.InferencePool) error {
+	// Check if ExtensionRef is specified.
+	if inferencePool.Spec.ExtensionRef == nil {
+		return nil // No extension reference to validate.
+	}
+
+	// Get the service name from ExtensionReference.
+	serviceName := inferencePool.Spec.ExtensionRef.Name
```
mathetake, line 130: same comment. It seems that `name` is required at the CEL level but I see no length requirement. I think we should fix the upstream spec? It's fine to have this validation at this point though.

internal/controller/inference_pool.go
```diff
@@ -0,0 +1,409 @@
...
+
+	// Service exists - validation passed.
+	return nil
+}
+
+// gatewayReferencesInferencePool checks if a Gateway references the given InferencePool through any routes.
+func (c *InferencePoolController) gatewayReferencesInferencePool(ctx context.Context, gateway *gwapiv1.Gateway, inferencePoolName string, inferencePoolNamespace string) bool {
+	// Check AIGatewayRoutes in the same namespace as the InferencePool that reference this Gateway.
+	var aiGatewayRoutes aigv1a1.AIGatewayRouteList
+	if err := c.client.List(ctx, &aiGatewayRoutes, client.InNamespace(inferencePoolNamespace)); err != nil {
```
mathetake, line 157: you should use the index as in `gateway.go`:

```go
	var routes aigv1a1.AIGatewayRouteList
	err := c.client.List(ctx, &routes, client.MatchingFields{
		k8sClientIndexAIGatewayRouteToAttachedGateway: fmt.Sprintf("%s.%s", gateway.Name, gateway.Namespace),
	})
	if err != nil {
		return ctrl.Result{}, err
	}
```

internal/controller/inference_pool.go
```diff
@@ -0,0 +1,409 @@
...
+
+	for _, route := range aiGatewayRoutes.Items {
+		// Check if this route references the Gateway.
+		if c.routeReferencesGateway(route.Spec.ParentRefs, gateway.Name, gateway.Namespace) {
+			// Check if this route references the InferencePool.
+			if c.routeReferencesInferencePool(&route, inferencePoolName) {
+				return true
+			}
+		}
+	}
```
mathetake: You are escaping the route unnecessarily

```suggestion
	for i := range aiGatewayRoutes.Items {
        route := &aiGatewayRoutes.Items[i]
		// Check if this route references the Gateway.
		if c.routeReferencesGateway(route.Spec.ParentRefs, gateway.Name, gateway.Namespace) {
			// Check if this route references the InferencePool.
			if c.routeReferencesInferencePool(route, inferencePoolName) {
				return true
			}
		}
	}
```

internal/controller/inference_pool.go
```diff
@@ -0,0 +1,409 @@
...
+
+	for _, route := range httpRoutes.Items {
+		// Check if this route references the Gateway.
+		if c.routeReferencesGateway(route.Spec.ParentRefs, gateway.Name, gateway.Namespace) {
+			// Check if this route references the InferencePool.
+			if c.httpRouteReferencesInferencePool(&route, inferencePoolName) {
+				return true
+			}
+		}
+	}
```
mathetake: same comment
```suggestion
	for i := range httpRoutes.Items {
        route := &httpRoutes.Items[i]
		// Check if this route references the Gateway.
		if c.routeReferencesGateway(route.Spec.ParentRefs, gateway.Name, gateway.Namespace) {
			// Check if this route references the InferencePool.
			if c.httpRouteReferencesInferencePool(route, inferencePoolName) {
				return true
			}
		}
	}
```

internal/controller/inference_pool.go
```diff
@@ -0,0 +1,409 @@
...
+
+	referencedGateways := make(map[string]*gwapiv1.Gateway)
+
+	// Check each Gateway to see if it references this InferencePool through routes.
+	for _, gw := range gateways.Items {
+		if c.gatewayReferencesInferencePool(ctx, &gw, inferencePool.Name, inferencePool.Namespace) {
+			gatewayKey := fmt.Sprintf("%s/%s", gw.Namespace, gw.Name)
+			referencedGateways[gatewayKey] = &gw
+		}
+	}
```
mathetake: same comment about escaping
```suggestion
	for i := range gateways.Items {
        gw := &gateways.Items[i]
		if c.gatewayReferencesInferencePool(ctx, gw, inferencePool.Name, inferencePool.Namespace) {
			gatewayKey := fmt.Sprintf("%s/%s", gw.Namespace, gw.Name)
			referencedGateways[gatewayKey] = gw
		}
	}
```

internal/controller/inference_pool.go
```diff
@@ -0,0 +1,409 @@
...
+				})
+			}
+		}
+
+		return requests
+	})
+}
+
+// routeEventHandler returns an event handler for AIGatewayRoute resources.
+func (c *InferencePoolController) routeEventHandler() handler.EventHandler {
```
mathetake: Let's be more explicit about the name to be consistent with httpRouteEventHandler below
```suggestion
// aiGatewayRouteEventHandler returns an event handler for AIGatewayRoute resources.
func (c *InferencePoolController) aiGatewayRouteEventHandler() handler.EventHandler {
```

mathetake (commented):
This is exciting. Thank you for taking a shot

tests/e2e/conformance_test.go
Xunzhuo: no there is an upstream bug, this testcase will impact other tests, so actually we can pass all the tests, but will enable this after upstream bug fixed

mathetake: could you put an upstream issue link here in the code comment?

internal/controller/inference_pool.go
Xunzhuo: we dont have cel for inference pool crd itself, I think, it might be an upstream cel rule

tests/e2e/conformance_test.go
Xunzhuo: done

internal/controller/inference_pool.go
Xunzhuo: we cannot list aigwroute by gateway namespace, inference pool can cross namespace with gateway, so if we need to find all refered aigwroute, we need to list it with inference pool ns

mathetake (approved):

PR #981, opened 2025-07-31
feat: add configurable HTTP request headers to metrics

### LINKED ISSUE #848: Flexible metrics label customization without building a custom extproc

surenraju-careem:
**Description**

This commit adds support for configuring additional metrics labels corresponding to request headers. Notably, this adds a new flag to the controller `metricsRequestHeaderLabelMapping` that will be validated and propagated to the extproc.

The flag is in the format of the comma separated key value pairs as in "header1:label1,header2:label2,..."

Fixes #848


surenraju-careem:
@mathetake I’ve moved this to a new PR, as the previous one had several merge conflicts due to many recent changes merged into main.

mathetake:
btw you didn't need to create a new PR just because there's a merge conflict. It's important to know the history of the discussion when people look at the merged commit linking to one PR, not the brand new PR without the previous discussion

examples/extproc_custom_metrics/main.go
```diff
@@ -69,6 +69,10 @@ func (m *myCustomChatCompletionMetrics) RecordTokenUsage(_ context.Context, inpu
 	m.logger.Info("RecordTokenUsage", "inputTokens", inputTokens, "outputTokens", outputTokens, "totalTokens", totalTokens)
 }
 
+func (m *myCustomChatCompletionMetrics) RecordTokenUsageWithHeaders(_ context.Context, headers map[string]string, inputTokens, outputTokens, totalTokens uint32, _ ...attribute.KeyValue) {
```
mathetake: can you instead of add this new method, simply chance RecordTokenUsage to accept "request_headers" (not just headers").

internal/controller/gateway_mutator.go
```diff
@@ -71,6 +73,15 @@ func (g *gatewayMutator) Default(ctx context.Context, obj runtime.Object) error
 	return nil
 }
 
+// buildMetricsHeaderLabelArgs builds the command line arguments for metrics header label mapping.
+func (g *gatewayMutator) buildMetricsHeaderLabelArgs() []string {
```
mathetake: Can you do all arg constructions here, not limited to header labels args?
```suggestion
func (g *gatewayMutator) buildExtProcArgs() []string {
```

cmd/extproc/extproc
mathetake, line 1: delete this please?

cmd/controller/main.go
```diff
@@ -54,6 +56,42 @@ func parsePullPolicy(s string) (corev1.PullPolicy, error) {
...
+	}
+
+	result := make(map[string]string)
+	pairs := strings.Split(s, ",")
+
+	for i, pair := range pairs {
+		pair = strings.TrimSpace(pair)
+		if pair == "" {
+			continue
+		}
```
mathetake: when does this happen and shouldn't we emit the error for such invalid pattern? At least can you put code comments here to explain the rationale?

cmd/controller/main.go
```diff
@@ -31,17 +32,18 @@ import (
...
+	extProcImagePullPolicy           corev1.PullPolicy
+	enableLeaderElection             bool
+	logLevel                         zapcore.Level
+	extensionServerPort              string
+	tlsCertDir                       string
+	tlsCertName                      string
+	tlsKeyName                       string
+	caBundleName                     string
+	envoyGatewayNamespace            string
+	metricsRequestHeaderLabelMapping string
```
mathetake, line 46: could you store map here already since you are unnecessarily calling `parseRequestHeaderLabelMapping` twice due to this?
```suggestion
	metricsRequestHeaderLabelMapping map[string]string
```

surenraju-careem: The reason I didn’t convert it to a map is that we would end up converting it back to a string to pass it to extProc. I didn’t see any value in converting it to a map unnecessarily.

mathetake: ah now i see that two `parseRequestHeaderLabelMapping` exists with exactly the same logic. Could you put `parseRequestHeaderLabelMapping` in `internalapi` and share the code?

mathetake: and using `string` makes sense here on the other hand. Thank you for explaining it

cmd/extproc/mainlib/main.go
```diff
@@ -85,6 +91,42 @@ func parseAndValidateFlags(args []string) (extProcFlags, error) {
 	return flags, errors.Join(errs...)
 }
 
+// parseRequestHeaderLabelMapping parses comma-separated key-value pairs for header-to-label mapping.
+// The input format is "header1:label1,header2:label2" where header names are HTTP request
+// headers and label names are Prometheus metric labels.
+// Example: "x-team-id:team_id,x-user-id:user_id".
+func parseRequestHeaderLabelMapping(s string) (map[string]string, error) {
```
mathetake: can you put this function into `internal/internalapi` and share the code between controller and extproc to remove the duplicate code and testsing

cmd/controller/main.go
surenraju-careem: I thought to handle scenarios like "x-team-id:team_id,x-user-id:user_id," or "x-team-id:team_id,,x-user-id:user_id".

But its okay to fail fast as well.

mathetake: yes, let's return error. I don't think we need to silence

cmd/extproc/mainlib/main.go
surenraju-careem: ok

internal/metrics/base_metrics.go
```diff
@@ -58,20 +60,30 @@ func (b *baseMetrics) SetBackend(backend *filterapi.Backend) {
...
 	attrs = append(attrs, extraAttrs...)
+
+	// Add header values as attributes based on the header mapping if headers are provided.
+	if headers != nil {
+		for headerName, labelName := range b.requestHeaderLabelMapping {
+			if headerValue, exists := headers[headerName]; exists {
+				attrs = append(attrs, attribute.Key(labelName).String(headerValue))
+			}
+		}
+	}
```
mathetake: Reading map on nil map doesn't panic so this can be simplified
```suggestion
	for headerName, labelName := range b.requestHeaderLabelMapping {
		if headerValue, exists := headers[headerName]; exists {
			attrs = append(attrs, attribute.Key(labelName).String(headerValue))
		}
	}
```

mathetake (approved):

mathetake:
Thank you for your multiple iterations and contribution!!! @surenraju-careem  🏅

PR #984, opened 2025-08-01
feat: support gcp anthropic streaming

alexagriffith:
**Description**
adding streaming translation for gcp anthropic / openai


internal/apischema/openai/openai.go
```diff
@@ -1004,7 +1004,7 @@ type ChatCompletionResponseChunkChoice struct {
 // https://platform.openai.com/docs/api-reference/chat/streaming#chat/streaming-choices
 type ChatCompletionResponseChunkChoiceDelta struct {
 	Content   *string                              `json:"content,omitempty"`
-	Role      string                               `json:"role"`
+	Role      string                               `json:"role,omitempty"`
```
alexagriffith, line 1011: I think we have a bug here actually and this should be omit if empty because it seems like we should only add role on initial chunk

https://platform.openai.com/docs/api-reference/chat_streaming/streaming

see also the example 

```
{"id":"chatcmpl-123","object":"chat.completion.chunk","created":1694268190,"model":"gpt-4o-mini", "system_fingerprint": "fp_44709d6fcb", "choices":[{"index":0,"delta":{"role":"assistant","content":""},"logprobs":null,"finish_reason":null}]}

{"id":"chatcmpl-123","object":"chat.completion.chunk","created":1694268190,"model":"gpt-4o-mini", "system_fingerprint": "fp_44709d6fcb", "choices":[{"index":0,"delta":{"content":"Hello"},"logprobs":null,"finish_reason":null}]}

....

{"id":"chatcmpl-123","object":"chat.completion.chunk","created":1694268190,"model":"gpt-4o-mini", "system_fingerprint": "fp_44709d6fcb", "choices":[{"index":0,"delta":{},"logprobs":null,"finish_reason":"stop"}]}
```

alexagriffith: I can make an issue

alexagriffith:
testupstream code coverage too low even though I added to it.... 🤔 ... will fix

internal/apischema/openai/openai.go
yuzisun: hmm on the openai schema it says mandatory
Screenshot 2025-08-02 at 6 14 27 AM

alexagriffith: it shows role only in first chunk here also https://platform.openai.com/docs/guides/streaming-responses?api-mode=chat#read_the_responses

gavrissh: Yes, in the OpenAI stream response, the role is set only in the first chunk. We have seen similar response with vLLM as well

alexagriffith: It says either role OR content can be set, so not both

alexagriffith: but its confusing, the docs dont say optional and https://github.com/openai/openai-go/blob/3fb5d8d68e5e7a74fa5d666550904d26d3f34bc8/chatcompletion.go#L659 the sdk shows role as just string, not omitempty or pointer

alexagriffith: additionally this adds to the confusion a bit, for tool_calls - where the other fields seem to be omit empty like id, type in the previous examples, these seem to be null and the go sdk doesn't omitempty or null (ptr) these 

https://platform.openai.com/docs/guides/function-calling?api-mode=chat#streaming

```
[{"index": 0, "id": "call_DdmO9pD3xa9XTPNJ32zg2hcA", "function": {"arguments": "", "name": "get_weather"}, "type": "function"}]
[{"index": 0, "id": null, "function": {"arguments": "{\"", "name": null}, "type": null}]
[{"index": 0, "id": null, "function": {"arguments": "location", "name": null}, "type": null}]
[{"index": 0, "id": null, "function": {"arguments": "\":\"", "name": null}, "type": null}]
[{"index": 0, "id": null, "function": {"arguments": "Paris", "name": null}, "type": null}]
[{"index": 0, "id": null, "function": {"arguments": ",", "name": null}, "type": null}]
[{"index": 0, "id": null, "function": {"arguments": " France", "name": null}, "type": null}]
[{"index": 0, "id": null, "function": {"arguments": "\"}", "name": null}, "type": null}]
null
```

.testcoverage.yml
```diff
@@ -23,5 +23,6 @@ exclude:
     - tests/internal/testenvironment/
     - tests/internal/testopeninference/openai_proxy.go
     - internal/testing/
+    - tests/internal/testupstreamlib/
```
alexagriffith: do we want this?

tests/internal/testupstreamlib/testupstream/main.go
```diff
@@ -257,25 +258,59 @@ func handler(w http.ResponseWriter, r *http.Request) {
...
+				}
+				logger.Println("response block sent:", string(block))
 			}
 
-			if f, ok := w.(http.Flusher); ok {
-				f.Flush()
-			} else {
-				panic("expected http.ResponseWriter to be an http.Flusher")
+		} else {
+			logger.Println("detected line-by-line stream, formatting as SSE")
```
yuzisun, line 290: remove the debugging line

.testcoverage.yml
mathetake: can you not add to this unless it's difficult? I don't think it's difficult to cover the lines you added

internal/extproc/translator/openai_gcpanthropic_stream.go
```diff
@@ -0,0 +1,357 @@
...
+type streamingToolCall struct {
+	ID         string
+	Name       string
+	InputJSON  bytes.Buffer
+	SentHeader bool
+}
+
+// AnthropicStreamParser manages the stateful translation of an Anthropic SSE stream
+// to an OpenAI-compatible SSE stream.
+type AnthropicStreamParser struct {
```
mathetake:
```suggestion
type anthropicStreamParser struct {
```

internal/extproc/translator/openai_gcpanthropic_stream.go
```diff
@@ -0,0 +1,357 @@
...
+	sseDataPrefix  = "data: "
+	sseDoneMessage = "[DONE]"
+)
+
+// streamingToolCall holds the state for a single tool call that is being streamed.
+type streamingToolCall struct {
+	ID         string
+	Name       string
+	InputJSON  bytes.Buffer
+	SentHeader bool
```
mathetake:
```suggestion
	id         string
	name       string
	inputJSON  bytes.Buffer
	sentHeader bool
```

internal/extproc/translator/openai_gcpanthropic_stream.go
```diff
@@ -0,0 +1,357 @@
...
+	sseEventPrefix = "event:"
+	sseDataPrefix  = "data: "
+	sseDoneMessage = "[DONE]"
+)
+
+// streamingToolCall holds the state for a single tool call that is being streamed.
+type streamingToolCall struct {
+	ID         string
+	Name       string
+	InputJSON  bytes.Buffer
```
mathetake: i don't see any reason to use `bytes.Buffer` which results in unnecessarily copying string twice from `anthropic.ContentBlockDeltaEvent.PartialJson` to `bytes.Buffer ([]byte)` and eventually into a string again via `bytes.Buffer.String()`.
```suggestion
	InputJSON  string
```

tests/internal/testupstreamlib/testupstream/main.go
mathetake: this is a testupstream and such logs are scattered many places, so i think we are good with keeping it

.testcoverage.yml
alexagriffith: I covered the lines and it was still failing - I added a test .. let me add another test

internal/extproc/translator/openai_gcpanthropic_stream.go
alexagriffith: did this in ide so I could use refactor tool but messed it up when I tried it with ID lol. should be good now though

mathetake:
i haven't reviewed the actual stream parser code yet... will do by the end of tomorrow 100%...

internal/extproc/translator/openai_gcpanthropic_stream.go
```diff
@@ -0,0 +1,353 @@
...
+		} else if strings.HasPrefix(line, sseDataPrefix) {
+			// This handles JSON data that might be split across multiple 'data:' lines
+			// by concatenating them (Anthropic's format).
+			data := strings.TrimSpace(strings.TrimPrefix(line, sseDataPrefix))
+			eventData.WriteString(data)
+		}
+	}
+
+	// After checking all lines, if we found an event, handle it.
+	if eventType != "" {
```
xiaolin593: In case of SSE event like below with pure data per SSE spec, eventType will be nil but eventData is not empty, `parseAndHandleEvent` basically ignore `eventData`?
```
data: some text

data: another message
data: with two lines
```

internal/extproc/translator/openai_gcpanthropic_stream.go
```diff
@@ -0,0 +1,353 @@
...
+		if err := p.writeChunk(finalEventBlock, &responseBodyBuilder); err != nil {
+			return nil, nil, LLMTokenUsage{}, err
+		}
+	}
+
+	// After the loop, if it's the end of the stream and the buffer still contains
+	// the final event (which has no trailing \n\n), process it now.
+	if endOfStream && p.buffer.Len() > 0 {
+		finalEventBlock := p.buffer.String()
+		p.buffer.Reset() // Clear the buffer as we are done with it.
```
xiaolin593: Why check `endOfStream & p.buffer.len() > 0` twice? If first block fails, second block will never hit as `p.buffer` is reset?

alexagriffith: ah good catch this was left over from a refactor

mathetake:
Could you apply a patch like this? I don't think we need to use strings.Builder which resulted in the heap escape of the builder itself (since &strings.Builder was passed to writeChunk) and unnecessary copy of the written body due to`mut := &extprocv3.BodyMutation_Body{Body: []byte(finalBody)}`.


```diff
diff --git a/internal/extproc/translator/openai_gcpanthropic_stream.go b/internal/extproc/translator/openai_gcpanthropic_stream.go
index e4b6499..cb0347c 100644
--- a/internal/extproc/translator/openai_gcpanthropic_stream.go
+++ b/internal/extproc/translator/openai_gcpanthropic_stream.go
@@ -52,7 +52,7 @@ func newAnthropicStreamParser(modelName string) *anthropicStreamParser {
        }
 }
 
-func (p *anthropicStreamParser) writeChunk(eventBlock string, builder *strings.Builder) error {
+func (p *anthropicStreamParser) writeChunk(eventBlock string, buf *[]byte) error {
        chunk, err := p.parseAndHandleEvent(eventBlock)
        if err != nil {
                return err
@@ -63,9 +63,10 @@ func (p *anthropicStreamParser) writeChunk(eventBlock string, builder *strings.B
                if err != nil {
                        return fmt.Errorf("failed to marshal stream chunk: %w", err)
                }
-               builder.WriteString(sseDataPrefix)
-               builder.Write(chunkBytes)
-               builder.WriteString("\n\n")
+               // Write the chunk to the response body builder.
+               *buf = append(*buf, sseDataPrefix...)
+               *buf = append(*buf, chunkBytes...)
+               *buf = append(*buf, '\n', '\n')
        }
        return nil
 }
@@ -79,14 +80,14 @@ func (p *anthropicStreamParser) Process(body io.Reader, endOfStream bool) (
                return nil, nil, LLMTokenUsage{}, fmt.Errorf("failed to read from stream body: %w", err)
        }
 
-       var responseBodyBuilder strings.Builder
+       mut := &extprocv3.BodyMutation_Body{Body: nil}
        for {
                eventBlock, remaining, found := bytes.Cut(p.buffer.Bytes(), []byte("\n\n"))
                if !found {
                        break
                }
 
-               if err := p.writeChunk(string(eventBlock), &responseBodyBuilder); err != nil {
+               if err := p.writeChunk(string(eventBlock), &mut.Body); err != nil {
                        return nil, nil, LLMTokenUsage{}, err
                }
 
@@ -98,7 +99,7 @@ func (p *anthropicStreamParser) Process(body io.Reader, endOfStream bool) (
                finalEventBlock := p.buffer.String()
                p.buffer.Reset()
 
-               if err := p.writeChunk(finalEventBlock, &responseBodyBuilder); err != nil {
+               if err := p.writeChunk(finalEventBlock, &mut.Body); err != nil {
                        return nil, nil, LLMTokenUsage{}, err
                }
        }
@@ -142,16 +143,16 @@ func (p *anthropicStreamParser) Process(body io.Reader, endOfStream bool) (
                        if err != nil {
                                return nil, nil, LLMTokenUsage{}, fmt.Errorf("failed to marshal final stream chunk: %w", err)
                        }
-                       responseBodyBuilder.WriteString(sseDataPrefix)
-                       responseBodyBuilder.Write(chunkBytes)
-                       responseBodyBuilder.WriteString("\n\n")
+                       // Write the final chunk to the response body.
+                       mut.Body = append(mut.Body, sseDataPrefix...)
+                       mut.Body = append(mut.Body, chunkBytes...)
+                       mut.Body = append(mut.Body, '\n', '\n')
                }
-               responseBodyBuilder.WriteString(sseDataPrefix + sseDoneMessage + "\n\n")
+               mut.Body = append(mut.Body, sseDataPrefix...)
+               mut.Body = append(mut.Body, sseDoneMessage...)
+               mut.Body = append(mut.Body, '\n', '\n')
        }
-
-       finalBody := responseBodyBuilder.String()
-       mut := &extprocv3.BodyMutation_Body{Body: []byte(finalBody)}
-
        return &extprocv3.HeaderMutation{}, &extprocv3.BodyMutation{Mutation: mut}, p.tokenUsage, nil
 }
```

internal/extproc/translator/openai_gcpanthropic_stream.go
```diff
@@ -0,0 +1,332 @@
...
+			// This handles JSON data that might be split across multiple 'data:' lines
+			// by concatenating them (Anthropic's format).
+			data := strings.TrimSpace(strings.TrimPrefix(line, sseDataPrefix))
+			eventData.WriteString(data)
+		}
+	}
+
+	// After checking all lines, if we found an event, handle it.
+	if eventType != "" && eventData.Len() > 0 {
+		return p.handleAnthropicStreamEvent(eventType, []byte(eventData.String()))
```
mathetake: this also causes unnecessary copy of the data, so let's use []byte for eventData

```go
func (p *anthropicStreamParser) parseAndHandleEvent(eventBlock []byte) (*openai.ChatCompletionResponseChunk, error) {
	var eventType string
	var eventData []byte

	lines := bytes.Split(eventBlock, []byte{'\n'})
	for _, line := range lines {
		// Per the SSE spec, lines starting with a colon are comments and should be ignored.
		if bytes.HasPrefix(line, []byte{':'}) {
			continue
		}

		if bytes.HasPrefix(line, sseEventPrefix) {
			eventType = string(bytes.TrimSpace(bytes.TrimPrefix(line, sseEventPrefix)))
		} else if bytes.HasPrefix(line, sseDataPrefix) {
			// This handles JSON data that might be split across multiple 'data:' lines
			// by concatenating them (Anthropic's format).
			data := bytes.TrimSpace(bytes.TrimPrefix(line, sseDataPrefix))
			eventData = append(eventData, data...)
		}
	}

	// After checking all lines, if we found an event, handle it.
	if eventType != "" && len(eventData) > 0 {
		return p.handleAnthropicStreamEvent(eventType, eventData)
	}

	return nil, nil
}
```

alexagriffith: good feedback thanks

internal/extproc/translator/openai_gcpanthropic_stream.go
```diff
@@ -0,0 +1,332 @@
...
+	}
+
+	mut := &extprocv3.BodyMutation_Body{Body: nil}
+	for {
+		eventBlock, remaining, found := bytes.Cut(p.buffer.Bytes(), []byte("\n\n"))
+		if !found {
+			break
+		}
+
+		if err := p.writeChunk(string(eventBlock), &mut.Body); err != nil {
```
mathetake: let's remove this copy as well
```diff
diff --git a/internal/extproc/translator/openai_gcpanthropic_stream.go b/internal/extproc/translator/openai_gcpanthropic_stream.go
index 9783a49..42d44ae 100644
--- a/internal/extproc/translator/openai_gcpanthropic_stream.go
+++ b/internal/extproc/translator/openai_gcpanthropic_stream.go
@@ -52,7 +52,7 @@ func newAnthropicStreamParser(modelName string) *anthropicStreamParser {
        }
 }
 
-func (p *anthropicStreamParser) writeChunk(eventBlock string, buf *[]byte) error {
+func (p *anthropicStreamParser) writeChunk(eventBlock []byte, buf *[]byte) error {
        chunk, err := p.parseAndHandleEvent(eventBlock)
        if err != nil {
                return err
@@ -86,7 +86,7 @@ func (p *anthropicStreamParser) Process(body io.Reader, endOfStream bool) (
                        break
                }
 
-               if err := p.writeChunk(string(eventBlock), &mut.Body); err != nil {
+               if err := p.writeChunk(eventBlock, &mut.Body); err != nil {
                        return nil, nil, LLMTokenUsage{}, err
                }
 
@@ -95,7 +95,7 @@ func (p *anthropicStreamParser) Process(body io.Reader, endOfStream bool) (
        }
 
        if endOfStream && p.buffer.Len() > 0 {
-               finalEventBlock := p.buffer.String()
+               finalEventBlock := p.buffer.Bytes()
                p.buffer.Reset()
 
                if err := p.writeChunk(finalEventBlock, &mut.Body); err != nil {
@@ -156,7 +156,7 @@ func (p *anthropicStreamParser) Process(body io.Reader, endOfStream bool) (
        return &extprocv3.HeaderMutation{}, &extprocv3.BodyMutation{Mutation: mut}, p.tokenUsage, nil
 }
 
-func (p *anthropicStreamParser) parseAndHandleEvent(eventBlock string) (*openai.ChatCompletionResponseChunk, error) {
+func (p *anthropicStreamParser) parseAndHandleEvent(eventBlock []byte) (*openai.ChatCompletionResponseChunk, error) {
        var eventType string
        var eventData []byte
```

internal/extproc/translator/openai_gcpanthropic_stream.go
```diff
@@ -0,0 +1,332 @@
...
+func (p *anthropicStreamParser) parseAndHandleEvent(eventBlock string) (*openai.ChatCompletionResponseChunk, error) {
+	var eventType string
+	var eventData []byte
+
+	lines := strings.Split(eventBlock, "\n")
+	for _, line := range lines {
+		// Per the SSE spec, lines starting with a colon are comments and should be ignored.
+		if strings.HasPrefix(line, ":") {
+			continue
+		}
```
mathetake: i think we can remove this check as a few lines below we only deal with the matching lines anyways?

mathetake (commented):
almos there and this is exciting !

internal/extproc/translator/openai_gcpanthropic_stream.go
alexagriffith: ok just then we need to convert every split case delimiter to a byte

internal/extproc/translator/openai_gcpanthropic_stream.go
alexagriffith: good catch

internal/extproc/translator/openai_gcpanthropic_stream.go
alexagriffith: one thing with eventType makes it kind of weird, we can use string and pass it in whereas we use bytes for the rest, or we can have parseAndHandleEvent use bytes then convert to string in handleAnthropicStreamEvent which is what I did, the case switch are constants, so it seems better to compare strings there? lmk what you think

internal/extproc/translator/openai_gcpanthropic_stream.go
```diff
@@ -0,0 +1,326 @@
...
+	extprocv3 "github.com/envoyproxy/go-control-plane/envoy/service/ext_proc/v3"
+
+	"github.com/envoyproxy/ai-gateway/internal/apischema/openai"
+)
+
+const (
+	sseEventPrefix = "event:"
+	sseDataPrefix  = "data: "
+	sseDoneMessage = "[DONE]"
+)
```
mathetake:
```suggestion
var (
	sseEventPrefix = []byte("event:")
	sseDataPrefix  = []byte("data: ")
	sseDoneMessage = []byte("[DONE]")
)
```

mathetake (commented):
one nit but then good to go!

mathetake (approved):
LGTM,  nice work!!

PR #983, opened 2025-07-31
site: v0.3 release notes

missBerg:
**Description**

v0.3 Release Notes.

**Notes to Reviewers:**
Currently in draft. Raising the Draft PR so others can help add notes and edit.

PR #988, opened 2025-08-02
controller: adds support for GatewayNamespace mode

### LINKED ISSUE #976: Support for GatewayNamespace mode
*Description*:
Gateway instance is not Running when Gateway instance deploy in GatewayNamespace mode, and just work in ControllerNamespace mode.
The about resource is created by AI controller in system namespace(envoy-gateway-system), example the mounted secret in the proxy pod.

*Environment*:
ai-gateway: v0.2.1
gateway: v1.4.2

*Logs*:
Image

mathetake:
**Description**

Closes #976

### DIFF
--- a/cmd/aigw/translate.go
+++ b/cmd/aigw/translate.go
@@ -160,10 +160,6 @@ func collectObjects(yamlInput string, out io.Writer, logger *slog.Logger) (
 	}
 }
 
-const (
-	envoyGatewayNamespace = "envoy-gateway-system"
-)
-
 // translateCustomResourceObjects translates the AI Gateway custom resources to Envoy Gateway and Kubernetes objects.
 func translateCustomResourceObjects(
 	ctx context.Context,
@@ -214,7 +210,7 @@ func translateCustomResourceObjects(
 		make(chan event.GenericEvent),
 	)
 	gwC := controller.NewGatewayController(fakeClient, fakeClientSet, logr.FromSlogHandler(logger.Handler()),
-		envoyGatewayNamespace, extProcUDSPath, "docker.io/envoyproxy/ai-gateway-extproc:latest",
+		extProcUDSPath, "docker.io/envoyproxy/ai-gateway-extproc:latest",
 	)
 	// Create and reconcile the custom resources to store the translated objects.
 	// Note that the order of creation is important as some objects depend on others.
--- a/internal/controller/controller.go
+++ b/internal/controller/controller.go
@@ -62,8 +62,6 @@ type Options struct {
 	// EnableLeaderElection enables leader election for the controller manager.
 	// Enabling this ensures there is only one active controller manager.
 	EnableLeaderElection bool
-	// EnvoyGatewayNamespace is the namespace where the Envoy Gateway system resources are deployed.
-	EnvoyGatewayNamespace string
 	// UDSPath is the path to the UDS socket for the external processor.
 	UDSPath string
 	// DisableMutatingWebhook disables the mutating webhook for the Gateway for testing purposes.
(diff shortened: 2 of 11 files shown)


manifests/charts/ai-gateway-helm/values.yaml
```diff
@@ -18,9 +18,6 @@ controller:
   logLevel: info
   nameOverride: ""
   fullnameOverride: "ai-gateway-controller"
-  # This is the namespace where the Envoy Gateway controller is installed.
-  # It is hence where the gateway pods are running and the AI Gateway controller will mutate the pods.
-  envoyGatewayNamespace: "envoy-gateway-system"
```
mathetake, line 23: Note: this to be noted in the release note #983

internal/controller/controller.go
```diff
@@ -85,10 +84,8 @@ func StartControllers(ctx context.Context, mgr manager.Manager, config *rest.Con
 
 	gatewayEventChan := make(chan event.GenericEvent, 100)
 	gatewayC := NewGatewayController(c, kubernetes.NewForConfigOrDie(config),
-		logger.WithName("gateway"), options.EnvoyGatewayNamespace, options.UDSPath, options.ExtProcImage)
+		logger.WithName("gateway"), options.UDSPath, options.ExtProcImage, false, uuid.NewString)
 	if err = TypedControllerBuilderForCRD(mgr, &gwapiv1.Gateway{}).
-		// We need the annotation change event to reconcile the Gateway referenced by AIGatewayRoutes.
-		WithEventFilter(predicate.Or(predicate.GenerationChangedPredicate{}, predicate.AnnotationChangedPredicate{})).
```
mathetake, line 93: this was a leftover from#629, so deleting it

Xunzhuo:
i think maybe we should deal with https://github.com/envoyproxy/ai-gateway/tree/main/manifests/envoy-gateway-config as well?

mathetake:
> i think maybe we should deal with https://github.com/envoyproxy/ai-gateway/tree/main/manifests/envoy-gateway-config as well?

no i don't think so. The code here should work regardless of the configuration there (deployment mode). GatewayNamespace mode is an experimental so i don't think i want to add a whole another e2e test for it but i am pretty sure it works according to the fact that the controller works without knowing EG's namespace.

cmd/aigw/translate.go
```diff
@@ -214,7 +210,9 @@ func translateCustomResourceObjects(
 		make(chan event.GenericEvent),
 	)
 	gwC := controller.NewGatewayController(fakeClient, fakeClientSet, logr.FromSlogHandler(logger.Handler()),
-		envoyGatewayNamespace, extProcUDSPath, "docker.io/envoyproxy/ai-gateway-extproc:latest",
+		extProcUDSPath, "docker.io/envoyproxy/ai-gateway-extproc:latest", true, func() string {
```
aabchoo, line 213: side note: should we make this version of extproc configurable?

internal/controller/gateway.go
```diff
@@ -474,3 +481,61 @@ func (c *GatewayController) annotateGatewayPods(ctx context.Context, gw *gwapiv1
...
+			egOwningGatewayNameLabel, gw.Name, egOwningGatewayNamespaceLabel, gw.Namespace),
+	})
+	if err != nil {
+		err = fmt.Errorf("failed to list pods: %w", err)
+		return
+	}
+	pods = ps.Items
+
+	var ds *appsv1.DeploymentList
+	ds, err = c.kube.AppsV1().Deployments("").List(ctx, metav1.ListOptions{
```
aabchoo: Create a variable for ListOptions since it's the same for all 3?

internal/controller/gateway.go
```diff
@@ -474,3 +481,61 @@ func (c *GatewayController) annotateGatewayPods(ctx context.Context, gw *gwapiv1
...
+			egOwningGatewayNameLabel, gw.Name, egOwningGatewayNamespaceLabel, gw.Namespace),
+	})
+	if err != nil {
+		err = fmt.Errorf("failed to list daemonsets: %w", err)
+		return
+	}
+	daemonSets = dss.Items
+
+	// We assume that all pods, deployments, and daemonsets are in the same namespace. Otherwise, it would be a bug in the EG
+	// or the disruptive configuration change of EG.
```
aabchoo, line 521: Why loop through them all if we assume it's all the same?

internal/controller/gateway.go
mathetake: good catch

internal/controller/gateway.go
mathetake: good catch

cmd/aigw/translate.go
mathetake: will do if someone asks for it but i doubt if anyone wants to control since at least for `aigw run`, it's not used at all

aabchoo:
Sanity Check: will data plane resources still be in `envoy-gateway-system` namespace  after this PR is merged and this feature is not enabled?

mathetake:
Yes won't change anything unless you explicitly enable the gateway namespace mode in EG config

PR #993, opened 2025-08-05
fix: adds missing index field in chat completions chunk

codefromthecrypt:
**Description**

according to the openai openai spec, there should be an index in the chat completions chunk data. This fixes it and avoids serializing nested empty structs.

See https://github.com/openai/openai-openapi/tree/manual_spec


yuzisun:
@codefromthecrypt @mathetake FYI, I am doing a massive refactoring to switch to use the official openai SDK types, see https://github.com/envoyproxy/ai-gateway/issues/995

codefromthecrypt:
@yuzisun cool will make a comment on your issue!

tests/extproc/testupstream_test.go
```diff
@@ -241,17 +241,17 @@ func TestWithTestUpstream(t *testing.T) {
 {"usage":{"inputTokens":41, "outputTokens":36, "totalTokens":77}}
 `,
 			expStatus: http.StatusOK,
-			expResponseBody: `data: {"choices":[{"delta":{"content":"","role":"assistant"}}],"object":"chat.completion.chunk"}
+			expResponseBody: `data: {"choices":[{"index":0,"delta":{"content":"","role":"assistant"}}],"object":"chat.completion.chunk"}
```
mathetake, line 244: so previously "index" was missing for aws/gcp backend responses vs with this commit, it will have zero by default always. i wonder if that would be an issue on the OpenAI SDK side (no idea ;)

mathetake: i see most examples set to zero https://github.com/openai/openai-openapi/blob/498c71ddf6f1c45b983f972ccabca795da211a3e/openapi.yaml so maybe no harm

codefromthecrypt: yeah it was a side effect of structs adding fields perhaps which maybe was to the point of @yuzisun wanting to use a canonical format auto-generated from their yaml (openai sdk -- stainless)

PR #996, opened 2025-08-05
test: add e2e test for traffic splitting with fallback

### LINKED ISSUE #987: test: traffic splitting + fallback test case
I think we lack the end to end tests for the following scnario:

At the top level, do the traffic splitting between backend A and backend B with the same priority. And both failed, do the fallback to backend-c. I _believe_ weights are used within the context of the same priority, so this should work but it would be nice to have an end-to-end test case for this.

```yaml
rules:
  - backendRefs:
      - name: backend-a
        priority: 0
        weight: 50
      - name: backend-b
        modelNameOverride: model-a
        priority: 0
        weight: 50
      - name: backend-c
        modelNameOverride: model-c
        priority: 1
    matches:
      - headers:
          - name: x-ai-eg-model
            type: Exact
            value: model-a
```

surenraju-careem:
**Description**

This commit adds an e2e test for traffic splitting with fallback functionality. The test validates that when primary backends (backend-a and backend-b) return 5xx errors, traffic properly falls back to the fallback backend (backend-c) using priority-based routing.

The test includes:
- Traffic distribution validation (50/50 weight split between backend-a and backend-b)
- Fallback mechanism testing using BackendTrafficPolicy with retry configuration

Closes #987

### DIFF
--- a/tests/e2e/traffic_splitting_fallback_test.go
+++ b/tests/e2e/traffic_splitting_fallback_test.go
@@ -30,7 +30,7 @@ func TestTrafficSplittingFallback(t *testing.T) {
 	t.Run("traffic-distribution", func(t *testing.T) {
 		// Test that traffic splitting configuration is working by making multiple requests
 		// and verifying they are distributed between backends A and B.
-		const requestCount = 500
+		const requestCount = 50
 		backendAResponses := 0
 		backendBResponses := 0
 


tests/e2e/traffic_splitting_fallback_test.go
```diff
@@ -0,0 +1,132 @@
...
+
+	const egSelector = "gateway.envoyproxy.io/owning-gateway-name=traffic-splitting-fallback"
+
+	// Wait for the gateway to be ready.
+	requireWaitForGatewayPodReady(t, egSelector)
+
+	t.Run("traffic-distribution", func(t *testing.T) {
+		// Test that traffic splitting configuration is working by making multiple requests
+		// and verifying they are distributed between backends A and B.
+		const requestCount = 50
```
mathetake, line 33: let's make more
```suggestion
		const requestCount = 500
```

tests/e2e/traffic_splitting_fallback_test.go
```diff
@@ -0,0 +1,132 @@
...
+
+		// Check that distribution is roughly 50/50 (within 20% tolerance).
+		backendARatio := float64(backendAResponses) / float64(requestCount)
+		backendBRatio := float64(backendBResponses) / float64(requestCount)
+
+		// Validate that both backends receive traffic within 50/50 distribution (30%-70% tolerance).
+		require.Greater(t, backendARatio, 0.3, "Backend A should receive at least 30% of traffic")
+		require.Less(t, backendARatio, 0.7, "Backend A should receive at most 70% of traffic")
+		require.Greater(t, backendBRatio, 0.3, "Backend B should receive at least 30% of traffic")
+		require.Less(t, backendBRatio, 0.7, "Backend B should receive at most 70% of traffic")
```
mathetake: can you use `require.InDelta` instead to simplify ?

tests/e2e/traffic_splitting_fallback_test.go
```diff
@@ -0,0 +1,132 @@
...
+
+		t.Logf("Traffic distribution: Backend A=%d (%.1f%%), Backend B=%d (%.1f%%)",
+			backendAResponses, backendARatio*100, backendBResponses, backendBRatio*100)
+	})
+
+	t.Run("fallback-backend-c", func(t *testing.T) {
```
mathetake, line 83: this case we shouldn't get any response from backend-a or -b right? can we add more assertions to verify that?

mathetake (commented):
Thank you for taking a shot!

tests/e2e/traffic_splitting_fallback_test.go
```diff
@@ -0,0 +1,129 @@
...
+
+	const egSelector = "gateway.envoyproxy.io/owning-gateway-name=traffic-splitting-fallback"
+
+	// Wait for the gateway to be ready.
+	requireWaitForGatewayPodReady(t, egSelector)
+
+	t.Run("traffic-distribution", func(t *testing.T) {
+		// Test that traffic splitting configuration is working by making multiple requests
+		// and verifying they are distributed between backends A and B.
+		const requestCount = 500
```
mathetake: i am so sorry but looks like 500 is too much lol. let me revert it to 50 and then good to go!
```suggestion
		const requestCount = 50
```

mathetake (approved):
Fantastic!!

PR #998, opened 2025-08-06
extproc: remove input api schema in filterapi config

yuzisun:
**Description**

This removes the input api schema field in the filter config as it shouldn't be a config and currently it is not really used. The input schema is already determined by the specific API endpoint being used by the Envoy AI Gateway, such as the unified OpenAI API at `v1/chat/completions` or later the Anthropic Messages API we are supporting at `anthropic/v1/messages`. This removes the confusion when we are going to support other vendor pass through APIs like Anthropic. We may need to consider deprecating the input `APISchema` defined on `AIGatewayRoute` which is also not really used.

Related Issue: #847


mathetake:
So this assumes that Anthropic, etc and other vendors' endpoint won't overlap with /v1/chat/completions etc? i think that makes sense at least for now

mathetake:
oh coverage is failing lol it seems it was right on the threshold before

mathetake (approved):

yuzisun:
> So this assumes that Anthropic, etc and other vendors' endpoint won't overlap with /v1/chat/completions etc? i think that makes sense at least for now

We are really supporting v1/chat/completion as the unified OpenAI compatible api which we do translation. For other vendor routes it needs to be prefixed with the vendor name so there is no conflict.

mathetake:
Totally makes sense!

mathetake:
let me take care of the coverage

PR #1001, opened 2025-08-06
feat: add otel tracing for aigw standalone mode

codefromthecrypt:
**Description**

This integrates OpenTelemetry tracing support using the OpenInference specification for span attributes. This applies to  the extproc router filter (e.g. incoming request to :1975/v1/chat/completions), regardless of upstream (inference backend like ollama, bedrock etc).

This first integrates with the CLI/Standalone mode (`aigw`), and I added docker instructions to cmd/aigw which work, and join traces with python.

Screenshot 2025-08-07 at 8 44 12 AM

This is configured with standard ENV variables which allow the `aigw` to be configured naturally, like litellm or others, as well allow our extproc tests to swap out simply for test config. This is intentionally conservative only if `OTEL_EXPORTER_OTLP_ENDPOINT` is set and `OTEL_SDK_DISABLED` is not set to true. https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#general-sdk-configuration

#791 

Once things start working here, I can update some demos like
https://github.com/elastic/testing-genai-applications/tree/main
and 
https://github.com/elastic/observability-examples/tree/main/inference-platforms

This doesn't integrate the with k8s gateway yet, because need to propagate ENV to the extproc filters, to enable otel, and there's no code to do that yet. It is also possible to change the configuration approach when applying to k8s to cli, just we should be extremely careful about doing so as there are many routine configurations, including below when we get into redaction. copy/hard-coding each standard env also as a flag or custom config property will be a lot of code to test and lead to maintenance.

This doesn't yet apply redaction as that would be too many things to support in the first PR. It should be followed up with https://github.com/Arize-ai/openinference/blob/main/spec/configuration.md

### DIFF
--- a/internal/tracing/openinference/chat_completion.go
+++ b/internal/tracing/openinference/chat_completion.go
@@ -28,17 +28,17 @@ func (ChatCompletionRecorder) StartParams(*openai.ChatCompletionRequest, []byte)
 	return "ChatCompletion", startOpts
 }
 
-// RecordRequest implements the same method as defined in tracing.Recorder.
+// RecordRequest implements the same method as defined in tracing.ChatCompletionRecorder.
 func (ChatCompletionRecorder) RecordRequest(span trace.Span, chatReq *openai.ChatCompletionRequest, body []byte) {
 	span.SetAttributes(buildRequestAttributes(chatReq, string(body))...)
 }
 
-// RecordChunk implements the same method as defined in tracing.Recorder.
+// RecordChunk implements the same method as defined in tracing.ChatCompletionRecorder.
 func (ChatCompletionRecorder) RecordChunk(trace.Span, int) {
 	// No-op for non-streaming responses.
 }
 
-// RecordResponse implements the same method as defined in tracing.Recorder.
+// RecordResponse implements the same method as defined in tracing.ChatCompletionRecorder.
 func (ChatCompletionRecorder) RecordResponse(span trace.Span, statusCode int, body []byte) {
 	bodyString := string(body)
 	if statusCode < 200 || statusCode >= 300 {
@@ -64,14 +64,14 @@ type ChatCompletionStreamingRecorder struct {
 	ChatCompletionRecorder
 }
 
-// RecordChunk implements the same method as defined in tracing.Recorder.
+// RecordChunk implements the same method as defined in tracing.ChatCompletionRecorder.
 func (ChatCompletionStreamingRecorder) RecordChunk(span trace.Span, chunkIdx int) {
 	if chunkIdx == 0 {
 		span.AddEvent("First Token Stream Event")
 	}
 }
 
-// RecordResponse implements the same method as defined in tracing.Recorder.
+// RecordResponse implements the same method as defined in tracing.ChatCompletionRecorder.
 // It converts SSE to JSON before calling the base implementation.
 func (ChatCompletionStreamingRecorder) RecordResponse(span trace.Span, statusCode int, body []byte) {
 	// Only convert if the body looks like SSE data (contains "data: " prefix).


internal/tracing/openinference/chat_completion.go
```diff
@@ -0,0 +1,87 @@
...
+	span.SetAttributes(buildRequestAttributes(chatReq, string(body))...)
+}
+
+// RecordChunk implements the same method as defined in tracing.Recorder.
+func (ChatCompletionRecorder) RecordChunk(trace.Span, int) {
+	// No-op for non-streaming responses.
+}
+
+// RecordResponse implements the same method as defined in tracing.Recorder.
+func (ChatCompletionRecorder) RecordResponse(span trace.Span, statusCode int, body []byte) {
```
mathetake, line 42: so this function essentially copies the entire body twice; `string(body)` and the unmarshaling into openai.ChatCompletionResponse. Would it be possible to reduce it?

internal/tracing/openinference/chat_completion.go
```diff
@@ -0,0 +1,87 @@
...
+	// Only convert if the body looks like SSE data (contains "data: " prefix).
+	if bytes.Contains(body, []byte("data: ")) {
+		// Convert SSE to a single completion response.
+		if jsonBody, convErr := convertSSEToJSON(body); convErr == nil && len(jsonBody) > 0 {
+			body = jsonBody
+		}
+	}
+
+	// Call the base implementation with the (possibly converted) body.
+	ChatCompletionRecorder{}.RecordResponse(span, statusCode, body)
```
mathetake, line 86: i don't know but the `body` here is in the unit of envoy's stream processing hence it can be any binary and even might not be valid UTF-8. I see inside ChatCompletionRecorder's RecordResponse it's converting it to string attribute which maybe troublesome?

internal/tracing/openinference/chat_completion.go
```diff
@@ -0,0 +1,87 @@
...
+// RecordChunk implements the same method as defined in tracing.Recorder.
+func (ChatCompletionStreamingRecorder) RecordChunk(span trace.Span, chunkIdx int) {
+	if chunkIdx == 0 {
+		span.AddEvent("First Token Stream Event")
+	}
+}
+
+// RecordResponse implements the same method as defined in tracing.Recorder.
+// It converts SSE to JSON before calling the base implementation.
+func (ChatCompletionStreamingRecorder) RecordResponse(span trace.Span, statusCode int, body []byte) {
```
mathetake, line 76: so the body chunk is coming in the unit of envoy's stream processing, not the SSE level. So this body might end in the middle and also might start in the middle of the even line. so the current impl might drop/ignore events when that happens. You would need to be stateful to be able to parse all events correctly.

mathetake:
I think i have an idea to resolve all the comments

internal/tracing/openinference/chat_completion.go
mathetake: also noticed that this is only called when endOfStream=true, so majority of body events are ignored

mathetake: hmm maybe i am not missing something... So calling this with the last event is what we want?

internal/extproc/chatcompletion_processor.go
```diff
@@ -89,17 +93,40 @@ func (c *chatCompletionProcessorRouterFilter) ProcessResponseHeaders(ctx context
...
+	var statusCode int
+	if upstream, ok := c.upstreamFilter.(*chatCompletionProcessorUpstreamFilter); ok {
+		if statusInt, _ := strconv.Atoi(upstream.responseHeaders[":status"]); statusInt > 0 {
+			statusCode = statusInt
+		}
+	}
+
+	// Record chunk timing for streaming responses.
+	if c.originalRequestBody.Stream {
+		c.span.RecordChunk()
```
mathetake, line 117: this can also be buggy due to the same reason as chunk here != events

mathetake:
let me refactor around the upstream filter etc to avoid duplicate parsing etc

mathetake:
I think i would like to unblock you so i will go ahead and merge and I will do the refactoring on a branch based off of the main

codefromthecrypt:
Thanks @mathetake, I'm glad it was easy to find the tension I also had when trying to write this, which was the chunk relationship and if there is a guarantee in the case of streaming a chunk falls on SSE lines or not. In chat completions, I assume it is *all text* regardless, just maybe not conveniently on a json boundary? Anyway, looking forward to your improvement esp for efficiency as I was guessing a bit and I can tell you know the real deal.

PR #1004, opened 2025-08-06
controller: not block Envoy pods before full reconciliation

### LINKED ISSUE #742: invalid secret content will cause envoyproxy cannot start-up

mathetake:
**Description**

Previously, there's a bit of race condition where the gateway pod mutation hook can be called before the secret config is created by the Gateway controller. That resulted in blocking the Envoy pods from starting up. This resolves the issue by doing lazy mutation until the secret is created.

Closes #742

### DIFF
--- a/internal/apischema/openai/openai.go
+++ b/internal/apischema/openai/openai.go
@@ -442,12 +442,14 @@ type ChatCompletionMessageToolCallFunctionParam struct {
 }
 
 type ChatCompletionMessageToolCallParam struct {
+	// Add this Index field. It is required for streaming.
+	Index *int `json:"index,omitempty"`
 	// The ID of the tool call.
-	ID string `json:"id"`
+	ID *string `json:"id"`
 	// The function that the model called.
 	Function ChatCompletionMessageToolCallFunctionParam `json:"function"`
 	// The type of the tool. Currently, only `function` is supported.
-	Type ChatCompletionMessageToolCallType `json:"type"`
+	Type ChatCompletionMessageToolCallType `json:"type,omitempty"`
 }
 
 type ChatCompletionResponseFormatType string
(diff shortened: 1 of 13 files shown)


aabchoo:
This is an admission hook right? Wouldn't mutatePod only trigger when kubectl api create/update/delete are made?

mathetake:
Yes. The controller will add annotations to the deployment template so the new pods will be created if the extproc container doesn't exist in the pod

mathetake:
so the comment about "trigger the mutation again" is misleading i agree -- "the mutation hook will be called on the new pods after the secret is created" is the correct comment

mathetake:
updated the comment

mathetake:
https://github.com/envoyproxy/ai-gateway/blob/92177ad4a4b06d5f52d502f9c13e620d8029e481/internal/controller/gateway.go#L456-L477

here!

PR #1005, opened 2025-08-06
feat: add more anthropic thinking support

alexagriffith:
**Description**
adding tests and more robust support for streaming with thinking


alexagriffith:
need to rebase once https://github.com/envoyproxy/ai-gateway/pull/984 is merged

PR #1007, opened 2025-08-07
feat: add more support for gcp anthropic thinking

alexagriffith:
**Description**
adding tests and more robust support for streaming with thinking

### DIFF
--- a/internal/extproc/translator/openai_gcpanthropic_stream.go
+++ b/internal/extproc/translator/openai_gcpanthropic_stream.go
@@ -204,7 +204,9 @@ func (p *anthropicStreamParser) handleAnthropicStreamEvent(eventType []byte, dat
 				if err != nil {
 					return nil, fmt.Errorf("failed to marshal tool use input: %w", err)
 				}
-				argsJSON = string(argsBytes)
+				if len(argsBytes) > 0 {
+					argsJSON = string(argsBytes)
+				}
 			}
 
 			// Store the complete input JSON in our state.
@@ -268,12 +270,16 @@ func (p *anthropicStreamParser) handleAnthropicStreamEvent(eventType []byte, dat
 				return nil, fmt.Errorf("received input_json_delta for unknown tool at index %d", event.Index)
 			}
 			index := int(event.Index)
+			args := ""
+			if len(event.Delta.PartialJSON) > 0 {
+				args = event.Delta.PartialJSON
+			}
 			delta := openai.ChatCompletionResponseChunkChoiceDelta{
 				ToolCalls: []openai.ChatCompletionMessageToolCallParam{
 					{
 						Index: &index,
 						Function: openai.ChatCompletionMessageToolCallFunctionParam{
-							Arguments: event.Delta.PartialJSON,
+							Arguments: args,
 						},
 					},
 				},
--- a/internal/extproc/translator/openai_gcpanthropic_stream.go
+++ b/internal/extproc/translator/openai_gcpanthropic_stream.go
@@ -204,9 +204,7 @@ func (p *anthropicStreamParser) handleAnthropicStreamEvent(eventType []byte, dat
 				if err != nil {
 					return nil, fmt.Errorf("failed to marshal tool use input: %w", err)
 				}
-				if len(argsBytes) > 0 {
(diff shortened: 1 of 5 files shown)


internal/extproc/translator/openai_gcpanthropic_stream.go
```diff
@@ -247,21 +270,22 @@ func (p *anthropicStreamParser) handleAnthropicStreamEvent(eventType []byte, dat
 				return nil, fmt.Errorf("received input_json_delta for unknown tool at index %d", event.Index)
 			}
 			index := int(event.Index)
+			args := ""
+			if len(event.Delta.PartialJSON) > 0 {
+				args = event.Delta.PartialJSON
+			}
```
mathetake: I think we can omit the if ?
```suggestion
			args := event.Delta.PartialJSON
```

internal/extproc/translator/openai_gcpanthropic_stream.go
```diff
@@ -209,12 +223,20 @@ func (p *anthropicStreamParser) handleAnthropicStreamEvent(eventType []byte, dat
...
+							Arguments: argsJSON,
 						},
 					},
 				},
 			}
 			return p.constructOpenAIChatCompletionChunk(delta, ""), nil
-		} else if event.ContentBlock.Type == string(constant.ValueOf[constant.Thinking]()) || event.ContentBlock.Type == string(constant.ValueOf[constant.RedactedThinking]()) {
+		}
+		if event.ContentBlock.Type == string(constant.ValueOf[constant.Thinking]()) {
+			delta := openai.ChatCompletionResponseChunkChoiceDelta{Content: ptr.To("")}
```
mathetake: could you define `var emptyStrPtr = ptr.To("")` globally and use it here so that we could avoid the allocation every time we reach this block

```suggestion
			delta := openai.ChatCompletionResponseChunkChoiceDelta{Content: emptyStrPtr}
```

mathetake (commented):
nice! I left nit comments but good to go!

internal/extproc/translator/openai_gcpanthropic_stream.go
alexagriffith: was testing something - removed it

internal/extproc/translator/openai_gcpanthropic_stream.go
alexagriffith: but im checking and in this code this is the only place we do it

alexagriffith: so maybe its not worth for now?

internal/extproc/translator/openai_gcpanthropic_stream.go
alexagriffith: just the test passed when I changed this locally but it was false positive bc it was skipped since i didn't reset the envoy command to use func-e locally

internal/extproc/translator/openai_gcpanthropic_stream.go
alexagriffith: ah i misunderstood your comment, yes can change

internal/extproc/translator/openai_gcpanthropic_test.go
```diff
@@ -211,6 +211,54 @@ func TestOpenAIToGCPAnthropicTranslatorV1ChatCompletion_RequestBody(t *testing.T
...
+		_, bm, err := translator.RequestBody(nil, thinkingReq, false)
+		require.NoError(t, err)
+		require.NotNil(t, bm)
+
+		body := bm.GetBody()
+		require.NotNil(t, body)
+
+		thinkingBlock := gjson.GetBytes(body, "thinking")
+		require.True(t, thinkingBlock.Exists(), "The 'thinking' field should exist in the request body")
+		require.True(t, thinkingBlock.IsObject(), "The 'thinking' field should be a JSON object")
```
xiaolin593, line 235: Can you add an example of `thinking` json object? Or do I miss it in your test case example?

alexagriffith: I do have it enabled and disabled thats just how the go anthropic sdk defines it

mathetake (approved):

PR #1009, opened 2025-08-07
fix: gcp auth with proxy settings

sukumargaonkar:
**Description**

**Issue:** GCP token rotator fails when http-proxy is specified.
**Reason:** The token rotator uses `google.golang.org/api/impersonate` to obtain the token. When proxyURL is provided in the form of an http-client (option.WithHTTPClient(...)), the `impersonate` library ignores the provided tokenSource ( option.WithTokenSource(...) )
**Fix:** This PR introduces `bearerAuthRoundTripper` which discovers proxyURL from env and injects the required auth header


internal/controller/rotators/gcp_oidc_token_rotator.go
```diff
@@ -291,6 +294,73 @@ func exchangeJWTForSTSToken(ctx context.Context, jwtToken string, wifConfig aigv
...
+	// to avoid connection pool fragmentation.
+	sharedTransport *http.Transport
+	// transportOnce ensures the shared transport is initialized only once.
+	transportOnce sync.Once
+)
+
+// getSharedTransportWithProxy returns a singleton HTTP transport configured with GCP proxy settings.
+// This prevents connection pool fragmentation by reusing the same transport across multiple
+// bearerAuthRoundTripper instances.
+func getSharedTransportWithProxy() (*http.Transport, error) {
```
xiaolin593: Could this be simplified as https://github.com/envoyproxy/ai-gateway/blob/main/internal/controller/tokenprovider/azure_token_provider.go#L54?

internal/controller/rotators/gcp_oidc_token_rotator.go
```diff
@@ -247,15 +248,17 @@ var _ stsTokenGenerator = exchangeJWTForSTSToken
 // exchangeJWTForSTSToken implements [stsTokenGenerator]
 // exchangeJWTForSTSToken exchanges a JWT token for a GCP STS (Security Token Service) token.
 func exchangeJWTForSTSToken(ctx context.Context, jwtToken string, wifConfig aigv1a1.GCPWorkloadIdentityFederationConfig, opts ...option.ClientOption) (*tokenprovider.TokenExpiry, error) {
-	proxyOpt, err := getGCPProxyClientOption()
+	roundTripper, err := NewBearerAuthRoundTripper("")
```
aabchoo: should the param be a token instead of an empty string

sukumargaonkar: This step doesn't pass the token via auth header, the empty string implies that the auth header will be skipped. added comment

internal/controller/rotators/gcp_oidc_token_rotator.go
```diff
@@ -291,6 +295,62 @@ func exchangeJWTForSTSToken(ctx context.Context, jwtToken string, wifConfig aigv
...
+	}
+	return sharedTransport, nil
+}
+
+type bearerAuthRoundTripper struct {
+	base  *http.Transport
+	token string
+}
+
+func NewBearerAuthRoundTripper(token string) (http.RoundTripper, error) {
```
mathetake: unexport

internal/controller/rotators/gcp_oidc_token_rotator.go
```diff
@@ -291,6 +295,62 @@ func exchangeJWTForSTSToken(ctx context.Context, jwtToken string, wifConfig aigv
...
+		sharedTransport = &http.Transport{Proxy: http.ProxyURL(gcpProxyURL)}
+	})
+
+	if err != nil {
+		return nil, err
+	}
+	return sharedTransport, nil
+}
+
+type bearerAuthRoundTripper struct {
```
mathetake: comment on what interface this implements just like elsewhere.

internal/controller/rotators/gcp_oidc_token_rotator.go
```diff
@@ -291,6 +295,62 @@ func exchangeJWTForSTSToken(ctx context.Context, jwtToken string, wifConfig aigv
...
+		return nil, err
+	}
+
+	return &bearerAuthRoundTripper{
+		base:  baseTransport,
+		token: token,
+	}, nil
+}
+
+func (rt *bearerAuthRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
```
mathetake, line 329: same comment on adding function doc comment about what this implements

internal/controller/rotators/gcp_oidc_token_rotator.go
```diff
@@ -291,6 +295,62 @@ func exchangeJWTForSTSToken(ctx context.Context, jwtToken string, wifConfig aigv
 	}, nil
 }
 
+var (
+	// sharedTransport is a singleton HTTP transport that gets reused across all bearerAuthRoundTripper instances
+	// to avoid connection pool fragmentation.
+	sharedTransport *http.Transport
+	// transportOnce ensures the shared transport is initialized only once.
+	transportOnce sync.Once
```
mathetake: i think this is overkill. can you initialize in the package init function `func init() { ... }` and delete this sync.Once

internal/controller/rotators/gcp_oidc_token_rotator.go
```diff
@@ -88,6 +87,24 @@ type gcpOIDCTokenRotator struct {
...
+func init() {
+	gcpProxyURL, err := getGCPProxyURL()
+	if err != nil {
+		initError = fmt.Errorf("error getting GCP proxy URL: %w", err)
+		// transport without proxy url.
+		sharedTransport = &http.Transport{}
+	} else {
+		sharedTransport = &http.Transport{Proxy: http.ProxyURL(gcpProxyURL)}
+	}
+}
```
mathetake: this init error thingy is not meaningful as it's not recoverable. When does this happen? I don't believe this is not expected at all and in that case we usually do panic.

Also, when proxy URL is not configured, let's not use transport.

```suggestion
// sharedGCPProxyTransport is a shared HTTP transport used for GCP API calls.
// It is initialized with the GCP proxy URL if provided in the environment variable.
var sharedGCPProxyTransport *http.Transport

func init() {
	gcpProxyURL, err := getGCPProxyURL()
	if err != nil {
		panic(fmt.Sprintf("failed to ..."))
	} else if gcpProxyURL != "" {
		sharedGCPProxyTransport = &http.Transport{Proxy: http.ProxyURL(gcpProxyURL)}
	}
}
```

internal/controller/rotators/gcp_oidc_token_rotator_test.go
```diff
@@ -871,34 +874,80 @@ func TestGetGCPProxyClientOption(t *testing.T) {
...
+
+			// Create a request to the test server.
+			req, err := http.NewRequest(http.MethodGet, server.URL, nil)
+			require.NoError(t, err)
+
+			// Execute the request through the round tripper.
+			resp, err := roundTripper.RoundTrip(req)
+			require.NoError(t, err)
+			require.Equal(t, http.StatusOK, resp.StatusCode)
+			resp.Body.Close()
```
mathetake: nit: can you do defer instead of inline? this won't be called if the assertion failed in the middle e.g. when status is not OK

internal/controller/rotators/gcp_oidc_token_rotator.go
sukumargaonkar: This will mainly happen when you set a bad en var as proxy url

True, it is not recoverable wasn't sure about panicking 

Will update to panic in case of bad proxy url and skip the transport ended proxy is not configured

mathetake: yeah, bad env shouldn't let the controller keep running, so panic and giving the early feedback to the provisioner would be ideal. Otherwise, they won't notice until sift through all the logs in the controller vs they will notice from the pod status.

sukumargaonkar: on second thoughts, should the controller fail due to one bad gcp-proxy-url env var
since other providers can still keep working when a bad env var is set for gcp-proxy-var, i would not want to crash the controller. 
Having an error log that the admin can alert on seems better.

thoughts @mathetake ?

mathetake: No don't think so. This is a startup configuration bug of the operator and shouldn't let it continue working. Imagine when it is correctly working with a correct env var. then they change the env var to the wrong one accidentally. Then they will end up breaking the GCP rotation without any prevention like that. Do you want to force everyone to setup the alert on the log line when we can just prevent it from happening?

mathetake: This makes me feel like we should do the same thing to all the other proxy env and do the initialization in the init function but we can do it later.

sukumargaonkar: fair, updated

internal/controller/rotators/gcp_oidc_token_rotator.go
```diff
@@ -88,6 +87,22 @@ type gcpOIDCTokenRotator struct {
 	stsTokenFunc stsTokenGenerator
 }
 
+// sharedTransport is a shared HTTP transport used for GCP API calls.
+// It is initialized with the GCP proxy URL if provided in the environment variable.
+var (
+	sharedTransport http.RoundTripper
+	initError       error
```
mathetake: Let's rename it to add GCP portion & remove initError
```suggestion
	sharedGCPTransport http.RoundTripper
```

mathetake (approved):

PR #1011, opened 2025-08-07
extproc: lift error status handling from translators 

mathetake:
**Description**

This lifts the error status code from each translators into the processors level so that we could share the same logic and won't need to copy paste anymore.


mathetake:
controller test failure is irrelevant and it seems GitHub+Cloudflare related incident

PR #1012, opened 2025-08-07
test: updates CRDs used in envtest

mathetake:
**Description**

This switches to use CRDs from EG's Chart to avoid manually maintaining GWAPI version as well as manually list necessary components.


mathetake:
extproc failure can be ignored as it seems groq is not available

Issue #1002, opened 2025-08-06
Prefix-based URL handling breaks extproc processing


siddharth1036:
*Description*:
>We're encountering a consistent issue with prefix-based URL handling in the Envoy AI Gateway setup. Specifically, requests with prefixed routes like /prefix/v1/chat/completions are causing problems with external processing.

*Problem*:
>The ext_proc filter does not receive an endpoint for prefixed paths such as /prefix/v1/chat/completions.
>This results in ext_proc receiving the orignal path, which leads to ext auth failure.

*Expected Behavior*:
>Either the rewrite should be processed before ext_proc, or there should be a native way for ext_proc to handle prefixed paths correctly.

*Additional Notes*:
>This is an inbound prefix for traffic landing on gateway which needs to be processed by extproc for model based routing.

Let us know if a workaround exists or if any changes are planned in the filter chain logic.

mathetake:
I see this is about the "input schema"'s base url configurability. This is somewhat relevant to the direction @yuzisun is thinking about: https://github.com/envoyproxy/ai-gateway/pull/998 
The context is that since we are looking to resolve the #948, passthroughs for major AI providers besides OpenAI, we are dropping the "input" configuration as in the PR ^^, and instead plan is to provide the prefix for the common AI provider.

For example,
* /anthropic/** -> treat all requests as anthropic format. If the backend is not anthropic, then do the translation from anthropic to whatever format. If not, just pass through
* /v1/** -> do the same thing for openai as above.

and i think we should be able to make these prefix configurable to resolve your problem as well as I think  using `/v1/` not `/openai/v1/` for openai seems not consistent. 

I propose the following:
* Deprecate the AIGatewayRoute.Spec.Schema field as per the discussion in #998 as in anyways it has not being used.
* Global configurations for the prefix per input provider like 
```diff
diff --git a/manifests/charts/ai-gateway-helm/values.yaml b/manifests/charts/ai-gateway-helm/values.yaml
index a5e4a3b..680c521 100644
--- a/manifests/charts/ai-gateway-helm/values.yaml
+++ b/manifests/charts/ai-gateway-helm/values.yaml
@@ -5,6 +5,15 @@
 
 # Default values for ai-gateway-helm.
 
+# This is a set of configurations for the Unified API capability.
+# Each input API format (e.g., OpenAI, Anthropic) should have a distinct prefix for its API endpoints.
+unified_api:
+  openai:
+    prefix: "/v1"
[...]

PR #1013, opened 2025-08-08
api: deprecate schema from AIGatewayRoute

mathetake:
**Description**
This removes the unused "schema" API from AIGatewayRoute as a preparation for transparent passthrough etc support for non-OpenAI SDK support as well as the prefix configuration #1002 

Follow up on #998 
Necessary for #1002


mathetake:
Let me remove the Grow credentials for now which will make the test cases skipped

PR #1018, opened 2025-08-08
deps: pin EG to v1.5 for the next release

mathetake:
**Description**

Envoy Gateway v1.5 has been released yesterday. This commit pins the version of EG in our dependency to the released version. It requires Go v1.24.6 so this also upgrades it from v1.24.4 to 1.24.6.


mathetake:
~why it's flaky ?!~ simply a bug

PR #1020, opened 2025-08-08
feat: make prefix of endpoints configurable 

### LINKED ISSUE #1002: Prefix-based URL handling breaks extproc processing

mathetake:
**Description**

This commit makes the "prefix" for the supported endpoints configurable via global configurations.

Closes #1002

### DIFF
--- a/internal/controller/ai_gateway_route_test.go
+++ b/internal/controller/ai_gateway_route_test.go
@@ -122,7 +122,7 @@ func TestAIGatewayRouterController_syncAIGatewayRoute(t *testing.T) {
 	fakeClient := requireNewFakeClientWithIndexes(t)
 	kube := fake2.NewClientset()
 	eventCh := internaltesting.NewControllerEventChan[*gwapiv1.Gateway]()
-	s := NewAIGatewayRouteController(fakeClient, kube, logr.Discard(), eventCh.Ch, "/v1")
+	s := NewAIGatewayRouteController(fakeClient, kube, logr.Discard(), eventCh.Ch, "/")
 	require.NotNil(t, s)
 
 	for _, backend := range []*aigv1a1.AIServiceBackend{
@@ -205,7 +205,7 @@ func Test_newHTTPRoute(t *testing.T) {
 			)
 			fakeClient := requireNewFakeClientWithIndexes(t)
 			eventCh := internaltesting.NewControllerEventChan[*gwapiv1.Gateway]()
-			s := NewAIGatewayRouteController(fakeClient, nil, logr.Discard(), eventCh.Ch, "/v1")
+			s := NewAIGatewayRouteController(fakeClient, nil, logr.Discard(), eventCh.Ch, "/")
 			httpRoute := &gwapiv1.HTTPRoute{
 				ObjectMeta: metav1.ObjectMeta{Name: "myroute", Namespace: ns},
 				Spec:       gwapiv1.HTTPRouteSpec{},
@@ -294,10 +294,11 @@ func Test_newHTTPRoute(t *testing.T) {
 					Name:  gwapiv1.ObjectName(getHostRewriteFilterName("myroute")),
 				},
 			}}
+			expPath := &gwapiv1.HTTPPathMatch{Value: ptr.To("/")}
 			expRules := []gwapiv1.HTTPRouteRule{
 				{
 					Matches: []gwapiv1.HTTPRouteMatch{
-						{Headers: []gwapiv1.HTTPHeaderMatch{{Name: "x-test", Value: "rule-0"}}},
+						{Headers: []gwapiv1.HTTPHeaderMatch{{Name: "x-test", Value: "rule-0"}}, Path: expPath},
 					},
 					BackendRefs: []gwapiv1.HTTPBackendRef{{BackendRef: gwapiv1.BackendRef{BackendObjectReference: gwapiv1.BackendObjectReference{Name: "some-backend1", Namespace: refNs}, Weight: ptr.To[int32](100)}}},
 					Timeouts:    &gwapiv1.HTTPRouteTimeouts{Request: &defaultTimeout},
@@ -305,7 +306,7 @@ func Test_newHTTPRoute(t *testing.T) {
 				},
 				{
 					Matches: []gwapiv1.HTTPRouteMatch{
-						{Headers: []gwapiv1.HTTPHeaderMatch{{Name: "x-test", Value: "rule-1"}}},
+						{Headers: []gwapiv1.HTTPHeaderMatch{{Name: "x-test", Value: "rule-1"}}, Path: expPath},
 					},
 					BackendRefs: []gwapiv1.HTTPBackendRef{
 						{BackendRef: gwapiv1.BackendRef{BackendObjectReference: gwapiv1.BackendObjectReference{Name: "some-backend2", Namespace: refNs}, Weight: ptr.To[int32](100)}},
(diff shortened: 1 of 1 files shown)


cmd/extproc/mainlib/main.go
```diff
@@ -50,6 +51,10 @@ type extProcFlags struct {
 	metricsPort                int        // HTTP port for the metrics server.
 	healthPort                 int        // HTTP port for the health check server.
 	metricsRequestHeaderLabels string     // comma-separated key-value pairs for mapping HTTP request headers to Prometheus metric labels.
+
+	// openAIPrefix is the OpenAI API prefix to be used for the external processor.
+	openAIPrefix string
```
gavrissh, line 55: Thanks for this change!!!

Thinking this might break for some edge cases. Let me know if it is correct. 
Our use case involves the same prefix being set across the input schema (currently openai, future anthropic, and many more). With the prefix annotation being specific to each input schema, every time a new input schema gets added, it might break for us without a patch.
cc: @johnugeorge @siddharth1036

mathetake: Not sure what you meant by "break". If an input schema is not existent, then adding "new annotation" won't be breaking anything, no?

Also, the whole point of this prefix per input schema is to enable the filter to choose the right "processor" without conflicts. How can we be sure that there won't be any conflict among providers. E.g. Are we sure that "/models" endpoint won't be used besides openai?

cc @yuzisun

mathetake: Also, "every time a new input schema gets added" won't frequently happen at all. At least, we are looking to add Anthropic but not sure for others for now. Like I am not sure how adding one annotation per release (~per three months if it's ever implemented) that won't break anything already working is an issue. Maybe I am missing something

mathetake: I saw your comment on making the default "/" configurable as well, and that makes me feel like i should give it more thinking to consolidate it with this one

gavrissh: Yes, I agree with the input schema specific change here. In issue #1002, we were hoping for more of a configurable default route change, something like adding this annotation "aigateway.envoyproxy.io/default-prefix" for all input schemas. Is that something that is possible?

So when these are set 
"aigateway.envoyproxy.io/default-prefix": "company-name"
"aigateway.envoyproxy.io/openai-prefix": "openai"

the accepted route will be - `https://api.mygateway.com/company-name/openai`

On the Anthropic case
"aigateway.envoyproxy.io/default-prefix": "company-name"
"aigateway.envoyproxy.io/anthropic-prefix": "anthropic"

the accepted route will be - `https://api.mygateway.com/company-name/anthropic`

johnugeorge: @mathetake To the above Gavrish's comment, can we also add a default prefix which will be a common root level prefix added to all the ai-gw routes? It can be empty by default so that current behaviour will be same.

mathetake: yeah that's what I am trying to think about together with this one!

mathetake:
I think the common root prefix as well as the schema level prefix could be a global option in the helm chart level. I think implementation-wise it's much simpler. We can start with it in v0.3 and later if someone asks for Gateway object-level granularity then we can revisit this annotation style.

mathetake:
@gavrissh @johnugeorge @yuzisun PTAL -- added two configuration layers; root prefix (which is added to all routes at HTTPRoute level) and the input-schema level configuration that comes after that root prefix. The latter is only for OpenAI now, but soon will be used for anthropic as well (see the comment here https://github.com/envoyproxy/ai-gateway/pull/1022#discussion_r2263935791).

The reason for this being global config, not per Gateway object, is to avoid the complexity in the impl. We can always revisit this if someone asks for more granularity like per-Gateway object level config.

manifests/charts/ai-gateway-helm/values.yaml
```diff
@@ -5,6 +5,24 @@
 
 # Default values for ai-gateway-helm.
 
+# Global configuration for the endpoints supported by the AI Gateway.
```
mathetake, line 8: note: taking a look at the doc here should be enough to do the discussion

gavrissh:
@mathetake Thanks a lot for this change!
I agree on the global config impl. The PR looks good to me.

mathetake:
i need to write/fix some tests before making this ready for review

Xunzhuo:
i got a question for this, why not add this in route level, since the prefix for different providers might be different? It looks like in a Gateway level configuration

mathetake:
Yes this is gateway level configuration and initially in the pr (see the early commit here), it's using the annotation on Gateway object, but switched to the global config as it's simpler per the comments here. The LLMCosts config also falls into the same category. 

I think I can revert back to the annotation style again as it's not that complex at the end of the day

mathetake:
> since the prefix for different providers might be different?

And your comment here seems a misunderstanding. You are talking about the backend (for translation) but this is about the input. It's happening before routing.

cmd/extproc/mainlib/main.go
```diff
@@ -171,9 +180,9 @@ func Main(ctx context.Context, args []string, stderr io.Writer) (err error) {
 	if err != nil {
 		return fmt.Errorf("failed to create external processor server: %w", err)
 	}
-	server.Register("/v1/chat/completions", extproc.ChatCompletionProcessorFactory(chatCompletionMetrics))
-	server.Register("/v1/embeddings", extproc.EmbeddingsProcessorFactory(embeddingsMetrics))
-	server.Register("/v1/models", extproc.NewModelsProcessor)
+	server.Register(path.Join(flags.openAIPrefix, "/chat/completions"), extproc.ChatCompletionProcessorFactory(chatCompletionMetrics))
```
aabchoo, line 183: Why don't we include the rootPrefix as part of the path?

mathetake: because it's included in as part of argument at extproc level, and extproc doesn't need to know the root.

aabchoo: will [:path](https://github.com/envoyproxy/ai-gateway/blob/ada7ac52574817dff6f66427bd8f4de8f099ed2a/internal/extproc/server.go#L104-L113) exclude the root prefix?

mathetake: No what i meant is the extproc level we already concat them.

image

aabchoo: 🤦 thank you for pointing that out

aabchoo (approved):

mathetake:
Note: I think this global config is not ideal to be honest -- there has to be "Gateway" level configuration CRD in this project which will also be able to configure "LLMCosts" API at that level. On the other hand, i think this will unblock the implementation of Anthropic input as well as the use case by @gavrissh for v0.3. We can always deprecate this and move to a better version as we iterate on the project over the time.

PR #1021, opened 2025-08-08
extproc: fixes modelNameOverride + forced stream_options

### LINKED ISSUE #1017: [BUG] modelname override + stream mutation is invalid
*Description*:

request:
```
curl -v http://x.x.x.x/v1/chat/completions -H "Authorization: Bearer token3" -H "Content-Type: application/json" -d '{
    "model": "tke/kimi-k2",
    "messages": [
        {"role": "user", "content": "hi there"}
    ],
    "stream": true
}'
```

response from llm

```
time=2025-08-08T08:29:29.553Z level=DEBUG msg="response body processing" request_id=2bc81a2e-75ba-4baa-8345-a7e092b611ca is_upstream_filter=false request="response_body:{body:\"{\\\"object\\\":\\\"error\\\",\\\"message\\\":\\\"[{'type': 'json_invalid', 'loc': ('body', 160), 'msg': 'JSON decode error', 'input': {}, 'ctx': {'error': 'Extra data'}}]\\\",\\\"type\\\":\\\"Bad Request\\\",\\\"param\\\":null,\\\"code\\\":400}\" end_of_stream:true} metadata_context:{}"
```

debug log in upstream filter:

```
{body:\"{\\n    \\\"model\\\": \\\"kimi-k2\\\",\\n    \\\"messages\\\": [\\n        {\\\"role\\\": \\\"user\\\", \\\"content\\\": \\\"hi there\\\"}\\n    ],\\n    \\\"stream\\\": true\\n,\\\"stream_options\\\":{\\\"include_usage\\\":true}}ue}}\"}}}
```

there is an extra `ue` at the end

we add include usage for stream at:

internal/extproc/chatcompletion_processor.go
```
// ProcessRequestBody implements [Processor.ProcessRequestBody].
func (c *chatCompletionProcessorRouterFilter) ProcessRequestBody(ctx context.Context, rawBody *extprocv3.HttpBody) (*extprocv3.ProcessingResponse, error) {
	model, body, err := parseOpenAIChatCompletionBody(rawBody)
	if err != nil {
		return nil, fmt.Errorf("failed to parse request body: %w", err)
	}
	if body.Stream && (body.StreamOptions == nil || !body.StreamOptions.IncludeUsage) && len(c.config.requestCosts) > 0 {
		// If the request is a streaming request and cost metrics are configured, we need to include usage in the response
		// to avoid the bypassing of the token usage calculation.
		body.StreamOptions = &openai.StreamOptions{IncludeUsage: true}
		// Rewrite the original bytes to include the stream_options.include_usage=true so that forcing the request body
		// mutation, which uses this raw body, will also result in the stream_options.include_usage=true.
		rawBody.Body, err = sjson.SetBytesOptions(rawBody.Body, "stream_options.include_usage", true, &sjson.Options{})
		if err != nil {
			return nil, fmt.Errorf("failed to set stream_options: %w", err)
		}
		c.forcedStreamOptionIncludeUsage = true
		// TODO: alternatively, we could just return 403 or 400 error here. That makes sense since configuring the
		// request cost metrics means that the gateway provisioners want to track the token usage for the request vs
		// setting this option to false means that clients are trying to escape that rule.
	}
```

mathetake:
**Description**

There was a bug in the translation of OpenAI->OpenAI layer where the body mutation is broken when forceBodyMutation = true. forceBodyMutation becomes true when stream_options is not present in the original body and we lacked the test case for it. 

Closes #1017

### DIFF
--- a/internal/extproc/translator/openai_openai.go
+++ b/internal/extproc/translator/openai_openai.go
@@ -36,7 +36,7 @@ type openAIToOpenAITranslatorV1ChatCompletion struct {
 }
 
 // RequestBody implements [OpenAIChatCompletionTranslator.RequestBody].
-func (o *openAIToOpenAITranslatorV1ChatCompletion) RequestBody(raw []byte, req *openai.ChatCompletionRequest, forceBodyMutation bool) (
+func (o *openAIToOpenAITranslatorV1ChatCompletion) RequestBody(original []byte, req *openai.ChatCompletionRequest, forceBodyMutation bool) (
 	headerMutation *extprocv3.HeaderMutation, bodyMutation *extprocv3.BodyMutation, err error,
 ) {
 	if req.Stream {
@@ -45,11 +45,10 @@ func (o *openAIToOpenAITranslatorV1ChatCompletion) RequestBody(raw []byte, req *
 	var newBody []byte
 	if o.modelNameOverride != "" {
 		// If modelName is set we override the model to be used for the request.
-		raw, err = sjson.SetBytesOptions(raw, "model", o.modelNameOverride, SJSONOptions)
+		newBody, err = sjson.SetBytesOptions(original, "model", o.modelNameOverride, SJSONOptions)
 		if err != nil {
 			return nil, nil, fmt.Errorf("failed to set model name: %w", err)
 		}
-		newBody = raw
 	}
 
 	// Always set the path header to the chat completions endpoint so that the request is routed correctly.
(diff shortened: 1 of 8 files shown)


yuzisun (approved):

internal/extproc/translator/openai_openai.go
```diff
@@ -65,8 +61,8 @@ func (o *openAIToOpenAITranslatorV1ChatCompletion) RequestBody(raw []byte, req *
 		},
 	}
 
-	if forceBodyMutation {
-		newBody = raw
+	if forceBodyMutation && len(newBody) == 0 {
```
yuzisun, line 64: Do we have the same issue in other translators ?

mathetake: well this is the openai->openai specific

PR #1022, opened 2025-08-08
feat: add native Anthropic API support for GCP integration

VarSuren:
**Description**

Allows to use Anthropic Native API with GCP Vertex.

Registered endpoint /v1/messages

Tested:

Streaming/non streaming

`
url -X POST http://localhost:1062/v1/messages \
  -H "x-ai-eg-selected-backend: gcp-anthropic-backend" \
  -d '{
    "model": "claude-3-7-sonnet@20250219",
    "max_tokens": 1024,
    "messages": [
      {
        "role": "user",
        "content": [{"type": "text", "text": "Hey Claude! Are you running on GCP?"}]
      }
    ],
    "temperature": 0.1,
    "stream": false
  }'
  `

Anthropic vs GCPAnthropic:

1. Anthropic API version is not equal to GCPAnthropic version i.e. 
  _Anthropic version passed in header of native API cannot be passed through to GCP. Configuration for that field is expected from yaml._

2. Model field that is passed in native API cannot be passed in the body to GCP.

3. I had to re-define root messages struct because go's implementation has no field stream ( in sdk those are two different methods )

4. I'm using `chatCompletionMetrics` in message factory as it's doing what we need and also use generic LLMTokenUsage but naming is misleading  maybe re-name it ?

**Open question**

<s>Official GCP SDK limit they way you can pass request i.e.

`content: "my content"` won't work with current set up in requires  `content : [{"text: "my content"}]`

As a workaround I suggest using plain Map we will loose validation and maybe translator will got a bit messier but it will allow to really do pass through  </s>

Using dictionary  as a struct as Claude code sending GO struct non compatible requests

TODO:

<s> verify integration with Claude code </s>

Claude code send to url + ?beta=true so I have to register both, didn't find a way to register dynamic URL.

Initial request from testing is fine regular gets dropped with connection issue but that's some internal configuration I'm not aware of, request itself is valid and can be used if you hit GCP directly.

Extproc config I used:

```
  - name: gcp-anthropic-backend
    schema:
      name: GCPAnthropic
      version: "vertex-2023-10-16"
```

### DIFF
--- a/tests/extproc/testupstream_test.go
+++ b/tests/extproc/testupstream_test.go
@@ -636,6 +636,73 @@ data: [DONE]
 			expStatus:       http.StatusOK,
 			expResponseBody: `{"choices":[{"message":{"content":"This is a test."}}]}`,
 		},
+		{
+			name:            "gcp-anthropicai - /anthropic/v1/messages",
+			backend:         "gcp-anthropicai",
+			path:            "/anthropic/v1/messages",
+			method:          http.MethodPost,
+			requestBody:     `{"model":"claude-3-sonnet","max_tokens":100,"messages":[{"role":"user","content":[{"type":"text","text":"Hello, just a simple test message."}]}],"stream":false}`,
+			expRequestBody:  `{"anthropic_version":"vertex-2023-10-16","max_tokens":100,"messages":[{"content":[{"text":"Hello, just a simple test message.","type":"text"}],"role":"user"}],"stream":false}`,
+			expHost:         "gcp-region-aiplatform.googleapis.com",
+			expPath:         "/v1/projects/gcp-project-name/locations/gcp-region/publishers/anthropic/models/claude-3-sonnet:rawPredict",
+			expHeaders:      map[string]string{"Authorization": "Bearer " + fakeGCPAuthToken},
+			responseStatus:  strconv.Itoa(http.StatusOK),
+			responseBody:    `{"id":"msg_123","type":"message","role":"assistant","stop_reason": "end_turn", "content":[{"type":"text","text":"Hello from native Anthropic API!"}],"usage":{"input_tokens":8,"output_tokens":15}}`,
+			expStatus:       http.StatusOK,
+			expResponseBody: `{"id":"msg_123","type":"message","role":"assistant","stop_reason": "end_turn", "content":[{"type":"text","text":"Hello from native Anthropic API!"}],"usage":{"input_tokens":8,"output_tokens":15}}`,
+		},
+		{
+			name:           "gcp-anthropicai - /anthropic/v1/messages - streaming",
+			backend:        "gcp-anthropicai",
+			path:           "/anthropic/v1/messages",
+			method:         http.MethodPost,
+			responseType:   "sse",
+			requestBody:    `{"model":"claude-3-sonnet","max_tokens":100,"messages":[{"role":"user","content":[{"type":"text","text":"Tell me a short joke"}]}],"stream":true}`,
+			expRequestBody: `{"anthropic_version":"vertex-2023-10-16","max_tokens":100,"messages":[{"content":[{"text":"Tell me a short joke","type":"text"}],"role":"user"}],"stream":true}`,
+			expHost:        "gcp-region-aiplatform.googleapis.com",
+			expPath:        "/v1/projects/gcp-project-name/locations/gcp-region/publishers/anthropic/models/claude-3-sonnet:streamRawPredict",
+			expHeaders:     map[string]string{"Authorization": "Bearer " + fakeGCPAuthToken},
+			responseStatus: strconv.Itoa(http.StatusOK),
+			responseBody: `event: message_start
+data: {"type": "message_start", "message": {"id": "msg_789", "usage": {"input_tokens": 8}}}
+
+event: content_block_start
+data: {"type": "content_block_start", "index": 0, "content_block": {"type": "text", "text": ""}}
+
+event: content_block_delta
+data: {"type": "content_block_delta", "index": 0, "delta": {"type": "text_delta", "text": "Why don't scientists trust atoms? Because they make up everything!"}}
+
+event: content_block_stop
+data: {"type": "content_block_stop", "index": 0}
+
+event: message_delta
+data: {"type": "message_delta", "delta": {"stop_reason": "end_turn"}, "usage": {"output_tokens": 15}}
+
+event: message_stop
+data: {"type": "message_stop"}
+
+`,
+			expStatus: http.StatusOK,
+			expResponseBody: `event: message_start
+data: {"type": "message_start", "message": {"id": "msg_789", "usage": {"input_tokens": 8}}}
+
+event: content_block_start
+data: {"type": "content_block_start", "index": 0, "content_block": {"type": "text", "text": ""}}
+
+event: content_block_delta
+data: {"type": "content_block_delta", "index": 0, "delta": {"type": "text_delta", "text": "Why don't scientists trust atoms? Because they make up everything!"}}
+
+event: content_block_stop
+data: {"type": "content_block_stop", "index": 0}
+
+event: message_delta
+data: {"type": "message_delta", "delta": {"stop_reason": "end_turn"}, "usage": {"output_tokens": 15}}
+
+event: message_stop
+data: {"type": "message_stop"}
+
+`,
+		},
 	} {
 		t.Run(tc.name, func(t *testing.T) {
 			require.Eventually(t, func() bool {


internal/extproc/messages_processor.go
```diff
@@ -0,0 +1,375 @@
...
+// ProcessResponseBody implements [Processor.ProcessResponseBody].
+func (c *messagesProcessorUpstreamFilter) ProcessResponseBody(ctx context.Context, body *extprocv3.HttpBody) (res *extprocv3.ProcessingResponse, err error) {
+	defer func() {
+		c.metrics.RecordRequestCompletion(ctx, err == nil, c.requestHeaders)
+	}()
+	var br io.Reader
+	var isGzip bool
+	switch c.responseEncoding {
+	case "gzip":
+		br, err = gzip.NewReader(bytes.NewReader(body.Body))
```
VarSuren: This is applicable to openAI python client as it uses httpx python object, plain curl doesn't care about those things we need to verify with Anthropic Python SDK but I assume it would need same removal. 

Bottomline: needs to be verified with SDK's but curl works fine

cmd/extproc/mainlib/main.go
```diff
@@ -159,6 +159,7 @@ func Main(ctx context.Context, args []string, stderr io.Writer) (err error) {
 	}
 	server.Register("/v1/chat/completions", extproc.ChatCompletionProcessorFactory(chatCompletionMetrics))
 	server.Register("/v1/embeddings", extproc.EmbeddingsProcessorFactory(embeddingsMetrics))
+	server.Register("/v1/messages", extproc.MessagesProcessorFactory(chatCompletionMetrics))
```
vdabravolski: maybe `anthropic/v1/message` for clean separation of vendor-specific APIs?

VarSuren: yeah makes sense, prolly make it configurable from yaml ? @yuzisun  your take ?

mathetake: that is exactly what we are working on it in #1020, so leaving it to us for that separation concern

yuzisun:
@VarSuren can you sign off the commits ?

internal/extproc/messages_processor.go
```diff
@@ -0,0 +1,375 @@
...
+
+	headerMutation, bodyMutation, tokenUsage, err := c.translator.ResponseBody(c.responseHeaders, br, body.EndOfStream)
+	if err != nil {
+		return nil, fmt.Errorf("failed to transform response: %w", err)
+	}
+	if bodyMutation != nil && isGzip {
+		if headerMutation == nil {
+			headerMutation = &extprocv3.HeaderMutation{}
+		}
+		// TODO: this is a hotfix, we should update this to recompress since its in the header
```
yuzisun: @VarSuren we should not need this logic here to remove content encoding as we already inserted the header mutation filter in the extension server, can you check the implementation in the chat completion processor? https://github.com/envoyproxy/ai-gateway/pull/818/files

VarSuren: sure lemme check

cmd/extproc/mainlib/main.go
mathetake: The PR landed. you can check the change there and add anthropic prefix config accordingly to do the separation

cmd/extproc/mainlib/main.go
```diff
@@ -165,6 +165,8 @@ func Main(ctx context.Context, args []string, stderr io.Writer) (err error) {
...
+	server.Register("/v1/messages?beta=true", extproc.MessagesProcessorFactory(chatCompletionMetrics))
```
yuzisun: Do we need to register this separately ? use case should still be able pass query parameters

internal/apischema/anthropic/anthropic.go
```diff
@@ -0,0 +1,33 @@
...
+// SPDX-License-Identifier: Apache-2.0
+// The full text of the Apache license is available in the LICENSE file at
+// the root of the repo.
+
+// Package anthropic contains Anthropic API schema definitions using the official SDK types.
+package anthropic
+
+// MessagesRequest represents a request to the Anthropic Messages API.
+// Uses a dictionary approach to handle any JSON structure flexibly.
+type MessagesRequest map[string]interface{}
```
yuzisun, line 11: what's the reason to change to the dict type , is it for efficiency?

VarSuren: GOSDK has different signatures not compatible with Claude code requests i.e. ```content: "hello I'm Claude"``` , go only expects ```content: [{"type": "text", "content": "Hey I'm Claude"}]```

cmd/extproc/mainlib/main.go
mathetake: yeah 100% this should be handled in one processor.

VarSuren: Usually yes, but I didn't find a way to do dynamic query

mathetake: I think we might need to fix `processorForPath` to trim query on processor search

internal/extproc/translator/anthropic_gcpanthropic.go
```diff
@@ -0,0 +1,159 @@
...
+	apiVersion        string
+	modelNameOverride string
+}
+
+// RequestBody implements [AnthropicMessagesTranslator.RequestBody] for Anthropic to GCP Anthropic translation.
+// This handles the transformation from native Anthropic format to GCP Anthropic format.
+func (a *anthropicToGCPAnthropicTranslator) RequestBody(_ []byte, body *anthropicschema.MessagesRequest, _ bool) (
+	headerMutation *extprocv3.HeaderMutation, bodyMutation *extprocv3.BodyMutation, err error,
+) {
+	log.Println("hit GCP translator")
```
yuzisun: remove the debug log line

cmd/extproc/mainlib/main.go
VarSuren: I initially did that, but that would have affected all endpoint, don't know whether other endpoint might use them in future

cmd/extproc/mainlib/main.go
VarSuren: thank you for more merge conflicts :)

internal/extproc/translator/anthropic_gcpanthropic.go
```diff
@@ -0,0 +1,159 @@
...
+	return
+}
+
+// ResponseHeaders implements [AnthropicMessagesTranslator.ResponseHeaders] for Anthropic to GCP Anthropic.
+func (a *anthropicToGCPAnthropicTranslator) ResponseHeaders(headers map[string]string) (
+	headerMutation *extprocv3.HeaderMutation, err error,
+) {
+	if status, ok := headers[":status"]; ok {
+		log.Printf("response status: %s", status)
+	}
```
yuzisun: remove the debug log line

internal/extproc/translator/anthropic_gcpanthropic.go
```diff
@@ -0,0 +1,159 @@
...
+	// For Anthropic to GCP Anthropic, no header transformation is needed.
+	return nil, nil
+}
+
+// ResponseBody implements [AnthropicMessagesTranslator.ResponseBody] for Anthropic to GCP Anthropic.
+// This is essentially a passthrough since both use the same Anthropic response format.
+func (a *anthropicToGCPAnthropicTranslator) ResponseBody(_ map[string]string, body io.Reader, endOfStream bool) (
+	headerMutation *extprocv3.HeaderMutation, bodyMutation *extprocv3.BodyMutation, tokenUsage LLMTokenUsage, err error,
+) {
+	log.Printf("hit translator - processing response body (endOfStream: %v)", endOfStream)
```
yuzisun: same here

internal/extproc/translator/anthropic_gcpanthropic.go
```diff
@@ -0,0 +1,159 @@
...
+	}
+
+	// Pass through the response body unchanged since both input and output are Anthropic format.
+	headerMutation = &extprocv3.HeaderMutation{}
+	setContentLength(headerMutation, bodyBytes)
+	bodyMutation = &extprocv3.BodyMutation{
+		Mutation: &extprocv3.BodyMutation_Body{Body: bodyBytes},
+	}
+
+	log.Println("response translation completed")
```
yuzisun: same here

cmd/extproc/mainlib/main.go
VarSuren: rebased and fixed

cmd/extproc/mainlib/main.go
VarSuren: but also think about it, you might want to have different processor for beta if say it has additional fields functionality

yuzisun (approved):

cmd/extproc/mainlib/main.go
```diff
@@ -86,6 +88,12 @@ func parseAndValidateFlags(args []string) (extProcFlags, error) {
 		"OpenAI endpoint prefix to be used for the external processor. This is used to route requests to the correct handler. "+
 			"Defaults to /v1, which is the standard OpenAI API prefix.",
 	)
+	fs.StringVar(&flags.anthropicPrefix,
+		"anthropicPrefix",
+		"/v1",
```
mathetake: can you have `/anthropic/v1` as a default? The whole point of this config is to have a separation. E.g. /models/ endpoint exists for both openai and anthropic.

Also you need to do the plumbing from helm config value to here.

VarSuren: sure thing, lemme address this

internal/extproc/translator/anthropic_gcpanthropic.go
VarSuren: addressed, btw what's the approach for this project, we don't use logs at all ?

cmd/extproc/mainlib/main.go
VarSuren: addressed, please verify

mathetake: helm option plumbing is still not done. you can refer to the PR above on what to do.

cmd/extproc/mainlib/main.go
yuzisun: Since for Anthropic we are mostly pass through and let cloud handle the validation, we should be fine with a single processor.

internal/extproc/translator/anthropic_gcpanthropic.go
yuzisun: We still want logging if needed but this one obviously is not :)

cmd/extproc/mainlib/main.go
mathetake: nvm, you won't need to do the hem plumbing in this PR!

internal/extproc/translator/anthropic_gcpanthropic.go
VarSuren: arguable but okay

mathetake:
We need extproc e2e test cases here to cover the code: https://github.com/envoyproxy/ai-gateway/blob/fb1f8757898ac5e828c3f2c69f71aeec3c1fbfca/tests/extproc/testupstream_test.go#L75

mathetake:
q: this doesn't support a real Anthropic backend yet right? I think that sounds like a must-have if we want to promote this as a "feature" in v0.3. How long do you think it would take to get it working in a follow up PR? (Also this needs a documentation as well 100%)

yuzisun:
> q: this doesn't support a real Anthropic backend yet right? I think that sounds like a must-have if we want to promote this as a "feature" in v0.3. How long do you think it would take to get it working in a follow up PR? (Also this needs a documentation as well 100%)

This is a real Anthropic backend just it is hosted on gcp, I think for 0.3 we are targeting to fully support gcp Gemini via chat completion and Anthropic models via either chat completion or native Anthropic message API. We can target for 0.4 to support first party Anthropic and AWS Anthropic via Anthropic message API.

mathetake:
sounds good. could you create the umbrella issue that tracks them like "Anthropic Input/SDK Support" that documents what works now and what doesn't as well as the plans for v0.4? Then this PR is good to go

mathetake:
oh i think we can reuse https://github.com/envoyproxy/ai-gateway/issues/847 -- can you post the updates assuming this PR lands? I haven't gotten around to reviewing all but will land it either i push the changes or as-is within today on west coast

yuzisun:
> oh i think we can reuse #847 -- can you post the updates assuming this PR lands? I haven't gotten around to reviewing all but will land it either i push the changes or as-is within today on west coast

Thanks! updated the issue

mathetake (approved):

PR #1026, opened 2025-08-09
controller: set allowModeOverride=true

mathetake:
**Description**

Regression from #963


mathetake:
i need to add a regression test case but hope this resolves the issue @Xunzhuo

Xunzhuo:
Cool, let us add an e2e in a follow up, thanks @mathetake !

PR #1027, opened 2025-08-10
feat: Adds OpenInference redaction config and documentation

codefromthecrypt:
**Description**

This simplifies OpenInference attribute recording while adding redaction configuration via standard ENV.

#791 

It also improves the aigw README until we have a better place to capture documentation like this. Likely, we should do that when k8s configuration is complete.

@mikeldking maybe we can swap notes here on your new docs at the openinference site or cross-link?


codefromthecrypt:
is codecov really looking for 100pct coverage? that's usually not a smart goal

cmd/extproc/mainlib/main.go
```diff
@@ -262,55 +254,6 @@ func startMetricsServer(lis net.Listener, logger *slog.Logger) (*http.Server, me
 	return server, meter
 }
 
-// noopTextMapPropagator is a no-op implementation of [propagation.TextMapPropagator].
-type noopTextMapPropagator struct{}
```
mathetake, line 266: i like less code in main.go

codefromthecrypt:
sweet thanks @mathetake and yeah I'm d'facto CODEOWNER on this so any issues mention me

PR #1029, opened 2025-08-11
controller: handles non-default listener filter chains

### LINKED ISSUE #1028: listener.defaultFilterChain can be nil which results in router filter not being assigned
*Description*:
>Describe the desired behavior, what scenario it enables and how it
would be used.

Currently, it is possible for users to deploy Gateway resources in namespaces other than envoy-gateway. However, the [post_translate_modify.go](https://github.com/envoyproxy/ai-gateway/blob/main/internal/extensionserver/post_translate_modify.go#L350-L360) logic appears to hardcode a prefix check for listener names starting with "envoy-gateway", which causes listeners defined in other namespaces to be skipped entirely during route configuration processing.

```
for _, listener := range listeners {
	if strings.HasPrefix(listener.Name, "envoy-gateway") {
		continue
	}
	routeConfigName := findListenerRouteConfig(listener)
	listenerNameToRouteName[listener.Name] = routeConfigName
	listenerNameToListener[listener.Name] = listener
}
```

As a result, valid Gateway and Listener resources deployed in other namespaces are silently ignored, even if they are correctly defined and referenced.

[optional *Relevant Links*:]

mathetake:
**Description**

Previously, we didn't handle the non-default listener filter chains in the extensionserver, which resulted in a bug where it doesn't exist. Notably, it can happen when TLS is configured on a listener. This fixes the extensionserver code so that it handles both default and non-default filter chains.

Closes #1028

### DIFF
--- a/cmd/aigw/testdata/translate_basic.out.yaml
+++ b/cmd/aigw/testdata/translate_basic.out.yaml
@@ -97,6 +97,8 @@ spec:
             - name: x-ai-eg-model
               type: Exact
               value: gpt-4o-mini
+          path:
+            value: "/"
       timeouts:
         request: 60s
     - backendRefs:
@@ -114,6 +116,8 @@ spec:
             - name: x-ai-eg-model
               type: Exact
               value: llama3-2-1b-instruct-v1
+          path:
+            value: "/"
       timeouts:
         request: 60s
     - backendRefs:
@@ -131,6 +135,8 @@ spec:
             - name: x-ai-eg-model
               type: Exact
               value: some-cool-self-hosted-model
+          path:
+            value: "/"
       timeouts:
         request: 60s
     - matches:
--- a/cmd/aigw/translate.go
+++ b/cmd/aigw/translate.go
@@ -206,7 +206,7 @@ func translateCustomResourceObjects(
 	aisbC := controller.NewAIServiceBackendController(fakeClient, fakeClientSet, logr.FromSlogHandler(logger.Handler()),
 		make(chan event.GenericEvent))
 	airC := controller.NewAIGatewayRouteController(fakeClient, fakeClientSet, logr.FromSlogHandler(logger.Handler()),
-		make(chan event.GenericEvent),
+		make(chan event.GenericEvent), "/",
 	)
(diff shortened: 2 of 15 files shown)


mathetake:
@Xunzhuo same change must be applied to the inference pool code as well. otherwise when the listener speaks TLS then filter chain won't be updated

mathetake:
ok went ahead and applies the same fix to inference pool stuff as well

internal/extensionserver/inferencepool.go
```diff
@@ -358,18 +358,32 @@ func searchInferencePoolInFilterChain(pool *gwaiev1a2.InferencePool, chain []*ht
...
-// Tries to find the route config name in the provided listener.
-func findListenerRouteConfig(listener *listenerv3.Listener) string {
+// findListenerRouteConfigs extracts route configuration names from the listener's filter chains.
+func findListenerRouteConfigs(listener *listenerv3.Listener) []string {
+	var names []string
 	// First, get the filter chains from the listener.
+	for _, filterChain := range listener.FilterChains {
+		httpConManager, _, err := findHCM(filterChain)
+		if err != nil {
+			continue // Skip this filter chain if it doesn't have an HTTP connection manager.
```
yuzisun: Is it possible that the filter chain does not have an http connection manager ?

internal/extensionserver/inferencepool.go
```diff
@@ -358,18 +358,32 @@ func searchInferencePoolInFilterChain(pool *gwaiev1a2.InferencePool, chain []*ht
 	return nil, -1, nil
 }
 
-// Tries to find the route config name in the provided listener.
-func findListenerRouteConfig(listener *listenerv3.Listener) string {
+// findListenerRouteConfigs extracts route configuration names from the listener's filter chains.
+func findListenerRouteConfigs(listener *listenerv3.Listener) []string {
```
yuzisun: This function is not specific to inference pool, can we move it out of this file ?

internal/extensionserver/inferencepool.go
mathetake: yes, if the listener is only configured with the l4 filters. Not sure if that happens to EG though.

PR #1054, opened 2025-08-11
e2e: add separate e2e target for infext

mathetake:
**Description**

This PR separates out the inference-extension related e2e tests from the other normal tests. The rationale is that it requires a separate Envoy Gateway config different than the default, so we would like to ensure that both works correctly on CI. Also, this will reduce the overall CI time.


tests/internal/e2elib/e2elib.go
mathetake, line 1: this is simply moved from e2e/e2e_test.go. so nothing interesting

tests/e2e/e2e_test.go
mathetake, line 1: this is simply moved to internal/e2elib. so nothing interesting

Xunzhuo:
Looks good!

PR #1063, opened 2025-08-12
controller: removes openAIPrefix config

mathetake:
**Description**

The configurability for "/v1" was introduced in #1020. However, it is unnecessary configuration no one asked for at the moment given that we keep the rootPrefix for the separation concern between AIGatewayRoutes vs HTTPRoutes.

This partially reverts #1020, and removes the config so that we can have a simpler config overall. We can revisit this if anyone asks for it later. If so, I think it will be a time to think about Gateway level CRD.

Follow up on #1020


cmd/controller/main.go
```diff
@@ -122,11 +121,6 @@ func parseAndValidateFlags(args []string) (flags, error) {
 		"/",
 		`The root prefix for all supported endpoints. Default is "/"`,
 	)
-	openAIPrefix := fs.String(
```
yuzisun, line 126: where does people configure if they want `/openai/v1/chat/completion` ?

mathetake: i think the main point was to wait to see if there's anyone like that ...

siddharth1036:
@mathetake the changes look good, should we also log the prefix in extproc which would make it easier for debugging? Also extproc still has openAIPrefix as the config variable name, should we rename it to avoid any confusion between controller rootPrefix and openAIPrefix from code perspective?

mathetake:
@siddharth1036 thank you for confirming that this looks good to you, and for review comments. I addressed them, and pushed the change!

cmd/controller/main.go
mathetake: as agreed by @Xunzhuo offline, let's remove this for now until someone asks for this configurability to reduce the complexity in our config and codebase. Two level indirection on the path configs seems not that user friendly

mathetake:
oh forgot the logging part

Xunzhuo (approved):

PR #1068, opened 2025-08-12
extproc: chat completion tracing support for all backends

mathetake:
**Description**

This commit is a follow up on the initial tracing implementation. There was a bug in vcr tests where streaming mode override is wrongly taken place at upstream filter level, which resulted in buffering all events at the router filter as opposed to the reality that they are streamed (not necessarily on the event boundary, as it's a tcp dataframe level), and that is why previously already passing the tests despite not taking care of the event boundaries. 

Notably, this makes the tracing do
* Support for all backends regardless of the schema vs previously only works for OpenAI backend.
* Correctly handling all streaming events vs previously assuming all events are available in buffer which is not true in end to end.
* Avoids unnecessary copies around the tracing; This utilizes the already-parsed data structures, so this makes the tracing zero-copy as long as the redaction is configured.


internal/tracing/openinference/openai/response_attrs.go
```diff
@@ -71,3 +71,7 @@ func buildResponseAttributes(resp *openai.ChatCompletionResponse, config *openin
 	}
 	return attrs
 }
+
+func buildResponseAttributesForChunk(chunk *openai.ChatCompletionResponseChunk, config *openinference.TraceConfig) []attribute.KeyValue {
+	panic("TODO")
```
mathetake: huge todo

internal/extproc/translator/openai_gcpanthropic_stream.go
```diff
@@ -73,9 +74,10 @@ func (p *anthropicStreamParser) writeChunk(eventBlock []byte, buf *[]byte) error
 
 // Process reads from the Anthropic SSE stream, translates events to OpenAI chunks,
 // and returns the mutations for Envoy.
-func (p *anthropicStreamParser) Process(body io.Reader, endOfStream bool) (
+func (p *anthropicStreamParser) Process(body io.Reader, endOfStream bool, span tracing.ChatCompletionSpan) (
 	*extprocv3.HeaderMutation, *extprocv3.BodyMutation, LLMTokenUsage, error,
 ) {
+	_ = span // TODO: add support for streaming chunks in tracing.
```
mathetake, line 80: i think i will punt this to a follow up as it seems a bit different than other translators (100% feasible though.)

internal/tracing/openinference/openai/chat_completion.go
```diff
@@ -61,42 +60,51 @@ func (r *ChatCompletionRecorder) RecordRequest(span trace.Span, chatReq *openai.
...
-	if bytes.Contains(body, []byte("data: ")) {
-		// Convert SSE to a single completion response.
-		if jsonBody, convErr := convertSSEToJSON(body); convErr == nil && len(jsonBody) > 0 {
-			body = jsonBody
+	// Set output attributes.
+	var attrs []attribute.KeyValue
+	attrs = buildResponseAttributesForChunk(chunk, r.traceConfig)
+
+	// TODO: what to do with the "output" attribute for streaming? Accumulate it? Right now the last chunk
+	// 	will overwrite the previous ones.
```
mathetake: I read the doc but not sure how the "output" should look like for streaming. There's no spec on it. @codefromthecrypt

internal/tracing/openinference/openai/response_attrs.go
```diff
@@ -71,3 +71,71 @@ func buildResponseAttributes(resp *openai.ChatCompletionResponse, config *openin
...
+
+	// Note: compound match here is from Python OpenInference OpenAI config.py.
+	if !config.HideOutputs && !config.HideOutputMessages {
+		for i, choice := range chunk.Choices {
+			delta := choice.Delta
+			if delta == nil {
+				continue
+			}
+			// TODO: OutputMessageAttribute and OutputMessageToolCallAttribute should take the index of the chunk, otherwise
+			//  the latter chunk will overwrite the previous ones.
```
mathetake: This also needs your input @codefromthecrypt

mathetake:
So the extproc tests vcr are passing except for the streaming which basically are related to two TODOs I tagged @codefromthecrypt 

After that, I think I will fix the unit tests and then all good

internal/tracing/openinference/openai/chat_completion.go
codefromthecrypt: there's code in sse. go which coalesces the SSE events into normal completion format. it would match https://github.com/envoyproxy/ai-gateway/blob/main/tests/internal/testopeninference/spans/chat-streaming.json and the old code should have already done that

internal/tracing/openinference/openai/response_attrs.go
codefromthecrypt: I think basically we need to move or revisit  this logic, renaming as necessary https://github.com/envoyproxy/ai-gateway/blob/main/internal/tracing/openinference/openai/sse_converter.go  as this matched the spans produced by openinference. If we need to record a bigger or different request to clarify behaviour, I can help with that if https://github.com/envoyproxy/ai-gateway/tree/main/tests/internal/testopenai#recording-new-cassettes and https://github.com/envoyproxy/ai-gateway/tree/main/tests/internal/testopeninference#recording-new-spans aren't clear. basically the easiest way to get onto the same page is to reproduce whatever is confusing as an openai cassette, then make a span from that and we have our gold state

mathetake:
so passed all the existing vcr tests, so the change is 100% compliant and not changing the semantics at all. Tomorrow i will polish the comment, naming etc as well as the busted unit tests. I think with this, we can confidently say "we have a support for openinference" 😎

tests/internal/testopeninference/spans/chat-streaming.json
```diff
@@ -24,7 +24,7 @@
...
+        "stringValue": "{\"choices\": [{\"message\": {\"annotations\":[], \"content\": \"Hello! How can I assist you today?\", \"role\": \"assistant\"}, \"index\": 0, \"finish_reason\": \"stop\"}], \"id\": \"chatcmpl-Bx5kRNsKBfYXdH2IeSETLB6DQk1kC\", \"created\": 1753423147, \"model\": \"gpt-4.1-nano-2025-04-14\", \"object\": \"chat.completion.chunk\", \"service_tier\": \"default\", \"system_fingerprint\": \"fp_479cfdfab2\", \"usage\": {\"completion_tokens\": 9, \"prompt_tokens\": 19, \"total_tokens\": 28, \"completion_tokens_details\": {\"accepted_prediction_tokens\": 0, \"audio_tokens\": 0, \"reasoning_tokens\": 0, \"rejected_prediction_tokens\": 0}, \"prompt_tokens_details\": {\"audio_tokens\": 0, \"cached_tokens\": 0}}}"
```
mathetake: note: annotations is not optional, so this needs to exist

tests/internal/testopeninference/spans/chat-streaming-detailed-usage.json
```diff
@@ -28,7 +28,7 @@
...
+        "stringValue":  "{\"choices\": [{\"message\": {\"annotations\":[], \"content\": \"Hello!\", \"role\": \"assistant\"}, \"index\": 0, \"finish_reason\": \"stop\"}], \"id\": \"chatcmpl-Bx5lfsiWaJAorkTpnbrjlnAnl0892\", \"created\": 1753423223, \"model\": \"gpt-4.1-nano-2025-04-14\", \"object\": \"chat.completion.chunk\", \"service_tier\": \"default\", \"system_fingerprint\": \"fp_38343a2f8f\", \"usage\": {\"completion_tokens\": 2, \"prompt_tokens\": 9, \"total_tokens\": 11, \"completion_tokens_details\": {\"accepted_prediction_tokens\": 0, \"audio_tokens\": 0, \"reasoning_tokens\": 0, \"rejected_prediction_tokens\": 0}, \"prompt_tokens_details\": {\"audio_tokens\": 0, \"cached_tokens\": 0}}}"
```
mathetake: note: annotations is not optional, so this needs to exist

mathetake:
Found another bug..

tests/internal/testopeninference/spans/chat-streaming-detailed-usage.json
codefromthecrypt: you are editing files that are created by python and will be overwritten when re-recorded. Maybe I should update the README to say that comparison is lenient? e.g. tests/internal/testopeninference/assertions.go already walks around things who we serialize differently in minor ways.

codefromthecrypt: I think this was an old find/replace issue. I will raise a PR with a re-recording of these spans. Looks like annotations are in there.

internal/apischema/openai/openai.go
```diff
@@ -977,7 +977,12 @@ type ChatCompletionResponseChoiceMessage struct {
 	ToolCalls []ChatCompletionMessageToolCallParam `json:"tool_calls,omitempty"`
 
 	// Annotations for the message, when applicable, as when using the web search tool.
-	Annotations []Annotation `json:"annotations,omitempty"`
+	//
+	// Note: do not use the value of slice, but instead use a pointer to the slice since otherwise,
```
codefromthecrypt: Maybe move this to a unit test that proves the round trip concern? Then you can shorten the comment and be sure it wont be undone unless someone deletes the test

mathetake:
Cool finally ready. Just in time for v0.3

missBerg (approved):

mathetake:
FYI @codefromthecrypt I must admit the tracing package API change/naming here might be awkward. We can always refactor them as it's internal package anyways. The important thing is that with this the tracing functionality is now perfect, which we want for v0.3. Let's do the refinement part in the followups... this is just the beginning!

PR #1074, opened 2025-08-13
feat: support bedrock reasoning content response

alexagriffith:
**Description**

support returning reasoning content from aws bedrock in openai. 
required adding aws bedrock fields that were missing and requires returning extra fields. since we are already refactoring to use openai go sdk, it makes sense to implement this here instead of adding it to our custom openai struct.


internal/extproc/translator/openai_awsbedrock.go
```diff
@@ -467,21 +513,20 @@ func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) bedrockStopReasonToOpenAI
 
 func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) bedrockToolUseToOpenAICalls(
 	toolUse *awsbedrock.ToolUseBlock,
-) *openai.ChatCompletionMessageToolCallParam {
+) openaigo.ChatCompletionMessageToolCall {
```
sukumargaonkar, line 516: switch to a pointer return?

internal/extproc/translator/openai_awsbedrock.go
```diff
@@ -582,40 +627,50 @@ func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) ResponseBody(_ map[string
...
-		Message: openai.ChatCompletionResponseChoiceMessage{
-			Role: bedrockResp.Output.Message.Role,
+	choice := CustomChatCompletionChoice{
+		Message: CustomChatCompletionMessage{
+			ChatCompletionMessage: openaigo.ChatCompletionMessage{
+				Role: openAIconstant.Assistant(bedrockResp.Output.Message.Role),
+			},
+			ExtraFields: make(map[string]interface{}),
 		},
-		FinishReason: o.bedrockStopReasonToOpenAIStopReason(bedrockResp.StopReason),
```
sukumargaonkar, line 609: can we set the finishReason inplace?

sukumargaonkar: same for Index

internal/extproc/translator/openai_awsbedrock.go
```diff
@@ -582,40 +627,50 @@ func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) ResponseBody(_ map[string
...
+
 	for _, output := range bedrockResp.Output.Message.Content {
-		if toolCall := o.bedrockToolUseToOpenAICalls(output.ToolUse); toolCall != nil {
-			choice.Message.ToolCalls = []openai.ChatCompletionMessageToolCallParam{*toolCall}
-		} else if output.Text != nil {
-			// For the converse response the assumption is that there is only one text content block, we take the first one.
-			if choice.Message.Content == nil {
-				choice.Message.Content = output.Text
+		switch {
+		case output.ToolUse != nil:
```
sukumargaonkar, line 663: `ToolUse` and `Text` cannot be non-nil at the same time, right?
if yes, can you add comment about it or point to docs about it

internal/extproc/translator/openai_awsbedrock.go
alexagriffith: we just dereference the ptr right after returning, what do u think is best to do?

Issue #1064, opened 2025-08-12
Install AIGW cli tool not working.


Dean-Coakley:
*Description*:
>What issue is being seen? Describe what should be happening instead of
the bug, for example: Envoy should not crash, the expected value isn't
returned, etc.

* Since https://github.com/envoyproxy/ai-gateway/commit/a1eea8f537d1b50f76f78c12f765b97618b8c376  cli of the `@main` version cannot be installed because of usage of `replace` in go.mod : https://github.com/envoyproxy/ai-gateway/blob/992ce5042096b791326d025668e3b1ed08745d79/go.mod#L5-L7

*Repro steps*:
> Include sample requests, environment, etc. All data and inputs
required to reproduce the bug.

```
$ go install github.com/envoyproxy/ai-gateway/cmd/aigw@main
go: github.com/envoyproxy/ai-gateway/cmd/aigw@main (in github.com/envoyproxy/ai-gateway@v0.2.1-0.20250812041027-992ce5042096):
        The go.mod file for the module providing named packages contains one or
        more replace directives. It must not contain directives that would cause
        it to be interpreted differently than if it were the main module.
```

For now a 'workaround'ish v0.2.1 can be installed `go install github.com/envoyproxy/ai-gateway/cmd/aigw@v0.2.1`

>**Note**: If there are privacy concerns, sanitize the data prior to
sharing.

*Environment*:
>Include the environment like gateway version, envoy version and so on.

*Logs*:
>Include the access logs and the Envoy logs.

Dean-Coakley:
@mathetake Can this be fixed now, or must we wait for next release? I would be open to sending a PR if you like

mathetake:
good catch. we need to fix it before v0.3. if you can send a patch, that would be awesome.

mathetake:
maybe `replace go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc v0.61.0 => go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc v0.60.0` this is tricky. Ultimately, i think we can compromise on forcing users to clone the repo or  publish the binary in the release. It's best if we can enable the plain go install though

Dean-Coakley:
[...]

PR #1081, opened 2025-08-14
docs: workaround for aigw CLI installation

mathetake:
**Description**

Currently, our go.mod has replace statements due to the very complex dependency resolution for otelgrpc package. Unfortunately, it causes the problem with go install command which cannot support the module with replace statements. https://github.com/golang/go/issues/44840

This adds a workaround for it temporarily for v0.3. We hope we can resolve the issue by v0.4.

Workaround for https://github.com/envoyproxy/ai-gateway/issues/1064

### DIFF
--- a/site/docs/cli/installation.md
+++ b/site/docs/cli/installation.md
@@ -9,7 +9,7 @@ To install the `aigw` CLI, run the following command (This may take a while):
 
 ```shell
 git clone https://github.com/envoyproxy/ai-gateway.git
-go install ./cmd/aigw@main
+go install ./cmd/aigw
 ```
 
 :::tip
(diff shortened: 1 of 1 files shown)


site/docs/cli/installation.md
```diff
@@ -19,6 +20,8 @@ For example, you can add the following line to your shell profile (e.g., `~/.bas
 ```sh
 export PATH=$PATH:$(go env GOPATH)/bin
 ```
+
```
aabchoo: nit: wouldn't the order of operations be explaining why go clone then go install?

aabchoo (approved):
🧨

mathetake:
Thanks for the look!!!

PR #1084, opened 2025-08-15
feat: vllm specic fields for embedding completion

hustxiayang:
**Description**

vLLM specific fields for embedding completions.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -7,7 +7,7 @@ on:
 jobs:
   docker_push:
     name: Push Docker Images
-    uses: ./.github/workflows/docker_builds_template.yaml
+    uses: ./.github/workflows/docker_build_job.yaml
     secrets: inherit
 
   release:


mathetake:
do you want to add documentation to https://github.com/envoyproxy/ai-gateway/blob/main/site/docs/capabilities/llm-integrations/vendor-specific-fields.md ?

hustxiayang:
@mathetake  This is not ready and used for discussion. Thanks a lot for your suggestions!

internal/apischema/openai/openai.go
```diff
@@ -1320,3 +1325,19 @@ type AnthropicVendorFields struct {
 	// https://docs.anthropic.com/en/api/messages#body-thinking
 	Thinking *anthropic.ThinkingConfigParamUnion `json:"thinking,omitzero"`
 }
+
+type vLLMEmbeddingVendorFields struct {
+	AddSpecialTokens param.Opt[bool] `json:"add_special_tokens,omitzero"`
```
mathetake, line 1330: can you not use openai sdk just for the optional.

hustxiayang: Thanks! Relevant questions: Currently, `chatcompletion` and `embedding` requests are manually maintained. Do you have plans to import https://github.com/openai/openai-go directly and embed new fields for different providers? I thought you might do this so just tried to make it consistent.

mathetake: > Do you have plans to import https://github.com/openai/openai-go directly and embed new fields for different providers

not a plan but some people started thinking about it, but not sure if we want to do that due to the loss of our control, lack of compliance in their OpenAPI (not AI) specification etc. See my old comment here https://github.com/envoyproxy/ai-gateway/blob/e33a5f3fa4ae5eb17b1034c7bcb3932e714a7bf2/internal/apischema/openai/openai.go#L6-L8

For example, they sometimes use the "undocumented" field etc. Given that SDK is auto-generated by some vendor (stainless), I am not 100% sure if we want to do that. Maintaining the schema for select a few endpoints is not that pain in other words.

edited: on the other hand, I am not completely opposed to it either. Just that the current state gives me the impression that using it seems like a cure worse than the disease. my 2c.

gavrissh: My understanding is that these added fields in the request body will be forwarded to the self-hosted vLLM intact, even without the changes of explicitly adding them to the API schema. 

vLLM supports many such fields in chat completions and embedding. Should this just be a documentation update highlighting that vLLM fields are supported, or an example?

mathetake: oh yeah you are right @gavrissh 

@hustxiayang since the translation is not happening currently for embedding endpoint, this PR doesn't do anything meaningful functionality-wise. In other words, OpenAI->OpenAI translation is pass-through, and no body mutation happens, hence adding vllm embedding specific field like your current change is ignored.

mathetake:
https://github.com/envoyproxy/ai-gateway/pull/1084#discussion_r2296188389 


We should do the documentation about the passthrough behavior maybe as per the comment by @gavrissh. Feel free to reopen the PR once you repurpose this to do the documentation on that. Thank you!

internal/apischema/openai/openai.go
hustxiayang: @mathetake Hi, sorry that did not have time for this pr. I do not think this is the case. This is not for the translation, but for request validation. I think @yuzisun is working on request validation for ai-gateway.

PR #1086, opened 2025-08-15
chore: fixes codecov exclusion on examples

mathetake:
**Description**

### DIFF
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -13,7 +13,7 @@ coverage:
         if_ci_failed: error
 ignore:
   - "cmd"
-  - "examples"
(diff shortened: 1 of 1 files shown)


mathetake:
hmm no idea on why recursive matching ignore doesn't take any effect on codecov. I see internal/testing/**/* in the ignore already on main branch vs I still can find these files on the UI which says they are not covered (https://app.codecov.io/gh/envoyproxy/ai-gateway?displayType=list&search=internal%2Ftesting). That's why the codecov bot gets angry about the project level coverage too low.

mathetake:
HELP WANTED

PR #1088, opened 2025-08-16
controller: do serverside apply on finalizer

mathetake:
**Description**
TODO


mathetake:
cc @arkodg thank you for the advice in another project

mathetake:
looks like somthing is broken ;)

PR #1092, opened 2025-08-18
docs: improve standalone README with an annotated Phoenix screenshot

codefromthecrypt:
**Description**

This improves the `aigw` quick start README by explaining the example a bit more, and in the case of OpenTelemetry, showing an annotated OpenTelemetry trace in [Phoenix](https://phoenix.arize.com).

phoenix

This also polishes some go versions in the example docker config that mismatched the project, and removed an unnecessary admin configuration (port) for standalone. This is distracting and no-longer needed since latest gateway uses latest func-e which no longer requires a user to explicitly add admin config.


cmd/aigw/testdata/ai-gateway-ollama-docker.yaml
```diff
@@ -37,14 +37,6 @@ spec:
   logging:
     level:
       default: error
-  bootstrap:
```
codefromthecrypt, line 40: this is the part latest [func-e](https://github.com/tetratelabs/func-e) (which runs envoy in standalone mode) obviates, as admin server is no longer a chore for the user. If it is there, it is used, otherwise an ephemeral admin server is.

PR #1094, opened 2025-08-18
cli: adds default buffer limit on aigw run

mathetake:
**Description**

This adds a default ClientTrafficPolicy to aigw run.


codefromthecrypt:
Thanks for fixing this, as it causes problems when you use a coding agent with an MCP server like Github's which implies several 10s of k just for the system prompt!

PR #1129, opened 2025-08-27
ci: removes PR title length check

mathetake:
**Description**

This removes the length check in the PR title. This has existed to enforce the single format of commit message but it was too restrict and has been annoying to be honest.


aabchoo (approved):
🔥

Issue #1127, opened 2025-08-27
test coverage flake/diff between linux and mac hosts


mathetake:
https://github.com/envoyproxy/ai-gateway/actions/runs/17260961162/job/48982170881

For some reason, the coverage is not reflecting the right number. I suspect something is wrong with cache but no idea.

mathetake:
maybe we have to add `-count=1`? idk

PR #1131, opened 2025-08-27
test: adds more tool call related test cases

mathetake:
**Description**

This commit adds additional test cases that involves tool calls and responses, which is a basic pattern in agent applications.

### DIFF
--- a/tests/extproc/real_providers_test.go
+++ b/tests/extproc/real_providers_test.go
@@ -210,10 +210,12 @@ func TestWithRealProviders(t *testing.T) {
 		}
 	})
 
-	t.Run("Bedrock uses tool in response", func(t *testing.T) {
+	t.Run("uses tool in response", func(t *testing.T) {
 		client := openai.NewClient(option.WithBaseURL(listenerAddress+"/v1/"), option.WithMaxRetries(0))
 		for _, tc := range []realProvidersTestCase{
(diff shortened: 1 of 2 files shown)


tests/extproc/real_providers_test.go
```diff
@@ -241,7 +243,6 @@ func TestWithRealProviders(t *testing.T) {
 								},
 							},
 						},
-						Seed:  openai.Int(0),
```
mathetake, line 244: it seems like gemini's OpenAI compat endpoint doesn't like the seed

mathetake:
Failure is https://github.com/envoyproxy/ai-gateway/issues/1127 so we can ignore it

tests/extproc/real_providers_test.go
mathetake: i hope this won't cause any flake. let's see.

tests/extproc/real_providers_test.go
```diff
@@ -281,7 +282,7 @@ func TestWithRealProviders(t *testing.T) {
 							t.Logf("Appended tool message: %+v", *toolMessage.OfTool) // Debug log.
 						}
 					}
-					if getWeatherCalled == false {
+					if !getWeatherCalled {
```
mathetake, line 285: 🙄

PR #1135, opened 2025-08-28
Update model name header incase of modelNameOverride for a backend

AyushSawant18588:
**Description**

This change ensures the model name used for observability and cost evaluation reflects the backend-resolved model (if overridden by routing) rather than the client-requested model.
Prefer the backend’s modelNameOverride when setting the metrics model via SetModel(...), so gen_ai.request.model (exported as gen_ai_request_model) reports the actual upstream model.
Sync c.requestHeaders[c.config.modelNameHeaderKey] to modelNameOverride in SetBackend(...) so any header-mapped labels and CEL-based request-cost evaluation use the backend-resolved model.
No network request headers are changed here; the update is to the processor’s internal requestHeaders map used for metrics/metadata and CEL. Translator logic continues to rewrite request bodies as appropriate for the chosen backend.

**Why:**
With traffic splitting and per-backend modelNameOverride, the previous behavior recorded the client-facing model (e.g., gpt-4o-mini) in metrics and CEL instead of the actual model used upstream (e.g., gemma-basic). This led to misleading metrics and misaligned cost calculations. The new behavior aligns metrics and costs with the effective backend model.


mathetake:
meeting summary: this makes sense!

mathetake (approved):
Nice! Thank you for your contribution

PR #1134, opened 2025-08-28
cli: fail and exit if ext-proc fails to start

nacx:
**Description**

When using the `aigw run` command, if the extproc fails to start, the command does not return and hangs forever without giving any feedback to the user unless the `--debug` flag is used.

This PR captures the extproc start errors and terminates the CLI process to give proper feedback to the user, printing the error and exiting with an error code:

```
$ ./out/aigw-darwin-arm64 run
looking up the latest patch for Envoy version 1.35
1.35.0 is already downloaded
starting: /tmp/envoy-gateway/versions/1.35.0/bin/envoy in run directory /tmp/envoy-gateway/runs/1756378634894657000
[2025-08-28 12:57:15.197][3950822][warning][config] [source/server/options_impl_platform_default.cc:9] CPU number provided by HW thread count (instead of cpuset).
2025/08/28 12:57:18 Error running: external processor run error: failed to start config watcher: failed to load initial config: failed to load config: cannot create backend auth handler: cannot retrieve AWS credentials: failed to refresh cached credentials, no EC2 IMDS role found, operation error ec2imds: GetMetadata, exceeded maximum number of attempts, 3, request send failed, Get "http://169.254.169.254/latest/meta-data/iam/security-credentials/": dial tcp 169.254.169.254:80: connect: host is down

$ echo $?
1
```

**Related Issues/PRs (if applicable)**

N/A

**Special notes for reviewers (if applicable)**

N/A

PR #1137, opened 2025-08-28
test: decouple run lifecycle tests from extproc internals

nacx:
**Description**

This improves the `aigw run` lifecycle test to not rely on implementation details of how extproc is executed.
The previous test relied on how extproc uses and binds the UDS, which couples the test to the extproc internals, making it error-prone and less maintainable.

This PR refactors the code so that the run tests can verify the lifecycle without relying on extproc internal details.

Related PR: https://github.com/envoyproxy/ai-gateway/pull/1134

N/A


mathetake (approved):
This makes sense to me

PR #1138, opened 2025-08-28
feat: support removal of sensitive/security headers to backend

siddharth1036:
**Description**
This commit introduces support for header mutation handling in the AIServiceBackend CRD and the extproc filter logic. This introduces a flexible and robust header mutation framework for backend integrations, modeled after the Kubernetes Gateway API. It adds support for structured header mutation ( set, remove) in the AIServiceBackend CRD and refactors the extproc filter logic to use this new approach. Sensitive headers configured by maintainer for backends will be removed before sending request upstream and restored as needed during retries and backend failover scenarios.

**Changes**
- Introduced a headerMutation field in the AIServiceBackend CRD, supporting set, and remove  header operations (Helm chart and YAML updated).
- Added a HeaderMutator utility to encapsulate header mutation logic
- Refactored extproc filter logic in chatcompletion_processor.go to:
-- Remove headers from requests when configured.
-- Restore headers from the original request when appropriate (e.g., during backend fallback).
-- Apply all header mutations in a structured, testable way.

**Why**
- Provides a clear, extensible, and maintainable approach to header mutation, aligned with Kubernetes Gateway API best practices.
- Ensures compliance and robustness for backend integrations in multi-backend and failover scenarios..

#1141 


api/v1alpha1/backendsecurity_policy.go
```diff
@@ -61,6 +61,12 @@ type BackendSecurityPolicySpec struct {
 	// +kubebuilder:validation:Enum=APIKey;AWSCredentials;AzureCredentials;GCPCredentials
 	Type BackendSecurityPolicyType `json:"type"`
 
+	// Sensitive headers will be removed while sending request to upstream.
+	// These headers will be added back to the request if request fails upstream
+	//
+	// +optional
+	SensitiveHeaders []string `json:"sensitiveHeaders,omitempty"`
```
mathetake: how about adding one additional indirection like HeaderMutation struct where we can have add, set and remove exactly like Gateway API's HTTPRoute level header mutation? https://github.com/kubernetes-sigs/gateway-api/blob/506bacdd62d6efada67dbed3e822be051a8eb469/apis/v1/httproute_types.go#L1039-L1105

mathetake: Then i think this should live in AIServiceBackend?

siddharth1036: sure @mathetake, makes sense, let me make the changes

siddharth1036: @mathetake I have made the changes, can you please review once and let me know if it looks good to you?

api/v1alpha1/ai_service_backend.go
```diff
@@ -72,6 +72,80 @@ type AIServiceBackendSpec struct {
...
+	//   my-header2: bar
+	//
+	// +optional
+	// +listType=set
+	// +kubebuilder:validation:MaxItems=16
+	Remove []string `json:"remove,omitempty"`
+}
+
+// HTTPHeader represents an HTTP Header name and value as defined by RFC 7230.
+type HTTPHeader struct {
```
mathetake: let's reuse the GatewayAPI's `gwapiv1.HTTPHeader`

siddharth1036: fixed

api/v1alpha1/ai_service_backend.go
```diff
@@ -72,6 +72,59 @@ type AIServiceBackendSpec struct {
 	// +optional
 	BackendSecurityPolicyRef *gwapiv1.LocalObjectReference `json:"backendSecurityPolicyRef,omitempty"`
 
+	// HeaderMutation defines the mutation of HTTP headers that will be applied to the request
+	// before sending it to the backend.
+	// +optional
+	HeaderMutation *HTTPHeaderMutation `json:"headerMutation,omitempty"`
```
aabchoo, line 78: Thoughts on using the envoy HeaderMutation instead?https://github.com/envoyproxy/go-control-plane/blob/1a04a01c30a3bb987b9a5268a6695fe7aad55255/envoy/service/ext_proc/v3/external_processor.pb.go#L1209-L1223

mathetake: well, it is an implementation detail (skipping the abstraction of GWAPI & EG) & the Envoy's API is not a type for CRD, so this would end up exploding the CRD definition here as well as it's not used in EG

mathetake:
Do you want to add an e2e test case?
* Edit the config here https://github.com/envoyproxy/ai-gateway/blob/main/tests/e2e/testdata/testupstream.yaml
* Add a new test case here https://github.com/envoyproxy/ai-gateway/blob/main/tests/e2e/testupstream_test.go#L33
  * You might need to add additional fields to case struct to verify "non expected headers", etc which uses https://github.com/envoyproxy/ai-gateway/blob/61fbd362d3962e70bf029431baa98c8f837e7f21/tests/internal/testupstreamlib/testupstream.go#L42

siddharth1036:
@mathetake  I have added the e2e test, can you please review the same once?

mathetake:
will do soon

tests/e2e/testupstream_test.go
```diff
@@ -75,6 +79,16 @@ func TestWithTestUpstream(t *testing.T) {
...
 				expResponseBody: `No matching route found. It is likely that the model specified your request is not configured in the Gateway.`,
 			},
+			{
+				name:               "openai-header-mutation",
+				modelName:          "some-cool-model",
+				expTestUpstreamID:  "primary",
+				expPath:            "/v1/chat/completions",
+				expHost:            "testupstream.default.svc.cluster.local",
+				fakeResponseBody:   `{"choices":[{"message":{"content":"This is a test."}}]}`,
+				nonexpectedHeaders: []string{"x-remove-header"},
```
mathetake, line 89: shouldn't this test case set `reqHeaders: []string{"x-remove-header"}` as well ?

siddharth1036: my bad, I think it got misplaced during refactoring, I have added it back.

internal/extproc/headermutator/header_mutator.go
```diff
@@ -0,0 +1,81 @@
...
+
+type HeaderMutator struct {
+	// getOrignalHeaders Callback to get removed sensitive headers from the router filter.
+	getOrignalHeaders func() map[string]string
+
+	// headerMutations is a list of header mutations to apply.
+	headerMutations *filterapi.HTTPHeaderMutation
+}
+
+func NewHeaderMutator(headerMutations *filterapi.HTTPHeaderMutation, getOrignalHeaders func() map[string]string) *HeaderMutator {
```
mathetake: why does this need to accept a closure?

siddharth1036: Actually I am using it get the headers in routerprocessor (the original request headers) to restore the headers on retries to different backends as the headermutator runs in upstream filter and doesn't have the context to router headers. The  closure also helps in decoupling it from the routeProcessor struct. I am not sure about storing a header map copy in this struct therefore added a callback to fetch it from rp.

mathetake: hmm. map is passed by a reference so i think simply passing map doesn't make any difference? passing a map doesn't result in copy in other words in Go. passing rp.requestHeaders directly to this function shouldn't differ from passing a closure like this one. Maybe I am missing something.

siddharth1036: Yeah it makes sense, let me make the change

siddharth1036: I have made the changes

mathetake (approved):
thanks!

PR #1140, opened 2025-08-29
test: fix bad recording of openai embeddings spans

codefromthecrypt:
**Description**

This fixes the test data that was unnoticed as bad even though the tests passed.

What happens is canonically all openai SDKs backfill an embedding `encoding_format`, so our requests didn't add that and this caused a request mismatch. `encoding_format` is backfilled to `base_64` when platform support exists for it, in python for example. This fixes the test fixture which silently succeeded and re-recorded all the data.

#696


tests/internal/testopeninference/spans/embeddings-unknown-model.json
```diff
@@ -1,72 +1,72 @@
...
-            "stringValue": "openai.ConflictError"
+          "key":  "exception.type",
+          "value":  {
+            "stringValue":  "openai.NotFoundError"
           }
         },
         {
-          "key": "exception.message",
-          "value": {
-            "stringValue": "TestOpenAI Error: Interaction out of date for POST /v1/embeddings. To re-record, delete /Users/codefromthecrypt/oss/ai-gateway/tests/internal/testopenai/cassettes/embeddings-unknown-model.yaml and re-run with OPENAI_API_KEY set."
```
codefromthecrypt, line 50: example of a request mismatch caused by python sending a different request (both), which the fixture now catches

Issue #995, opened 2025-08-05
switch to use the official openai go sdk types


yuzisun:
*Description*:
OpenAI now supports an official go SDK and 1.0.0 was released back in May, we can switch to use the types from there instead of maintaining the bespoke decoding logic in envoy ai gateway. 

https://github.com/openai/openai-go/releases?page=3

[optional *Relevant Links*:]
>Any extra documentation required to understand the issue.

codefromthecrypt:
this sounds good, but one caution I would make is that for the official openai upstream this makes more sense than the dozens emulating it. as these types are auto-generated (stainless sdk does that), it might end up adding too much or drifting "canonical openai backends" or our router, over a product decision driven by openai the platform. So, if we do this, proceed with care not to accidentally paint the larger openai ecosystem into a corner.

PR #1147, opened 2025-08-29
refactor: using openaigo SDK for non-stream response

alexagriffith:
**Description**
first part of refactor to use openaigo sdk chatcompletion for non-stream responses. 
includes updates to all code and tests that are affected.
added custom struct to allow for custom fields like obfuscation that are not in the openaigo sdk at the moment.


codefromthecrypt:
Thanks for the PR, @alexagriffith! I read the issue leading to this #995 and having personally dealt with API drift, can see value in doing better than we are now.

However, I want to cite some trade-offs. While it cuts bespoke code, it comes at a cost of control and flexibility:

### Pros from #995
- Offloads upkeep for core features we support (chat completions, embeddings, models).
- Leverages official types for better reliability on official OpenAI (the company) interactions generated with the Stainless SDK.

### Potential Downsides
- **Coupling Risks:** SDK is OpenAI-specific and auto-generated via Stainless; could break compat with dozens of emulators (Ollama, Grok, etc.), adding workarounds/tech debt.
- **Limited Control:** No contrib rights on auto-generated code; reliant on OpenAI's fixes, with evidence of slow responses (e.g., unresolved obfuscation mismatch in [openai/openai-go#484](https://github.com/openai/openai-go/issues/484)) indicating potential staffing gaps.
- **Bloat and Scope:** Pulls unnecessary deps ([go.mod L5-10](https://github.com/openai/openai-go/blob/main/go.mod#L5-L10)) and features beyond our narrow proxy needs, increasing dependency surface and potential for unrelated issues like build bloat.
- **Instability:** History of breaking changes in SDK requires fast-following; extra vigilance with tests (e.g., VCR) to catch drifts in marshalling.
- **Performance:** We have custom code around pointers etc to prevent re-marshalling JSON, and probably we don't have benchmarks to assess if current version of the openai SDK, much less a future one, will cause a potentially large regression (CPU memory burnt on duplicate JSON marshaling of the same data)
- **Drift Management:** Easier now with AI tools (Claude/Goose) for auto-updates via traffic recordings; SDK adds vendor unknowns like future usage call backs or platform assumptions that don't fit proxies.
- **Irreversibility:** Hard to revert due to refactoring; plus, as a company SDK, it may evolve with implicit changes we can't anticipate.

Alternative: Maybe there's a smarter way to handle marshaling in our existing types that eases maintenance without impairing portability or our ability to directly address gaps? Worth exploring more or do you feel this is the only way out, warts and all?

yuzisun:
@codefromthecrypt thanks so much for the detailed feedback and for raising these thoughtful points!
I agree with some of the downsides, but it has been a growing pain to maintain our own types and it is still not fully compatible with openai. I wrote some of the bespoke union type decoding logic which I am not very happy with and openai go sdk has better handling for that. Regarding the scope their sdk deps seem small and our project already included the same dependencies.  Happy to discuss with you more to make a decision weighting these pros and cons.

codefromthecrypt:
Had a good chat with @yuzisun and here's my 2p

For Envoy AI Gateway's multi-provider proxy needs, consider code generation (e.g., via ogen) from a forked/curated OpenAI OpenAPI YAML spec as an alternative to SDK adoption—this fixes official spec issues, enables custom extensions for performance and compatibility, and avoids coupling risks while automating tedious maintenance like union decoding. Compared to official/community SDKs (e.g., openai-go or sashabaranov/go-openai), it offers better flexibility and control; start hybrid if full gen feels big. Just an option to ponder, not a must-do.

If you want a really long answer version, I have this, too, and also what we do tactically vs strategically can be different answers!

codefromthecrypt:
if we continue down the current path, probably worth raising an upstream issue on openai-go to stop serializing empty/zero values because it might break folks who are not expecting these values, and also wastes bandwidth.

mathetake:
Not following the discussion fully, but the conflicts with the main branch seem huge as well as this doesn't looks an active PR, so let me close one (I am sweeping the PR list now). Feel free to reopen if that's necessary!

Issue #1130, opened 2025-08-27
Document how to setup Access Logs including model name, modelNameOverride, token usage, etc


mathetake:
*Description*:

Since we have these information in the dynamic metadata, it would be straightforward to include them in access logs. However, it's not that trivial for those who are new to envoy .

nacx:
I'm working on this and will open a PR in short. @mathetake can you assign this to me?

PR #1148, opened 2025-08-30
docs: document how to include AI metadata in access logs

nacx:
**Description**

Adds a page documenting how to include the AI Gateway metadata (model selection, token usage, etc.) in the Envoy Access logs. It also preconfigures it in the default configuration for the `aigw run` command.

Fixes: https://github.com/envoyproxy/ai-gateway/issues/1130

N/A


cmd/aigw/ai-gateway-default-resources.yaml
```diff
@@ -90,6 +90,49 @@ spec:
...
+              protocol: "%PROTOCOL%"
+              response_code: "%RESPONSE_CODE%"
+              response_flags: "%RESPONSE_FLAGS%"
+              response_code_details: "%RESPONSE_CODE_DETAILS%"
+              connection_termination_details: "%CONNECTION_TERMINATION_DETAILS%"
+              upstream_transport_failure_reason: "%UPSTREAM_TRANSPORT_FAILURE_REASON%"
+              bytes_received: "%BYTES_RECEIVED%"
+              bytes_sent: "%BYTES_SENT%"
+              duration: "%DURATION%"
+              x-envoy-upstream-service-time: "%RESP(X-ENVOY-UPSTREAM-SERVICE-TIME)%"
```
yuzisun, line 124: also good to add the `x-request-tx-duration-ms` this is the preprocessing time before hitting the upstream

yuzisun:
very nice doc! thanks @nacx

cmd/aigw/ai-gateway-default-resources.yaml
nacx: The fields I put here were the default ones (as far as I found), with just the addition of the `genai_*` ones. The main idea is to illustrate how to pull the AIGW info from the metadata, and limit this guide to that.
The fields below the genai are the default ones (I've just added them to preserve the original access logs with the addition of the AI metrics), and I'd rather to keep them as-is, to avoid confusion and have the guide focused only on the AI bits, with no "side changes". WDYT?

yuzisun: I think the preprocessing time is critical at least for us, user wants to know how much time is spent on the ai gateway extproc vs upstream. Our doc should be tailored for the AI traffic needs beyond envoy defaults. cc @johnugeorge @gavrissh

nacx: That's fair, but this is a different issue. We're now talking about EAIGW needing a different set of defaults for access logs, because the current ones lack important info.

I think it's better to keep the scope of this PR to explain how to populate the filter metadata info in the access logs, and discuss separately what the default access logs config should be for EAIGW.

Issue #1121, opened 2025-08-25
docs: update with instructions on how to authenticate Phoenix (for genai traces)


codefromthecrypt:
*Description*:

Right now, our examples and docs suggest running arize phoenix without authentication as it is just a quick start:

* https://aigateway.envoyproxy.io/docs/capabilities/observability/tracing 
  * from https://github.com/envoyproxy/ai-gateway/blob/main/site/docs/capabilities/observability/tracing.md#install-phoenix-for-llm-observability
* https://github.com/envoyproxy/ai-gateway/blob/main/cmd/aigw/docker-compose-otel.yaml#L30

These should be updated with at least a note of how to configure them with normal otel ENV variables as defined here
* https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/

We should verify this, but according to Phoenix docs, it seems like just setting these ENV variables:

possibly having a file somewhere .env.otel.phoenix could help?
```
OTEL_EXPORTER_OTLP_ENDPOINT=http://your_phoenix:6006
OTEL_EXPORTER_OTLP_HEADERS="Authorization=Bearer YOUR_API_KEY"
OTEL_RESOURCE_ATTRIBUTES=project.name=YOUR_PHOENIX_PROJECT_NAME
# Phoenix does not support OTLP metrics or logs
OTEL_METRICS_EXPORTER=none
OTEL_LOGS_EXPORTER=none
```

https://arize.com/docs/phoenix/tracing/how-to-tracing/setup-tracing/custom-spans

cc @mikeldking

codefromthecrypt:
one thing might be helpful are these instructions here which are for elasticstack, but similar to phoenix works with pure otel config. Notably this includes how to do things with k8s https://github.com/elastic/elasticsearch-labs/tree/main/example-apps/chatbot-rag-app#run-with-kubernetes

ion-elgreco:
@codefromthecrypt since we are passing it into extraProcEnvVars it only takes in name=value, so the API key will be visible in the string, is there a workaround for that?

mathetake:
i think enhancing extraProcEnvVars template to be able to support secretRef as well in addition to name=value makes sense. @ion-elgreco do you want to take a shot?

ion-elgreco:
@mathetake sure! I'll see what I can do later this week

codefromthecrypt:
thanks for taking this forward, folks!

codefromthecrypt:
[...]

PR #1150, opened 2025-09-01
feat: secretKeyRef support in extraProcEnvVars

ion-elgreco:
**Description**
Allows references secrets in the ext_proc env vars as a value: `secretKeyRef {"name":"secret_name", "key": "secret_key"}`

Complete example:

`OTEL_EXPORTER_OTLP_HEADERS=secretKeyRef {"name":"secret_name", "key": "secret_key"}`

This should also enable layering secrets with k8s, so you can configure you values.yaml this way for aigw, where secret_key is just the plain secret and then OTEL_EXPORTER_OTLP_HEADERS adds Bearer as prefix, while taking $SECRET_HEADER from the env var:

```yaml
extProc:
  extraEnvVars:
    - name: OTEL_EXPORTER_OTLP_ENDPOINT
      value: http://phoenix-svc.phoenix.svc.cluster.local:6006
    - name: SECRET_HEADER
      value: "secretKeyRef {\"name\":\"secret_name\", \"key\": \"secret_key\"}"
    - name: OTEL_EXPORTER_OTLP_HEADERS
      value: "Bearer $SECRET_HEADER"
```

- closes https://github.com/envoyproxy/ai-gateway/issues/1121

cc @codefromthecrypt, @mathetake

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -29,21 +29,35 @@ jobs:
         with:
           username: ${{ vars.DOCKERHUB_USERNAME }}
           password: ${{ secrets.DOCKERHUB_PASSWORD }}
+
       - name: Push Helm chart
         run: |
           make helm-push HELM_CHART_VERSION=${HELM_CHART_VERSION}
           make helm-push HELM_CHART_VERSION=${HELM_CHART_VERSION_WITHOUT_V}
 
+      - name: Build CLI release binaries
+        run: |
+          make build.aigw GOOS_LIST="linux" GOARCH_LIST="amd64 arm64"
+          # amd64 for macos is not available on archive-envoy for the last few table versions
+          # which in turn is used by EG's standalone mode used by aigw, so we skip it
+          make build.aigw GOOS_LIST="darwin" GOARCH_LIST="arm64"
+
       - name: Create a release candidate
         if: ${{ contains(github.ref, '-rc') }}
         run: |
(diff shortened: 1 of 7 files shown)


codefromthecrypt:
ps thanks for working on this

ion-elgreco:
> Thanks, did you consider alternatives like
> 
> 1. Introduce another flag dedicated for secretRef
> 2. Add support for `valueFrom` at extraEnvVars of **helm level** to match the Kubernetes API
> 3. Pass the secretRefs from the helm into the new dedicated flag for secretRef
> 4. Pass the secretRef **as-is** into the pod template
> 
> That way, we could avoid conflating everything in the comma separate stuff like the current one. Manual parsing like this seems an error-prone and looks clunky to be honest.
> 
> wdyt?

I didn't want to change to much beyond what I did now, since I didn't know how big of a change was acceptable.

Option 2 would definetely solve the error prone issue, because we would know how valueFrom gets converted into a string and decoded again.

The main issue is we have to go through this string structure, not sure if i'ts possible otherwise since I don't know the codebase well enough, but can definetly work another approach

mathetake:
@ion-elgreco let me know if you have any intent to continue this PR or cool with closing this one since it's been a while since last activity

ion-elgreco:
@mathetake yes, I would like to but have to find some time

mathetake:
This will be covered in a cleaner way in #1623, so i will close !

PR #1155, opened 2025-09-02
feat: add bedrock reasoning for non-stream

alexagriffith:
**Description**

support returning reasoning content from aws bedrock in openai.
required adding aws bedrock fields that were missing and requires returning extra fields. since we are already refactoring to use openai go sdk, it makes sense to implement this here instead of adding it to our custom openai struct.


internal/apischema/openai/openai.go
```diff
@@ -1330,3 +1336,12 @@ type AnthropicVendorFields struct {
 	// https://docs.anthropic.com/en/api/messages#body-thinking
 	Thinking *anthropic.ThinkingConfigParamUnion `json:"thinking,omitzero"`
 }
+
+// ResponseVendorFields contains extra vendor-specific fields to add to the openai response.
+type ResponseVendorFields struct {
```
sukumargaonkar: lets rename to be vendor specific. `AnthropicResponseVendorFields`

internal/extproc/translator/openai_awsbedrock.go
```diff
@@ -82,6 +82,14 @@ func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) RequestBody(_ []byte, ope
 		bedrockReq.InferenceConfig.StopSequences = stopSequence
 	}
 
+	// Handle Anthropic vendor fields if present. Currently only supports thinking fields.
+	if openAIReq.AnthropicVendorFields != nil && openAIReq.Thinking != nil {
```
sukumargaonkar, line 86:
```suggestion
	if openAIReq.AnthropicVendorFields != nil && openAIReq.AnthropicVendorFields.Thinking != nil {
```

internal/extproc/translator/openai_awsbedrock.go
```diff
@@ -82,6 +82,14 @@ func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) RequestBody(_ []byte, ope
...
+		if bedrockReq.AdditionalModelRequestFields == nil {
+			bedrockReq.AdditionalModelRequestFields = make(map[string]interface{})
+		}
+		bedrockReq.AdditionalModelRequestFields["thinking"] = openAIReq.Thinking
```
sukumargaonkar, line 90:
```suggestion
		bedrockReq.AdditionalModelRequestFields["thinking"] = openAIReq.AnthropicVendorFields.Thinking
```

alexagriffith: it mmaps to that anyways but can be explicit because could cause issues in future

alexagriffith: actually linter complains - i think its ok as is?

internal/apischema/openai/openai.go
```diff
@@ -1338,3 +1344,12 @@ type AnthropicVendorFields struct {
 	// https://docs.anthropic.com/en/api/messages#body-thinking
 	Thinking *anthropic.ThinkingConfigParamUnion `json:"thinking,omitzero"`
 }
+
+// AnthropicResponseVendorFields contains extra vendor-specific fields to add to the openai response.
+type AnthropicResponseVendorFields struct {
```
yuzisun: @alexagriffith @sukumargaonkar `ReasoningContent` is AWS Bedrock specific not Anthropic right?

yuzisun: Bedrock's reasoningContent is different from anthropic's.
https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_ReasoningContentBlock.html
https://docs.anthropic.com/en/docs/build-with-claude/extended-thinking#how-extended-thinking-works

alexagriffith: woops was a mixup

PR #1156, opened 2025-09-02
cli: allow setting the Envoy version in standalone mode

nacx:
**Description**

Allow configuring the version of Envoy to use when running in standalone mode with `aigw run`:

```
 $ ./out/aigw-darwin-arm64 run --envoy-version 1.35.0
1.35.0 is already downloaded
starting: /tmp/envoy-gateway/versions/1.35.0/bin/envoy in run directory /tmp/envoy-gateway/runs/1756854499474593000

$ ./out/aigw-darwin-arm64 run --envoy-version 1.34.2
downloading https://archive.tetratelabs.io/envoy/download/v1.34.2/envoy-v1.34.2-darwin-arm64.tar.xz
starting: /tmp/envoy-gateway/versions/1.34.2/bin/envoy in run directory /tmp/envoy-gateway/runs/1756854634015259000
```

Depends on: https://github.com/envoyproxy/gateway/pull/6891
Keeping it as draft until this PR is merged.

N/A


go.mod
```diff
@@ -524,3 +524,5 @@ tool (
 	sigs.k8s.io/controller-tools/cmd/controller-gen
 	sigs.k8s.io/kind
 )
+
+replace github.com/envoyproxy/gateway => github.com/nacx/gateway v0.0.0-20250902232433-3185d71f97ff
```
nacx, line 528: This needs to be removed when the EG PR is merged.

go.mod
```diff
@@ -524,3 +524,5 @@ tool (
 	sigs.k8s.io/controller-tools/cmd/controller-gen
 	sigs.k8s.io/kind
 )
+
+replace github.com/envoyproxy/gateway => github.com/nacx/gateway v0.0.0-20250903102036-59ada94d34a9
```
nacx, line 528: This needs to be removed when the EG PR is merged.

go.mod
```diff
@@ -255,7 +254,9 @@ require (
 	sigs.k8s.io/apiserver-network-proxy/konnectivity-client v0.31.2 // indirect
 	sigs.k8s.io/json v0.0.0-20241014173422-cfa47c3a1cc8 // indirect
 	sigs.k8s.io/kubectl-validate v0.0.5-0.20241223122011-eb064d2f92d5 // indirect
-	sigs.k8s.io/mcs-api v0.2.0 // indirect
+	sigs.k8s.io/mcs-api v0.3.0 // indirect
 	sigs.k8s.io/randfill v1.0.0 // indirect
 	sigs.k8s.io/structured-merge-diff/v4 v4.7.0 // indirect
 )
+
+replace github.com/envoyproxy/gateway => github.com/nacx/gateway v0.0.0-20250909074754-a7edf6c24f62
```
nacx: This needs to be removed when the EG PR is merged.

mathetake:
i can't think of the situation where users want to use a specific stable version of Envoy as our impls will anyways be tied with one or two version. What do you all think about simply not exposing it but just internally specifying the stable version we want to run with?

mathetake:
iow i think it's an implementation detail in at least aigw  run?

codefromthecrypt:
I think it would be nice to have the ability to update. `ENVOY_VERSION` makes sense as it is what func-e uses for the same purpose. so we can hard-code the known working and allow override with `ENVOY_VERSION`.

we can use the "last known envoy" approach like func-e to pin the version so that it is the same as what is in the Docker build, but also allow it to be overridden. We would do a test like func-e that proves the last known is up to date.

so, in `GenerateOpenAIConfig` we can default EnvoyVersion to the last known and then update it with ENV like all the rest of the stuff.

mathetake:
sg

nacx:
I've rebased the PR and updated it to use the `ENVOY_VERSION` env var instead of a cli flag. Note that since we already have https://github.com/envoyproxy/gateway/pull/6891, the right envoy version will be picked by default unless it's explicitly set.
```
ENVOY_VERISON=1.35.0 ./out/aigw-darwin-arm64 run
```

cmd/aigw/run.go
```diff
@@ -219,11 +219,13 @@ func run(ctx context.Context, c cmdRun, o runOpts, stdout, stderr io.Writer) err
 			return
 		}
 		debugLogger.Info("Found Envoy admin server", "adminPort", envoyAdmin.Port())
-		pollEnvoyReady(ctx, debugLogger, envoyAdmin, 2*time.Second)
-		c.AdminPort = envoyAdminPort // write back for testing
-		// Print a status message without any timestamp formatting
-		startDuration := time.Since(start).Round(100 * time.Millisecond)
-		_, _ = fmt.Fprintf(stderr, "Envoy AI Gateway listening on http://localhost:%d (admin http://localhost:%d) after %v\n", listenerPort, envoyAdmin.Port(), startDuration)
+		if err = pollEnvoyReady(ctx, debugLogger, envoyAdmin, 2*time.Second); err != nil {
```
nacx, line 222: This is a small side change I spotted, to avoid printing "envoy ready" if users cancel (Ctrl^C) the process while the polling was ongoing.

nacx:
Failing tests are fixed in https://github.com/envoyproxy/ai-gateway/pull/1326

cmd/aigw/run_test.go
```diff
@@ -470,7 +470,7 @@ func TestPollEnvoyReady(t *testing.T) {
 		t.Cleanup(func() { callCount = 0 })
 		envoyAdmin, err := aigw.NewEnvoyAdminClient(ctx, os.Getpid(), adminPort)
 		require.NoError(t, err)
-		pollEnvoyReady(ctx, l, envoyAdmin, 50*time.Millisecond)
+		require.ErrorIs(t, pollEnvoyReady(ctx, l, envoyAdmin, 50*time.Millisecond), context.DeadlineExceeded)
```
codefromthecrypt, line 473: Be careful as if i am reading correctly this is now using require in an t.cleanup which is sadly something the linter wont pick up but iiuc will exit the runtime

codefromthecrypt:
Ps unless we want to persist this to a file, there will be a simpler way in a pending change to func-e.

We might consider not shipping this I mean as there will be another way to set the same thing and also align it by default to the docker baked version.

PR #1160, opened 2025-09-04
extproc: fixed issues in recording request completion metric

AyushSawant18588:
**Description**
This PR fixes and tests the request completion metric recording across chat completions, embeddings, and messages processors. 
- Prevents double recording of successful requests by only recording in SetBackend on error (no success recording in SetBackend)
- Ensures failures are recorded on:
   - translator errors
   - non-2xx upstream responses
- Ensures successes are recorded only when EndOfStream is received for streaming responses
- Prevents double recording by suppressing defer paths when we explicitly record on non-2xx
- Adds unit tests to verify the above behaviors using mock metrics

**Key updates:**

- Chat completions: refine defer logic; record non-2xx failures immediately; success only at EOS
- Embeddings: mirror chat completion behavior (non-2xx immediate failure; EOS success)
- Messages: record failure on translator error; success only at EOS

**Tests:**
internal/extproc/chatcompletion_processor_test.go
internal/extproc/embeddings_processor_test.go
internal/extproc/messages_processor_test.go
These tests assert:
failure on translator error: RequireRequestFailure
failure on non-2xx: RequireRequestFailure
not completed mid-stream: RequireRequestNotCompleted
success at end-of-stream: RequireRequestSuccess


internal/extproc/chatcompletion_processor.go
```diff
@@ -335,6 +345,9 @@ func (c *chatCompletionProcessorUpstreamFilter) ProcessResponseBody(ctx context.
 			}
 			c.span.EndSpanOnError(code, b)
 		}
+		// Record failure for non-2xx status codes and suppress defer.
+		c.metrics.RecordRequestCompletion(ctx, false, c.requestHeaders)
+		recorded = true
```
nacx: What about just doing:
```go
forceErrorMetric = true
```
And simplify the `defer` statement? This way we still keep the metric recording at one single place?

mathetake:
Thank you for your contribution @AyushSawant18588 💯

PR #1162, opened 2025-09-04
api: fix reasoning and response format fields

yuzisun:
**Description**
The defined `reasoning` struct is from response API, chat completion only supports `reasoning_effort`. Add the verbosity field and the missing json_object on `response_format`.


internal/apischema/openai/openai.go
```diff
@@ -673,13 +671,16 @@ type ChatCompletionRequest struct {
 	// Docs: https://platform.openai.com/docs/api-reference/chat/create#chat-create-presence_penalty
 	PresencePenalty *float32 `json:"presence_penalty,omitempty"` //nolint:tagliatelle //follow openai api
 
-	// Reasoning
-	// o-series models only
-	// refs: https://platform.openai.com/docs/api-reference/responses/create#responses-create-reasoning
-	Reasoning *Reasoning `json:"reasoning,omitempty"`
```
gavrissh, line 679: I see this field was present only in OpenAI `v1/responses` and not in `v1/chat/completions`. 
Will it break any translation that was using this?

yuzisun: no translation is using this, it was added by mistake

mathetake (approved):

PR #1170, opened 2025-09-08
extproc: fixed issues in recording token latency metric

AyushSawant18588:
**Description**

outputTokenLatency was incorrect and sometimes missing/duplicated. Currently dividing the tiny gap between the last and second last stream chunk by the total output tokens and assuming usage appears in the final chunk for all providers (not true for Bedrock/Anthropic), which produced misleading values and occasional double-recording.

This change fixes it by:
- Recording outputTokenLatency exactly once at end-of-stream, as an average since the first token: (now − firstTokenTime) / totalOutputTokens.
- Tracking totalOutputTokens as the maximum cumulative value observed across chunks, so it works whether usage appears mid-stream or only once.

This makes the metric work the same for all providers and show the correct latency of token generation.


AyushSawant18588:
> @AyushSawant18588 hey, could you rewrite the PR description so that it focuses on describing what was the issue and how this is fixed? You don't need to explain every single one of change in the PR (i guess you are using the coding agent though it's understandable as to why the description looks like this).
> 
> Especially describe the problem at the very first line of the description. It's so hard to read through the wall of text you put in there to get the idea of the bug supposedly being fixed in there. The PR description will become the commit message as well (see CONTRIBUTING.md) and imagine someone is blaming the line of code changed here and see the current PR description. It would be so hard for them to understand the why of the change vs when you simply describe the problem by your word in the commit message.

Thanks for the feedback. I’ve updated the PR description to lead with the problem and a concise fix summary.

mathetake (approved):
@AyushSawant18588 Thank you for rewriting the PR description. LGTM!

PR #1173, opened 2025-09-08
feat: add bedrock reasoning stream support

alexagriffith:
**Description**
adding reasoning support for aws bedrock streaming

### DIFF
--- a/internal/extproc/translator/anthropic_gcpanthropic.go
+++ b/internal/extproc/translator/anthropic_gcpanthropic.go
@@ -6,6 +6,7 @@
 package translator
 
 import (
+	"bytes"
 	"encoding/json"
 	"fmt"
 	"io"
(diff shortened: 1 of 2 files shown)


internal/apischema/awsbedrock/awsbedrock.go
```diff
@@ -413,8 +404,9 @@ type ConverseStreamEvent struct {
 // ConverseStreamEventContentBlockDelta is defined in the AWS Bedrock API:
 // https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_ContentBlockDelta.html
 type ConverseStreamEventContentBlockDelta struct {
-	Text    *string            `json:"text,omitempty"`
-	ToolUse *ToolUseBlockDelta `json:"toolUse,omitempty"`
+	Text             *string            `json:"text,omitempty"`
+	ToolUse          *ToolUseBlockDelta `json:"toolUse,omitempty"`
+	ReasoningContent *ReasoningContentBlock
```
mathetake: missing json tag?

alexagriffith: ah thank you good catch

internal/apischema/openai/openai_test.go
```diff
@@ -1611,3 +1611,65 @@ func TestChatCompletionResponseChunkChoiceDelta_annotations_round_trip(t *testin
...
+				Value: []ChatCompletionAssistantMessageParamContent{
+					{
+						Type: ChatCompletionAssistantMessageParamContentTypeText,
+						Text: ptr.To("hello from array"),
+					},
+				},
+			},
+		},
+		{
+			// This is the specific case your CI is likely missing.
```
mathetake: let's delete 🤖 comment

alexagriffith: hahaha woops!!

internal/apischema/openai/openai.go
```diff
@@ -419,6 +428,8 @@ type ChatCompletionAssistantMessageParamContent struct {
 	Refusal *string `json:"refusal,omitempty"`
 	// The text content.
 	Text *string `json:"text,omitempty"`
+
+	RedactedContent []byte `json:"redactedContent,omitempty"`
```
mathetake: I am understanding correctly that this is the AWS-only vender specific param right? shouldn't we need the same structure as the `*GCPVertexAIVendorFields` inlined at the top level here?

Or should we somehow define it in the toplevel `AWSVendorFields` on ChatCompletionRequest struct? In other words, I think we haven't decided on what to do with non-top level vendor specific fields? cc @johnugeorge @sukumargaonkar @yuzisun

yuzisun: This field is at the message content level, I think we can allow adding extra fields the same way as top level fields . If you check the openai types they also allow adding extra fields at non-top level.
https://github.com/openai/openai-go/blob/main/chatcompletion.go#L270

mathetake: sg

internal/extproc/translator/openai_awsbedrock.go
```diff
@@ -663,10 +702,7 @@ func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) extractAmazonEventStreamE
 	for {
 		msg, err := dec.Decode(r, nil)
 		if err != nil {
-			// When failed, we stop processing the events.
-			// Copy the unread bytes to the beginning of the buffer.
-			copy(o.bufferedBody, o.bufferedBody[lastRead:])
-			o.bufferedBody = o.bufferedBody[:len(o.bufferedBody)-int(lastRead)]
+			o.bufferedBody = o.bufferedBody[lastRead:]
```
yuzisun, line 703: @alexagriffith is this change necessary ?

yuzisun: it seems ok as the overall effect is to "truncate" the buffer, leaving only the unread data at the beginning of the slice. This is an efficient way to manage a buffer without needing to re-allocate memory and copy the data. @mathetake are you good with this?

PR #1174, opened 2025-09-09
controller: isolate invalid backends in config propagation

### LINKED ISSUE #1122: Handle filter config secret update to extproc gracefully
*Description*:
When reconciliation fails due to individual AIServiceBackend or BackendSecurityPolicy error, filterconfig will not be pushed to extproc which causes healthy routes are not processed and not registered in Envoy config due to current [reconcileFilterConfigSecret](https://github.com/envoyproxy/ai-gateway/blob/e8503c2cca0c39f1c7d78da92c9c1c8468308b7e/internal/controller/gateway.go#L182) implementation sequently loop through and [early error return](https://github.com/envoyproxy/ai-gateway/blob/e8503c2cca0c39f1c7d78da92c9c1c8468308b7e/internal/controller/gateway.go#L222). For instance, GCP backendsecuritypolicy misconfigured or failure should not cause OpenAI route not propagated.

Ideal solution is they should be handled independently.

mathetake:
**Description**

This changes the logic in the config construction in Gateway object controller where we collect all the backends as well as the attached BSP in a way that invalid backend/bsp won't block others to be reconciled.

Fixes #1122

### DIFF
--- a/internal/apischema/openai/openai.go
+++ b/internal/apischema/openai/openai.go
@@ -471,20 +471,40 @@ type ChatCompletionMessageToolCallParam struct {
 
 type ChatCompletionResponseFormatType string
 
+// Constants for the different response formats.
 const (
-	ChatCompletionResponseFormatTypeJSONObject ChatCompletionResponseFormatType = "json_object"
-	ChatCompletionResponseFormatTypeJSONSchema ChatCompletionResponseFormatType = "json_schema"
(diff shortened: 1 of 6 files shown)


mathetake:
cc @xiaolin593

xiaolin593 (approved):
LGTM. Thanks for making the change!

PR #1177, opened 2025-09-10
fix: the response_format type to make it compatible with OpenAI API

hustxiayang:
**Description**

The `json_object` should be { "type": "json_object"}  according to OpenAI's API.
The codes is similar to `ChatCompletionContentPartUserUnionParam`

### DIFF
--- a/internal/apischema/openai/openai.go
+++ b/internal/apischema/openai/openai.go
@@ -479,7 +479,6 @@ const (
 )
 
 // Only one field can be non-empty.
-// TODO: use omitempty or omitzero?
 type ChatCompletionResponseFormatUnion struct {
 	OfText       *ChatCompletionResponseFormatTextParam       `json:",omitempty,inline"`
 	OfJSONSchema *ChatCompletionResponseFormatJSONSchema      `json:",omitempty,inline"`
--- a/internal/apischema/openai/openai.go
+++ b/internal/apischema/openai/openai.go
@@ -471,20 +471,41 @@ type ChatCompletionMessageToolCallParam struct {
 
 type ChatCompletionResponseFormatType string
 
+// Constants for the different response formats.
 const (
-	ChatCompletionResponseFormatTypeJSONObject ChatCompletionResponseFormatType = "json_object"
-	ChatCompletionResponseFormatTypeJSONSchema ChatCompletionResponseFormatType = "json_schema"
 	ChatCompletionResponseFormatTypeText       ChatCompletionResponseFormatType = "text"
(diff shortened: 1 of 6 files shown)


internal/apischema/openai/openai_test.go
```diff
@@ -257,7 +333,7 @@ func TestOpenAIChatCompletionMessageUnmarshal(t *testing.T) {
 		},
 		{
 			name: "response_format",
-			in:   []byte(`{ "model": "azure.gpt-4o", "messages": [ { "role": "user", "content": "Tell me a story" } ], "response_format": { "type": "json_schema", "json_schema": { "name": "math_response", "schema": { "type": "object", "properties": { "step": "test_step" }, "required": [ "steps"], "additionalProperties": false }, "strict": true } } }`),
+			in:   []byte(`{ "model": "azure.gpt-4o", "messages": [ { "role": "user", "content": "Tell me a story" } ], "response_format": { "type": "json_schema", "json_schema": { "name": "math_response", "schema": { "type": "object", "properties": { "step": {"type": "string"} }, "required": [ "steps"], "additionalProperties": false }, "strict": true } } }`),
```
mathetake, line 336: FYI @codefromthecrypt maybe we are missing some edge case recording of OpenAI request, but not sure 100%. maybe

hustxiayang: @mathetake I did not expect someone would catch it as it should not affect the results

mathetake: oh right

internal/apischema/openai/openai.go
```diff
@@ -502,6 +522,71 @@ type ChatCompletionResponseFormatJSONSchema struct {
...
+	var responseFormatPart map[string]any
+	if err := json.Unmarshal(data, &responseFormatPart); err != nil {
+		return err
+	}
+
+	var typeValue string
+	var ok bool
+	if typeValue, ok = responseFormatPart["type"].(string); !ok {
+		return fmt.Errorf("response format does not have type")
+	}
```
mathetake, line 558: so i understand this follows the same pattern in this file. Having said that, I would like to switch to use `gjson.Get()` instead of unmarshaling twice here and the below for specific union type. 

Would you mind in a follow up PR to refactor around it?

mathetake (approved):
Thanks!

internal/apischema/openai/openai.go
hustxiayang: Yeah, we should change for it and `ChatCompletionContentPartUserUnionParam`. Thanks!

PR #1178, opened 2025-09-10
refactor: remove unncessary unmarshaling

hustxiayang:
**Description**
Do not need to unmarshaling twice.

discussed in https://github.com/envoyproxy/ai-gateway/pull/1177

@mathetake `ChatCompletionMessageParamUnion`'s implementation is bit different than `ChatCompletionResponseFormatUnion` and `ChatCompletionContentPartUserUnionParam`. Should we make them consistent?


mathetake:
> @mathetake ChatCompletionMessageParamUnion's implementation is bit different than ChatCompletionResponseFormatUnion and ChatCompletionContentPartUserUnionParam. Should we make them consistent?

yeah we should make it consistent 100%. can you do that in another follow up PR?

mathetake (approved):

internal/apischema/openai/openai.go
```diff
@@ -142,15 +143,15 @@ type ChatCompletionContentPartUserUnionParam struct {
...
-	var chatContentPart map[string]any
-	if err := json.Unmarshal(data, &chatContentPart); err != nil {
-		return err
-	}
-	var contentType string
-	var ok bool
-	if contentType, ok = chatContentPart["type"].(string); !ok {
-		return fmt.Errorf("chat content does not have type")
+
+	typeResult := gjson.Get(string(data), "type")
```
mathetake: Can you use `GetBytes` instead to avoid copying the entire body bytes (which invalidates the point of this refactoring)
```suggestion
	typeResult := gjson.GetBytes(data, "type")
```

hustxiayang: thanks and updated!

mathetake (approved):

hustxiayang:
> yeah we should make it consistent 100%. can you do that in another follow up PR?
Sure, would do that

PR #1188, opened 2025-09-12
controller: do not annotate pods when extproc not present

mathetake:
**Description**

It seems there there's very slight edge/race condition in our configuration propagation logic. Previously, we annotated pods with the uuid regardless of whether or not the pod has the extproc inserted via mutating webhook. However, since the pod annotation happens before the deployment/daemonset annotation, there's a slight race condition between k8s checking the pod-template diff with the actual pod state. As a result, "sometimes", not always, the extproc container insertion won't happen without another manual trigger during the Gateway pod initialization. That's the best guess I have for what I have observed internally.

This fixes the annotation logic so that it will add annotation only to pod or daemonset/deployment so that pod mutation happens 100%.

### DIFF
--- a/internal/controller/gateway.go
+++ b/internal/controller/gateway.go
@@ -398,9 +398,9 @@ func (c *GatewayController) annotateGatewayPods(ctx context.Context,
 	uuid string,
 ) error {
 	allPodsExtProcConfigured := true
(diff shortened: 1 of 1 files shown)


mathetake:
was a red herring

PR #1189, opened 2025-09-12
fix: streamline gzip handling in response processing

sukumargaonkar:
**Description**
Bug: Token Usage not extracted for gzipped http-responses from upstream when endpoint is `v1/messages`

Root Cause: The `messages_processor.go` doensn't decompress response body (if gzipped) before extracting token usage. see [1].

Corresponding code in `chatcompletion_processor.go`(See [2]) which handles gzip decompressions.

This PR introduces utility functions for handling gzipped http-responses and adds them to `chatcompletion_processor.go`, `embeddings_processor.go` and `messages_processor.go`

[1]: https://github.com/envoyproxy/ai-gateway/blob/e767da5559fadf89a01d2168eea262b48a1ab6a4/internal/extproc/messages_processor.go#L254

[2]: https://github.com/envoyproxy/ai-gateway/blob/e767da5559fadf89a01d2168eea262b48a1ab6a4/internal/extproc/chatcompletion_processor.go#L330-L340


internal/extproc/util.go
```diff
@@ -5,6 +5,63 @@
...
+
+	extprocv3 "github.com/envoyproxy/go-control-plane/envoy/service/ext_proc/v3"
+)
+
+// ContentDecodingResult contains the result of content decoding operation.
+type ContentDecodingResult struct {
+	Reader    io.Reader
+	IsEncoded bool
+	Encoding  string
+}
```
mathetake: no reason to export

sukumargaonkar: mybad,
unexported

internal/extproc/util.go
```diff
@@ -5,6 +5,63 @@
...
+	"io"
+
+	extprocv3 "github.com/envoyproxy/go-control-plane/envoy/service/ext_proc/v3"
+)
+
+// contentDecodingResult contains the result of content decoding operation.
+type contentDecodingResult struct {
+	reader    io.Reader
+	isEncoded bool
+	encoding  string
```
mathetake: i don't think we need this variable in main code?

internal/extproc/util.go
```diff
@@ -5,6 +5,63 @@
...
+}
+
+// decodeContentIfNeeded decompresses the response body based on the content-encoding header.
+// Currently supports gzip encoding, but can be extended to support other encodings in the future.
+// Returns a reader for the (potentially decompressed) body and metadata about the encoding.
+func decodeContentIfNeeded(body []byte, contentEncoding string) (*contentDecodingResult, error) {
```
mathetake: i would return non pointer to avoid the allocation

PR #1195, opened 2025-09-15
extproc: fixed issues in recording token usage metric

AyushSawant18588:
**Description**
This PR fixes incorrect token usage metrics for streaming chat completions.
Issue:
- For OpenAI streaming responses, the translator captured the first usage it saw and then stopped parsing further chunks due to bufferingDone. 
- The processor recorded token usage on every chunk. Since most chunks have no usage, totals stayed low.

Fix:
Translator (OpenAI → OpenAI):
- Always read and parse streaming chunks and return the latest usage seen in the current batch.
- Stop relying on bufferingDone to halt parsing after first usage; keep scanning to pick up final cumulative usage.

Processor (chatcompletion_processor.go):
- In streaming mode, treat translator-reported usage as cumulative.
- Update c.costs only when a non-zero usage is returned, and record TokenUsage exactly once at end-of-stream with the final totals.


internal/extproc/chatcompletion_processor.go
```diff
@@ -378,21 +378,32 @@ func (c *chatCompletionProcessorUpstreamFilter) ProcessResponseBody(ctx context.
...
-	c.costs.TotalTokens += tokenUsage.TotalTokens
+	if c.stream {
+		// For streaming, translators report cumulative usage; keep the latest totals.
+		if tokenUsage != (translator.LLMTokenUsage{}) {
+			c.costs = tokenUsage
+		}
+	} else {
+		// Non-streaming: single-shot totals.
+		c.costs = tokenUsage
+	}
```
mathetake, line 391: I think simply removing `+` would be fine?

internal/extproc/translator/openai_openai.go
```diff
@@ -174,9 +173,7 @@ func (o *openAIToOpenAITranslatorV1ChatCompletion) extractUsageFromBufferEvent(s
 				OutputTokens: uint32(usage.CompletionTokens), //nolint:gosec
 				TotalTokens:  uint32(usage.TotalTokens),      //nolint:gosec
 			}
-			o.bufferingDone = true
```
mathetake, line 177: can you remove `bufferingDone` field from the struct? i think it's not used anymore

internal/extproc/chatcompletion_processor.go
AyushSawant18588: I added this check just as a safeguard to prevent wiping costs to 0 on chunks without usage as skipping zero-value structs will preserve the last returned cumulative totals. Last stream chunk having usage might not be true for all providers. What do you say?

internal/extproc/translator/openai_openai.go
AyushSawant18588: done

internal/extproc/chatcompletion_processor.go
mathetake: ah makes sense

mathetake (approved):

PR #1199, opened 2025-09-15
feat(metrics): updates to latest otel spec

codefromthecrypt:
**Description**

This commit updates the metrics implementation to align with the latest OpenTelemetry Semantic Conventions for Generative AI systems (per https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-metrics/).

Key changes include:
- Renaming system-related attributes to provider-based ones (e.g., `gen_ai.system.name` to `gen_ai.provider.name`) and updating backend values accordingly.
- Removing recording of "total" token usage metrics, as the spec only defines "input" and "output" token types to avoid double-counting.
- Refining latency metrics for chat completions: using `time.Duration` for precision, recording time-to-first-token (TTFT) on the first chunk, and calculating time-per-output-token as `(request_duration - TTFT) / (output_tokens - 1)` only at end-of-stream if output_tokens > 1. Notably, the spec wants you to amortize tokens after the first token, not the first chunk of tokens.
- For embeddings, only input tokens are recorded since there are no output tokens.


internal/extproc/chatcompletion_processor.go
```diff
@@ -402,7 +402,7 @@ func (c *chatCompletionProcessorUpstreamFilter) ProcessResponseBody(ctx context.
 	c.costs.TotalTokens += tokenUsage.TotalTokens
 
 	// Update metrics with token usage.
-	c.metrics.RecordTokenUsage(ctx, tokenUsage.InputTokens, tokenUsage.OutputTokens, tokenUsage.TotalTokens, c.requestHeaders)
+	c.metrics.RecordTokenUsage(ctx, tokenUsage.InputTokens, tokenUsage.OutputTokens, c.requestHeaders)
```
codefromthecrypt, line 387: ps folks who are overriding `RecordTokenUsage` to supply own impl can add input + output as they wish. That said there is discussing in otel that this is not the *total* amount of tokens anyway as there are other types like reasoning.

internal/metrics/chat_completion_metrics.go
```diff
@@ -95,19 +97,24 @@ func (c *chatCompletion) RecordTokenLatency(ctx context.Context, tokens uint32,
 	}
 
 	// Record once at end-of-stream using average from first token.
-	if endOfStream && c.totalOutputTokens > 0 {
-		firstTokenTime := c.requestStart.Add(time.Duration(c.timeToFirstToken * float64(time.Second)))
-		c.interTokenLatency = time.Since(firstTokenTime).Seconds() / float64(c.totalOutputTokens)
-		c.metrics.outputTokenLatency.Record(ctx, c.interTokenLatency, metric.WithAttributeSet(attrs))
+	// Per OTEL spec: time_per_output_token = (request_duration - time_to_first_token) / (output_tokens - 1).
```
codefromthecrypt, line 100: this is the main logic change, which agree isn't terribly sensible spec, but it is what the spec says to do.


(by not sensible, I don't really agree with counting time between tokens when it is more important the time between chunks. However, we say we are following the spec, so it is less confusing to do that vs not)

mathetake: yeah this feels not that sensible. I mean i am getting how `-1` makes this more spec compliant vs the tokens arrive in chunk so we cannot guarantee that `c.firstTokenSent = true` happens when only one token is generated but more than a few tokens could have been generated by that time. On the other hand, token usage info (at least for OpenAI) is not available at that point, so we have no means to know exactly how many tokens got generated on the first callback to this.. 

I think in anyways in practice this won't be an issue so I think we are good with `-1` yeah

internal/metrics/base_metrics.go
```diff
@@ -47,13 +47,13 @@ func (b *baseMetrics) SetModel(model string) {
 }
 
 // SetBackend sets the name of the backend to be reported in the metrics according to:
-// https://opentelemetry.io/docs/specs/semconv/attributes-registry/gen-ai/#gen-ai-system
+// https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-metrics/
 func (b *baseMetrics) SetBackend(backend *filterapi.Backend) {
 	switch backend.Schema.Name {
```
mathetake, line 52: now i am wondering why we only have this special casing for openai and awsbedrock lol

internal/metrics/chat_completion_metrics.go
codefromthecrypt: Right, I think what this does is shore things up to the spec, then if we want to deviate (like someone actually uses it and says.. hey wait..) we can document what alt behavior we do possibly using a different metric name. For now, it is simpler being <1.0 to just let that play out.

tests/extproc/testupstream_test.go
```diff
@@ -921,7 +921,7 @@ data: {"type": "message_stop"}
 		body, err := io.ReadAll(resp.Body)
 		require.NoError(t, err)
 		require.Containsf(t, string(body),
-			`gen_ai_server_time_per_output_token_seconds_bucket{gen_ai_operation_name="chat",gen_ai_request_model="something",gen_ai_system_name="aws.bedrock",otel_scope_name="envoyproxy/ai-gateway"`,
+			`gen_ai_server_time_per_output_token_seconds_bucket{gen_ai_operation_name="chat",gen_ai_provider_name="aws.bedrock",gen_ai_request_model="something",otel_scope_name="envoyproxy/ai-gateway"`,
```
mathetake, line 924: this is a breaking change btw so we need to document that in a release note later

codefromthecrypt:
yep the two release note affecting things are:
1. spec change `gen_ai.system.name` -> `gen_ai.provider.name` (would impact any dashboards)
2. removal of the `total` metric (there are problems of double-counting and the fact there are [many other token usage types](https://ai.pydantic.dev/api/usage/#pydantic_ai.usage.RunUsage.output_tokens) resulting in total being not just in the spec, but not correct either)

PR #1200, opened 2025-09-15
extproc: added gen_ai.response.model label in metrics

AyushSawant18588:
**Description**
This change adds gen_ai.response.model (https://opentelemetry.io/docs/specs/semconv/registry/attributes/gen-ai/) label to metrics so we can distinguish between the model requested by the client and the model that actually generated the response (e.g., after a backend override). Previously, the single gen_ai.request.model label could reflect an overridden value, making it hard to attribute usage accurately to the original request vs. the final routed model.

This PR:
- Adds a new metric label gen_ai.response.model.
- Keeps gen_ai.request.model as the original model from the client request body.
- Sets gen_ai.response.model to the effective model that generated the response (backend ModelNameOverride when present; otherwise the original request model).
- Updates extproc processors so, on ProcessRequestHeaders, the request label comes from the parsed body and the response label from the effective header.
- Refactors base metrics plumbing:
   - Introduces requestModel and responseModel in base metrics, and emits both labels on all relevant histograms.
   - Changes SetModel signature to SetModel(requestModel, responseModel).


mathetake:
so we discussed this before in the community meeting and the consensus was that 
* We should add metrics with the final model name since if you think about who primarily looks at the metrics, it's platform onwners/engineers and they don't care about the actual user input to get the insights of the perf etc & We do want to be careful about the cardinality
* The original model name is available with x-ai-eg-model header so if you really want the metrics related to the original model name, then you can aggregate the access logs with the header name


so I think we won't merge this unless we miss something else

AyushSawant18588:
> so we discussed this before in the community meeting and the consensus was that
> 
> * We should add metrics with the final model name since if you think about who primarily looks at the metrics, it's platform onwners/engineers and they don't care about the actual user input to get the insights of the perf etc & We do want to be careful about the cardinality
> * The original model name is available with x-ai-eg-model header so if you really want the metrics related to the original model name, then you can aggregate the access logs with the header name
> 
> so I think we won't merge this unless we miss something else

Primary motivation for this PR is to correctly use the [gen_ai.request.model](https://opentelemetry.io/docs/specs/semconv/registry/attributes/gen-ai/)
 and [gen_ai.response.model](https://opentelemetry.io/docs/specs/semconv/registry/attributes/gen-ai/)
 labels as defined by the OpenTelemetry conventions.

Current issue: at present gen_ai.request.model ends up showing the final model name (when modelNameOverride is set), not the original model the request was sent to. This was introduced in [envoyproxy/ai-gateway#1135](https://github.com/envoyproxy/ai-gateway/pull/1135).

What OTel specifies:
- gen_ai.request.model should represent the name of the GenAI model the request is made to.
- gen_ai.response.model should represent the name of the GenAI model that generated the response.

Corrected behaviour with this PR:
- When modelNameOverride is not set, both labels will show the same model name.
- When modelNameOverride is set, gen_ai.request.model will show the original model name and gen_ai.response.model will show the final model that produced the response.

In scenarios where an AI Gateway route is configured with provider fallback or traffic splitting, metrics at the overall setup level become clearer. gen_ai.request.model will consistently show the original model name, giving platform owners/engineers a more accurate view of request distribution.

Also in modelNameOverride cases, the x-ai-eg-model header no longer holds the original model name it is updated to the final model name in SetBackend. So the labels in this PR are a straightforward way to see the original request model in metrics.

So this was the intent behind the PR. What do you think about this approach?

mathetake:
> Also in modelNameOverride cases, the x-ai-eg-model header no longer holds the original model name it is updated to the final model name in SetBackend. So the labels in this PR are a straightforward way to see the original request model in metrics.

i don't think so. where we override the header?

mathetake:
but I think i misunderstood the intent of this PR; and I see these two exist in the spec, and it makes sense to me.

Let me review again later

AyushSawant18588:
> > Also in modelNameOverride cases, the x-ai-eg-model header no longer holds the original model name it is updated to the final model name in SetBackend. So the labels in this PR are a straightforward way to see the original request model in metrics.
> 
> i don't think so. where we override the header?

here: https://github.com/envoyproxy/ai-gateway/blob/2d52d718af36b7607db3fc2c636cd76483e64546/internal/extproc/chatcompletion_processor.go#L450

mathetake:
That's only in extproc. Envoy's access log is different

mathetake (approved):
Thanks!

PR #1204, opened 2025-09-16
feat(metrics): support otel and console metric export

### LINKED ISSUE #1100: otel native metrics export
*Description*:

Right now, time to first token and inter-token latency are recorded with otel libraries, but only available by prometheus scraping. This experience is not ideal for sites who use push model backends. They need to setup an otel pump just to scrape prometheus metrics and push them into their backend. Since they were recorded with otel in the first place, this experience gap should be easy to fill without limiting the ability to also consume via prom.

Let's make sure natural otel ENV like we use in tracing also can be applied to metrics, and without a custom binary. Key use case is otel native backends like Elastic Stack which can accept both traces and metrics in one config.

One point to remember is not all otel backends accept otel metrics. For example, phoenix only accepts traces and derives metrics from them. This is a normal concern addressed in standard otel ENV (e.g. overriding the metrics endpoint to a different host, or just disabling them), but we may want to document this.

codefromthecrypt:
**Description**

This extends otel ENV config to metrics, when enabled exports instead of a prometheus listener. This allows metrics in otel native systems like elastic stack, otel tui and otherwise, without a prometheus pump. This also allows you to do ad-hoc configuration by using `console` config.

Here's an example of standalone using otel-tui https://github.com/ymtdzzz/otel-tui:
tui-metrics
tui-trace

Updated cmd/aigw examples:
- Add otel-tui service to show native otel w/o prometheus
- Use profile-based env file selection (.env.otel.${COMPOSE_PROFILES:-console})
- Set console as default for immediate debugging without external dependencies

Documentation:
- Document separate OTLP endpoints for traces/metrics and exporter types

Phoenix users: Set OTEL_METRICS_EXPORTER=none as Phoenix only supports traces

Fixes #1100

### DIFF
--- a/internal/metrics/metrics_test.go
+++ b/internal/metrics/metrics_test.go
@@ -6,228 +6,214 @@
 package metrics
 
 import (
+	"bytes"
 	"context"
+	"io"
+	"net/http"
+	"net/http/httptest"
 	"testing"
+	"testing/synctest"
+	"time"
 
-	promregistry "github.com/prometheus/client_golang/prometheus"
 	"github.com/stretchr/testify/require"
 	"go.opentelemetry.io/otel/metric"
-	"go.opentelemetry.io/otel/metric/noop"
 	sdkmetric "go.opentelemetry.io/otel/sdk/metric"
+	"go.opentelemetry.io/otel/sdk/metric/metricdata"
 )
 
-// TestNewMetricsFromEnv_DefaultServiceName tests that the service name.
-// defaults to "ai-gateway" when OTEL_SERVICE_NAME is not set.
-func TestNewMetricsFromEnv_DefaultServiceName(t *testing.T) {
-	tests := []struct {
-		name              string
-		env               map[string]string
-		expectServiceName string
-		expectNoop        bool
-	}{
-		{
-			name: "default service name when OTEL_SERVICE_NAME not set",
-			env: map[string]string{
-				"OTEL_METRICS_EXPORTER": "console",
-			},
-			expectServiceName: "ai-gateway",
-		},
-		{
-			name: "OTEL_SERVICE_NAME overrides default",
-			env: map[string]string{
-				"OTEL_METRICS_EXPORTER": "console",
-				"OTEL_SERVICE_NAME":     "custom-service",
-			},
-			expectServiceName: "custom-service",
-		},
-		{
-			name:              "prometheus default when no metrics exporter set",
-			env:               map[string]string{},
-			expectServiceName: "ai-gateway",
-		},
-		{
-			name: "disabled when OTEL_SDK_DISABLED is true",
-			env: map[string]string{
-				"OTEL_SDK_DISABLED":     "true",
-				"OTEL_METRICS_EXPORTER": "console",
-			},
-			expectNoop: true,
-		},
-		{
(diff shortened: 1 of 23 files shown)


codefromthecrypt:
Tested with ElasticStack here, so I think we're good https://github.com/elastic/observability-examples/pull/86

codefromthecrypt:
triple checked the README everything is smooth

codefromthecrypt:
`make test-extproc` is flakey.. trying to figure out why (sometimes passes)

codefromthecrypt:
taking to draft because I think it isn't exposing prom by default as it was before. even if things work I suspect we want to default to that and make sure everything works exactly as before unless you set an override

codefromthecrypt:
hopefully last change fixes the flake where other tests making a request were required to have the one that flaked pass (race condition)

codefromthecrypt:
Something else is flakey not sure if I will be able to solve in this PR

mathetake (approved):
Reviewed the code and seems straightforward vs unlocking a new use case which is exciting!

mathetake:
```
2025/09/16 07:22:47 httptest.Server blocked in Close after 5 seconds, waiting for connections:
  *net.TCPConn 0xc000610650 127.0.0.1:40452 in state active
panic: test timed out after 10m0s
	running tests:
		TestOtelOpenAIChatCompletions_propagation (9m37s)
```

this seems flaky @codefromthecrypt

codefromthecrypt:
@mathetake refactored the prom metrics test to not be flakey `TestOtelOpenAIChatCompletions_propagation` I'm racing locally and if can find something will do another PR

codefromthecrypt:
ok 🤞

internal/metrics/metrics_test.go
```diff
@@ -0,0 +1,393 @@
...
+					output := stdout.String()
+					require.Empty(t, output, "no console output expected")
+				}
+			})
+		})
+	}
+}
+
+// TestNewMetricsFromEnv_NetworkExporters tests OTLP and other network-based exporters.
+// We CANNOT use synctest here because it creates a "bubble" where goroutines are isolated
```
codefromthecrypt, line 165: ps I intentionally left this in as it is something I will discuss in the golang meetup tomorrow ;)

tests/extproc/vcr/prometheus_metrics_test.go
```diff
@@ -0,0 +1,117 @@
...
+	"github.com/prometheus/common/expfmt"
+	"github.com/stretchr/testify/require"
+
+	"github.com/envoyproxy/ai-gateway/tests/internal/testopenai"
+)
+
+// TestPrometheusMetrics verifies that metrics are properly exported via Prometheus
+// when processing a chat completion request. This test uses the default configuration
+// which exposes metrics on the /metrics endpoint.
+func TestPrometheusMetrics(t *testing.T) {
```
codefromthecrypt, line 25: @hustxiayang fyi once this lands, feel free to refactor it or rename it to TestPromOpenAIChatCompletions so you can have a second TestPromOpenAIEmbeddings that verifies the token things aren't there (or remain not there). another way is to make this a table test, but the parameters may be difficult enough to warrant a separate test for embeddings. not sure.. anyway hope this base test helps!

mathetake (approved):

PR #1206, opened 2025-09-16
init

hustxiayang:
**Description**

It does not make sense to include `firstTokenLatency` and `outputTokenLatency` metrics for embedding models. Actually, embedding models are not generative models, thus need to separate them.

### DIFF
--- a/internal/metrics/base_metrics.go
+++ b/internal/metrics/base_metrics.go
@@ -17,7 +17,7 @@ import (
 
 // baseMetrics provides shared functionality for AI Gateway metrics implementations.
 type baseMetrics struct {
-	metrics                   *genAI
+	metrics                   *apiMetrics
 	operation                 string
 	requestStart              time.Time
 	model                     string
@@ -26,9 +26,9 @@ type baseMetrics struct {
 }
 
 // newBaseMetrics creates a new baseMetrics instance with the specified operation.
-func newBaseMetrics(meter metric.Meter, operation string, requestHeaderLabelMapping map[string]string) baseMetrics {
+func newBaseMetrics(meter metric.Meter, operation string, requestHeaderLabelMapping map[string]string, isGen bool) baseMetrics {
 	return baseMetrics{
-		metrics:                   newGenAI(meter),
+		metrics:                   newAPIMetrics(meter, isGen),
 		operation:                 operation,
 		model:                     "unknown",
 		backend:                   "unknown",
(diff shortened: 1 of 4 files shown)


codefromthecrypt:
I think this PR is no longer necessary as even if the metrics are there it is only used in chatCompletion not embeddings. double check me if I'm wrong. Regardless we can test this better soon.

codefromthecrypt:
after #1204 we have an easy way to ad-hoc test for this (though without it we can still hit the /metrics endpoint and look) if you run the CLI (standalone mode), you can do `OTEL_METRICS_EXPORTER=console OTEL_METRIC_EXPORT_INTERVAL=100 aigw ...` then make whatever the requests you want. the metrics will show up on the console after 100ms.

hustxiayang:
@codecov-commenter When we deploy the embedding models, we got metrics like `"prompt_tokens":7,"total_tokens":7,"completion_tokens":0,"prompt_tokens_details":null` and our users complain that the metrics are not correct as it does not make any sense to see `completion_tokens:0` as embedding models are not even generative. That' the motivation of this PR.

codefromthecrypt:
@hustxiayang so I think the main way to get on the same page is to discuss things in terms of what code is in main right now, not what 0.3 did.

As far as I can tell, there is no "first token" latency metrics for embeddings in main any more, which is a part of this PR. If there were tests added to this PR you would be able to verify what I'm saying as true or false.

Also, "total" is removed now, so I think that is also something that the above discussion is out of date with.

Separately, there seems to be a concern if tokens should be a metric for the embeddings endpoint. Perhaps the naming should change but token is a billing and rate-limiting metric regardless of whether we are talking LLM or embeddings. For example, if we remove tokens completely from the embeddings endpoint we will remove our ability to know or control based on things folks are billed on https://docs.langdock.com/api-endpoints/embedding/openai-embedding

I would suggest that first rebase this on main, then use tests as a way to prove the before and after behavior of what this PR aims to do. Without it, we get stuck in interpretations which may no longer be valid.

Make sense?

codefromthecrypt:
meanwhile I will refactor the tests so that embeddings metrics are tested. it is possible we aren't settng the operation name correctly per https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-metrics/#metric-gen_aiclienttokenusage

so for chat completions gen_ai.operation.name=chat, and embeddings, gen_ai.operation.name=embeddings

If this isn't set correctly then the metrics would be conflated and nonsense, so if that's the case it is very much a bug!

codefromthecrypt:
https://github.com/envoyproxy/ai-gateway/pull/1219 verifies the old code indeed split metrics on op name.

mathetake:
Let's reopen if this needs to be revived from the conflicts!

PR #1216, opened 2025-09-18
fix: support reason content string for qwen models

### LINKED ISSUE #1214: Unmarshal Error when response fields that have conflicting name with one set in VendorFields
*Description*:
Noticed unmarshal error when a provider returns `reasoningContent` field in the response, but not of type `AWSBedrockReasoningContent` 

https://github.com/envoyproxy/ai-gateway/blob/main/internal/apischema/openai/openai.go#L1525

*Environment*:
latest main

*Logs*:
`"response_code":500,"response_code_details":"ext_proc_error_gRPC_error_2{error_processing_request_message:_cannot_process_response_body:_failed_to_transform_response:_failed_to_unmarshal_body:_json:_cannot_unmarshal_string_into_Go_struct_field_ChatCompletionResponseChoiceMessage.choices.message.AWSBedrockResponseVendorFields.reasoning_content_of_type_openai.AWSBedrockReasoningContent{via_upstream}}"`

Expected Behaviour:
Do not throw an error if VendorField names collide

Proposal:
Use `any` type for these fields and typecast it in the translation layer. This way, it will avoid a collision with the vendor fields

yuzisun:
**Description**
Support qwen reasoning content with string type in the chat completion response. 
https://qwen.readthedocs.io/en/latest/deployment/vllm.html#thinking-non-thinking-modes

This is an example reasoning response
```json
{
  "id": "chatcmpl-e371ae95-b978-418d-ac5c-06c20b6d4b50",
  "object": "chat.completion",
  "created": 1758219085,
  "model": "dsp.qwen3-30b-a3b",
  "choices": [
    {
      "index": 0,
      "message": {
        "role": "assistant",
        "content": "\n\nThe capital of France is **Paris**. It is a major city in northern France, known for its iconic landmarks such as the Eiffel Tower, the Louvre Museum, and Notre-Dame Cathedral. Paris serves as the political, economic, and cultural center of the country.",
        "refusal": null,
        "annotations": null,
        "audio": null,
        "function_call": null,
        "tool_calls": [],
        "reasoning_content": "\nOkay, the user is asking for the capital of France. Let me think. I know that France is a country in Europe. The capital is the city where the government is located. From what I remember, Paris is the capital. But wait, I should double-check to make sure I'm not confusing it with another city. Sometimes people mix up capitals with other major cities. For example, I know that London is the capital of the UK, and Berlin is the capital of Germany. So France's capital should be Paris. Also, I recall that Paris is famous for landmarks like the Eiffel Tower and the Louvre Museum. That makes sense. I don't think there's any other city that's the capital. Maybe someone might think it's Marseille or Lyon, but those are just major cities. Yeah, I'm pretty confident the answer is Paris. Let me just confirm with a quick check. Yes, according to all sources, Paris is the capital of France. So the answer should be Paris.\n"
      },
      "logprobs": null,
      "finish_reason": "stop",
      "stop_reason": null
    }
  ],
  "service_tier": null,
  "system_fingerprint": null,
  "usage": {
    "prompt_tokens": 14,
    "total_tokens": 280,
    "completion_tokens": 266,
    "prompt_tokens_details": null
  },
  "prompt_logprobs": null,
  "kv_transfer_params": null
}
```

fixes #1214

### DIFF
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -45,6 +45,8 @@ type flags struct {
 	metricsRequestHeaderLabels string
 	rootPrefix                 string
 	extProcExtraEnvVars        string
+	// extProcMaxRecvMsgSize is the maximum message size in bytes that the gRPC server can receive.
+	extProcMaxRecvMsgSize int
 }
 
 // parsePullPolicy parses string into a k8s PullPolicy.
@@ -127,6 +129,11 @@ func parseAndValidateFlags(args []string) (flags, error) {
 		"",
 		"Semicolon-separated key=value pairs for extra environment variables in extProc container. Format: OTEL_SERVICE_NAME=ai-gateway;OTEL_TRACES_EXPORTER=otlp",
 	)
+	extProcMaxRecvMsgSize := fs.Int(
+		"extProcMaxRecvMsgSize",
+		512*1024*1024,
+		"Maximum message size in bytes that the gRPC server can receive for extProc. Default is 512MB.",
+	)
 
 	if err := fs.Parse(args); err != nil {
 		err = fmt.Errorf("failed to parse flags: %w", err)
@@ -180,6 +187,7 @@ func parseAndValidateFlags(args []string) (flags, error) {
 		metricsRequestHeaderLabels: *metricsRequestHeaderLabels,
 		rootPrefix:                 *rootPrefix,
 		extProcExtraEnvVars:        *extProcExtraEnvVars,
+		extProcMaxRecvMsgSize:      *extProcMaxRecvMsgSize,
 	}, nil
 }
 
@@ -255,6 +263,7 @@ func main() {
 		MetricsRequestHeaderLabels: flags.metricsRequestHeaderLabels,
 		RootPrefix:                 flags.rootPrefix,
 		ExtProcExtraEnvVars:        flags.extProcExtraEnvVars,
+		ExtProcMaxRecvMsgSize:      flags.extProcMaxRecvMsgSize,
 	}); err != nil {
 		setupLog.Error(err, "failed to start controller")
 	}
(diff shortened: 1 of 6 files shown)


internal/apischema/openai/openai.go
```diff
@@ -1516,13 +1516,39 @@ type AnthropicVendorFields struct {
...
-	// See https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_ReasoningContentBlock.html for more information.
-	ReasoningContent *AWSBedrockReasoningContent `json:"reasoning_content,omitzero"`
+// ReasoningContentUnion content regarding the reasoning that is carried out by the model.
+// Reasoning refers to a Chain of Thought (CoT) that the model generates to enhance the accuracy of its final response.
+type ReasoningContentUnion struct {
+	Value any
+}
+
+func (r *ReasoningContentUnion) UnmarshalJSON(data []byte) error {
+	// for qwen model it returns the reason content as a string
```
aabchoo:
```suggestion
	// For qwen model it returns the reason content as a string.
```

internal/apischema/openai/openai.go
```diff
@@ -1516,13 +1516,39 @@ type AnthropicVendorFields struct {
...
+	var str string
+	err := json.Unmarshal(data, &str)
+	if err == nil {
+		r.Value = str
+		return nil
+	}
+	// See https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_ReasoningContentBlock.html for more information
```
aabchoo:
```suggestion
	// See https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_ReasoningContentBlock.html for more information.
```

aabchoo (approved):
looks good overall -- thanks for adding this

mathetake:
non blocking comment but could be done in followup or ignore: 

So does this mean that the response format will differ depending on the backend the request actually reached? I think we could do better like in openai->openai passthrough translator to be consistent about the format of reasoning content?

yuzisun:
> non blocking comment but could be done in followup or ignore:
> 
> So does this mean that the response format will differ depending on the backend the request actually reached? I think we could do better like in openai->openai passthrough translator to be consistent about the format of reasoning content?

essentially this PR is to handle the openai -> openai case which is currently failing when decoding the response mostly for extracting the usage because the vllm returns the reasoning content as simple string type while aws defines a more complex type. We could do better without decoding the entire ChatCompletionResponse.

PR #1219, opened 2025-09-19
feat(metrics): use authoritative model from response and test with otel

### LINKED ISSUE #1217: flaky test: TestOtelMetrics_ChatCompletions/chat-streaming-detailed-usage
https://github.com/envoyproxy/ai-gateway/actions/runs/17839411487/job/50724755727
```
2025-09-18T19:37:03.8932560Z --- FAIL: TestOtelMetrics_ChatCompletions (21.55s)
2025-09-18T19:37:03.9045850Z     test_environment.go:106: Starting test environment with ports: upstream=49303, extproc=49304, envoyListener=49307, envoyAdmin=49308
2025-09-18T19:37:03.9137100Z     test_environment.go:147: Checking connections to all services in the test environment
2025-09-18T19:37:03.9241040Z     test_environment.go:190: Successfully connected to extProcMetrics on port 49305
2025-09-18T19:37:03.9343730Z     test_environment.go:190: Successfully connected to envoyAdmin on port 49308
2025-09-18T19:37:03.9445560Z     test_environment.go:190: Successfully connected to extProc on port 49304
2025-09-18T19:37:03.9548550Z     test_environment.go:190: Successfully connected to upstream on port 49303
2025-09-18T19:37:03.9652110Z     test_environment.go:190: Successfully connected to envoyListener on port 49307
2025-09-18T19:37:03.9752710Z     test_environment.go:153: All services are up and running
2025-09-18T19:37:03.9855730Z     --- FAIL: TestOtelMetrics_ChatCompletions/chat-streaming-detailed-usage (1.14s)
2025-09-18T19:37:03.9955400Z         otel_metrics_test.go:133: 
2025-09-18T19:37:04.0059060Z             	Error Trace:	/Users/runner/work/ai-gateway/ai-gateway/tests/extproc/vcr/otel_metrics_test.go:133
2025-09-18T19:37:04.0160150Z             	            				/Users/runner/work/ai-gateway/ai-gateway/tests/extproc/vcr/otel_metrics_test.go:104
2025-09-18T19:37:04.0261840Z             	Error:      	"0" is not greater than "0"
2025-09-18T19:37:04.0365130Z             	Test:       	TestOtelMetrics_ChatCompletions/chat-streaming-detailed-usage
```

### LINKED ISSUE #1218: metrics: response.model label should use the model name in the response body if available
*Description*:

When the model name is available in the response body, e.g. OpenAI (vs AWS doesn't return it), we should consider it as the most concrete/ authoritative model name to be used as a `response.model` label. If not available, it should be safe to assume that the model name giving to the upstream is the one that actually generated the response, so in that case what the initial PR #1200 has done should be OK.

The change should be like the following; Change the translator's response handling method to return the model name (optional), and then put it as a final response.model label if it is not empty string

```diff
diff --git a/internal/extproc/chatcompletion_processor.go b/internal/extproc/chatcompletion_processor.go
index dc3be544..3f0ac0cd 100644
--- a/internal/extproc/chatcompletion_processor.go
+++ b/internal/extproc/chatcompletion_processor.go
@@ -366,10 +366,14 @@ func (c *chatCompletionProcessorUpstreamFilter) ProcessResponseBody(ctx context.
                }, nil
        }
 
-       headerMutation, bodyMutation, tokenUsage, err := c.translator.ResponseBody(c.responseHeaders, decodingResult.reader, body.EndOfStream, c.span)
+       headerMutation, bodyMutation, tokenUsage, respModelName, err := c.translator.ResponseBody(c.responseHeaders, decodingResult.reader, body.EndOfStream, c.span)
        if err != nil {
                return nil, fmt.Errorf("failed to transform response: %w", err)
        }
+       if respModelName != "" {
+               // Update the response model if the translator extracted a model name from the response body.
+               c.metrics.SetModel("", respModelName)
+       }
 
        // Remove content-encoding header if original body encoded but was mutated in the processor.
        headerMutation = removeContentEncodingIfNeeded(headerMutation, bodyMutation, decodingResult.isEncoded)
diff --git a/internal/extproc/translator/translator.go b/internal/extproc/translator/translator.go
index ac8c8ca7..d053846a 100644
--- a/internal/extproc/translator/translator.go
+++ b/internal/extproc/translator/translator.go
@@ -56,10 +56,13 @@ type OpenAIChatCompletionTranslator interface {
        //      - `body` is the response body either chunk or the entire body, depending on the context.
        //      - This returns `headerMutation` and `bodyMutation` that can be nil to indicate no mutation.
        //  - This returns `tokenUsage` that is extracted from the body and will be used to do token rate limiting.
+       //      - `respModelName` is the model name extracted from the response body if available. For example, AWS Bedrock doesn't
+       //        return the model name in the response headers vs OpenAI does.
        ResponseBody(respHeaders map[string]string, body io.Reader, endOfStream bool, span tracing.ChatCompletionSpan) (
                headerMutation *extprocv3.HeaderMutation,
                bodyMutation *extprocv3.BodyMutation,
                tokenUsage LLMTokenUsage,
+               modelName string,
                err error,
        )
 
diff --git a/internal/metrics/base_metrics.go b/internal/metrics/base_metrics.go
index afc153c8..9d70f135 100644
--- a/internal/metrics/base_metrics.go
+++ b/internal/metrics/base_metrics.go
@@ -47,8 +47,12 @@ func (b *baseMetrics) StartRequest(_ map[string]string) {
 
 // SetModel sets the model for the request.
 func (b *baseMetrics) SetModel(requestModel, responseModel string) {
-       b.requestModel = requestModel
-       b.responseModel = responseModel
+       if responseModel != "" {
+               b.requestModel = requestModel
+       }
+       if requestModel != "" {
+               b.responseModel = responseModel
+       }
 }
```

codefromthecrypt:
**Description**

This changes our code so that authoritative request and response models are used consistently between metrics and tracing. Specifically, the real response model is used, and is documented how it can be different from the request model and the original received by the router.

- **Original Model** = The model name extracted from the incoming request body. e.g. "gpt-5"
- **Request Model** (sent to backend) = Either the Original Model OR the overridden model if configured. e.g. "gpt-5-nano"
- **Response Model** (from backend) = The model the backend used which might be more specific than the input. e.g. "gpt-5-nano-2025-08-07"

At a high level, this ensures the Request and Response Models are coherently used for metrics:
- Decouples `SetModel` into separate `SetRequestModel` (request/user-provided) and `SetResponseModel` (actual/response-provided), improving accuracy for virtualized/aliased models (e.g., OpenAI's `gpt-4o` -> `gpt-4o-2024-11-20`).
- Updates processors/translators to set response model from API responses where available.
- Adds OTEL metrics verification for chat/embeddings with and without model overrides.

This also backfills a lot of documentation, notably using type aliases declared in the same file. This is an effort to make it less difficult to understand the flow.
  - new `internalapi.ModelNameHeaderKey` - Header key for model name, used in routing/metrics
    - consolidate constants into  `internalapi.ModelNameHeaderKeyDefault` - to hold the default "x-ai-eg-model" for `internalapi.ModelNameHeaderKey`
  - new `internalapi.ModelNameOverride` - Backend-specific model override (e.g., llama3 → us.meta.llama3-v1:0)
  - new `internalapi.OriginalModel` - Model name extracted from the incoming request body before any virtualization applies.
  - new `internalapi.RequestModel` - Model name sent to provider (may be overridden)
  - new `internalapi.ResponseModel` - Actual model that generated response (e.g., gpt-5-nano-2025-08-07)
  - Enhances model virtualization docs with provider-specific behaviors and examples.

fixes #1217
fixes #1218


codefromthecrypt:
I have to fly so if green consider this ready for review ;)

internal/extproc/chatcompletion_processor.go
```diff
@@ -371,6 +369,12 @@ func (c *chatCompletionProcessorUpstreamFilter) ProcessResponseBody(ctx context.
 		return nil, fmt.Errorf("failed to transform response: %w", err)
 	}
 
+	// Update response model in metrics
+	// Set response model from actual response when present
+	if tokenUsage.ResponseModel != "" {
+		c.metrics.SetResponseModel(tokenUsage.ResponseModel)
+	}
```
AyushSawant18588: fallback to e.requestHeaders[e.config.modelNameHeaderKey] header is missed here

internal/extproc/translator/openai_openai.go
```diff
@@ -134,9 +134,10 @@ func (o *openAIToOpenAITranslatorV1ChatCompletion) ResponseBody(_ map[string]str
 		return nil, nil, tokenUsage, fmt.Errorf("failed to unmarshal body: %w", err)
 	}
 	tokenUsage = LLMTokenUsage{
-		InputTokens:  uint32(resp.Usage.PromptTokens),     //nolint:gosec
-		OutputTokens: uint32(resp.Usage.CompletionTokens), //nolint:gosec
-		TotalTokens:  uint32(resp.Usage.TotalTokens),      //nolint:gosec
+		InputTokens:   uint32(resp.Usage.PromptTokens),     //nolint:gosec
+		OutputTokens:  uint32(resp.Usage.CompletionTokens), //nolint:gosec
+		TotalTokens:   uint32(resp.Usage.TotalTokens),      //nolint:gosec
+		ResponseModel: resp.Model,
```
AyushSawant18588: Does not seem right to add ResponseModel in LLMTokenUsage, Can be returned separately from this function

internal/extproc/chatcompletion_processor.go
```diff
@@ -227,14 +227,12 @@ func (c *chatCompletionProcessorUpstreamFilter) ProcessRequestHeaders(ctx contex
...
-	respModel := c.requestHeaders[c.config.modelNameHeaderKey]
-	// Request label should reflect the original user-provided model; fall back to header if body is unavailable.
-	reqModel := respModel
+	// Set the request model for metrics. This should reflect the original user-provided model.
+	reqModel := c.requestHeaders[c.config.modelNameHeaderKey]
 	if c.originalRequestBody != nil && c.originalRequestBody.Model != "" {
 		reqModel = c.originalRequestBody.Model
 	}
-	c.metrics.SetModel(reqModel, respModel)
+	c.metrics.SetModel(reqModel)
```
AyushSawant18588: One issue with not setting response model here itself is that when recording Request Duration metric incases of failure in any of the process functions till ProcessResponseBody the gen_ai.response.model label will be empty

codefromthecrypt: right but in error cases it can be wrong, correct? I suspect this is why it is optional in the otel spec. Putting an incorrect value in for the response model can cause problems.

codefromthecrypt: wrong for example, it is a 400 or 404, so the response model is invalid.

internal/extproc/translator/openai_openai.go
codefromthecrypt: good catch. I'll remove it!

internal/extproc/chatcompletion_processor.go
codefromthecrypt: another thing, and I clarified in the docs, that I think a lot of people misunderstand "request model" could not be a model at all, as many including azure either virualize or ignore the request model field. If we put unverified information into the response model field, we are tainting the only authoritative field we have. I believe doing this would be a mistake, but maybe you have information I'm not aware of and we should document. Here's what I've documented so far https://github.com/envoyproxy/ai-gateway/pull/1219/files#diff-b7fd69078accb792965b2ac3406b30ae4abed781c106df68d254edf969170430

codefromthecrypt:
@AyushSawant18588 marking as a draft. bugs aside (thanks for catch) I realize that I didn't document well enough the problem of recording unauthoritative data as the response model in metrics.

I'll re-open this after revising the docs, so that it is easier to get on the same page. This way, if what ends up merged is revisited later, even a day after, the rationale for doing this was captured at some point not in GitHub comments, which are usually never read again.

Bear with me as I'm traveling so might get to this today (sat morning my time) or mondayish

codefromthecrypt:
ok I have gone through and documented what was missing before which can explain why models are different and how our processing affects the one used for the request. I need to audit the changes themselves carefully when back to work Monday or tuesday before taking out of draft.

internal/extproc/chatcompletion_processor.go
```diff
@@ -550,8 +555,10 @@ func buildDynamicMetadata(config *processorConfig, costs *translator.LLMTokenUsa
 		metadata[rc.MetadataKey] = &structpb.Value{Kind: &structpb.Value_NumberValue{NumberValue: float64(cost)}}
 	}
 
-	if modelNameOverride != "" {
-		metadata["model_name_override"] = &structpb.Value{Kind: &structpb.Value_StringValue{StringValue: modelNameOverride}}
+	// Add the actual request model that was used (after any backend overrides were applied).
+	// At this point, the header contains the final model that was sent to the upstream.
```
codefromthecrypt, line 557: after tracing the code, I added comments like this to explain what we are doing as they also explain why we can rely on it (and ignore fields named modelNameOverride which at some point are redundant)

internal/extproc/messages_processor_test.go
```diff
@@ -616,7 +619,69 @@ func TestMessages_ProcessRequestHeaders_SetsRequestAndResponseModels(t *testing.
...
+	// Should use body model as request model
+	require.Equal(t, "body-model", mm.requestModel)
+	// Response model is not set until we get actual response
+	require.Empty(t, mm.responseModel)
+}
+
+// TestMessages_ProcessResponseBody_UsesActualResponseModelOverHeaderOverride verifies that
+// the actual response model from the API response is used for metrics, not the header override.
+// This is important because the backend may return a more specific model version than what was
+// requested (e.g., "claude-3-opus-20240229" instead of "claude-3-opus").
```
codefromthecrypt, line 632: TODO worst example: use azure or openai instead of;)

internal/metrics/embeddings_metrics.go
```diff
@@ -23,8 +23,18 @@ type embeddings struct {
 type EmbeddingsMetrics interface {
 	// StartRequest initializes timing for a new request.
 	StartRequest(headers map[string]string)
-	// SetModel sets the model the request. This is usually called after parsing the request body .
-	SetModel(requestModel, responseModel string)
+	// SetModel sets the model from the request. This is usually called after parsing the request body.
+	// Example: text-embedding-3-small
+	SetModel(requestModel string)
+	// SetResponseModel sets the model that ultimately generated the response.
```
codefromthecrypt, line 30: I hope this comment helps in the future. Last year when working on otel genai metrics spec, it was really confusing why the response model could be different, but this is the short answer why

AyushSawant18588:
Thanks for providing these provider-specific virtualization examples. I see how the model name in the request can differ from the model name in the response (for example, sending a request to an alias like gpt-4o might yield a response from the actual model gpt-4o-2024-11-20). However, I have a concern regarding metrics and observability:

If we start populating gen_ai.response.model with the actual provider model name (the versioned or provider-specific identifier) instead of the user-defined alias from modelNameOverride, it could become difficult for users to query and interpret metrics. From a user’s perspective, they know the unified alias they used (e.g. “claude-4-sonnet” or “gpt-4o”), and they will likely try to filter or aggregate metrics by that alias. If the metrics are labeled with a different, provider-specific name that the user never directly interacted with, how would they know what to query for? This seems to create an observability gap: the label in the metrics no longer matches the model identifier the user is familiar with.
Also this can inflate label cardinality as provider point-releases roll (each timestamped version becomes a new time series).

Given this concern, how should we handle this use case after the change? Is there a recommended approach to ensure users can still query metrics by the model names (aliases) they use in practice?

codefromthecrypt:
@AyushSawant18588, this is a great question that essentially asks, "Who is querying the metrics?" You're correct that in some cases, end users will primarily query metrics, while in other deployments, system administrators will be the ones doing so. The frontend/backend perspective is exactly why we have a router, so this is unavoidable.

That said, spans and metrics must have primary data around the metric they record. In this case, the backend/upstream. Spans must record upstream data, as this is the data actually sent to the system, especially when guardrails are involved later. Metrics like streaming metrics and TTFT (time to first token) are inherently tied to the upstream. Therefore, the data recorded should primarily reflect what flows in and out of the backend. So, how do we address this tension? The most troubled way is to "fight over a field" and record incoherent or asymetric data in a field. When an attribute is defined for a purpose, use it for that, and be consistent in the key request/response parity.

I believe if you used azure you would feel differently about the response model issue. Literally, there is no way to know the actual model in use if you were the user, unless that field is read. There is no risk of metrics splitting worse in openai on response model than other rarely changing attributes. There is no zero sum game on this because the response model is written for a specific purpose which is capturing exactly this concern. If someone were to audit all the attributes and felt the response model is something that is causing issues, it could be filtered out. I strongly believe that using the response model as a sort of bucket to put data that is not about the response model is a change I wouldn't make. I would insist you make that change on your own after this PR and take ownership of that decision. Even if I disagree you can still do that, just I wouldn't personally be associated with it.

What I think is happening is we are over focusing on a zero sum game. It is possible to add additional attributes to avoid "fighting over a field" and using it for a purpose it was not defined for, or creating an additional metric.

I don't really have time to go back and forth on this topic. So, if you feel strongly that this PR which makes the code coherent must also have a separate field for the router model, that it must be in the same PR as this cleanup. I'll do it. However, I think in doing open source for over 15 years it is best to be able to do coherent cleanup changes separate from behavioral ones.

codefromthecrypt:
@mathetake I think more people should be involved about the customization of the genai metrics. I see the value in retaining the routing model in the data. However, I don't think it should happen in this PR nor are we better off doing it that way. Holding cleanups captive is not a practice we want to promote in this project, where there are still a lot of cleanup left to do. It discourages people from doing cleanups, and also makes it very difficult to understand change.

So, my suggesting forward is we keep this PR to following the spec https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-metrics/#generative-ai-client-metrics and not abusing any fields. Immediately after I will raise a PR to pencil in an additional attribute for the routing model, and even do the research for what others do about this. Then, that PR is easier to review and more than one person besides me can easily participate in that discussion as it is a change from a coherent state vs an incoherent one. SG?

codefromthecrypt:
TODOs for me:

1. finish the work I had this in draft for (basically I put panics in the code to make sure all code is coherent. backfill tests and then remove the panic on empty)
2. audit the code and docs carefully so that there is nothing incoherent in this PR. That in all cases either we use the authoritative **response** model for the response (not the request), or we pass the request because there is no such field
3. raise a draft PR of adding an attribute to clarify the requested model, similar to litellm who have an attribute for exactly this reason

once I do all that, I'll take it out of draft, and hope this is sensible why I am pretty dug in on this. We have to catalog what we are doing and what the spec says (put the response model into the response field). This does not prevent us from making a dialect of the spec which is tuned for proxies. Just one step at a time.

hope this makes sense!

mathetake:
> @mathetake I think more people should be involved about the customization of the genai metrics. I see the value in retaining the routing model in the data. However, I don't think it should happen in this PR nor are we better off doing it that way. Holding cleanups captive is not a practice we want to promote in this project, where there are still a lot of cleanup left to do. It discourages people from doing cleanups, and also makes it very difficult to understand change.
> 
> So, my suggesting forward is we keep this PR to following the spec https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-metrics/#generative-ai-client-metrics and not abusing any fields. Immediately after I will raise a PR to pencil in an additional attribute for the routing model, and even do the research for what others do about this. Then, that PR is easier to review and more than one person besides me can easily participate in that discussion as it is a change from a coherent state vs an incoherent one. SG?

Sounds good and reasonable. I think the discussion here should get more people involved as you mentioned as well as this including the original vs modelOverride in metrics was discussed in the past community meeting and the consensus was that we should stick to the overridden model as a "request" model" which I think we should revisit in that follow up PR. To be clear, at the time of the discussion, we were not aware of the label "response.model" and the discussion was solely around which we should put in request.model label, the model in the original request or the overridden one. 

Since it's not like we are cutting new release tomorrow or soon, I believe we have plenty of time to land something that everyone feels right, not limited to you and @AyushSawant18588!

internal/extproc/chatcompletion_processor.go
codefromthecrypt: moved to capture this: https://github.com/envoyproxy/ai-gateway/issues/1224

It is tricky to make tests for this edge case, but after this PR is done, if someone wants to do more "best efforts" it would be easier.

codefromthecrypt:
here's the final analysis on the router model thing. I'll think about the naming overnight and then put the PR up tomorrow.

| Gateway              | Metric Name                                               | Original Model           | Request Model          | Response Model          | Override Support                                  | Notes                                                                                     |
|----------------------|-----------------------------------------------------------|--------------------------|------------------------|-------------------------|---------------------------------------------------|-------------------------------------------------------------------------------------------|
| **AgentGateway**     | [`gen_ai_client_token_usage`][agentgateway-metrics]       | `gen_ai_request_model`   | N/A                    | `gen_ai_response_model` | Config file per-provider                          | No attribute for overridden model sent to backend                                         |
| **Envoy AI Gateway** | [`gen_ai.client.token.usage`][ai-gateway-metrics]         | `gen_ai.request.model`   | N/A                    | `gen_ai.response.model` | Backend config override                           | No attribute for overridden model sent to backend                                         |
| **Bifrost**          | [`bifrost_upstream_requests_total`][bifrost-metrics]      | N/A                      | `model`                | N/A                     | N/A                                               |                                                                                           |
| **Higress**          | [`ai-statistics`][higress-metrics]                        | N/A                      | `gen_ai.request.model` | N/A                     | Provider-specific in ai-proxy                     |                                                                                           |
| **Kong AI Gateway**  | [`ai.*.meta`][kong-metrics]                               | N/A                      | `request_model`        | `response_model`        | Provider-specific                                 | `response_model` falls back to `request_model` if extraction fails                        |
| **Labring AI Proxy** | [`summary_minutes`][labring-metrics]                      | N/A                      | `model`                | N/A                     | Channel-based model mapping                       | `model` attribute contains `ActualModel` (post-mapping); `OriginModel` not in metrics     |
| **LiteLLM**          | [`litellm_deployment_success_responses`][litellm-metrics] | `requested_model`        | `litellm_model_name`   | N/A                     | Config file model aliasing and deployment mapping | `requested_model` contains model group; `litellm_model_name` contains specific deployment |
| **Llama Stack**      | [`prompt_tokens`][llama-stack-metrics]                    | N/A                      | `model_id`             | N/A                     | N/A                                               | `model_id` and `provider_id` attributes in all token metrics                              |

---

[agentgateway-metrics]: https://github.com/agentgatewayai/agentgateway/blob/main/crates/agentgateway/src/telemetry/metrics.rs#L31
[ai-gateway-metrics]: https://github.com/envoyproxy/ai-gateway/blob/main/internal/metrics/genai.go#L14
[bifrost-metrics]: https://github.com/maximhq/bifrost/blob/main/plugins/telemetry/main.go#L37
[higress-metrics]: https://github.com/alibaba/higress/blob/main/plugins/wasm-go/extensions/ai-statistics/main.go#L78
[kong-metrics]: https://github.com/Kong/kong/blob/master/kong/llm/plugin/shared-filters/serialize-analytics.lua#L53
[labring-metrics]: https://github.com/labring/aiproxy/blob/main/core/model/summary-minute.go#L15
[litellm-metrics]: https://github.com/BerriAI/litellm/blob/main/litellm/proxy/proxy_server.py#L1180
[llama-stack-metrics]: https://github.com/meta-llama/llama-stack/blob/main/llama_stack/core/routers/inference.py#L150

## Legend
- **Original Model** = The model name extracted from the incoming request body
- **Request Model** (sent to backend) = Either the Original Model OR the overridden model if configured
- **Response Model** (from backend) = The actual model the backend used
- **Override Support** = Mechanism for overriding/translating the model (N/A if not supported)

codefromthecrypt:
in creating the draft PR over this, I found a logic bug as the request model is set too early. I'll put another test here to make sure it is the correct value

codefromthecrypt:
OK took a while to backfill functional tests for the metrics, and also assemble the [research](https://github.com/envoyproxy/ai-gateway/pull/1219#issuecomment-3323855101). I've updated the description to match the changes.

NEXT STEP: Raise a draft PR over this branch which adds "gen_ai.original.model" attribute to metrics, so that one can aggregate differently based on what the router received (original) vs the request sent to the backend.

I am going with "original" for now, as I notice a lot of "original" in our codebase and also some envoy headers use that term for the same thing. It is a bikeshed I can manage on the other PR if contested.

codefromthecrypt:
FYI: before taking this out of draft, the OriginalModel constant will be removed from the commit here (as that will be in the subsequent PR which allows us to decouple debates about it from this PR)

codefromthecrypt:
https://github.com/envoyproxy/ai-gateway/pull/1230 takes the "original' concept so let's get this in as it is blocking a lot of things including embeddings tracing!

api/v1alpha1/shared_types.go
```diff
@@ -64,12 +64,6 @@ const (
 	APISchemaGCPAnthropic APISchema = "GCPAnthropic"
 )
 
-const (
-	// AIModelHeaderKey is the header key whose value is extracted from the request by the ai-gateway.
-	// This can be used to describe the routing behavior in HTTPRoute referenced by AIGatewayRoute.
-	AIModelHeaderKey = "x-ai-eg-model"
```
mathetake, line 70: This is not really an "internal" as it's used in the user-defined AIGatewayRules. For example, some user (I assume BB) programmatically generates the AIGatewayRoute where they define the rule using this constant. In other words, this could be also a breaking change, though not that serious since it's not like k8s will be broken.

So I would suggest that keep this and make the alias in internalapi where we document for the internal codebase for this PR. wdyt?

internal/extproc/chatcompletion_processor.go
```diff
@@ -227,14 +227,12 @@ func (c *chatCompletionProcessorUpstreamFilter) ProcessRequestHeaders(ctx contex
...
 	c.metrics.StartRequest(c.requestHeaders)
-	// Response label at this stage defaults to the effective header model; backend override (if any) will be applied in SetBackend.
-	respModel := c.requestHeaders[c.config.modelNameHeaderKey]
-	// Request label should reflect the original user-provided model; fall back to header if body is unavailable.
-	reqModel := respModel
-	if c.originalRequestBody != nil && c.originalRequestBody.Model != "" {
+	// Set the request model for metrics from the original model or override if applied.
+	reqModel := c.requestHeaders[c.config.modelNameHeaderKey]
+	if reqModel == "" && c.originalRequestBody != nil && c.originalRequestBody.Model != "" {
 		reqModel = c.originalRequestBody.Model
```
mathetake: I can't think of the situation where `c.originalRequestBody != nil && c.originalRequestBody.Model != ""` is not true ... (it shouldn't exist in the current main in the first place...)

internal/extproc/chatcompletion_processor.go
```diff
@@ -550,8 +555,10 @@ func buildDynamicMetadata(config *processorConfig, costs *translator.LLMTokenUsa
 		metadata[rc.MetadataKey] = &structpb.Value{Kind: &structpb.Value_NumberValue{NumberValue: float64(cost)}}
 	}
 
-	if modelNameOverride != "" {
-		metadata["model_name_override"] = &structpb.Value{Kind: &structpb.Value_StringValue{StringValue: modelNameOverride}}
+	// Add the actual request model that was used (after any backend overrides were applied).
+	// At this point, the header contains the final model that was sent to the upstream.
+	if actualModel := requestHeaders[config.modelNameHeaderKey]; actualModel != "" {
```
mathetake: not sure when this will be false...

internal/extproc/chatcompletion_processor.go
mathetake: ah sorry nvm

internal/extproc/translator/openai_awsbedrock.go
```diff
@@ -20,23 +20,27 @@ import (
...
 // NewChatCompletionOpenAIToAWSBedrockTranslator implements [Factory] for OpenAI to AWS Bedrock translation.
-func NewChatCompletionOpenAIToAWSBedrockTranslator(modelNameOverride string) OpenAIChatCompletionTranslator {
+func NewChatCompletionOpenAIToAWSBedrockTranslator(modelNameOverride internalapi.ModelNameOverride) OpenAIChatCompletionTranslator {
 	return &openAIToAWSBedrockTranslatorV1ChatCompletion{modelNameOverride: modelNameOverride}
 }
 
-// openAIToAWSBedrockTranslator implements [Translator] for /v1/chat/completions.
+// openAIToAWSBedrockTranslator translates OpenAI Chat Completions API requests to AWS Bedrock Converse API.
+// Note: This uses the Converse API directly, not Bedrock's OpenAI-compatible API:
+// https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_Converse.html
```
mathetake, line 34: 👍

internal/extproc/translator/openai_openai.go
```diff
@@ -116,28 +129,42 @@ func (o *openAIToOpenAITranslatorV1ChatCompletion) ResponseHeaders(map[string]st
...
+			return nil, nil, tokenUsage, o.requestModel, fmt.Errorf("failed to read body: %w", err)
 		}
 		o.buffered = append(o.buffered, buf...)
 		tokenUsage = o.extractUsageFromBufferEvent(span)
+		// Use stored streaming response model, fallback to request model for non-compliant backends
+		if o.streamingResponseModel != "" {
+			responseModel = o.streamingResponseModel
+		} else {
+			responseModel = o.requestModel
+		}
```
mathetake: I like using cmp.Or
```suggestion
		responseModel = cmp.Or(o.streamingResponseModel, o.requestModel)
```

internal/extproc/translator/openai_openai.go
```diff
@@ -116,28 +129,42 @@ func (o *openAIToOpenAITranslatorV1ChatCompletion) ResponseHeaders(map[string]st
...
 	tokenUsage = LLMTokenUsage{
 		InputTokens:  uint32(resp.Usage.PromptTokens),     //nolint:gosec
 		OutputTokens: uint32(resp.Usage.CompletionTokens), //nolint:gosec
 		TotalTokens:  uint32(resp.Usage.TotalTokens),      //nolint:gosec
 	}
+	responseModel = resp.Model
+	// Fallback to request model for test or non-compliant OpenAI backends
+	if responseModel == "" {
+		responseModel = o.requestModel
+	}
```
mathetake:
```suggestion
	// Fallback to request model for test or non-compliant OpenAI backends
	responseModel = cmp.Or(resp.Model, o.requestModel)
```

internal/extproc/translator/translator.go
```diff
@@ -56,10 +57,12 @@ type OpenAIChatCompletionTranslator interface {
 	// 	- `body` is the response body either chunk or the entire body, depending on the context.
 	//	- This returns `headerMutation` and `bodyMutation` that can be nil to indicate no mutation.
 	//  - This returns `tokenUsage` that is extracted from the body and will be used to do token rate limiting.
+	//  - This returns `responseModel` that is the model name from the response (may differ from request model).
```
mathetake, line 60: 👍

internal/filterapi/filterconfig.go
```diff
@@ -36,7 +39,7 @@ type Config struct {
 	// the "calculated" cost in the filter metadata at the end of the response body processing.
 	LLMRequestCosts []LLMRequestCost `json:"llmRequestCosts,omitempty"`
 	// ModelNameHeaderKey is the header key to be populated with the model name by the filter.
-	ModelNameHeaderKey string `json:"modelNameHeaderKey"`
+	ModelNameHeaderKey internalapi.ModelNameHeaderKey `json:"modelNameHeaderKey"`
```
mathetake, line 42: side note: I was looking to drop this ModelNameHeaderKey configurability from here there's no reason to make it so... let's think about it later ...

mathetake (approved):
LGTM except for some minor comments!

internal/filterapi/filterconfig.go
codefromthecrypt: yeah I also would prefer to delete this and just re-use the value constant. would be a lovely post cleanup!

api/v1alpha1/shared_types.go
codefromthecrypt: yeah good point. I didn't notice this was not internal

codefromthecrypt:
ok 🤞 tests pass. thanks for the cleanup suggestions, @mathetake!

PR #1222, opened 2025-09-22
fix: aws bedrock reasoning content delta for streaming

yuzisun:
**Description**
Fix bedrock reasoning content delta type which is different from the non-streaming case.
https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_ReasoningContentBlockDelta.html
```text
data: {"choices":[{"index":0,"delta":{"content":"","role":"assistant"}}],"object":"chat.completion.chunk"}
data: {"choices":[{"index":0,"delta":{"role":"assistant","reasoning_content":{"text":"I nee"}}}],"object":"chat.completion.chunk"}
data: {"choices":[{"index":0,"delta":{"role":"assistant","reasoning_content":{"text":"d to calculate 27"}}}],"object":"chat.completion.chunk"}
data: {"choices":[{"index":0,"delta":{"role":"assistant","reasoning_content":{"text":" * 453"}}}],"object":"chat.completion.chunk"}
data: {"choices":[{"index":0,"delta":{"role":"assistant","reasoning_content":{"text":"."}}}],"object":"chat.completion.chunk"}
```

### DIFF
--- a/internal/metrics/metrics_test.go
+++ b/internal/metrics/metrics_test.go
@@ -427,3 +427,37 @@ func TestNewMetricsFromEnv_ErrorHandling(t *testing.T) {
 		})
 	}
 }
+
+// TestNewMetricsFromEnv_OTLPHeaders tests that OTEL_EXPORTER_OTLP_HEADERS
+// is properly handled by the autoexport package.
+func TestNewMetricsFromEnv_OTLPHeaders(t *testing.T) {
+	expectedAuthorization := "ApiKey test-key-123"
+	actualAuthorization := make(chan string, 1)
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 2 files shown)


internal/apischema/awsbedrock/awsbedrock.go
```diff
@@ -267,6 +267,17 @@ type ReasoningContentBlock struct {
 	RedactedContent []byte `json:"redactedContent,omitempty"`
 }
 
+// ReasoningContentBlockDelta contains the reasoning trace for the inference that the model ran.
+// See: https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_ReasoningContentBlockDelta.html
+type ReasoningContentBlockDelta struct {
+	// The reasoning that the model used to return the output.
+	Text string `json:"text"`
```
mathetake: according to the document, it looks like the text is also optional? 
```suggestion
	Text string `json:"text,omitzero"`
```

Issue #1085, opened 2025-08-15
add tracing support for the embeddings router filter


codefromthecrypt:
*Description*:

@Xunzhuo asked me about tracing support for embeddings, and here's info from @mikeldking (phoenix lead)

> We do have some ways to evaluate embeddings, notably finding outliers (e.g. toxic content, out of bounds questions) as well as EDA (question space exploration)
> But the benefit of logging embeddings does come at a storage cost.
> In general if you're doing things like query expansion or agentic RAG, I can see it being useful to see the text getting embedded.

![](https://files.slack.com/files-pri/T5CALEVRN-F09ADSV0N4U/image.png)

![](https://files.slack.com/files-pri/T5CALEVRN-F09AC115P54/image.png)

![](https://files.slack.com/files-pri/T5CALEVRN-F09A4AY7JSF/image.png)

It makes sense to add support for the embeddings endpoint, by recording openai responses, openinference spans, and creating the impl like we did for chat completions.

mathetake:
Let's aim for v0.4!!!

codefromthecrypt:
Cool. I found 1 or 2 embedding requests so far that cause hiccups even before tracing occurs. I wont add tracing before 0.4 but will figure out if there is a bug somewhere and report back today.

codefromthecrypt:
in one case it seems the OpenAI platform in an embedding error sends a response with 1 character less than the Content-Length it sets. I think that's not something we'll likely correct (clients may struggle, but same bug can happen in any SaaS or http service). Unless I find anything else, probably fine on embeddings.

codefromthecrypt:
here are the two glitches and backfill of all test data without adding tracing yet. This ensures RAG apps that use token embeddings don't crash on the 0.3 release https://github.com/envoyproxy/ai-gateway/pull/1087

codefromthecrypt:
about the content-length issue, it seems typical for extproc to adjust this, so I think the thing I found when upstream openai sends a buggy reply (content-length longer than the content) we can just fix this (by not crashing on EOF as long as we have valid json). Will give a go.

codefromthecrypt:
the content-length situation isn't specific to embeddings as it is reproducible generally. I'm not sure we can affect it without changing the filter modes, as the crash seems to occur in envoy after response headers and before the body hook is sent back to the extproc. So, I added a new issue and we can forget about it here https://github.com/envoyproxy/ai-gateway/issues/1091

codefromthecrypt:
[...]

PR #1232, opened 2025-09-25
chore: update to latest openinference embedding schema

codefromthecrypt:
**Description**

This updates to the latest openinference semantic conventions for CreateEmbeddings requests, and re-records all spans. This also removes the ModelO3Mini as OpenAI themselves recently removed it.

#1085


codefromthecrypt:
wait.. there's a huge data here.. lemme look which

codefromthecrypt:
ah right because the span data is decoded vectors, which ends up really big in json due to otel json encoding not being efficient:

```json
      "key": "embedding.embeddings.4.embedding.vector",
      "value": {
        "arrayValue": {
          "values": [
            {
              "doubleValue": 0.030946535989642143
            },
            {
              "doubleValue": 0.007837203331291676
            },
```

mathetake (approved):
LGTM-- I confirmed that this does not have any functional change, which is aligned with the PR description. Pure tests updates and hardening sounds good to me

PR #1233, opened 2025-09-25
feat(aigw): generate config from OpenAI environment variables

### LINKED ISSUE #1211: Generate aigw config on-demand with OpenAI ENV variables
*Description*:

Right now, we have a default config with a lot of options in it for public clouds.

However, for openai compatible this doesn't work as well.  This is why there is a https://github.com/envoyproxy/ai-gateway/blob/main/cmd/aigw/ai-gateway-local.yaml file, which defaults to ollama and is overridable for plain text openai compatible services via OPENAI_HOST and OPENAI_PORT. I have used this to test with  other stacks personally.

I think something like this is more natural and easier for end users, and fits in well with other systems like LiteLLM:

```bash
# for a single openai platform backend
$ OPENAI_API_KEY=sk-your-key aigw run
# for a compatible service like ollama
$ OPENAI_BASE_URL=http://localhost:11434/v1 OPENAI_API_KEY=unused aigw run
```

This makes configuration disappear.

codefromthecrypt:
This commit implements automatic configuration generation for aigw when OpenAI environment variables are set, making it easier to get started with OpenAI-compatible services without writing YAML configuration.

When OPENAI_API_KEY is set and no config file is provided, aigw will automatically generate appropriate configuration based on:
- OPENAI_API_KEY: API key for authentication (required)
- OPENAI_BASE_URL: Base URL for the backend (defaults to https://api.openai.com/v1)

Key features:
- Automatic localhost to 127.0.0.1.nip.io conversion for Docker/K8s
- TLS detection based on URL scheme (https) e.g., Tetrate Agent Router Service: https://api.router.tetrate.ai/v1
- Support for custom API path prefixes (technically this is the path prefix) e.g., OpenRouter: https://openrouter.ai/api/v1 e.g., Llama Stack: http://localhost:8321/v1/openai/v1
- Clean YAML output (omits version field when it's "v1")

This simplifies common use cases:
- OpenAI: OPENAI_API_KEY=sk-your-key aigw run
- Ollama: OPENAI_BASE_URL=http://localhost:11434/v1 OPENAI_API_KEY=unused aigw run

Here's an example:
```bash
$  OPENAI_API_KEY=sk-not-tellin go run . run
1.35.3 is already downloaded
starting: /tmp/envoy-gateway/versions/1.35.3/bin/envoy in run directory /tmp/envoy-gateway/runs/1758789738364346000
[2025-09-25 16:42:18.391][48486493][warning][config] [source/server/options_impl_platform_default.cc:9] CPU number provided by HW thread count (instead of cpuset).
{"bytes_received":124,"bytes_sent":795,"connection_termination_details":null,"downstream_local_address":"127.0.0.1:1975","downstream_remote_address":"127.0.0.1:62691","duration":3635,"genai_backend_name":"default/openai/route/aigw-run/rule/0/ref/0","genai_model_name":"gpt-5-nano","genai_model_name_override":"gpt-5-nano","genai_tokens_input":21,"genai_tokens_output":268,"method":"POST","response_code":200,"start_time":"2025-09-25T08:43:07.390Z","upstream_cluster":"httproute/default/aigw-run/rule/0","upstream_host":"162.159.140.245:443","upstream_local_address":"192.168.0.108:62692","upstream_transport_failure_reason":null,"user-agent":"curl/8.14.1","x-envoy-origin-path":"/v1/chat/completions","x-request-id":"9daf1c85-f75e-4c88-a90c-f9cccf85970c"}
```

Fixes #1211


codefromthecrypt:
@anuraaga how is this? no more config and also no need to do something different just to switch your openai endpoint to something not pre-canned.

codefromthecrypt:
once this is in, I'll update https://github.com/elastic/observability-examples/tree/main/inference-platforms/aigw which turns it from the worst example in that dir to the best one.

internal/envgen/config.yaml.tmpl
```diff
@@ -0,0 +1,194 @@
...
+kind: AIServiceBackend
+metadata:
+  name: openai
+  namespace: default
+spec:
+  timeouts:
+    request: 3m
+  schema:
+    name: OpenAI
+{{- if and .Version (ne .Version "v1") }}
```
anuraaga:
```suggestion
{{- if .Version }}
```

I guess this is already handled in config.go

cmd/aigw/run.go
```diff
@@ -247,18 +248,27 @@ func pollEnvoyReadiness(ctx context.Context, l *slog.Logger, addr string, interv
...
-// similar to `envsubst`. If the path is empty the default config is returned.
+// readConfig reads and returns the configuration as a string from the specified path,
+// substituting environment variables similar to envsubst. If the path is empty and
+// OPENAI_API_KEY is set, it generates the configuration from OpenAI environment variables.
+// If the path is empty and no OPENAI_API_KEY is set, it returns the default configuration.
+// An error is returned if the file cannot be read or if config generation fails.
 func readConfig(path string) (string, error) {
 	if path == "" {
+		if os.Getenv("OPENAI_API_KEY") != "" {
+			config, err := envgen.GenerateOpenAIConfig()
```
nacx, line 259: The current default configuration file already includes an OpenAI backend that is used by default if no explicit header is set to force the backend, and includes other backends taht are enabled as well if the credentials are present.

What are your thoughts on just keeping that file and letting it honor the `OPENAI_BASE_URL` variable, instead of creating an OpenAI-only config if that env var is set? This way we can keep the current behavior but with other backends still enabled as well (helping users more easily get started with those as well), while allowing to cleanly use a custom openai-compatible backend?

site/docs/cli/run.md
```diff
@@ -16,6 +16,27 @@ Not only does it help in testing the configuration, but it is also useful in a l
...
+```
+
+### Supported OpenAI Environment Variables
+
+| Variable          | Required | Default                     | Description                                          |
+|-------------------|----------|-----------------------------|------------------------------------------------------|
+| `OPENAI_API_KEY`  | Yes      | -                           | API key for authentication (use "unused" for Ollama) |
+| `OPENAI_BASE_URL` | No       | `https://api.openai.com/v1` | Base URL of your OpenAI-compatible backend           |
+
 ## Default Proxy
```
nacx, line 40: Related to my comment above about generating an OpenAI-only file, this "default proxy" section is no longer true with this change. Depending on what is decided on the previous comment we may need to adjust these docs as well.

internal/envgen/testdata/llamastack.yaml
```diff
@@ -0,0 +1,176 @@
...
+kind: AIServiceBackend
+metadata:
+  name: openai
+  namespace: default
+spec:
+  timeouts:
+    request: 3m
+  schema:
+    name: OpenAI
+    version: "v1/openai/v1"
```
anuraaga, line 125: Heh - well assuming this is what we wanted

anuraaga (approved):
Pretty awesome

cmd/aigw/run.go
codefromthecrypt: When we get to adding the other OPENAI env variables, such as OPENAI_ORG_ID OPENAI_PROJECT_ID, I think the value of that other file quickly approaches zero.

internal/envgen/testdata/llamastack.yaml
codefromthecrypt: yeah it is a horribly named field

site/docs/cli/run.md
codefromthecrypt: strong no on adding tech debt to make sure a config that should have been in the examples folder in the first place isn't impacted. I'll revise the PR comments

codefromthecrypt:
@nacx @mathetake as I think the three of us represent the largest population of users of the CLI without an argument, please read the *Reviewer's Notes* section carefully. I will own moving the monolithic file to an example after this PR so that there's no risk of a future person expecting it to work by default anymore. I can also add https support for the file arg if one-step is required: if anyone is able to use that file, they can reach the internet.

nacx:
Thanks for adding the notes and motivation. I agree on moving the monolithic config to examples and agree that would help clean things up and clean the docs.
The PR LGTM. Thx!

codefromthecrypt:
after this, finishing up the standard OpenAI ENV (e.g. for Azure) will be doable. At elastic, this was the more prolific endpoint so that's important especially as it is a bit odd in how it works (deployment instead of a model)

Personally, the more important thing I would like to demonstrate with this "quick start" mode is all 3 otel signals. Adding support for otel logs based on ENV variable (remapping the access logger) will make it possible to send directly to elastic cloud with no scraping or log forwarding. Even if a real cluster needs more work than this, it will be great to show the only proxy that supports all three: logging, metrics and tracing OOB (afaik)

internal/envgen/testdata/llamastack.yaml
mathetake: impending refactoring & deprecation

mathetake (approved):
LGTM assuming moving the current default.yaml into example as well as removal of `--show-default` flag in a follow up PR?

mathetake:
On second thought, i think we can just remove the current default config. It overlaps with the existing example configs since there's nothing specific to `aigw run` vs k8s. I think that's clearer anyways  (examples/basic provides already multiple routes to multiple providers example)

codefromthecrypt:
good point @mathetake https://github.com/envoyproxy/ai-gateway/pull/1235

PR #1234, opened 2025-09-25
deps: use Go 1.25.1 & update libs

mathetake:
**Description**

This updates the Go version used here as well as some Go dependencies.


go.mod
```diff
@@ -1,35 +1,36 @@
 module github.com/envoyproxy/ai-gateway
 
-go 1.25
+// Explicitly specify the Go patch version to be able to purge the CI cache correctly.
+go 1.25.1
```
mathetake, line 4: Note: not sure exactly why but apparently the CI didn't automatically use 1.25.1 hence Trivy reported some CVE (which is not applicable to this project though) on main branch.

tests/extproc/vcr/prometheus_metrics_test.go
```diff
@@ -43,7 +44,7 @@ func TestPrometheusMetrics(t *testing.T) {
 
 	// Wait for metrics to be available with exponential backoff.
 	// Parse the Prometheus metrics.
-	parser := expfmt.TextParser{}
+	parser := expfmt.NewTextParser(model.UTF8Validation)
```
mathetake, line 47: without this, the latest prometheus library panics

PR #1236, opened 2025-09-27
feat(aigw): consolidate admin server into a single port

codefromthecrypt:
**Description**

This consolidates the extproc admin server from two separate ports (metricsPort and healthPort) into a single adminPort (default 1064) serving both /metrics and /health endpoints. The `aigw run` and the new `aigw healthcheck` commands now take --admin-port flags, propagating this to extproc.

Previously, extproc exposed metrics on port 1064 and health checks on port 1065. Now both are served from a single HTTP server on port 1064, simplifying configuration and Docker HEALTHCHECK setup.

The Docker HEALTHCHECK is now functional, using the new healthcheck subcommand which polls the admin server's /health endpoint. This /health endpoint proxies to the extproc gRPC health check, returning HTTP 200 when SERVING. A new subcommand (`aigw healthcheck`) was needed as the image is distroless (has no curl etc). 

When aigw run starts Envoy, it polls for readiness using either the Envoy admin address (when configured in EnvoyProxy bootstrap) or by checking if the Gateway listener port accepts TCP connections. This ensures startup completes reliably with minimal Gateway configurations.

Gateway validation now requires at least one listener to be configured.


codefromthecrypt:
also polished to remove transient errors from the `make e2e` log at least for test upstream which is usually the one that fails when we change things

codefromthecrypt:
flake.. looking at it..

```
Handling connection for 40963
    token_ratelimit_test.go:124: Response: status=200, body={"status":"success","data":{"resultType":"vector","result":[]}}
    token_ratelimit_test.go:153: 
        	Error Trace:	/home/runner/work/ai-gateway/ai-gateway/tests/e2e/token_ratelimit_test.go:153
        	            				/opt/hostedtoolcache/go/1.25.1/x64/src/runtime/asm_amd64.s:1693
        	Error:      	[]string(nil) does not contain "input"
        	Test:       	Test_Examples_TokenRateLimit
    token_ratelimit_test.go:110: 
        	Error Trace:	/home/runner/work/ai-gateway/ai-gateway/tests/e2e/token_ratelimit_test.go:110
        	Error:      	Condition never satisfied
        	Test:       	Test_Examples_TokenRateLimit
--- FAIL: Test_Examples_TokenRateLimit (143.13s)
```

codefromthecrypt:
Wasn't a flake just a drift. Now gone

examples/monitoring/monitoring.yaml
```diff
@@ -43,17 +43,17 @@ data:
             target_label: kubernetes_pod_name
           # "metrics" is the port name for Envoy Proxy's prometheus metrics.
           # https://github.com/envoyproxy/gateway/blob/eb7266c042f4573d7e67c9ea02a19e0deff6e17f/internal/infrastructure/kubernetes/proxy/resource.go#L92
-          # "aigw-metrics" is the port name for AI Gateway filter's prometheus metrics.
+          # "aigw-admin" is the port name for AI Gateway filter's prometheus metrics.
           - source_labels: [__meta_kubernetes_pod_container_port_name]
-            regex: "metrics|aigw-metrics"
+            regex: "metrics|aigw-admin"
```
mathetake, line 48: so this could also be a breaking change... I think having two names pointing to the same port and removing it after the next release won't hurt. let me do that

internal/controller/gateway_mutator.go
```diff
@@ -219,9 +217,9 @@ func (g *gatewayMutator) mutatePod(ctx context.Context, pod *corev1.Pod, gateway
 		Image:           g.extProcImage,
 		ImagePullPolicy: g.extProcImagePullPolicy,
 		Ports: []corev1.ContainerPort{
-			{Name: "aigw-metrics", ContainerPort: extProcMetricsPort},
+			{Name: "aigw-admin", ContainerPort: extProcAdminPort},
```
mathetake, line 220: so this could also be a breaking change. I think having two names pointing to the same port and removing it after the next release won't hurt. let me do that

Dockerfile
```diff
@@ -35,4 +35,9 @@ COPY --from=envoy-downloader /tmp/envoy-gateway /tmp/envoy-gateway
 COPY ./out/${COMMAND_NAME}-${TARGETOS}-${TARGETARCH} /app
 
 USER nonroot:nonroot
+
+# The healthcheck subcommand performs an HTTP GET to localhost:1064/healthlthy
+HEALTHCHECK --interval=10s --timeout=5s --start-period=5s --retries=3 \
+    CMD ["/app", "healthcheck"]
```
mathetake: this feels not quite right as this image is also used with controller as well as is ignored in k8s.

mathetake: added clarifying comment

internal/controller/gateway_mutator.go
codefromthecrypt: good idea thanks!

Dockerfile
codefromthecrypt: ah cool I didn't realize that this was a shared Dockerfile. thanks

PR #1239, opened 2025-09-28
feat: support configurable inferencepool

Xunzhuo:
**Description**


site/docs/capabilities/inference/aigatewayroute-inferencepool.md
```diff
@@ -433,6 +433,71 @@ AIGatewayRoute provides rich metrics for InferencePool usage:
 - **Token consumption**: Monitor token usage and costs
 - **Endpoint performance**: Detailed metrics per inference endpoint
 
+## InferencePool Configuration Annotations
+
+InferencePool supports configuration annotations to customize the external processor behavior:
```
yuzisun, line 438: I think this is generally useful for other AIGatewayRoute as well ? maybe should be a property there instead of an annotation on the InferencePool.

internal/extensionserver/post_cluster_modify.go
```diff
@@ -19,8 +19,8 @@ import (
 )
 
 // clusterRefInferencePool generates a unique reference for an InferencePool cluster.
-func clusterRefInferencePool(namespace, name, serviceName string, servicePort uint32) string {
-	return fmt.Sprintf("%s/%s/%s/%d", namespace, name, serviceName, servicePort)
+func clusterRefInferencePool(namespace, name, serviceName string, servicePort uint32, bodyMode string, allowModeOverride string) string {
+	return fmt.Sprintf("%s/%s/%s/%d/%s/%s", namespace, name, serviceName, servicePort, bodyMode, allowModeOverride)
```
yuzisun, line 23: why do we need to add the configuration overrides to the reference name?

site/docs/capabilities/inference/aigatewayroute-inferencepool.md
Xunzhuo: Annotation on inference pool makes this work both in aigwroute and httproute

internal/extensionserver/post_cluster_modify.go
Xunzhuo: The inference pool configuration is generating in post translation hook, in there is no inferenepool level extension resource, so it can only read the metadata pre generated in cluster hook and route hook, constructed the configuration of inference pool extproc

site/docs/capabilities/inference/aigatewayroute-inferencepool.md
Xunzhuo: the extension server doesn't care if it is from aigwroute or httproute, both can work : )

yuzisun (approved):

mathetake:
This makes sense!

PR #1248, opened 2025-09-29
fix: adds upgrade testing & graceful termination workaround

### LINKED ISSUE #1060: Add "Upgrade Testing" after v0.3
* Create a directory `tests/e2e-upgrade` and the file `tests/e2e-upgrade/e2e_upgrade_test.go`
* Adds a make recipe `make test-e2e-upgrade`
* In the tests suite
    * Setup the k8s environment as usual except that instead of installing Envoy AI Gateway via helm locally, install `registry-1.docker.io/envoyproxy/ai-gateway-helm --version 0.3.0`. This might require changes in tests/internal/e2elib.
    * Adds a test case where continuously making requests to a gateway with some test manifest with testupstream server etc. (You can reference tests/e2e/testdata directory.
    * While making requests, upgrade Envoy AI Gateway using the local helm under menifests/charts directory, and restart the controller.
    * During the upgrade, ensures all requests are successful for 2 minutes.

### LINKED ISSUE #1241: Request fails during ai gateway upgrade
*Description*:
During upgrade we saw requests failed with the following errors, seems like envoy is able to accept requests but ai gateway extproc container uds socket is not ready yet.

```
ext_proc_error_gRPC_error_14{upstream_connect_error_or_disconnect/reset_before_headers._reset_reason:_re
mote_connection_failure,_transport_failure_reason:_immediate_connect_error:_No_such_file_or_directory|remot
e_address:/etc/ai-gateway-extproc-uds/run.sock
```

*Repro steps*:
- Continuously sending requests to envoy ai gateway 
- Upgrade envoy ai gateway helm chart
- Observed 500 errors during upgrade

mathetake:
**Description**

This adds an end-to-end upgrade testing where we are supposed to test two test scenarios: simply rolling upgrade and control plane upgrade while keep making requests and verify no requests are dropped. What we found is that, as reported in #1241, there's a slight gap between Envoy stop receiving requests and extproc termination, hence users might experience requests being dropped during upgrade.

The fundamental fix is to set extproc sidecar container in the k8s API sense, but it's only available after k8s v1.33 by default. So, this adds a common workaround to sleep before the context cancelation. The workaround fix is verified to work in the newly added e2e upgrade tests.

After this is merged, adding k8s-version detection and enabling sidecar by default automatically as well as backporting the fix to v0.3 would be necessary. After that, we can enable the control-plane upgrade variant of the test case that is currently disabled in this commit.

Closes #1241
Closes #1060

### DIFF
--- a/tests/e2e-upgrade/upgrade_test.go
+++ b/tests/e2e-upgrade/upgrade_test.go
@@ -83,14 +83,14 @@ func TestUpgrade(t *testing.T) {
 			initFunc: func(ctx context.Context) string {
 				const previousEnvoyAIGatewayVersion = "v0.3.0"
 				const kindClusterName = "envoy-ai-gateway-cp-upgrade"
-				require.NoError(t, e2elib.SetupAll(t.Context(), kindClusterName, e2elib.AIGatewayHelmOption{
+				require.NoError(t, e2elib.SetupAll(ctx, kindClusterName, e2elib.AIGatewayHelmOption{
 					ChartVersion: previousEnvoyAIGatewayVersion,
 				}, false, false))
 				return kindClusterName
 			},
 			runningAfterUpgrade: 2 * time.Minute, // Control plane pods take longer to restart and roll out new Envoy pods.
 			upgradeFunc: func(ctx context.Context) {
-				require.NoError(t, e2elib.InstallOrUpgradeAIGateway(t.Context(), e2elib.AIGatewayHelmOption{}))
+				require.NoError(t, e2elib.InstallOrUpgradeAIGateway(ctx, e2elib.AIGatewayHelmOption{}))
 			},
 		},
 	} {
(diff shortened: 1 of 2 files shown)


mathetake:
OK great, got to reproduce the failure

tests/e2e-upgrade/upgrade_test.go
```diff
@@ -0,0 +1,298 @@
...
+				cmd := e2elib.Kubectl(ctx, "patch", "deployment", "-n", e2elib.EnvoyGatewayNamespace,
+					deploymentName, "--type=json", "-p",
+					`[{"op":"add","path":"/spec/template/metadata/annotations/upgrade-timestamp","value":"`+uuid.NewString()+`"}]`)
+				require.NoError(t, cmd.Run(), "failed to patch deployment")
+			},
+		},
+		{
+			name: "control-plane upgrade",
+			// TODO: Enable after the zero-downtime upgrade fix is in 0.3.x.
+			skip: true,
```
mathetake, line 82: this will be enabled after we backport the fix

yuzisun (approved):

PR #1251, opened 2025-09-30
feat: increase timeout for inferencepool extproc

Xunzhuo:
**Description**

This PR increased timeout from 5s for inferencepool extproc, preventing timeout from LLM

### DIFF
--- a/internal/extensionserver/inferencepool.go
+++ b/internal/extensionserver/inferencepool.go
@@ -310,7 +310,7 @@ func buildHTTPFilterForInferencePool(pool *gwaiev1a2.InferencePool) *extprocv3.E
 			ResponseTrailerMode: extprocv3.ProcessingMode_SEND,
 		},
 		AllowModeOverride: allowModeOverride,
-		MessageTimeout:    durationpb.New(5 * time.Second),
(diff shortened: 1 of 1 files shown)


internal/extensionserver/inferencepool.go
```diff
@@ -310,7 +310,7 @@ func buildHTTPFilterForInferencePool(pool *gwaiev1a2.InferencePool) *extprocv3.E
 			ResponseTrailerMode: extprocv3.ProcessingMode_SEND,
 		},
 		AllowModeOverride: allowModeOverride,
-		MessageTimeout:    durationpb.New(5 * time.Second),
+		MessageTimeout:    durationpb.New(300 * time.Second),
```
zirain, line 313: what about make it configurable instead of using a const value?

Xunzhuo: it is not easier to configure for inference pool, since all configuarable params passed from metadata added from cluster/route level hook, but listener level extproc injected in post translate. 

Let us get this default in, and configure out more general ways to config.

missBerg (approved):

Xunzhuo:
tracked in https://github.com/envoyproxy/ai-gateway/issues/1252

PR #1255, opened 2025-10-01
ci: removes http forward race in e2e upgrade test by sharing one

codefromthecrypt:
**Description**

Hopefully helps with `TestUpgrade` flakiness

### DIFF
--- a/tests/e2e-upgrade/upgrade_test.go
+++ b/tests/e2e-upgrade/upgrade_test.go
@@ -56,7 +56,7 @@ func (tp testPhase) String() string {
 //     to the latest version. This simulates a scenario where both the control plane
 //     and data plane are being updated.
 //
-// In both test cases,g we continuously make requests to the Envoy Gateway and ensure
+// In both test cases, we continuously make requests to the Envoy Gateway and ensure
 // that no requests fail during the upgrade process.
 func TestUpgrade(t *testing.T) {
 	for _, tc := range []struct {


codefromthecrypt:
cc @mathetake 🤞

codefromthecrypt:
will iterate more as it didn't deflake quite yet

mathetake:
Maybe the original PR failed to remove all root causes..

mathetake:
Thank you for taking a look!!

codefromthecrypt:
Thor's hammer on this one ;)

codefromthecrypt:
GitHub server error

tests/e2e-upgrade/upgrade_test.go
```diff
@@ -36,7 +56,7 @@ const egSelector = "gateway.envoyproxy.io/owning-gateway-name=upgrade-test"
 //     to the latest version. This simulates a scenario where both the control plane
 //     and data plane are being updated.
 //
-// In both test cases, we continuously make requests to the Envoy Gateway and ensure
+// In both test cases,g we continuously make requests to the Envoy Gateway and ensure
```
mathetake: 😄 
```suggestion
// In both test cases, we continuously make requests to the Envoy Gateway and ensure
```

tests/internal/e2elib/e2elib.go
```diff
@@ -570,76 +574,293 @@ func RequireWaitForPodReady(t *testing.T, selector string) {
...
+		resp, doErr = http.DefaultClient.Do(req)
+		isStale := isStaleConnectionError(doErr)
+		if doErr == nil {
+			respBody, err = io.ReadAll(resp.Body)
+			_ = resp.Body.Close()
+			if err != nil {
+				return nil, fmt.Errorf("failed to read response body: %w", err)
+			}
+			// Empty 500 indicates broken upstream (stale forward during rollout).
+			if resp.StatusCode == http.StatusInternalServerError && len(respBody) == 0 {
```
mathetake, line 808: on second though, so i think this might be overkilling as the whole point of this test is not to drop requests, hence no 500 from Envoy at all during the upgrade... 

That's why I think i chose to create a forward inside the loop to test that once the connection established then do not drop the request. But on the other hand, it still had a slight gap if:  port-forward established->rollout happening hence the pod marked terminated -> then try making requests... Having said that though, it still should result in connection error rather than 500 returned by envoy, so I think we should drop `resp.StatusCode == http.StatusInternalServerError` and if that causes the flake, then it's not a flake but more about a graceful termination bug as discussed in the original #1248. Not 100% sure though...


let me verify that the upgrade test still fails without the fix in #1248

mathetake: yeah this is overkill as I have observed that the test passes even without the fix in #1248

```
diff --git a/cmd/extproc/main.go b/cmd/extproc/main.go
index b81b4266..b97053c2 100644
--- a/cmd/extproc/main.go
+++ b/cmd/extproc/main.go
@@ -11,7 +11,6 @@ import (
        "os"
        "os/signal"
        "syscall"
-       "time"
 
        "github.com/envoyproxy/ai-gateway/cmd/extproc/mainlib"
 )
@@ -34,7 +33,7 @@ func main() {
                // This can be removed after the floor of supported k8s versions is larger than 1.32.
                //
                // 15s should be enough to propagate the readiness info to the load balancer for most cases.
-               time.Sleep(15 * time.Second)
+               //time.Sleep(15 * time.Second)
                log.Printf("shutting down the server now")
                cancel()
        }()
```

mathetake: i got a fix

mathetake: https://github.com/envoyproxy/ai-gateway/pull/1259

mathetake:
BTW

```
--- FAIL: TestPortForwarder_ConcurrentRestarts (0.02s)
    forwarder_test.go:274: 
        	Error Trace:	/Users/runner/work/ai-gateway/ai-gateway/tests/internal/e2elib/forwarder_test.go:274
        	Error:      	Not equal: 
        	            	expected: 2
        	            	actual  : 4
        	Test:       	TestPortForwarder_ConcurrentRestarts
        	Messages:   	expected initial start plus one coordinated restart
```

This is flaky

PR #1259, opened 2025-10-01
test: use metallb in upgrade test instead of port-forward

mathetake:
**Description**

This is a follow up on #1248 and #1255. This switches to use metallb instead of port-forwarding multiple times to avoid the complexity/race as well as to make the situation closer to the real world scenario where type: LoadBalancer Service is utilized to access the gateway rather than directly accessing pod ips via port-forward.


mathetake:
OK at least there's no flake. Next is to try reverting the fix in #1248 and see if the test is actually failing
image

mathetake:
cool reliably failing
image

mathetake:
cool reverting it back passed all tests again!

PR #1266, opened 2025-10-02
cli: removes "aigw translate" command

mathetake:
**Description**

`aigw translate` was introduced at the early stage of the project where the control plane was simple. At the time, it simply translates the CRD into EG's corresponding resources and that was it. However, as the project has evolved since then, now the majority of the work has been done in the extension server, meaning that the mere translation result from our CRD into EG resource is almost useless. It's still useful for the developers/contributors on this project internally as well as the underlying translation function is used in `aigw run`, so the implementation is still necessary but we should remove it from `aigw` CLI as it's confusing to the users who usually are not interested in such implementation details.


codefromthecrypt (approved):
sgtm!

cmd/aigw/main.go
```diff
@@ -77,7 +68,6 @@ func main() {
 //   - tf is the function to call to translate the AI Gateway resources to Envoy Gateway and Kubernetes resources. Mainly for testing.
```
aabchoo:
```suggestion
```

aabchoo (approved):
Just need to remove that one line!

cmd/aigw/main.go
mathetake: good catch man

PR #1268, opened 2025-10-03
feat: add AzureApiKey auth and auto-configure CLI

codefromthecrypt:
**Description**

This implements native Azure OpenAI API key authentication, matching practice suggested in their consoles, and OpenAI SDK norms. This is in addition to the OAuth which was there before. Specifically, the `AzureAPIKey` policy type injects the "api-key" header.

This allows us to auto-configure the `aigw` CLI with environment variables for azure similar to how the official OpenAI SDK bundles in ENV autoconfiguration both for OpenAI compatible and also the Azure dialect. This allows us to be a drop-in replacement for python SDK based proxies or client-side configuration.

https://github.com/openai/openai-python/blob/main/src/openai/lib/azure.py

I was motivated to do this due to an Elastic conference next week. Elastic sites tend to use Azure with ApiKey auth, as reflected in their canonical demo: https://github.com/elastic/elasticsearch-labs/blob/main/example-apps/chatbot-rag-app/env.example#L27-L36

** CLI ENV based auto-configuration ** 

`aigw run` uses the same ENV as the OpenAI SDK, with a heuristic to prioritizing Azure when present to decide between dialects:

For example, if "AZURE_OPENAI_API_KEY" is set, "AZURE_OPENAI_ENDPOINT" is read
Otherwise, normal "OPENAI_API_KEY" and "OPENAI_ENDPOINT" are.

Since both normal OpenAI as well as Azure OpenAI have headers for billing and access control, I also added them to auto-configuration, just using normal header manipulation: "OPENAI_ORG_ID" "OPENAI_PROJECT_ID"

** Example usage **

Example `aigw run` where you can see the routing works as the upstream is using MS conventions including the api key query,
```bash
$ AZURE_OPENAI_ENDPOINT=https://adria-mg8sjkj6-eastus2.cognitiveservices.azure.com/ AZURE_OPENAI_API_KEY=NOT_TELLIN OPENAI_API_VERSION=2024-12-01-preview ./aigw run
1.35.3 is already downloaded
starting: /tmp/envoy-gateway/versions/1.35.3/bin/envoy in run directory /tmp/envoy-gateway/runs/1759449568156171000
[2025-10-03 07:59:28.269][62295512][warning][config] [source/server/options_impl_platform_default.cc:9] CPU number provided by HW thread count (instead of cpuset).
{"bytes_received":120,"bytes_sent":35,"connection_termination_details":null,"downstream_local_address":"127.0.0.1:1975","downstream_remote_address":"127.0.0.1:54348","duration":1,"genai_backend_name":null,"genai_model_name":null,"genai_model_name_override":null,"genai_tokens_input":null,"genai_tokens_output":null,"method":"POST","response_code":404,"start_time":"2025-10-03T00:00:36.667Z","upstream_cluster":null,"upstream_host":null,"upstream_local_address":null,"upstream_transport_failure_reason":null,"user-agent":"curl/8.7.1","x-envoy-origin-path":"/chat/completions","x-request-id":"3c95a3d0-e784-48e8-9e30-28bd532aaaf3"}
{"bytes_received":120,"bytes_sent":665,"connection_termination_details":null,"downstream_local_address":"127.0.0.1:1975","downstream_remote_address":"127.0.0.1:54351","duration":2503,"genai_backend_name":"default/openai/route/aigw-run/rule/0/ref/0","genai_model_name":"gpt-5-nano","genai_model_name_override":"gpt-5-nano","genai_tokens_input":12,"genai_tokens_output":204,"method":"POST","response_code":200,"start_time":"2025-10-03T00:00:45.218Z","upstream_cluster":"httproute/default/aigw-run/rule/0","upstream_host":"20.62.58.5:443","upstream_local_address":"192.168.0.108:54352","upstream_transport_failure_reason":null,"user-agent":"curl/8.7.1","x-envoy-origin-path":"/openai/deployments/gpt-5-nano/chat/completions?api-version=2024-12-01-preview","x-request-id":"a5dceec3-6af8-4019-830b-1eff286370c0"}
{"bytes_received":120,"bytes_sent":683,"connection_termination_details":null,"downstream_local_address":"127.0.0.1:1975","downstream_remote_address":"127.0.0.1:54355","duration":2050,"genai_backend_name":"default/openai/route/aigw-run/rule/0/ref/0","genai_model_name":"gpt-5-nano","genai_model_name_override":"gpt-5-nano","genai_tokens_input":12,"genai_tokens_output":145,"method":"POST","response_code":200,"start_time":"2025-10-03T00:01:19.564Z","upstream_cluster":"httproute/default/aigw-run/rule/0","upstream_host":"20.62.58.5:443","upstream_local_address":"192.168.0.108:54356","upstream_transport_failure_reason":null,"user-agent":"curl/8.7.1","x-envoy-origin-path":"/openai/deployments/gpt-5-nano/chat/completions?api-version=2024-12-01-preview","x-request-id":"8cfc5abd-47a7-498c-a64a-4a3d486b75c7"}
```

Sample request where you can see the microsoft headers in the reply. My deployment happens to be named the same as the model.
```bash
$ curl -v http://localhost:1975/v1/chat/completions \
>   -H "Authorization: Bearer fake" \
>   -H "Content-Type: application/json" \
>   -d '{
>     "model": "gpt-5-nano",
>     "messages": [
>       {"role": "user", "content": "Say hello in one sentence."}
>     ]
>   }'
* Host localhost:1975 was resolved.
* IPv6: ::1
* IPv4: 127.0.0.1
*   Trying [::1]:1975...
* connect to ::1 port 1975 from ::1 port 54354 failed: Connection refused
*   Trying 127.0.0.1:1975...
* Connected to localhost (127.0.0.1) port 1975
> POST /v1/chat/completions HTTP/1.1
> Host: localhost:1975
> User-Agent: curl/8.7.1
> Accept: */*
> Authorization: Bearer fake
> Content-Type: application/json
> Content-Length: 120
> 
* upload completely sent off: 120 bytes
< HTTP/1.1 200 OK
< content-length: 683
< content-type: application/json
< apim-request-id: e662d601-4dd2-488d-b04e-6a06b4933dd1
< strict-transport-security: max-age=31536000; includeSubDomains; preload
< x-content-type-options: nosniff
< x-ms-region: East US 2
< x-ratelimit-remaining-requests: 98
< x-ratelimit-limit-requests: 100
< x-ratelimit-remaining-tokens: 99986
< x-ratelimit-limit-tokens: 100000
< azureml-model-session: d022-20250930175838
< x-accel-buffering: no
< x-ms-rai-invoked: true
< x-request-id: 2a886a75-cd6b-4f30-adb1-2d61ba49eb67
< x-ms-deployment-name: gpt-5-nano
< date: Fri, 03 Oct 2025 00:01:20 GMT
< 
{"choices":[{"content_filter_results":{},"finish_reason":"stop","index":0,"logprobs":null,"message":{"annotations":[],"content":"Hello, nice to meet you!","refusal":null,"role":"assistant"}}],"created":1759449680,"id":"chatcmpl-CMNWaRPY95oUgzSR5uk94rOydUTFC","model":"gpt-5-nano-2025-08-07","object":"chat.completion","prompt_filter_results":[{"prompt_index":0,"content_filter_results":{}}],"system_fingerprint":null,"usage":{"completion_tokens":145,"completion_tokens_details":{"accepted_prediction_tokens":0,"audio_tokens":0,"reasoning_tokens":128,"rejected_prediction_tokens":0},"prompt_tokens":12,"prompt_tokens_details":{"audio_tokens":0,"cached_tokens":0},"total_tokens":157}}
* Connection #0 to host localhost left intact
```

Testing took some thinking to avoid a lot of configuration. The approach taken besides unit tests was to backfill one azure request ("azure-chat-basic") into the VCR package, mainly to ensure the translation occurs. Technically, we send the router request (normal chat completions "chat-basic") with a "X-Cassette-Name: azure-chat-basic". This ensures the azure upstream/backend is used (testopenai). When an azure request is received it is processed per azure rules, even if the body is basically identical from the normal OpenAI services. There were some glitches as some serialization policy created empty arrays, and I handled that.


api/v1alpha1/backendsecurity_policy.go
```diff
@@ -71,6 +73,11 @@ type BackendSecurityPolicySpec struct {
 	// +optional
 	AWSCredentials *BackendSecurityPolicyAWSCredentials `json:"awsCredentials,omitempty"`
 
+	// AzureAPIKey is a mechanism to access Azure OpenAI backend(s). The API key will be injected into the api-key header.
+	//
+	// +optional
+	AzureAPIKey *BackendSecurityPolicyAzureAPIKey `json:"azureAPIKey,omitempty"`
```
mathetake, line 79: what's your take on instead of having this, adding a configuration knob in the existing `BackendSecurityPolicyAPIKey` to specify the header key that defaults to "authorization" ? I think "bearer: " portion makes it tricky though...

codefromthecrypt: not a fan tbh. I'd rather switch to ad hoc mutation than clutter unrelated auth approaches. the thing is the other one is used for oauth and this has nothing to do with oauth.

codefromthecrypt:
sorry messed up the vcr tests.. lemme

api/v1alpha1/backendsecurity_policy.go
mathetake: oh i meant here https://github.com/envoyproxy/ai-gateway/blob/bb2a476eeb7fe14495a64ced1396c65a07ea2cd3/api/v1alpha1/backendsecurity_policy.go#L94 but clunky anyways i agree

codefromthecrypt: so I think if we wanted to do a better job with docs, we could and I can help. My guess is this was called ApiKey because maybe the first use was OpenAI e.g. `OPENAI_API_KEY`. In their docs (and some others do the same thing) say that the bearer token is where to put the "api key"

In the case of OpenAI, they literally name this azure api key and it corresponds to `AZURE_OPENAI_API_KEY`. Since it is both a different header and not substitutable as a bearer token, I think we have a choice to unsupport it (make it ad-hoc) or support it. If we do, I think this naming is the most intuitive.

In their [code](https://github.com/openai/openai-python/blob/main/src/openai/lib/azure.py#L192-L193) (in the azure.py which is implicitly azure) they do call it `api_key`, just externally you need to know this is azure.

That's my thinking anyway.

mathetake (approved):
Looks good

PR #1273, opened 2025-10-03
test: enables cp upgrade test case

mathetake:
**Description**

This is a follow up on #1248, and enables the skipped test case by using the old commit hash. We will pin it later to the next version which is v0.4.0..


mathetake:
nice picture aaron

PR #1274, opened 2025-10-03
refactor: deprecates unneeded fields from filter api

mathetake:
**Description**

This deprecates two unnecessary fields from the filter api

* metadataNamespace
* modelNameHeaderKey

They exited due to the very early stage's vision to make the AI Gateway extproc workable independent of the control plane hence the customizability, but that is not our goal anymore.

We cannot just drop them as the older version expects these two fields, which results in 404 responses during the upgrade.


mathetake:
oh right just as expected this will create an instant 404 during the control plane upgrade (So glad the test is working as intended anyways!). We need to apply a patch to default to the constants before dropping the field.

internal/controller/gateway.go
```diff
@@ -181,7 +181,9 @@ func (c *GatewayController) reconcileFilterConfigSecret(
 ) error {
 	// Precondition: aiGatewayRoutes is not empty as we early return if it is empty.
 	ec := &filterapi.Config{UUID: uuid}
+	// TODO: Drop this after v0.4.0.
 	ec.ModelNameHeaderKey = internalapi.ModelNameHeaderKeyDefault
+	ec.MetadataNamespace = aigv1a1.AIGatewayFilterMetadataNamespace
```
mathetake, line 186: note: this makes it backward compatible

codefromthecrypt (approved):
mo fields mo problems

thanks!

internal/controller/gateway.go
mathetake: and it is ensured in the upgrade test case added in #1273

internal/filterapi/filterconfig.go
```diff
@@ -23,22 +22,22 @@ import (
 
 // DefaultConfig is the default configuration that can be used as a
 // fallback when the configuration is not explicitly provided.
-var DefaultConfig = fmt.Sprintf(`
-schema:
-  name: OpenAI
-modelNameHeaderKey: %s
-`, internalapi.ModelNameHeaderKeyDefault)
+var DefaultConfig = ``
```
aabchoo, line 25: Why remove schema?

mathetake: it's deprecated way before #1013 and this was remaining piece

missBerg (approved):

aabchoo (approved):

PR #1275, opened 2025-10-04
feat: mcp cleanups and CLI can now do MCP+LLM at once

codefromthecrypt:
**Description**

Before, the MCP configuration argument to the standalone mode was not compatible with existing OpenAI ENV vars. This fixes it, so that you can have a single agent origin without yaml. 

For example:
```
OPENAI_API_KEY=sk-xxxx aigw run --mcp-json '{"mcpServers":{"kiwi":{"type":"http","url":"https://mcp.kiwi.com"}}}'
```

**Details**

**MCP config generator bug**

maybeGenerateResourcesForMCPGateway unconditionally added resources for MCP which caused configuration give-ups. fixed and backfilled tests

**docker mcp examples**

wrote examples in the same style as chat, embeddings etc into `cmd/aigw/docker-compose*.yaml` so that they are easy to invoke.

**Envoy Gateway techdebt workarounds**

- hard-codes a shared directory for envoy downloads, so changed CI to write to /tmp/envoy-gateway
- hard-codes a cert directory as well, which causes other tests to fail when one writes certs to it. deleting that on startup
- hard-codes os.Stdout for envoy when launched in standalone mode, so disable debug even though we could capture otherwise and only report on fail.
- has broken CLI args passed to envoy which makes admin server detection fail, fall back to poll the listen port instead

**Fix either/or with MCP config**

The main issue before this change was that you couldn't do both LLM and MCP in auto-config.

This PR refactors the MCP configuration system and introduces a unified `autoconfig` package that supports OpenAI-only, MCP-only, or combined OpenAI+MCP configurations from environment variables and JSON.

The new `autoconfig` package replaces the old `envgen` package with a more flexible architecture, generates complete AI Gateway YAML from environment variables and JSON configuration, and handles both OpenAI providers (OpenAI, Azure OpenAI) and MCP server configurations.

**naming consistency and organization**

- mcpServers "tools" -> "allowedTools" for internal consistency and also consistency with gemini
- `MCPNameSelector` → `MCPToolSelector` as there are other entities (tools, prompts, resources)
- `nameSelector` → `toolSelector` similar to above
- `cmd/aigw/*example*` -> `examples/mcp/` for better organization and discovery
- `test-e2e-mcp` -> `test-e2e-aigw` and similar like`tests/e2e-aigw`

**admin check hardening**

I also enhanced Envoy admin discovery in `internal/aigw/admin.go` with a new `EnvoyAdminClient` interface for checking Envoy readiness, automatic discovery of Envoy admin port from subprocess arguments using `gopsutil`. However, due to tech debt in Envoy Gateway which was fixed but not released, we need a fallback to a listener port check.

**MCP code issues**

- bug where we wrote MCP resources even when there were none caused blocking loop on Envoy waiting for them
- backfilled tests to ensure test-coverage would pass

**aigw run test deflake**
- rewrote to share shutdown code, and work around cert state leftovers due to hard-coding in envoy gateway.

**ollama hardening**

- changed CI to keep a log which could explain if 403 in tests have anything to do with ollama
- refactored CI to use a larger context size as default is known to not be big enough for goose

**goose deflakes**

- larger model as `qwen3:0.6b` is known to not be viable for agents
- rewrote the examples and goose tests to be clear what happened on error give up

**CLI debug propagation**

- `aigw --debug` now propagates to envoy debug which makes it possible to debug many problems
- due to envoy-gateway not propagating stdout/stderr from cobra to func-e, this cannot be left on in tests

**envoy CI downloads in unit tests**

- some tests flake because they are downloading envoy, so added steps to do that in CI. this uses the hard-coded directory in envoy gateway.

**documentation backfill**

I also noticed a lack of documentation, so I spent a lot of time on that, too
- `site/docs/capabilities/mcp/index.md` - New MCP guide
- New `site/docs/capabilities/mcp/index.md` explaining MCP integration in detail
- `site/docs/cli/run.md` - MCP configuration docs
- Updated `site/docs/cli/run.md` with new `--mcp-json` flag documentation
- `cmd/aigw/README.md` - Add MCP quick start
- `cmd/aigw/docker-compose*.yaml` - Add MCP configuration

**test fixture consolidation**

several tools consolidated into internal/testing notably `RequireEventuallyNoError` needed to know what the last error is on failure


codefromthecrypt:
I removed the translation test from examples/mcp as that's covered in in the autoconfig package, e.g. internal/autoconfig/testdata/openai-context7-github.yaml

codefromthecrypt:
TestMCP_standalone_oauth is flakey and probably needs to be refactored in a later PR

cmd/aigw/README.md
```diff
@@ -63,8 +69,8 @@ Here are values we use for Ollama:
 [docker-compose-otel.yaml](docker-compose-otel.yaml) includes OpenTelemetry
 metrics and tracing.
 
-All profiles below use at least these two Docker services:
-- **aigw** (port 1975): Envoy AI Gateway CLI (standalone mode) with OTEL tracing
+All profiles below use at least these Docker services:
+- **aigw** (port 1975): Envoy AI Gateway CLI with OpenAI endpoints at `/v1/*` and MCP endpoint at `/mcp`
```
mathetake, line 73: nice

internal/filterapi/mcpconfig.go
```diff
@@ -39,19 +39,20 @@ type MCPBackend struct {
...
-	// Include is a list of strings to include. If specified, only the strings in this list are included.
+// MCPToolSelector filters tools using include patterns with exact matches or regular expressions.
+type MCPToolSelector struct {
+	// Include is a list of tool names to include. Only the specified tools will be available.
 	Include []string `json:"include,omitempty"`
 
-	// IncludeRegex is a list of RE2-compatible regular expressions that, when matched, include the string.
+	// IncludeRegex is a list of RE2-compatible regular expressions that, when matched, include the tool.
+	// Only tools matching these patterns will be available.
+	// TODO: regex is almost completely absent in the MCP ecosystem, consider removing this for simplicity.
```
mathetake, line 55: +1

codefromthecrypt:
Ps after this change I want to rename /docs maybe to /designs as I think it is more intuitive and helps people look elsewhere for usage docs

codefromthecrypt:
Excellent notes on the CommandContext and t.Context ordering issue @mathetake ! Very instructive

codefromthecrypt:
taking some time to refactor admin detection hopefully it helps with the example tests, that or we can move that package out of the `make test-coverage` package...

codefromthecrypt:
fingers crossed the refactored admin ready code improves things (also added a secondary check for open ports since in some cases the admin is ephemeral, so we should check both fixed envoy admin port and also the fixed gateway port)

codefromthecrypt:
got it I think.. the examples test was getting into `test-coverage` which runs -race.

We already learned not to race integration tests. I moved it to the e2e for mcp with goose

codefromthecrypt:
Still one flake left. Will report back in a few hours

codefromthecrypt:
the admin detection will fail still because even though an uberbug fix is merged it wasn't released, so we'll rely on the fallback logic until this is not the case https://github.com/envoyproxy/gateway/pull/7006

cmd/aigw/run_test.go
```diff
@@ -59,15 +63,16 @@ func TestRun(t *testing.T) {
...
 	defer func() { cancel(); <-done }()
 
+	client := openai.NewClient(option.WithBaseURL("http://localhost:1975/v1/"))
+	chatReq := openai.ChatCompletionNewParams{
+		Messages: []openai.ChatCompletionMessageParamUnion{
+			openai.UserMessage("Say this is a test"),
+		},
+		Model: ollamaModel,
+	}
 	t.Run("chat completion", func(t *testing.T) {
```
codefromthecrypt, line 81: @mathetake for me, not sure about CI, but I routinely cannot get this to pass on main or this branch as the config always ends up as a 503 error. In docker it is fine 🤷

codefromthecrypt: e.g. If I build and run it is fine.. I noticed in this PR the MCP one is flaking, but in my laptop that one passes, but the non MCP one (chat completion) flakes (same here as main)!

```bash
$ OPENAI_BASE_URL=http://localhost:11434/v1 OPENAI_API_KEY=unused out/aigw-darwin-arm64 run                                                                                        
1.35.3 is already downloaded
starting: /tmp/envoy-gateway/versions/1.35.3/bin/envoy in run directory /tmp/envoy-gateway/runs/1759660125781489000
[2025-10-05 18:28:45.797][67801183][warning][config] [source/server/options_impl_platform_default.cc:9] CPU number provided by HW thread count (instead of cpuset).
{"bytes_received":120,"bytes_sent":97,"connection_termination_details":null,"downstream_local_address":"127.0.0.1:1975","downstream_remote_address":"127.0.0.1:56480","duration":6,"genai_backend_name":null,"genai_model_name":"gpt-5-nano","genai_model_name_override":null,"genai_tokens_input":null,"genai_tokens_output":null,"method":"POST","response_code":404,"start_time":"2025-10-05T10:29:39.079Z","upstream_cluster":"httproute/default/aigw-run/rule/0","upstream_host":"127.0.0.1:11434","upstream_local_address":"127.0.0.1:56481","upstream_transport_failure_reason":null,"user-agent":"curl/8.7.1","x-envoy-origin-path":"/v1/chat/completions","x-request-id":"f0da5c53-92dd-4587-9f67-0cbb337f4490"}
{"bytes_received":122,"bytes_sent":317,"connection_termination_details":null,"downstream_local_address":"127.0.0.1:1975","downstream_remote_address":"127.0.0.1:56485","duration":835,"genai_backend_name":"default/openai/route/aigw-run/rule/0/ref/0","genai_model_name":"qwen2.5:0.5b","genai_model_name_override":"qwen2.5:0.5b","genai_tokens_input":35,"genai_tokens_output":10,"method":"POST","response_code":200,"start_time":"2025-10-05T10:30:02.006Z","upstream_cluster":"httproute/default/aigw-run/rule/0","upstream_host":"127.0.0.1:11434","upstream_local_address":"127.0.0.1:56486","upstream_transport_failure_reason":null,"user-agent":"curl/8.7.1","x-envoy-origin-path":"/v1/chat/completions","x-request-id":"886a5b10-d30b-49fb-ae2e-a80b58f8d813"}
```

codefromthecrypt:
latest flake was once again due to integration tests (such as the ones flaking) not being in the e2e. 

```
2025-10-05T10:28:02.920Z	ERROR	infrastructure	host/proxy_infra.go:103	failed to run envoy	{"runner": "infrastructure", "error": "error untarring https://archive.tetratelabs.io/envoy/download/v1.35.3/envoy-v1.35.3-darwin-arm64.tar.xz: context canceled"}
--- FAIL: TestRunMCP (77.37s)
```

We really should scrub hard and in no case run anything that launches envoy except in a `make ...e2e...` mode not just to avoid racing, but also to avoid the downloading of envoy issue since the e2e workflow caches it up front.

codefromthecrypt:
I updated standalone mode to propagate the debug flag to the autoconfig, so that envoy firehose of debug happens.

The root problem was that MCP configuration was being added unconditionally, so even if you had no MCP servers at all, and glitching the test. I fixed that and backfilled the unit test.

codefromthecrypt:
ok the reason I lost another day was because envoy gateway hard-codes its certs directory and you have to manually clean it up or else another test will fall over it. I added that to the list of tech debt in the PR desc

codefromthecrypt:
I backfilled coverage gaps in the MCP code which were failing on test-coverage 🤞

codefromthecrypt:
raised https://github.com/envoyproxy/gateway/pull/7142 about the debug log glitch in EG

cmd/aigw/docker-compose.yaml
```diff
@@ -99,3 +114,24 @@ services:
...
+      - --method
+      - tools/call
+      - --tool-name
+      - kiwi__search-flight
+      - --tool-arg
+      - flyFrom=NYC
+      - --tool-arg
+      - flyTo=LAX
+      - --tool-arg
+      - departureDate=15/12/2025
```
nacx, line 137: [nit] the Kiwi flight search tool fails when dates in the past are passed to it (and probably a date far away in the future will produce no results). Should we use a different MCP server, like the `dreamtap` you used before, where we know there will always be an answer? (same comment applies to the otel docker-compose and tests that are actually using it)

tests/extproc/extproc_test.go
```diff
@@ -134,15 +134,15 @@ func TestMain(m *testing.M) {
...
+func startTestEnvironment(t testing.TB, extprocConfig string, okToDumpLogOnFailure, extProcInProcess bool) *testenvironment.TestEnvironment {
 	return testenvironment.StartTestEnvironment(t,
 		requireUpstream, map[string]int{"upstream": 8080},
 		extprocBin, extprocConfig, nil, envoyConfig, okToDumpLogOnFailure, extProcInProcess, 120*time.Second,
 	)
 }
 
 // requireUpstream starts the external processor with the given configuration.
-func requireUpstream(t testenvironment.TestingT, out io.Writer, ports map[string]int) {
+func requireUpstream(t testing.TB, out io.Writer, ports map[string]int) {
```
mathetake, line 145: oh i didn't know that!

mathetake (approved):
I reviewed all change and confirmed that essentially any change here is about MCP added in a few commits ago. I like the unified configuration now in aigw run command that host both LLM and MCP altogether in a single endpoint.

Note: I think we still have a few places (CommandContext(t.Context()) not relevant in this PR so i think we can do the further clean up in a follow up!

cmd/aigw/docker-compose.yaml
codefromthecrypt: We can use a calendar to pick next week meanwhile

PR #1277, opened 2025-10-06
ci: format yaml with go-prettier and remove yamllint 

### LINKED ISSUE #1265: Migrate to yamllint alternative that can actually autofix
*Description*:

Pretty much what the title says. yamllint cannot do the autofix (and will never support), so I think it makes sense to find an alternative that can actually do the autofix.

https://github.com/adrienverge/yamllint/issues/62

anuraaga:
**Description**

Executes go-prettier to format yaml files in `make format` and removes yamllint.

Fixes #1265 

yamllint has many checks that aren't related to formatting and naturally not autofixable, one I see in this repo is key duplicates [1]. To have extremely great yaml files, it can be best to apply both a formatter and yamllint, but my interpretation of the issue is formatting is enough, so I removed yamllint.

#### Formatting differences

- Single space before comments instead of at least two spaces [2]. Prettier does not allow this to be configured, having the "minimal configuration for formatting" philosophy. If this is important, it is a blocker to using prettier. Note, people may have issues if they "generally prefer two spaces", but also to allow multiple lines to have aligned comments which might be more of a problem, looking through the PR you'll find some cases where they are now not aligned. FWIW, I believe this will apply to any autoformatting of yaml (or code) out there since they almost always lose whitespace information when parsing.
- Indentation is always applied. yamllint only checked for indentation consistency
- Quotes are always double. yamllint is not currently configured to check it. Prettier does have a configuration option for single quote, then quotes are always single. I found enough usages of both in this codebase to not know which to pick and stuck with the default. Let me know if single is better
- A soft line-length limit is applied, meaning when the meaning of the yaml wouldn't change because of a line-length break, it is broken. The default is 80 chars but this can be configured, let me know

One particular point to call out - k8s has it's own yaml formatting style that is different from what most people like, notably it doesn't indent lists. Prettier always indents lists - as this repo has both k8s yaml files and envoy, CI, etc yaml files, if it's a problem to format all files including k8s with indented lists, it's a blocker to using prettier. That being said, looking through this diff, I see many k8s files already formatted in "non-k8s" format, with indented lists.

#### About go-prettier [3]

Want to be upfront about go-prettier's architecture - it uses the standard prettier codebase but runs it with quickjs compiled to Wasm with wazero. quickjs is an interpreter without any code generation - it works pretty well for simple scripts but definitely has reduced performance vs Node, especially for highly recursive code like prettier's which is the bane of interpreters. I notice with this PR, checking takes 5s with go-prettier vs 1s for prettier. It may be worth consolidating on Go for a smallish difference - note for reference, before adding the missing `manifests/charts/ai-gateway-crds-helm` to ignore, there was a monster file being formatted that by itself took 15s, while 1s on Node.

If it seems an issue now, or later, then adding an assumption that Node is installed and replacing `go tool prettier` with `npx prettier` is a 1 line change, can do it now if there are any concerns with using quickjs.

BTW, this PR only applies to yaml, but it could be considered for Markdown if any interest. In that case, go-prettier has a unique feature of applying `go fmt` to any Go codeblocks in the Markdown.

1. https://github.com/envoyproxy/ai-gateway/blob/main/.yamllint#L44
2. https://github.com/envoyproxy/ai-gateway/blob/main/.yamllint#L26
3. https://github.com/wasilibs/go-prettier

/cc @mathetake


.prettierignore
```diff
@@ -0,0 +1,11 @@
...
+# SPDX-License-Identifier: Apache-2.0
+# The full text of the Apache license is available in the LICENSE file at
+# the root of the repo.
+
+manifests/charts/ai-gateway-helm
+manifests/charts/ai-gateway-crds-helm
+tests/internal/testopenai/cassettes
+tests/internal/testopeninference/spans
+cmd/aigw/testdata
+internal/envgen/testdata
```
anuraaga: I added this folder since I think it shouldn't be formatted for the same reason as aigw/testdata

PR #1281, opened 2025-10-06
mcp: apply tool filter in tool calls as well

nacx:
**Description**

Currently, the tool call handler doesn't check the tool selectors, so an agent can call any tool, even the ones that are not in the list tools result.

N/A

N/A

### DIFF
--- a/internal/mcpproxy/handlers_test.go
+++ b/internal/mcpproxy/handlers_test.go
@@ -47,7 +47,7 @@ func newTestMCPProxyWithTracer(t tracingapi.MCPTracer) *MCPProxy {
 			backendListenerAddr: "http://test-backend",
 			routes: map[filterapi.MCPRouteName]*mcpProxyConfigRoute{
 				"test-route": {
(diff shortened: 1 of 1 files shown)


internal/mcpproxy/handlers_test.go
```diff
@@ -47,6 +47,9 @@ func newTestMCPProxyWithTracer(t tracingapi.MCPTracer) *MCPProxy {
 			backendListenerAddr: "http://test-backend",
 			routes: map[filterapi.MCPRouteName]*mcpProxyConfigRoute{
 				"test-route": {
+					toolSelectors: map[filterapi.MCPBackendName]*nameSelector{
```
mathetake: drift
```suggestion
					toolSelectors: map[filterapi.MCPBackendName]*toolSelector{
```

nacx: yes, I have to rebase after the last merge, will do in a bit

mathetake (approved):

PR #1284, opened 2025-10-06
mcp: dorps the tight 16 backendRefs on MCPRoute

mathetake:
**Description**
wip


mathetake:
cc @nacx @codefromthecrypt

api/v1alpha1/mcp_route.go
```diff
@@ -77,7 +77,7 @@ type MCPRouteSpec struct {
 	//
 	// +kubebuilder:validation:Required
 	// +kubebuilder:validation:MinItems=1
-	// +kubebuilder:validation:MaxItems=32
+	// +kubebuilder:validation:MaxItems=256
```
mathetake, line 80: 256 * 200 bytes per backend = 0.5mb < 2mb k8s limit and i think it should work any use in practice given Goose recommends 25 tools etc

mathetake: btw we cannot drop this altogether as it will be rejected by k8s

codefromthecrypt (approved):
Good, as I fell into this yesterday!!

does this affect autoconfig generation in any way? if so, please update so it is not creating invalid config layout.. I guess it is not because we have a couples tests that uses auto-config mcp..

mathetake:
> does this affect autoconfig generation in any way? 

it won't affect autoconfig, so no worries!

api/v1alpha1/mcp_route.go
codefromthecrypt: yeah the ideal would be a pre-flight validation that it is under size (not under a count), but ack that's hard to implement especially as there are no size based constraints we could use.

codefromthecrypt (approved):

missBerg (approved):

examples/mcp/README.md
```diff
@@ -237,13 +237,13 @@ Once you have everything running you can start the agent by passing a prompt fil
...
 ```shell
-$ uv run --exact -q --env-file <environment file>> agent.py /path/to/prompt.txt
+$ uv run --exact -q --env-file .env agent.py /path/to/prompt.txt
 ```
 
 or
 
 ```shell
-$ uv run --exact -q --env-file <environment file>> agent.py <<EOF
+$ uv run --exact -q --env-file .env agent.py << EOF
```
mathetake, line 246: this was a drift in #1290 being merged while not updated to the latest main. I can do this in another PR but i think it is harmless for this specifically

tests/e2e/mcp_route_test.go
```diff
@@ -36,6 +36,7 @@ func TestMCP(t *testing.T) {
 	t.Cleanup(func() {
 		_ = e2elib.KubectlDeleteManifest(t.Context(), manifest)
 	})
+	maniyBackendsRouteToolNames := requireCreateMCPManyBackends(t)
```
nacx, line 39: [typo nit] `manyBackendsRouteToolNames`

PR #1288, opened 2025-10-07
fix: make mcp span names informative

codefromthecrypt:
**Description**

Before everything was vaguely "mcp.request" which is not untrue, but not helpful. There are plenty of low-cardinality choices for span or metric names and the `req.Method` is the most intuitive. This uses the same naming conventions as mcp-go

image

### DIFF
--- a/internal/mcpproxy/handlers.go
+++ b/internal/mcpproxy/handlers.go
@@ -234,6 +234,7 @@ func (m *MCPProxy) servePOST(w http.ResponseWriter, r *http.Request) {
 			}
 			err = m.handleSetLoggingLevel(ctx, s, w, msg, p, span)
 		case "ping":
+			// Ping is intentionally not traced as it's a lightweight health check.
 			err = m.handlePing(ctx, w, msg)
 		case "prompts/list":
 			p := &mcp.ListPromptsParams{}
--- a/internal/testing/testotel/collector.go
+++ b/internal/testing/testotel/collector.go
@@ -133,17 +133,6 @@ func (o *OTLPCollector) TakeSpan() *tracev1.Span {
 	}
 }
 
-// DrainSpans drains all pending spans from the channel.
-func (o *OTLPCollector) DrainSpans() {
-	for {
-		select {
-		case <-o.spanCh:
-			// Discard the span
-		case <-time.After(otlpTimeout):
-			return // No more spans available
-		}
-	}
-}
 
 // DrainMetrics returns metrics or nil if none were recorded.
 func (o *OTLPCollector) DrainMetrics() *metricsv1.ResourceMetrics {
--- a/internal/testing/testotel/collector.go
+++ b/internal/testing/testotel/collector.go
@@ -133,6 +133,18 @@ func (o *OTLPCollector) TakeSpan() *tracev1.Span {
 	}
 }
 
+// DrainSpans drains all pending spans from the channel.
+func (o *OTLPCollector) DrainSpans() {
+	for {
+		select {
+		case <-o.spanCh:
+			// Discard the span
+		case <-time.After(otlpTimeout):
+			return // No more spans available
+		}
+	}
+}
+
 // DrainMetrics returns metrics or nil if none were recorded.
 func (o *OTLPCollector) DrainMetrics() *metricsv1.ResourceMetrics {
 	select {
--- a/internal/testing/testotel/collector.go
+++ b/internal/testing/testotel/collector.go
@@ -133,7 +133,6 @@ func (o *OTLPCollector) TakeSpan() *tracev1.Span {
 	}
 }
 
-
 // DrainMetrics returns metrics or nil if none were recorded.
 func (o *OTLPCollector) DrainMetrics() *metricsv1.ResourceMetrics {
 	select {
--- a/tests/extproc/mcp/env.go
+++ b/tests/extproc/mcp/env.go
@@ -17,7 +17,6 @@ import (
 
 	"github.com/modelcontextprotocol/go-sdk/mcp"
 	"github.com/stretchr/testify/require"
-	v1 "go.opentelemetry.io/proto/otlp/trace/v1"
 
 	"github.com/envoyproxy/ai-gateway/internal/filterapi"
 	"github.com/envoyproxy/ai-gateway/internal/testing/testotel"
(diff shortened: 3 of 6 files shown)


codefromthecrypt:
Screenshot 2025-10-07 at 9 42 20 AM

codefromthecrypt:
changed to re-use MCP go naming conventions as that also matches the openinference names of these spans.

codefromthecrypt:
now envoy ai gateway generated spans look similar!

Screenshot 2025-10-07 at 11 24 54 AM

codefromthecrypt:
🤞 last commit deflakes

codefromthecrypt:
wasn't enough to deflake I am flying to NYC now. appreciate if someone can take over otherwise will wait a while

mathetake:
unified trace looking coooooooooooooooooooool

nacx (commented):
Nice!

codefromthecrypt:
I will rewrite the test so that not all tests also do tracing because it is too hairy ;)

codefromthecrypt:
ok brutal work, but hopefully deflaked

tests/extproc/mcp/env.go
```diff
@@ -241,16 +240,18 @@ func (m *mcpEnv) newSession(t *testing.T) *mcpSession {
...
+		"mcp.client.name":    "demo-http-client",
+		"mcp.client.title":   "",
+		"mcp.client.version": "0.1.0",
+	})
 
 	// **NOTE*** Do not add any direct access to the in-memory server session. Otherwise, the tests will result in
 	// not being able to run in end-to-end tests. The test code must solely operate through the client side sessions.
 	t.Cleanup(func() {
-		require.NoError(t, ret.session.Close())
+		_ = ret.session.Close()
```
codefromthecrypt, line 254: should be a linter for this somewhere. @anuraaga come across any that notice when accidental use of require.RequreXX n a cleanup? t.LogXX is fine, but require is bad

tests/extproc/mcp/mcp_test.go
```diff
@@ -31,76 +32,67 @@ const (
...
+	{name: "ResourceSubscribe", testFn: testResourceSubscribe},
+	{name: "ResourceListChangeNotifications", testFn: testResourceListChangeNotifications},
+	{name: "ListRootsAndChangeRoots", testFn: testListRootsAndChangeRoots},
+	{name: "SamplingCreateMessage", testFn: testSamplingCreateMessage},
+	{name: "Elicit", testFn: testElicit},
+	{name: "NotificationCancelled", testFn: testNotificationCancelled},
+	{name: "Complete", testFn: testComplete},
+}
+
 func TestMCP(t *testing.T) {
```
codefromthecrypt, line 64: I broke these up vs having triple nested subtests

tests/extproc/mcp/mcp_test.go
```diff
@@ -497,6 +545,8 @@ func testResourceListChangeNotifications(t *testing.T, m *mcpEnv) {
 		})
 		require.NoError(t, err)
 		require.False(t, res.IsError)
+		// Consume the span from the cleanup operation.
+		_ = m.collector.TakeSpan()
```
codefromthecrypt, line 551: this bit hard

tests/extproc/mcp/mcp_test.go
```diff
@@ -625,6 +694,10 @@ func testNotificationCancelled(t *testing.T, m *mcpEnv) {
 	select {
 	case <-time.After(time.Microsecond * 500):
 		cancel()
+		// Wait for the goroutine to complete so its span doesn't leak to the next test.
+		<-doneCh
```
codefromthecrypt, line 700: this also bit hard ;)

codefromthecrypt:
most of why this test is so hard is we share resources including otel everywhere in this file. It would probably be better to have two tests at some point. one that does all this behavioural stuff. The other simply tests each op creates the span expected. Combining both made this a huge amount of work to fix

internal/testing/testotel/collector.go
```diff
@@ -133,6 +133,18 @@ func (o *OTLPCollector) TakeSpan() *tracev1.Span {
...
+func (o *OTLPCollector) DrainSpans() {
+	for {
+		select {
+		case <-o.spanCh:
+			// Discard the span
+		case <-time.After(otlpTimeout):
+			return // No more spans available
+		}
+	}
+}
```
mathetake: this seems unused 
```suggestion
```

internal/mcpproxy/handlers.go
```diff
@@ -234,6 +234,7 @@ func (m *MCPProxy) servePOST(w http.ResponseWriter, r *http.Request) {
 			}
 			err = m.handleSetLoggingLevel(ctx, s, w, msg, p, span)
 		case "ping":
+			// Ping is intentionally not traced as it's a lightweight health check.
```
mathetake, line 237: cc @nacx

codefromthecrypt:
one sec I found I didn't assert our span events.

mathetake:
you have my one sec!

codefromthecrypt:
k done :)

tests/extproc/mcp/mcp_test.go
```diff
@@ -876,13 +879,59 @@ func requireMCPSpan(t *testing.T, span *tracev1.Span, expectedName string, addit
...
+				// exception messages are substring match due to IPs, etc.
+				actualMsg := attr.Value.GetStringValue()
+				require.Contains(t, actualMsg, exceptionMessage)
+				attr.Value = &commonv1.AnyValue{Value: &commonv1.AnyValue_StringValue{StringValue: exceptionMessage}}
+			}
+			filteredAttrs = append(filteredAttrs, attr)
+		}
+		event.Attributes = filteredAttrs
+	}
+	require.Empty(t, cmp.Diff(expectedEvents, span.Events, protocmp.Transform()))
```
codefromthecrypt, line 934: ps this is my favorite trick in tests!

codefromthecrypt:
thanks for the merge. since latest commit is pushed to docker, the new elastcisearch example can see this new data policy! https://github.com/elastic/observability-examples/pull/90

codefromthecrypt:
Screenshot 2025-10-08 at 7 04 15 AM

tests/extproc/mcp/env.go
anuraaga: There's a close one but no cigar

https://github.com/Antonboom/testifylint?tab=readme-ov-file#go-require

Probably not too hard to adapt that to apply to both cases though

codefromthecrypt: Cc @nacx

PR #1290, opened 2025-10-07
ci: format markdown files with prettier

anuraaga:
**Description**

Formats markdown files with go-prettier in `make format`

This is a strawman's proposal to format markdown files with go-prettier. Only accept it if you think it's an improvement, it took almost no time to prepare this PR either way.

The main advantage of this IMO is formatting of code blocks

- Go code is all properly indented, with all tabs
- YAML gets formatted in the same way as the example yaml files
- Even the bash scripts get some minor cleanups

This actually affects the rendered result - all other changes are for the markdown source code and no-op in the rendered result. Because markdown often has multiple ways of doing something, like bolding and list bullets, and prettier is opinionated, it picks one and applies it everywhere, and this can't be configured.

I also found that bash code blocks with blanklines get indented, so the editorconfig check was failing for it. There isn't any need for editorconfig check for autoformatted files so I just disabled that one.

Anyways feel free to accept or reject :)

/cc @mathetake


anuraaga:
Nevermind - somehow it seems to take an insane amount of memory. I'll see if this is an issue on my end.

anuraaga:
Yeah realized I was forgetting to close the wasm module, looks like CI works fine now (before it was basically unlimited memory growth)

missBerg:
Let me take a look, I want to see if we can include a VSCode extension plugin setup so people making site edits have this in saving in VSCode

anuraaga:
Thanks for the context! For reference I have this setting in all my repos to format with prettier on save.

https://github.com/wasilibs/go-prettier/blob/main/go-prettier.code-workspace#L23

However this uses official prettier, which does not format Go and Bash code blocks like go-prettier does. IDE formatting matches ci 90% but commands may be required in the end for reproducing the formatting of Go blocks.

If there is a goal for the formatting to be completely replicable with just VSCode, it would be an understandable reason to avoid this PR.

mathetake:
> If there is a goal for the formatting to be completely replicable with just VSCode, it would be an understandable reason to avoid this PR.

i don't think we have such requirement, so good to go

PR #1291, opened 2025-10-07
fix: makes gRPC's MaxRecvMsgSize for the extension server

owayss:
**Description**
Makes go's gRPC `MaxRecvMsgSize` configurable for the extension server, whose default value is 4MB https://pkg.go.dev/google.golang.org/grpc#MaxRecvMsgSize

When the gRPC message payload from the xDS server exceeds 4MB, it is rejected by the extension server and the below errors is observed on envoy-gateway:
```
2025-10-07T10:26:44.625Z        ERROR   xds     runner/runner.go:245    failed to translate xds ir      {"runner": "xds", "error": "rpc error: code = ResourceExhausted desc = grpc: received message larger than max (6683527 vs. 4194304)"}
2025-10-07T10:26:44.625Z        ERROR   watchable       message/watchutil.go:86 observed an error       {"runner": "xds", "error": "rpc error: code = ResourceExhausted desc = grpc: received message larger than max (6683527 vs. 4194304)"}
```
This PR makes it configurable via the command-line flag `maxRecvMsgSize` and keeps its default at 4MB.

Fixes https://github.com/envoyproxy/ai-gateway/issues/1293
Related PR to make gRPC's message sizes configurable, though that was for the external processor gRPC server: https://github.com/envoyproxy/ai-gateway/pull/1213

Note that this applies to the xDS resource messages sent from the xDS server to the extension server. The payload of these messages increases with the size/count of the config resources.

### DIFF
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -49,6 +49,8 @@ type flags struct {
 	extProcExtraEnvVars            string
 	// extProcMaxRecvMsgSize is the maximum message size in bytes that the gRPC server can receive.
 	extProcMaxRecvMsgSize int
+	// maxRecvMsgSize is the maximum message size in bytes that the gRPC extension server can receive.
+	maxRecvMsgSize int
 }
 
 // parsePullPolicy parses string into a k8s PullPolicy.
@@ -146,6 +148,11 @@ func parseAndValidateFlags(args []string) (flags, error) {
 		512*1024*1024,
 		"Maximum message size in bytes that the gRPC server can receive for extProc. Default is 512MB.",
 	)
+	maxRecvMsgSize := fs.Int(
+		"maxRecvMsgSize",
+		4*1024*1024,
+		"Maximum message size in bytes that the gRPC extension server can receive. Default is 4MB.",
+	)
 
 	if err := fs.Parse(args); err != nil {
 		err = fmt.Errorf("failed to parse flags: %w", err)
@@ -215,6 +222,7 @@ func parseAndValidateFlags(args []string) (flags, error) {
 		rootPrefix:                     *rootPrefix,
 		extProcExtraEnvVars:            *extProcExtraEnvVars,
 		extProcMaxRecvMsgSize:          *extProcMaxRecvMsgSize,
+		maxRecvMsgSize:                 *maxRecvMsgSize,
 	}, nil
 }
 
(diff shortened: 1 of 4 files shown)


mathetake:
omg this is really a good catch!

cmd/controller/main.go
```diff
@@ -146,6 +148,11 @@ func parseAndValidateFlags(args []string) (flags, error) {
 		512*1024*1024,
 		"Maximum message size in bytes that the gRPC server can receive for extProc. Default is 512MB.",
 	)
+	maxRecvMsgSize := fs.Int(
+		"maxRecvMsgSize",
+		4*1024*1024,
```
mathetake, line 153: can you make the default more reasonable considering the Envoy config size in general? I think 64MB should be good (it's so random though...) since extension servers are not taking untrusted users' requests but only EG is the client plus we don't want to fail AIGW anyways 
```suggestion
		64*1024*1024,
		"Maximum message size in bytes that the gRPC extension server can receive. Default is 64MB.",
```

owayss: @mathetake sounds good too. My rationale for keeping the the default as low as `4MB` (that's gRPC's package default) was to avoid running into OOM and crash the extension server if the container memory limits were not adjusted beforehand, and instead have it only reject the large config (as is the current behaviour now) so users can increase this parameter along with their container resources accordingly.
Let me know what you think.

mathetake: yeah that makes sense too. Yeah i think we are good to go with 4mb anyways as long as we provide the configuration knob, it should be good

mathetake (commented):
@owayss before merging, could you add a helm chart flag to configure this one?

* https://github.com/envoyproxy/ai-gateway/blob/fe78de10f013fb20509927583625e1ad765c8b23/manifests/charts/ai-gateway-helm/templates/deployment.yaml#L44
* https://github.com/envoyproxy/ai-gateway/blob/fe78de10f013fb20509927583625e1ad765c8b23/manifests/charts/ai-gateway-helm/values.yaml#L37

then good to go!

owayss:
> @owayss before merging, could you add a helm chart flag to configure this one?
> 
>     * https://github.com/envoyproxy/ai-gateway/blob/fe78de10f013fb20509927583625e1ad765c8b23/manifests/charts/ai-gateway-helm/templates/deployment.yaml#L44
> 
>     * https://github.com/envoyproxy/ai-gateway/blob/fe78de10f013fb20509927583625e1ad765c8b23/manifests/charts/ai-gateway-helm/values.yaml#L37
> 
> 
> then good to go!

thanks, done in https://github.com/envoyproxy/ai-gateway/pull/1291/commits/45620fd69595e93168588892bb3ef9dd94709005

PR #1294, opened 2025-10-07
mcp: configure heartbeats to be less chatty

nacx:
**Description**

We were emitting heartbeats every 1s and that was very chatty, especially when we're generating traces for each heartbeat. This PR:

* Makes the heartbeat configurable and allows even disabling it by setting the `MCP_PROXY_HEARTBEAT_INTERVAL=0`.
* Changes the default heartbeat interval to 1 minute.
* Eagerly sends an initial ping to tackle the Goose issue. With this, heartbeats can be disabled and Goose works fine as well.

N/A

N/A

### DIFF
--- a/examples/mcp/agent.py
+++ b/examples/mcp/agent.py
@@ -21,6 +21,7 @@
 #     "openai-agents",
 #     "httpx",
 #     "mcp",
+#     "elastic-opentelemetry",
 #     "openinference-instrumentation-openai-agents",
 #     "opentelemetry-instrumentation-httpx",
 #     "openinference-instrumentation-mcp",
@@ -42,32 +43,28 @@ from agents import (
     OpenAIProvider,
     RunConfig,
     Runner,
-    Tool,
 )
-from agents.mcp import MCPServerStreamableHttp, MCPUtil
+from agents.mcp import MCPServer, MCPServerStreamableHttp, MCPUtil
 
-
-async def run_agent(prompt: str, model_name: str, tools: list[Tool]):
-    model = OpenAIProvider(use_responses=False).get_model(model_name)
-    agent = Agent(name="Assistant", model=model, tools=tools)
-    result = await Runner.run(
-        starting_agent=agent,
-        input=prompt,
-        run_config=RunConfig(workflow_name="envoy-ai-gateway"),
-    )
-    print(result.final_output)
+# Uncomment the following lines to enable agent verbose logging
+# from agents import enable_verbose_stdout_logging
+# enable_verbose_stdout_logging()
 
 
 async def main(prompt: str, model_name: str, mcp_url: str):
-    if not mcp_url:
-        await run_agent(prompt, model_name, [])
-        return
-
-    async with MCPServerStreamableHttp({"url": mcp_url,"timeout": 300.0},cache_tools_list=True) as server:
-        tools = await server.list_tools()
-        util = MCPUtil()
-        tools = [util.to_function_tool(tool, server, False) for tool in tools]
-        await run_agent(prompt, model_name, tools)
+    async with MCPServerStreamableHttp(
+        name="Envoy AI Gateway MCP",
+        params={"url": mcp_url, "timeout": 300},
+        cache_tools_list=True,
+    ) as server:
+        model = OpenAIProvider(use_responses=False).get_model(model_name)
+        agent = Agent(name="Assistant", model=model, mcp_servers=[server])
+        result = await Runner.run(
+            starting_agent=agent,
+            input=prompt,
+            run_config=RunConfig(workflow_name="Envoy AI Gateway Example"),
+        )
+        print(result.final_output)
 
 
 if __name__ == "__main__":


internal/mcpproxy/session.go
```diff
@@ -218,6 +239,27 @@ func (s *session) streamNotifications(ctx context.Context, w http.ResponseWriter
...
 
+// defaultHeartbeatInterval is the default heartbeat interval if the MCP_HEARTBEAT_INTERVAL environment variable is not set or invalid.
+// This is set as a variable and not a constant to let the tests modify it.
+var defaultHeartbeatInterval = 1 * time.Minute
+
+// getHeartbeatInterval returns the heartbeat interval configured via the MCP_HEARTBEAT_INTERVAL environment variable.
+// If the environment variable is not set or invalid, it returns the default value of 1 minute.
+// This value is intentionally hidden under an environment variable as it is unclear if it is generally useful.
+func (s *session) getHeartbeatInterval() time.Duration {
+	heartbeatIntervalStr := os.Getenv("MCP_PROXY_HEARTBEAT_INTERVAL")
```
zirain: would it be better to do this when proxy init?

nacx: For now, since this setting is experimental, I wanted to have it as isolated as possible, without promoting a "proper config knob" elsewhere in the codebase. I think it is OK to read the env var every time (performance is negligible) until we have more clarity on the right defaults and if this is a useful setting or not.

mathetake: then i would put this at least in func init() right above this function

```

func Getenv(key string) (value string, found bool) {
	copyenv()
	if len(key) == 0 {
		return "", false
	}

	envLock.RLock()
	defer envLock.RUnlock()

```

Getenv takes a global lock which I don't think we want in the request path as much as possible

nacx: Done

nacx:
The `TestMCP/custom_write_timeout/ToolCountDown/custom_write_timeout` test is failing in this pull request.
What the test "tried to test" is the reconnection from the MCP client, by setting a small `writeTimeout = 2*time.Second` on the server, and assumes clients would reconnect and resume the connection.

This execution path, though, **was never exercised** because the heartbeats were every 1 second, causing the server to never reach the `writeTimeout`, and client reconnections never happened -> the test is not testing what it's trying to test with the custom `writeTimeout`.

This PR changes the default heartbeat interval to 1 minute (effectively making heartbeats not happen for this test, as it uses a 2-second writeTimeout), and this should trigger the desired behavior. However, reaching the `writeTimeout` **doesn't seem to close the connection** and the MCP client reconnection logic is not triggered, and the test simply times out.

I've added more logs to highlight the situation:

```
=== RUN   TestMCP/custom_write_timeout/ToolCountDown/custom_write_timeout
2025/10/08 11:01:52 MCP SERVER connection [new] 127.0.0.1:55919 -> 127.0.0.1:55901
2025/10/08 11:01:52 MCP SERVER connection [active] 127.0.0.1:55919 -> 127.0.0.1:55901
2025/10/08 11:01:52 MCP SERVER connection [idle] 127.0.0.1:55919 -> 127.0.0.1:55901
2025/10/08 11:01:52 MCP SERVER connection [new] 127.0.0.1:55922 -> 127.0.0.1:55901
2025/10/08 11:01:52 MCP SERVER connection [active] 127.0.0.1:55922 -> 127.0.0.1:55901
2025/10/08 11:01:52 MCP SERVER connection [idle] 127.0.0.1:55922 -> 127.0.0.1:55901
2025/10/08 11:01:52 MCP SERVER connection [active] 127.0.0.1:55919 -> 127.0.0.1:55901
    env.go:247: created new MCP session with ID  lqciHIDFLBX30JywisPpjmH2LVBWbfPwZUKvnMyfAFOjUyDJtUR9PEwMqVDkpcMGYh+yqBwEEcSvOh/hCUrfHwyXD93wpXEltW2ez68zr3zTapBjPc/bnZUZVEMOsDjx+ZWz9Qnmqb5Ts1jez2XrAJj5UD4M9MEBJ7Rx/vCDxsmEE/wuDXhuznSczVwFovGUg3tXhqDBy5+BnHNHQZYIIhuzjM7Xiw== , first span:  trace_id:"\x8eQ\xda\xdcJ\xa8\x0f\xf5i\x07\xe9!\xb4!ZN"  span_id:"+b{w񩦼"  flags:256  name:"mcp.request"  kind:SPAN_KIND_CLIENT  start_time_unix_nano:1759914112798257000  end_time_unix_nano:1759914112812874667  attributes:{key:"mcp.protocol.version"  value:{string_value:"2025-06-18"}}  attributes:{key:"mcp.transport"  value:{string_value:"http"}}  attributes:{key:"mcp.request.id"  value:{string_value:"{1}"}}  attributes:{key:"mcp.method.name"  value:{string_value:"initialize"}}  attributes:{key:"mcp.client.name"  value:{string_value:"demo-http-client"}}  attributes:{key:"mcp.client.title"  value:{string_value:""}}  attributes:{key:"mcp.client.version"  value:{string_value:"0.1.0"}}  events:{time_unix_nano:1759914112800489000  name:"route to backend"  attributes:{key:"mcp.backend.name"  value:{string_value:"dumb-mcp-backend"}}  attributes:{key:"mcp.session.id"  value:{string_value:"5GNMQPZNVPFOATH7QJASXVPHOG"}}  attributes:{key:"mcp.session.new"  value:{bool_value:true}}}  events:{time_unix_nano:1759914112800603000  name:"route to backend"  attributes:{key:"mcp.backend.name"  value:{string_value:"default-mcp-backend"}}  attributes:{key:"mcp.session.id"  value:{string_value:"Q5DATCXHFRXX54YW4N7J25A32U"}}  attributes:{key:"mcp.session.new"  value:{bool_value:true}}}  status:{code:STATUS_CODE_OK}
2025/10/08 11:01:52 MCP SERVER connection [new] 127.0.0.1:55929 -> 127.0.0.1:55901
2025/10/08 11:01:52 MCP SERVER connection [active] 127.0.0.1:55929 -> 127.0.0.1:55901
2025/10/08 11:01:52 MCP SERVER connection [idle] 127.0.0.1:55929 -> 127.0.0.1:55901
2025/10/08 11:01:52 MCP SERVER connection [active] 127.0.0.1:55929 -> 127.0.0.1:55901
2025/10/08 11:01:52 MCP SERVER connection [idle] 127.0.0.1:55929 -> 127.0.0.1:55901
    mcp_test.go:275: Receive progress n=0
    mcp_test.go:275: Receive progress n=1
    mcp_test.go:275: Receive progress n=2
    mcp_test.go:275: Receive progress n=3
    mcp_test.go:272: timeout waiting for progress notification
2025/10/08 11:02:04 MCP SERVER connection [active] 127.0.0.1:55929 -> 127.0.0.1:55901
2025/10/08 11:02:04 MCP SERVER connection [idle] 127.0.0.1:55929 -> 127.0.0.1:55901
--- FAIL: TestMCP/custom_write_timeout/ToolCountDown/custom_write_timeout (11.70s)
2025/10/08 11:02:04 MCP SERVER connection [closed] 127.0.0.1:55922 -> 127.0.0.1:55901
2025/10/08 11:02:04 MCP SERVER connection [closed] 127.0.0.1:55929 -> 127.0.0.1:55901
```

The server write timeout is reached at the time of `Receive progress n=3`, but the connection is not closed, and the client doesn't attempt to reconnect. The test fails, and we can see the connection closing as part of the test teardown, where we call `session.Close()`.

With this:

* The existing test is not adding any value because it was not testing what it apparently was pretending to test.
* I haven't found a way to force a disconnect in the tests to refactor them in a way that we could exercise the path the test pretended to exercise.

@mathetake What would be the best path forward? Do we want to hold this PR until we figure out how to test reconnections? (I'd rather tackle this in another PR as the test issues are not really related to the feature in this PR; this PR just showcased a deficiency in a test).

tests/extproc/mcp/mcp_test.go
```diff
@@ -79,15 +79,6 @@ func TestMCP_forceJSONResponse(t *testing.T) {
...
 }
 
-func TestMCP_customWriteTimeout(t *testing.T) {
-	env := requireNewMCPEnv(t, false, 2*time.Second, defaultMCPPath)
-	for _, tc := range tests {
-		t.Run(tc.name+"/custom_write_timeout", func(t *testing.T) {
-			tc.testFn(t, env)
-		})
-	}
-}
```
nacx, line 89: Discussed with @mathetake in the context of https://github.com/envoyproxy/ai-gateway/pull/1294#issuecomment-3380610707 and decided to remove the test for now.

Using `writeTimeout` to achieve what this test tried to achieve is not the right approach and this test is doing nothing now, so it makes no sense to preserve it. Reconnection tests will need to be tackled differently.

PR #1296, opened 2025-10-07
ci: skip test_e2e_aigw on doc chaneg & fix docker job requirement

mathetake:
**Description**

Previously, test_e2e_aigw was not skipped unlike others when there's only doc changes in the PR. Also, it was not required in the docker build&push on the main branch, so this commit fixes these two issues.

### DIFF
--- a/.github/workflows/build_and_test.yaml
+++ b/.github/workflows/build_and_test.yaml
@@ -328,6 +328,7 @@ jobs:
   test_e2e_aigw:
     needs: changes
     name: E2E Test for aigw CLI
+    if: ${{ needs.changes.outputs.code == 'true' }}
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


mathetake:
perfect image

.github/workflows/build_and_test.yaml
```diff
@@ -370,15 +371,7 @@ jobs:
...
-    needs:
-      [
-        unittest,
-        test_crdcel,
-        test_controller,
-        test_extproc,
-        test_e2e,
-        test_e2e_inference_extension,
-      ]
+    needs: [ci-required]
```
aabchoo, line 374: Do you need to specify `ci-required.needs`?

Also the list in ci-required has two additional elements `test_e2e_aigw` and `test_e2e_upgrade`. Is that intended?

mathetake: > Do you need to specify ci-required.needs?

you mean to change here to `needs: [ci-required.needs]`? Then no. If you mean adding test_e2e_aigw to `ci-required.needs` then it's already there.

> Also the list in ci-required has two additional elements test_e2e_aigw and test_e2e_upgrade. Is that intended?

yes that's how to block the PR being merged (the comment on ci-required job).

aabchoo (approved):

PR #1297, opened 2025-10-07
translator: add marshal/unmarshal for tool choice

aabchoo:
**Description**

Previously, `openAIReq.ToolChoice` would be understood as type `map[string]interface{}` in this code block [1] when given body field `"tool_choice":{"function":{"name":"qa_theme_extractor"},"type":"function"}}`. Adding specific marshaling/unmarshaling logic to fix this issue. Wanted to use openai-go but there is an issue with the tool choice unmarshal [2].

Issue was not caught because tests sent were written in the proper type. Adding a test that replicates the issue by having the code marshal the tool choice.

1. https://github.com/envoyproxy/ai-gateway/compare/aaron/tool-choice-union-type?expand=1#diff-6da52452dbe8387dcea4abc098618db9120d7d8b3071bf502d60530912158737L144-L178
2. https://github.com/openai/openai-go/issues/520


examples/mcp/README.md
```diff
@@ -184,7 +184,7 @@ Once authenticated, Claude will confirm:
 
 ```
 > /mcp mcp-gateway
-  ⎿  Authentication successful. Connected to mcp-gateway.
+  ⎿  Authentication successful. Connected to mcp-gateway.
```
mathetake: oh this is weird... sorry for the distraction 


@anuraaga go-prettier bug? idk

mathetake: @aabchoo can you confirm this change is caused by `make precommit` or it's just a bad merge?

aabchoo: This initial [commit](https://github.com/envoyproxy/ai-gateway/pull/1297/commits/036309471d02acf28d4c8c39e9b397626697d142) was when I ran `make precommit. Not sure how the file was modified since I didn't touch it

That lead to merge conflicts because `README.md` was changed. I copied+pasted the most recent version of that file to avoid having this show up in my changed file which is how we got here with this strange discrepancy

mathetake: so in other words this is specific change is not caused by make precommit then. can you revert this? 

image

yuzisun (approved):

PR #1299, opened 2025-10-08
fix: response_format schema type

### LINKED ISSUE #1298: schema type `any` in response_format field produces different LLM response
_Description_:

ChatCompletionResponseFormatJSONSchemaJSONSchema's Schema field with type `any` makes `json.Marshal` sort response_format's json field. Different order of fields will force LLM to produce different result

xiaolin593:
**Description**

Fix response_format schema type with raw byte to avoid sorting behavior by `json/encode` Go library. 

Fixes #1298

### DIFF
--- a/tests/internal/testopenai/cassettes.go
+++ b/tests/internal/testopenai/cassettes.go
@@ -26,7 +26,6 @@ const (
 
 	// CassetteChatBasic is the canonical OpenAI chat completion request.
 	CassetteChatBasic Cassette = iota
-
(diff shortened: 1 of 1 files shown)


tests/internal/testopenai/cassettes.go
mathetake, line 1: could you revert the changes in this file which is irrelevant ?

mathetake (commented):
nit but LGTM

PR #1309, opened 2025-10-08
deps: upgrade EG Go dependency

mathetake:
**Description**

This upgrades the Envoy Gateway dependency to the latest main branch commit. Note that the Go level EG dependency is essentially **only used in aigw run** standalone mode. That is because in the k8s environment, the only code path where we use EG is to generate EG configurations, hence the data structures/control plane/extension server gRPC API which we should assume is stable. Moreover, we still have the e2e test with v1.5, so by no means this implies  AIGW main branch cannot work with v1.5. On the other hand, this brings a lot of bug fixes and enhancement to the standalone mode.

### DIFF
--- a/manifests/charts/ai-gateway-crds-helm/templates/aigateway.envoyproxy.io_aigatewayroutes.yaml
+++ b/manifests/charts/ai-gateway-crds-helm/templates/aigateway.envoyproxy.io_aigatewayroutes.yaml
@@ -94,7 +94,7 @@ spec:
                               Claims lists the names of resources, defined in spec.resourceClaims,
                               that are used by this container.
 
-                              This is an alpha field and requires enabling the
+                              This field depends on the
                               DynamicResourceAllocation feature gate.
 
                               This field is immutable. It can only be set for containers.
--- a/manifests/charts/ai-gateway-crds-helm/templates/aigateway.envoyproxy.io_backendsecuritypolicies.yaml
+++ b/manifests/charts/ai-gateway-crds-helm/templates/aigateway.envoyproxy.io_backendsecuritypolicies.yaml
@@ -301,7 +301,6 @@ spec:
                               By default, its unset.
                             properties:
                               sameSite:
-                                default: Strict
                                 enum:
                                 - Lax
                                 - Strict
@@ -400,6 +399,12 @@ spec:
                             required:
                             - headers
                             type: object
+                          disableTokenEncryption:
+                            description: |-
+                              Disable token encryption. When set to true, both the access token and the ID token will be stored in plain text.
+                              This option should only be used in secure environments where token encryption is not required.
+                              Default is false (tokens are encrypted).
+                            type: boolean
                           forwardAccessToken:
                             description: |-
                               ForwardAccessToken indicates whether the Envoy should forward the access token
@@ -799,8 +804,7 @@ spec:
                                                 items:
                                                   description: HTTPStatus defines
                                                     the http status code.
-                                                  exclusiveMaximum: true
-                                                  maximum: 600
+                                                  maximum: 599
                                                   minimum: 100
                                                   type: integer
                                                 type: array
@@ -1333,8 +1337,7 @@ spec:
                                             items:
                                               description: HTTPStatus defines the
                                                 http status code.
-                                              exclusiveMaximum: true
-                                              maximum: 600
+                                              maximum: 599
                                               minimum: 100
                                               type: integer
                                             type: array
@@ -1741,7 +1744,6 @@ spec:
                               By default, its unset.
                             properties:
                               sameSite:
-                                default: Strict
                                 enum:
                                 - Lax
                                 - Strict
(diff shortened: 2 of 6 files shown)


go.mod
```diff
@@ -3,8 +3,6 @@ module github.com/envoyproxy/ai-gateway
 // Explicitly specify the Go patch version to be able to purge the CI cache correctly.
 go 1.25.1
 
-replace go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc v0.61.0 => go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc v0.60.0
```
mathetake, line 6: fyi we got `go run github.com/envoyproxy/ai-gateway/cmd/aigw@main run` back

nacx: Niiiiiice!!

mathetake:
hmm mcp got broken

mathetake:
```
<< npx @modelcontextprotocol/inspector@v0.16.0 --cli http://localhost:1975/mcp --transport http --method 'tools/list'
Need to install the following packages:
@modelcontextprotocol/inspector@0.16.0
Ok to proceed? (y) y

Failed to connect to MCP server: Error POSTing to endpoint (HTTP 404): unsupported path: /mcp
Failed to connect to MCP server: Error POSTing to endpoint (HTTP 404): unsupported path: /mcp

```

so looks like the mcp request reaches the ai extproc... i guess some behavioral change around extensionserver

cmd/aigw/translate.go
```diff
@@ -72,10 +72,10 @@ func translate(ctx context.Context, paths []string, output, stderr io.Writer) er
 		mustWriteObj(&secret.TypeMeta, &secret, output)
 	}
 	for _, secret := range originalSecrets {
-		mustWriteObj(nil, secret, output)
+		mustWriteObj(&secret.TypeMeta, secret, output)
```
mathetake, line 75: a bug somehow surfaced with the upgrade

internal/extensionserver/post_translate_modify.go
```diff
@@ -625,10 +625,6 @@ func (s *Server) isRouteGeneratedByAIGateway(route *routev3.Route) bool {
 	if s.isStandAloneMode {
 		// In stand-alone mode, we don't have annotations to check, so instead use the name prefix.
 		for _, resource := range resources.GetListValue().Values {
-			kind, ok := resource.GetStructValue().Fields["kind"]
-			if !ok || kind.GetStringValue() != "HTTPRoute" {
-				continue
-			}
```
mathetake, line 631: also a bug that surfaced with the upgrade

go.mod
```diff
@@ -19,9 +17,9 @@ require (
 	github.com/cncf/xds/go v0.0.0-20250501225837-2ac532fd4443
 	github.com/coreos/go-oidc/v3 v3.15.0
 	github.com/docker/docker v28.4.0+incompatible
-	github.com/envoyproxy/gateway v1.5.0
-	github.com/envoyproxy/go-control-plane v0.13.5-0.20250622153809-434b6986176d
-	github.com/envoyproxy/go-control-plane/envoy v1.32.5-0.20250622153809-434b6986176d
+	github.com/envoyproxy/gateway v0.5.0-rc.1.0.20251008155728-197fd10aeb03
```
aabchoo, line 20: Downgrade?

mathetake: so i think i told you before a few months ago but EG's main branch has this weird format ....

codefromthecrypt:
EXCELLENT: so this has the fix @nacx made which was very important. func-e inside EG creates an implicit admin server on demand when there is no fixed port one in the yaml. Before, the EG code was broken in how it passed args to cmd Exec (space encoding arg/value in the args slice which is odd worked at all). I can tell by running with `--debug` on the latest docker image that this bug is now fixed. Not only can you tell the admin server, but it is used to check status of the child process.


```
aigw         | time=2025-10-09T00:37:55.526Z level=INFO msg="Discovered Envoy admin adminPort" envoyAdminPort=34703
--snip--
aigw         | time=2025-10-09T00:37:59.530Z level=INFO msg="Waiting for Envoy to be ready..." err="unexpected status code: 503"
--snip--
aigw         | [2025-10-09 00:38:11.529][21][debug][http] [source/common/http/conn_manager_impl.cc:2031] [Tags: "ConnectionId":"8","StreamId":"16872216129129905875"] Codec completed encoding stream.
aigw         | time=2025-10-09T00:38:11.529Z level=INFO msg="Envoy is ready!"
```

internal/extensionserver/post_translate_modify.go
arkodg: @zirain this relates to https://github.com/envoyproxy/gateway/pull/7128#discussion_r2409011455, can you help triage this

mathetake: Yeah won't be an issue for us but sure I think it's a regression on EG side

zirain: @mathetake can you raise an issue in EG?

arkodg: Raised https://github.com/envoyproxy/gateway/issues/7183

zirain: @mathetake this might due to https://github.com/kubernetes-sigs/controller-runtime/pull/3229.

it's not from EG side but controller-time.

```
The fakeclient currently differs from the liveclient in that if a
structured object is created that has typemeta set, it will retain that.
In contrast to that, the liveclient always strips it.

This change makes the fakeclient strip it just like the live client.
```

PR #1310, opened 2025-10-08
test(mcp): drops learn-microsoft as it's flaky 

mathetake:
**Description**

Learn Microsoft MCP server is flaky in the sense that it's unreliable implementation: it _sometimes_ return 405 for exactly the same requests which previously returns 200. This removes our test dependence on it.


mathetake:
That's just a string manipulation test so no harm but good point!

PR #1311, opened 2025-10-08
feat: imagepull secrets for extproc image

johnugeorge:
**Description**
Support imagepull secrets for extproc image


johnugeorge:
@mathetake Done

PR #1312, opened 2025-10-08
refactor: enables peformance gocritic

mathetake:
**Description**


mathetake:
Copilot agent added a bunch of garbage as well so i need to clean up...

PR #1315, opened 2025-10-09
feat: support "CachedInputToken" type in "llmRequestCosts"

everpeace:
**Description**

Many AI providers have recently supported prompt caching on the provider side. Cached token prices are significantly cheaper than normal token processing. For example, in Open AI, cached token price is 10x cheaper than normal token [1]. Thus, Envoy AI Gateway would like to take cached token count into account in calculating `llmRequestCosts` in `AIGatewayRequestCosts`.

Moreover, for self-hosted LLMs, cached tokens can drastically reduce GPU usage. Thus, in such a case, users would like to care about cached token usage in `llmRequestCosts`.

**API Change**

This PR supports `CachedInputToken` in `spec.llmRequestCosts` in `AIGatewayRoute` API:

```yaml
kind: AIGatewayRoute
spec:
  llmRequestCosts:
    type: InputToken
    type: OutputToken
    type: TotalToken
  #
  # [NEW!] This PR adds this
  #
  - metadataKey: llm_cached_input_token
    type: CachedInputToken
  #
  # [NEW!] "cached_input_tokens" variable can be avaialble in CEL
  #
  - metadataKey: llm_request_cost
    type: CEL
    CEL: |-
      (input_tokens - cached_input_tokens) + cached_input_tokens * 0.1 + output_tokens
```

And correspondingly, `gen_ai.token.type="cached_input"` label in `gen_ai.client.token.usage` metrics.

↓The cached input token value extraction from the various provider responses was already implemented by #1276 

~`CachedInputToken` value is derived from:~
- ~Open AI: `.usage.prompt_tokens_details.cached_tokens` field [2]~
- ~Anthropic: `.usage.cache_read_input_tokens` field [3]~
- ~Gemini AI: `.usageMetadata.cachedContentTokenCount` field[4]~
- ~Bedrock: `.usage.cacheReadInputTokens` field[5]~

None

It's my first contribution to Envoy AI Gateway. I'm not so confident in 
- translators changes because I did not try all the providers actually, but I only tried OpenAPI (via vLLM's Open AI Compatible server), and
- test coverage

Please feel free to give your feedback.

1: https://openai.com/api/pricing/
2: https://platform.openai.com/docs/guides/prompt-caching
3: https://docs.claude.com/en/docs/build-with-claude/prompt-caching
4: https://ai.google.dev/api/generate-content#UsageMetadata
5: https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_TokenUsage.html


mathetake:
totally makes sense. let me review soon!

everpeace:
Let me fix extproc test 🙇

everpeace:
I pushed several commits to fix the ext_proc test and the uncovered code.

mathetake:
I realized that the majority of the work here overlaps with https://github.com/envoyproxy/ai-gateway/pull/1276 by @yuzisun  , so I will get it merged before this one then i think this will be more focused on the control plane & API

mathetake:
https://github.com/envoyproxy/ai-gateway/pull/1276 landed -- could you rebase on the latest main branch?

everpeace:
> I realized that the majority of the work here overlaps with https://github.com/envoyproxy/ai-gateway/pull/1276 by @yuzisun , so I will get it merged before this one then i think this will be more focused on the control plane & API

Wow, I didn't notice open-telemetry/semantic-conventions#1276 had been opened already. The PR acutally covered the most of part which I'm not so conficent. Thanks @yuzisun!!

> https://github.com/envoyproxy/ai-gateway/pull/1276 landed -- could you rebase on the latest main branch?

I rebaesd my branch to the latest main branch. So, this PR mainly contains API, control plane, metrics, tests(unit, extproc and e2e), and docs.

api/v1alpha1/shared_types.go
```diff
@@ -111,6 +113,8 @@ type LLMRequestCostType string
 const (
 	// LLMRequestCostTypeInputToken is the cost type of the input token.
 	LLMRequestCostTypeInputToken LLMRequestCostType = "InputToken"
+	// LLMRequestCostTypeCachedInputToken is the cost type of the cached input token.
+	LLMRequestCostTypeCachedInputToken LLMRequestCostType = "CachedInputToken"
```
mathetake, line 117: CachedInputToken vs CachedToken is my question.. thoughts? @yuzisun @codefromthecrypt  OpenAI uses `cached_tokens` and Anthropic has `cache_creation_input_tokens`. Given that we will use this API for both OpenAI endpoints as well as Anthropic /messages endpoint, so I would like to explore as neutral a name as possible

everpeace: > Anthropic has cache_creation_input_tokens

nits: `CachedInputToken` is derived from `cache_read_input_tokens` rather than `cache_creation_input_tokens`.

https://github.com/envoyproxy/ai-gateway/blob/498319505f0bb34f1851bb5d33b2db160363a83c/internal/extproc/translator/anthropic_gcpanthropic.go#L160-L167

internal/metrics/chat_completion_metrics.go
```diff
@@ -71,13 +71,17 @@ func (c *chatCompletion) StartRequest(headers map[string]string) {
...
+func (c *chatCompletion) RecordTokenUsage(ctx context.Context, inputTokens, cachedInputTokens, outputTokens uint32, requestHeaders map[string]string) {
 	attrs := c.buildBaseAttributes(requestHeaders)
 
 	c.metrics.tokenUsage.Record(ctx, float64(inputTokens),
 		metric.WithAttributeSet(attrs),
 		metric.WithAttributes(attribute.Key(genaiAttributeTokenType).String(genaiTokenTypeInput)),
 	)
+	c.metrics.tokenUsage.Record(ctx, float64(cachedInputTokens),
+		metric.WithAttributeSet(attrs),
+		metric.WithAttributes(attribute.Key(genaiAttributeTokenType).String(genaiTokenTypeCachedInput)),
```
mathetake, line 83: so this "cached_token" is not defined in the otel semantics currently. https://github.com/open-telemetry/semantic-conventions/issues/1959  thoughts? cc @codefromthecrypt

mathetake: i think I would suggest you modify this PR not to include the cached read stuff in the metrics but simply only define it in the metadata where we have full control over what to put. That should be enough to enforce rate limit as well as include the token number in the access logs.

PR #1316, opened 2025-10-09
aigw: refactors logging so there is basic status output

### LINKED ISSUE #1304: aigw: better envoy launch error handling
Right now, if you have an invalid config envoy will crash and there will be no output, and the process will hang. If you supply `—debug`, you can see why it crashed but it still hangs. 

```
2025-10-08T07:57:23.453-0400	ERROR	provider	file/file.go:100	failed to reload resources initially	{"runner": "provider", "error": "failed to load resources from file /var/folders/qy/4hghk2xs5wd4jhjtg5ffxdwr0000gn/T/aigw-run/envoy-ai-gateway-resources/config.yaml: local validation error: spec.telemetry.accessLog.settings[0].format.omit_empty_values: Invalid value: value provided for unknown field"}
time=2025-10-08T07:57:26.318-04:00 level=WARN msg="Falling back to Envoy listener adminPort check" err="timeout waiting for child processes: no child process found"
time=2025-10-08T07:57:28.319-04:00 level=INFO msg="Waiting for Envoy to be ready..." err="dial tcp 127.0.0.1:1975: connect: connection refused"
```

A couple improvements we should make:

1. Ensure we always see ERROR from envoy in aigw output, as this explains failures. So, reconfigure so we never have zero output on ERROR.
2. Ensure when envoy crashes, we crash
  - If possible, with the way cobra is working, have its failures cancel our cancelable context to end the goroutines
  - If not, refactor out the subprocess detection code in the admin-lookup helper and crash unconditionally of admin server if the envoy subprocess died.

codefromthecrypt:
**Description**

Before, there was no feedback at all unless you turn on `--debug`. Choosing what to enable is difficult as the EG layer creates known errors that confuse, and our layer we log too much debug as info.

This buys time by making a simple status message and verifying it in tests.

.e.g.
```bash
$ out/aigw-darwin-arm64 run --mcp-json '{
  "mcpServers": {
    "kiwi": {
      "type": "http",
      "url": "https://mcp.kiwi.com"
    }
  }
}'
Envoy AI Gateway listening on http://localhost:1975 (admin http://localhost:51333) after 16.8s
```

Note, it most importantly only shows when (significantly later) envoy is actually ready. If you send requests before this, it will fail. Future PRs can improve logging.

Fixes #1304

### DIFF
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -47,6 +47,7 @@ type flags struct {
 	spanRequestHeaderAttributes    string
 	rootPrefix                     string
 	extProcExtraEnvVars            string
+	extProcImagePullSecrets        string
 	// extProcMaxRecvMsgSize is the maximum message size in bytes that the gRPC server can receive.
 	extProcMaxRecvMsgSize int
 	// maxRecvMsgSize is the maximum message size in bytes that the gRPC extension server can receive.
@@ -143,6 +144,11 @@ func parseAndValidateFlags(args []string) (flags, error) {
 		"",
 		"Semicolon-separated key=value pairs for extra environment variables in extProc container. Format: OTEL_SERVICE_NAME=ai-gateway;OTEL_TRACES_EXPORTER=otlp",
 	)
+	extProcImagePullSecrets := fs.String(
+		"extProcImagePullSecrets",
+		"",
+		"Semicolon-separated list of image pull secret names for extProc container. Format: my-registry-secret;another-secret",
+	)
 	extProcMaxRecvMsgSize := fs.Int(
 		"extProcMaxRecvMsgSize",
 		512*1024*1024,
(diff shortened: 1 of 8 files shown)


codefromthecrypt:
real provider flaked waiting for envoy to listen which is unrelated and we can't diagnose due to being cautious due to real credentials involved. Once we refactor to have proper info vs debug messages internally I think we should be able to dump logs (as info scope would never have anything that could be a credential)... for now, just kick again

```
	env := startTestEnvironment(t, string(configBytes),
		// Do not dump the log by default since it "might" contain sensitive information.
		// On CI, they should be redacted by GHA automatically, but it would be better to not log them at all just in case.
		// Note: This test won't run on CI for fork PRs.
		false, false)
```

codefromthecrypt:
opened https://github.com/envoyproxy/ai-gateway/issues/1317 about epic start latency as even if we have an upstream issue in EG it is manifesting here

PR #1322, opened 2025-10-09
fix: request memory corruption

johnugeorge:
**Description**

Currently, there is subtle request memory corruption due to sjson ReplaceInPlace usage. 

sjson modifies the original buffer when ReplaceInPlace is set. This causes request memory corruption during  sjson.SetBytesOption calls to replace string(eg: model overwrite).  This issue is always consistently reproducible in fallback mode. It happens in other cases as well but not evident as original request is not reused. 

Steps to reproduce:
1. Create a AIGWRoute openai-llama-fb-uep with fallback enabled with local Provider(model: llama) and openai provider(model: gpt-3.5-turbo). 
2. Make local provider down to simulate fallback
3. Send chat completions request 
{"model":"openai-llama-fb-uep","messages":[{"role":"user","content":"Hello world"}],"max_tokens":10, "stream":true}'

Result: 

OpenAI throws error " time=2025-10-09T17:43:09.703Z level=INFO msg="Backend error response" processor=chat-completion isUpstreamFilter=true cluster="" status_code=400 response_body="{\n    \"error\": {\n        \"message\": \"We could not parse the JSON body of your request. (HINT: This likely means you aren't using your HTTP library correctly. The OpenAI API expects a JSON payload, but what was sent was not valid JSON. If you have trouble figuring out how to fix this, please contact us through our help center at help.openai.com.)\",\n        \"type\": \"invalid_request_error\",\n        \"param\": null,\n        \"code\": null\n    }\n}\n"

On debugging, actual request sent to OpenAI is a corrupt json
{\"model\": \"gpt-3.5-turbo\", \"messages\": [{\"role\": \"user\", \"content\": \"Hello world\"}], \"max_tokens\": 10}x_tokens\": 10}

Attached a simple go program to reproduce the memory corruption behaviour created using sjson.SetBytesOption

@mathetake  Is there a reason why ReplaceInPlace is set to true ?

Ref: https://github.com/user-attachments/files/22804340/test_sjson.go.txt

### DIFF
--- a/internal/extproc/translator/translator.go
+++ b/internal/extproc/translator/translator.go
@@ -218,7 +218,7 @@ type LLMTokenUsage struct {
 // sjsonOptions are the options used for sjson operations in the translator.
 var sjsonOptions = &sjson.Options{
 	Optimistic: true,
-	// Note: DO NOT set ReplaceInPlace to true since at the translation layer, it must be
-	// ensured that the original body is not modified, i.e. the operation must be idempotent.
+	// Note: DO NOT set ReplaceInPlace to true since at the translation layer, which might be called multiple times per retry,
+	// it must be ensured that the original body is not modified, i.e. the operation must be idempotent.
 	ReplaceInPlace: false,
 }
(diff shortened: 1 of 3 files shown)


mathetake:
spent a few hours but finally managed to write up a regression test... i think it's good to go

PR #1263, opened 2025-10-02
cotroller: run extproc as a sidecar container if k8s > v1.32

mathetake:
**Description**

PR #1323, opened 2025-10-10
fix: check sidecar container for rolling update

yuzisun:
**Description**
Fix previous PR's missing check for sidecar containers when handling envoy gateway deployment/daemonset rolling update. https://github.com/envoyproxy/ai-gateway/pull/1263


mathetake:
GOOD CATCH!!!

yuzisun:
> GOOD CATCH!!!

Our envoy gateway pods spent whole day doing rolling updates lol

PR #1332, opened 2025-10-10
fix: correctly URL encode ARNs when using inference profiles on Bedrock

adam-weber:
**Description**
This PR fixes URL encoding of model names in AWS Bedrock request paths to properly handle inference profile ARNs containing special characters.

AWS Bedrock inference profiles use ARN format that includes forward slashes:`arn:aws:bedrock:region:account:inference-profile/model-name`.

Previously, these forward slashes were not URL encoded in the request path, causing invalid URLs. This fix uses `url.PathEscape()` to properly encode special characters in model names before constructing the path.

The change ensures that ARNs like:
`arn:aws:bedrock:us-east-1:123456789012:inference-profile/us.anthropic.claude-3-5-sonnet-20241022-v2:0`

are correctly encoded in the path as:
`/model/arn:aws:bedrock:us-east-1:123456789012:inference-profile%2Fus.anthropic.claude-3-5-sonnet-20241022-v2:0/converse`

  - Updated existing tests to account for URL encoding behavior
  - Added minimal test coverage specifically for inference profile ARNs
  - The fix maintains backward compatibility with model IDs
  - Only characters that need URL encoding (like `/`) are affected

### DIFF
--- a/internal/extproc/translator/openai_awsbedrock_test.go
+++ b/internal/extproc/translator/openai_awsbedrock_test.go
@@ -11,6 +11,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"net/url"
 	"strconv"
 	"strings"
 	"testing"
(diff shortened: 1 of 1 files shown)


missBerg:
Thanks @adam-weber for raising this PR

mathetake:
so i dug into AWS Go SDK and it looks like it's not necessarily matching the URL encoder it seems:

* https://github.com/aws/aws-sdk-go-v2/blob/a58de27396ac71a567d928fe1641a56333809d58/service/bedrockruntime/serializers.go#L211C22-L211C62
* https://github.com/aws/smithy-go/blob/0ccf069f0a011a44ee83477748e637a0e3898d7e/encoding/httpbinding/path_replace.go#L49
* https://github.com/aws/smithy-go/blob/0ccf069f0a011a44ee83477748e637a0e3898d7e/encoding/httpbinding/path_replace.go#L82-L93

where only the forward slashes are encoded. Having said that though, the change here should be harmless regardless

adam-weber:
Thanks for the quick look, @mathetake. You're right, the AWS SDK uses a more selective encoding approach, whereas `url.PathEscape()` is more comprehesive.

For the primary use case (AWS Bedrock inference profile ARNs), both approaches produce identical results. The difference would only appear with characters like `#` that aren't found in AWS model identifiers. With that said - as you noted, this should be harmless when it encodes additional chars for the URL.

yuzisun:
@adam-weber We have model names with slash too and did not run into issues, just want to make sure you have tested and not breaking anything.

adam-weber:
@yuzisun Interesting. I did test both the merged code and manually encoding the URL before sending it. This was definitely the issue with Amazon Bedrock Application Inference Profiles. The issue is on the Bedrock API endpoint side, not within the Envoy AI framework.

PR #1341, opened 2025-10-12
feat(aigw): replace custom admin monitor with one in func-e

codefromthecrypt:
**Description**

This replaces our custom envoy admin monitor code now that func-e has the same functionality.

functionality upstreamed (and more) here https://github.com/tetratelabs/func-e/pull/483

### DIFF
--- a/.github/workflows/build_and_test.yaml
+++ b/.github/workflows/build_and_test.yaml
@@ -363,7 +363,8 @@ jobs:
           FUNC_E_HOME: /tmp/envoy-gateway # hard-coded directory in EG
       - name: Install Goose
         env:
-          GOOSE_VERSION: v1.8.0
+          GOOSE_VERSION: v1.10.0
+          OS: Linux
         run: |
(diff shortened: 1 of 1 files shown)


cmd/aigw/run.go
```diff
@@ -214,65 +214,40 @@ func run(ctx context.Context, c cmdRun, o runOpts, stdout, stderr io.Writer) err
 	}
 	server.SetArgs([]string{"server", "--config-path", egConfigPath})
 
-	// Start a monitoring goroutine to poll Envoy's readiness. This starts
-	// before the server to ensure we don't miss the readiness window.
-	go func() {
```
codefromthecrypt, line 219: we no longer need a separate goroutine, as the awaiting admin ready is happening on the func-e goroutine, which existed before, just we had no hook to reuse it.

codefromthecrypt:
going to re-use this PR for another cleanup around directory names wrt func-e defaults

Issue #1335, opened 2025-10-11
tracing: Headers to tracing (span) attributes


abolishgenocidenow:
_Description_:

Do you think it makes sense to have a way to enrich the attributes of each generated span from a specified headers?

For example: If the request has: `X-Tracing-Enrichment-User-Region` (so operator can specify a prefix, here is ` X-Tracing-Enrichment`) becomes `user.region` in the span attribute. This is currently very helpful for us to do some tracing export routing. At the tracing export router side, the implementor has an option (read: opportunity) to scrub it before sending it to the actual sink.

nacx:
This can be already configured, for metrics and traces, with these two controller config values:
https://github.com/envoyproxy/ai-gateway/blob/main/manifests/charts/ai-gateway-helm/values.yaml#L46

This is not accounted for yet in MCP metrics/traces. I'll get that fixed. @mathetake can you assign this to me?

PR #1342, opened 2025-10-12
mcp: add configured header attribtues to metrics and spans

nacx:
**Description**

The LLM metrics and trace providers allow for the configuration of custom headers to be included as metric attributes and span attributes. The MCP implementation, however, did not honor those values.

This PR updates the MCP metrics and tracer configurations to also take this configuration into account.
* The tracer method to create a span now receives the request headers.
* To avoid having to propagate request headers all the way down inside the MCP Proxy so that they can be taken into account when recording metrics, I've opted to instantiate an MCPProxy on each request, and have it set up with the request attributes from the beginning. This allows reusing the metrics implementation in a clean way.

Fixes https://github.com/envoyproxy/ai-gateway/issues/1335
Fixes https://github.com/envoyproxy/ai-gateway/issues/1348

N/A

### DIFF
--- a/internal/metrics/mcp_metrics.go
+++ b/internal/metrics/mcp_metrics.go
@@ -14,6 +14,8 @@ import (
 	mcpsdk "github.com/modelcontextprotocol/go-sdk/mcp"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/metric"
+
+	"github.com/envoyproxy/ai-gateway/internal/lang"
 )
 
 // nolint: godot
@@ -337,8 +339,8 @@ func (m *mcp) withDefaultAttributes(params mcpsdk.Params, attrs ...attribute.Key
 	all := make([]attribute.KeyValue, 0, len(m.defaultAttributes)+len(m.requestHeaderAttributeMapping)+len(attrs))
 	all = append(all, m.defaultAttributes...)
 	if params != nil {
-		for k, v := range params.GetMeta() {
-			if target, ok := m.requestHeaderAttributeMapping[k]; ok {
+		for src, target := range m.requestHeaderAttributeMapping {
+			if v := lang.CaseInsensitiveValue(params.GetMeta(), src); v != "" {
 				all = append(all, attribute.String(target, fmt.Sprintf("%v", v)))
 			}
 		}
--- a/internal/tracing/mcp.go
+++ b/internal/tracing/mcp.go
@@ -8,11 +8,7 @@ package tracing
 import (
 	"context"
 	"fmt"
-	"maps"
 	"net/http"
-	"slices"
-	"sort"
-	"strings"
 
 	"github.com/modelcontextprotocol/go-sdk/jsonrpc"
 	"github.com/modelcontextprotocol/go-sdk/mcp"
@@ -21,6 +17,7 @@ import (
 	"go.opentelemetry.io/otel/propagation"
 	"go.opentelemetry.io/otel/trace"
 
+	"github.com/envoyproxy/ai-gateway/internal/lang"
 	tracing "github.com/envoyproxy/ai-gateway/internal/tracing/api"
 )
 
@@ -90,7 +87,7 @@ func (m mcpTracer) StartSpanAndInjectMeta(ctx context.Context, req *jsonrpc.Requ
 		// Check if the attribute is present in the metadata first, as this is the common place to add custom attributes
 		// in MCP requests. Fall back to headers if not found in metadata.
 		// If the attribute is not found there, check if there is any custom header to map.
-		if metaValue := caseInsensitiveValue(param.GetMeta(), srcName); metaValue != "" {
+		if metaValue := lang.CaseInsensitiveValue(param.GetMeta(), srcName); metaValue != "" {
 			attrs = append(attrs, attribute.String(targetName, metaValue))
 		} else if headerValue := headers.Get(srcName); headerValue != "" { // this is case-insensitive
 			attrs = append(attrs, attribute.String(targetName, headerValue))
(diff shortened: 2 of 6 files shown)


internal/metrics/mcp_metrics.go
```diff
@@ -121,16 +124,19 @@ type MCPMetrics interface {
...
-	methodCount            metric.Float64Counter
-	initializationDuration metric.Float64Histogram
-	capabilitiesNegotiated metric.Float64Counter
-	progressNotifications  metric.Float64Counter
+	requestDuration           metric.Float64Histogram
+	methodCount               metric.Float64Counter
+	initializationDuration    metric.Float64Histogram
+	capabilitiesNegotiated    metric.Float64Counter
+	progressNotifications     metric.Float64Counter
+	requestHeaderLabelMapping map[string]string // maps HTTP headers to metric label names.
```
codefromthecrypt: please use the word attribute as this is otel and not just prom. label is a prom thing and applied at export

internal/mcpproxy/handlers.go
```diff
@@ -1215,7 +1214,13 @@ func parseParamsAndMaybeStartSpan[paramType mcp.Params](ctx context.Context, m *
 		return nil, err
 	}
 
-	span := m.tracer.StartSpanAndInjectMeta(ctx, req, p)
+	// TODO: headers with multiple values
```
codefromthecrypt: curious why not pass `http.Header` type instead of converting it to a map like this?

codefromthecrypt:
another way out would be to do the HTTP tier in case some attribute maps apply, just don't close out MCP mapping until there's another attribute for request meta -> span attributes. Doing things at the HTTP layer will apply attributes to multiple traces (to to the JSON-RPC messaging layer), which is different than normal HTTP where one request is a single span.

https://github.com/modelcontextprotocol/modelcontextprotocol/pull/414 is the details about why trace propagation is at meta level, and so likely should be other things like session.id.

nacx:
I thought about the `_meta` field as well, but I don't know if it would be useful in practice to achieve the custom attribute thing.
Enriching LLM metrics with custom headers today, is mostly about initializing their agents with the headers they want. Something like:

```python
client = AsyncOpenAI(base_url="...", api_key="...", default_headers=CUSTOM_HEADERS)
provider = OpenAIProvider(openai_client=client)
```

So, for people already using custom headers for LLM, initializing the MCP client using the same configuration looks like something reasonable that will just work with current setups, as the clients are already accepting custom headers.

```python
mcp = MCPServerStreamableHttp(name="Envoy AI Gateway MCP",
        params={"url": mcp_url, "timeout": 300, "headers": CUSTOM_HEADERS},
        cache_tools_list=True,
)
```

Having the tracing implementation extract these custom attributes from the `_meta` field would require agent/mcp SDKs to populate that information there, and AFAIK this is done specifically for the tracing information, but not ina. more general fashion today.

codefromthecrypt:
opened https://github.com/envoyproxy/ai-gateway/issues/1348 so we don't have to go around on this one. It is ok to skip if you don't feel it is straightforward to implement and wait for someone to complain. Just it takes several months sometimes for an aigw release. Meanwhile any late request can use the docker images built from main branch anyway!

internal/mcpproxy/handlers.go
nacx: Mostly to keep a similar interface to the chat completions one: https://github.com/envoyproxy/ai-gateway/blob/main/internal/tracing/api/api.go#L81
But I can change to `http.Header` if we think that'd be better.

nacx:
@codefromthecrypt in https://github.com/envoyproxy/ai-gateway/pull/1342/commits/1420dbb39eb798fa7ce732e6ffccb13e7eaddf92 I'm reading custom attributes from the metadata as well, those taking precedence over the HTTP header ones. WDYT?

internal/metrics/mcp_metrics.go
nacx: Done. I've also added a commit to rename it in the chat completion and embeddings metrics implementations

internal/tracing/mcp.go
```diff
@@ -81,9 +81,14 @@ func (m mcpTracer) StartSpanAndInjectMeta(ctx context.Context, req *jsonrpc.Requ
 	attrs = append(attrs, getMCPParamsAsAttributes(param)...)
 
 	// Apply header-to-attribute mapping if configured.
-	for headerName, attrName := range m.headerAttributes {
-		if headerValue, ok := headers[headerName]; ok {
-			attrs = append(attrs, attribute.String(attrName, headerValue))
+	for srcName, targetName := range m.attributeMappings {
+		// Check if the attribute is present in the metadata first, as this is the common place to add custom attributes
```
codefromthecrypt, line 87: great idea!

codefromthecrypt:
fyi I am working on goose to do propagation at the moment and will probably try to get it to propagate its session id when done

mathetake:
is this ready or on hold?

nacx:
This is ready! :)
With the latest changes, it also tackles https://github.com/envoyproxy/ai-gateway/issues/1348

internal/mcpproxy/handlers.go
```diff
@@ -1215,7 +1219,13 @@ func parseParamsAndMaybeStartSpan[paramType mcp.Params](ctx context.Context, m *
...
+	headerMap := make(map[string]string, len(headers))
+	for header := range headers {
+		headerMap[header] = headers.Get(header)
+	}
+
```
mathetake: non blocking nit comment (as i think the impact is negligible): maybe skip this header copying when the header-to-attribute thingy is not configured at the startup time

internal/mcpproxy/handlers.go
codefromthecrypt: it isn't too important just I must have started a problem with copying in maps? if we don't need to copy we shouldn't especailly as headers are case insentive in http

internal/tracing/mcp.go
```diff
@@ -75,6 +80,18 @@ func (m mcpTracer) StartSpanAndInjectMeta(ctx context.Context, req *jsonrpc.Requ
 	}
 	attrs = append(attrs, getMCPParamsAsAttributes(param)...)
 
+	// Apply header-to-attribute mapping if configured.
+	for srcName, targetName := range m.attributeMappings {
+		// Check if the attribute is present in the metadata first, as this is the common place to add custom attributes
+		// in MCP requests. Fall back to headers if not found in metadata.
+		// If the attribute is not found there, check if there is any custom header to map.
+		if metaValue, ok := param.GetMeta()[srcName]; ok {
```
codefromthecrypt: we need to make sure that we do downcase based comparison as unlike http these are case sensitive, but we should clobber anyway.

codefromthecrypt: or any other normal way to do case insensitive map (probably need to scan and decide to pick first or last even though maps aren't ordered in go..)

codefromthecrypt:
fyi for testing https://github.com/block/goose/pull/5165

internal/tracing/mcp.go
mathetake: @nacx could you address this adrian's comment

nacx: Done!

PR #1346, opened 2025-10-13
feat: implement otel tracing and metrics for completions endpoint

### LINKED ISSUE #1231: Support v1/completions Endpoint
*Description*:
Envoy currently lacks support for the `v1/completions` endpoint. While this is fine for many cloud LLM integrations, self-hosted workflows (e.g., LoRA) often depend on `v1/completions`. Adding support would enable those use cases.

[optional *Relevant Links*:]
>[vLLM docs on serving LoRA with Open-AI compatible.](https://docs.vllm.ai/en/latest/features/lora.html#serving-lora-adapters)

codefromthecrypt:
**Description**

This completes metrics and tracing and updates docs. So, now /v1/completions is complete!

Screenshot 2025-10-12 at 10 16 18 PM

Fixes #1231


internal/tracing/openinference/openai/request_attrs.go
```diff
@@ -97,14 +98,91 @@ func buildRequestAttributes(chatRequest *openai.ChatCompletionRequest, body stri
 						}
 					}
 				}
-			} else {
-				// For other message types, use the simple extraction.
-				content := extractMessageContent(msg)
-				if content != "" {
-					if config.HideInputText {
-						content = openinference.RedactedValue
+			case msg.OfAssistant != nil:
```
codefromthecrypt, line 101: this was tech debt earlier I fixed in the same change, we skipped structured results for non-string cases except user messages

codefromthecrypt:
will fix coverage tomorrow morning US time

codefromthecrypt:
coverage is fixed

internal/metrics/genai.go
```diff
@@ -24,13 +24,14 @@ const (
...
 
-	genaiOperationChat      = "chat"
-	genaiOperationEmbedding = "embeddings"
-	genaiProviderOpenAI     = "openai"
-	genaiProviderAWSBedrock = "aws.bedrock"
-	genaiTokenTypeInput     = "input"
-	genaiTokenTypeOutput    = "output"
-	genaiErrorTypeFallback  = "_OTHER"
+	genaiOperationChat       = "chat"
+	genaiOperationCompletion = "completion"
```
mathetake, line 28: completion or completions? the request path seems /completions

codefromthecrypt: /chat/completions also ends in s ;)

codefromthecrypt: right now, this matches the openinference spec. That said, it might make sense to name it with an s since you techically can send more than one completion in the same request. however, I think that sort of rename would be easy if we did it later vs this PR and I would want to make sure it is coherent with the spec. make sense?

mathetake: sg

codefromthecrypt: fwiw otel spec calls it text_completion which of course the word text isn't in the path ;) wild out there

Issue #1334, opened 2025-10-11
controller: Watch specific namespaces


abolishgenocidenow:
_Description_:

Envoy Gateway has this `config.envoyGateway.provider.kubernetes.watch.namespaces` to make the controller to watch only a specific set of namespaces. Can we have this in here too?

To have a similar feature with Envoy Gateway's `config.envoyGateway.provider.kubernetes.watch.namespaces`.

nacx:
@mathetake can you assign this to me?

dpunkturban:
Hey everyone,

thanks a lot for the super quick fix!  
It's always a great experience working with you, the responsiveness and care you put into the Envoy AP(I) Gateway are really impressive.  

Every time we stumble across something in the Envoy API or AI Gateway, you’re incredibly fast to respond and often even quicker to fix it.

Really appreciate it - thanks again!

cc @arkodg

PR #1351, opened 2025-10-13
feat: allow to configure the namespaces to watch

nacx:
**Description**

Allows configuring the namespaces to watch by the controller, to minimize the permissions and resources needed. The following properties have been added to the Helm values:

```yaml
controller:
  # Configuration for how the Kubernetes controllers watch the different resources.
  watch:
    # Namespaces to watch. An empty list means to watch all namespaces.
    # Default is an empty list, to watch all namespaces.
    namespaces: []
    # Sync timeout for the Kubernetes cache. If the cache is not synced within this time, the controller will exit.
    # Default is 2 minutes.
    cacheSyncTimeout: 2m
```

Fixes https://github.com/envoyproxy/ai-gateway/issues/1334

N/A

### DIFF
--- a/.github/workflows/build_and_test.yaml
+++ b/.github/workflows/build_and_test.yaml
@@ -351,7 +351,8 @@ jobs:
       - uses: docker/setup-buildx-action@v3
       - run: make test-e2e-namespaced
         env:
-          # We only need to test the upgrade from the latest stable version of EG.
+          # We only need to test with the latest stable version of EG, since these e2e tests
+          # do not depend on the EG version.
           EG_VERSION: v1.5.0
 
   test_e2e_aigw:


.github/workflows/build_and_test.yaml
```diff
@@ -325,6 +325,42 @@ jobs:
...
+    # https://github.com/actions/runner-images/issues/17#issuecomment-1971073406
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - name: latest
+            envoy_gateway_version: v0.0.0-latest
+          - name: v1.5.0
+            envoy_gateway_version: v1.5.0
```
mathetake: maybe run this only on v1.5.0 (latest stable EG) to reduce CI pressure a little since this should't be affected by EG?

nacx: sounds good

nacx: Done

Issue #1331, opened 2025-10-10
Flaky test: TestGatewayAPIInferenceExtension/InferencePoolResolvedRefsCondition/InferencePool_should_show_Accepted:True_by_parents_and_be_routable_via_multiple_HTTPRoutes


mathetake:
https://github.com/envoyproxy/ai-gateway/actions/runs/18415103930/job/52476853866

```
=== RUN   TestGatewayAPIInferenceExtension/InferencePoolResolvedRefsCondition/InferencePool_should_show_Accepted:True_by_parents_and_be_routable_via_multiple_HTTPRoutes
    inferencepool_resolvedrefs_condition.go:75: Waiting for InferencePool gateway-conformance-app-backend/primary-inference-pool to be Accepted by a parent Gateway (Reason: Accepted)
    inferencepool_resolvedrefs_condition.go:75: InferencePool gateway-conformance-app-backend/primary-inference-pool successfully has condition Type=Accepted, Status=True, Reason='Accepted'
    inferencepool_resolvedrefs_condition.go:75: InferencePool gateway-conformance-app-backend/primary-inference-pool is Accepted by a parent Gateway (Reason: Accepted)
    inferencepool_resolvedrefs_condition.go:76: Waiting for InferencePool gateway-conformance-app-backend/primary-inference-pool to be Accepted by a parent Gateway (Reason: Accepted)
    helpers.go:94: Error fetching InferencePool gateway-conformance-app-backend/primary-inference-pool (type: *v1.InferencePool): context deadline exceeded. Retrying.
    inferencepool_resolvedrefs_condition.go:76: 
        	Error Trace:	/home/runner/go/pkg/mod/sigs.k8s.io/gateway-api-inference-extension@v1.0.1/conformance/utils/kubernetes/helpers.go:157
        	            				/home/runner/go/pkg/mod/sigs.k8s.io/gateway-api-inference-extension@v1.0.1/conformance/utils/kubernetes/helpers.go:264
        	            				/home/runner/go/pkg/mod/sigs.k8s.io/gateway-api-inference-extension@v1.0.1/conformance/tests/inferencepool_resolvedrefs_condition.go:76
        	Error:      	timed out or condition not met for InferencePool gateway-conformance-app-backend/primary-inference-pool to have condition Type=Accepted, Status=True, Reason='Accepted'. Polling error: context deadline exceeded. Last error during fetching: context deadline exceeded.
        	            	Last observed InferencePool status:
[...]

PR #1352, opened 2025-10-13
ci: add missing EG_VERSION env var in inference e2e

nacx:
**Description**

Add the missing environment variable to force the EG version to use in the inference e2e tests.

@mathetake @Xunzhuo am I reading it wrong or this env var was missing and the tests were running twice against the same EG version?

N/A

N/A


nacx:
I lack the knowledge to fix the test failure if it is legit... @Xunzhuo could you take a look?

mathetake:
https://github.com/envoyproxy/ai-gateway/issues/1331 same flake so we can ignore

Issue #1347, opened 2025-10-13
Fix MCP test flakes for good with VCR


codefromthecrypt:
_Description_:

VCR tests like testopenai allow us to record and replay interactions. We should have interactions for tools, so that they don't flake and break the build so often like this

```
    --- FAIL: TestPublicMCPServers/tools/list (0.45s)
        publicmcp_test.go:74: tools/list response: &{Meta:map[] NextCursor: Tools:[0xc000ca5bd0 0xc000ca5cc0 0xc000ca5e50 0xc00113c0a0 0xc00113c2d0 0xc00113c3c0 0xc00113c780 0xc00113caf0 0xc00113cd20 0xc00113cfa0 0xc00113d270]}
```

mathetake:
I had to disable GitHub MCP server tests in https://github.com/envoyproxy/ai-gateway/pull/1659 due to https://github.com/github/github-mcp-server/issues/1610  cc @nacx

codefromthecrypt:
yeah I think sometime soon I need to put cycles on this one..

PR #1354, opened 2025-10-13
test: use aigw also for llm in goose e2e

codefromthecrypt:
**Description**

Before, our goose test only went through aigw for MCP instead of proving single agent origin works, by also using it for LLM. Now.. we're good.

### DIFF
--- a/tests/e2e-aigw/goose_test.go
+++ b/tests/e2e-aigw/goose_test.go
@@ -8,6 +8,7 @@ package e2emcp
 import (
 	"encoding/json"
 	"fmt"
+	"os"
 	"os/exec"
 	"path/filepath"
 	"strings"
@@ -18,7 +19,10 @@ import (
 )
 
 func TestMCPGooseRecipe(t *testing.T) {
-	startAIGWCLI(t, aigwBin, "run", "--debug", "aigw_config.yaml")
+	startAIGWCLI(t, aigwBin, []string{
+		"OPENAI_BASE_URL=http://127.0.0.1:11434/v1",
+		"OPENAI_API_KEY=unused",
+	}, "run", "--debug", "--mcp-json", `{"mcpServers":{"kiwi":{"type":"http","url":"https://mcp.kiwi.com"}}}`)
 
 	for _, tc := range []gooseRecipeTestCase{
 		{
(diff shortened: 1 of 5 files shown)


tests/e2e-aigw/goose_test.go
```diff
@@ -52,6 +56,12 @@ func TestMCPGooseRecipe(t *testing.T) {
 				cmd := exec.CommandContext(t.Context(), "goose", args...)
 				cmd.Stdout = buffers[0]
 				cmd.Stderr = buffers[1]
+				// Point Goose at aigw also for LLM access
+				cmd.Env = append(os.Environ(),
+					"OPENAI_HOST=http://127.0.0.1:1975",
+					"OPENAI_BASE_PATH=v1/chat/completions",
```
codefromthecrypt, line 62:
```suggestion
```

tests/e2e-aigw/testdata/kiwi_recipe.yaml
```diff
@@ -42,9 +54,4 @@ parameters:
     input_type: string
     requirement: required
     description: flight date in MM/DD/YYYY format
-    default: "01/10/2025"
-
-settings:
-  goose_provider: ollama
-  goose_model: qwen3:0.6b
-  temperature: 0.0
```
mathetake, line 49: so temperature is important to at least make it less flaky.. is it possible to still specify it?

codefromthecrypt: if you switch to a model like gpt-5 it totally doesn't use it at all. so if we add temperature we have to be careful to remove it if we need to use gpt-5-nano etc.

codefromthecrypt: e.g. it will crash saying the only supported value is 1

mathetake: hah sounds good as-is

codefromthecrypt:
I will put temperature back with a big warning that it makes the recipe incompatible with gpt-5 line which is fine for our test

codefromthecrypt:
I put it back with the warning just because openai is so very unpredictable in their product choices such as dropping temperature!

mathetake:
nooo still flakyyyyyyy

codefromthecrypt:
this will always be flakey. see https://github.com/envoyproxy/ai-gateway/issues/1347 for non-flakey future with VCR

nothing we can run in CI is possible to be unflakey as even premium models are not 100pct on agent leaderboards.

I don't think we should run something like this in CI or ever expect it to not fail. We are on a [fool's errand](https://en.wikipedia.org/wiki/Fool%27s_errand).

https://www.tbench.ai/leaderboard

mathetake:
yeah maybe let's disable goose test altogether for now

PR #1357, opened 2025-10-13
mcp: make session encryption seed rotatable

mathetake:
**Description**

MCP Proxy component has a command line argument to specify the seed for session encryption. However, previously it was not possible to rotate it since the change was atomic. In other words, once you change it, the sessions encrypted with the old seed will fail. This fixes it by allowing users to specify the fallback seed which is only used for decryption when the "primary" is failing to decrypt. Also, this adds a missing helm char value to specify the seed and its plumbing to the extproc container.


manifests/charts/ai-gateway-helm/values.yaml
```diff
@@ -165,8 +165,15 @@ controller:
...
   # from xDS (envoy-gateway).
   # This value should be increased in setups where count/complexity of xDS
   # resources (configuration) is big. Defaults to 4MB.
   # maxRecvMsgSize: "4194304"
+
+  # mcpSessionEncryptionSeed is an arbitrary string seed used to derive the MCP session encryption key.
+  #	Do not include commas as they are used as separators. You can optionally pass "fallback" seed after the first one to allow for key rotation.
+  #	For example: "new-seed,old-seed-for-fallback". The fallback seed is only used for decryption.
+  #
+  # mcpSessionEncryptionSeed: "some-random-string,with-one-optional-comma-separated-fallback-seed"
```
nacx: For better docs and clarity, I would leave this property uncommented with the default value, and add a note here that users should change this to a secure value of their own.

mathetake: https://github.com/envoyproxy/ai-gateway/pull/1357/commits/8906b5dd7bac3055f3c3a50643c3b39edf36869e done

PR #1358, opened 2025-10-13
feat: include model information in ChatCompletion responses for GCP Anthropic and VertexAI

sukumargaonkar:
**Description**
Ensure the model name from response body is passed on to openAI response body

### DIFF
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -55,9 +55,10 @@ type flags struct {
 	// extProcMaxRecvMsgSize is the maximum message size in bytes that the gRPC server can receive.
 	extProcMaxRecvMsgSize int
 	// maxRecvMsgSize is the maximum message size in bytes that the gRPC extension server can receive.
-	maxRecvMsgSize   int
-	watchNamespaces  []string
-	cacheSyncTimeout time.Duration
+	maxRecvMsgSize           int
+	mcpSessionEncryptionSeed string
+	watchNamespaces          []string
+	cacheSyncTimeout         time.Duration
 }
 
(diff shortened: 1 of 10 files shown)


sukumargaonkar:
forced pushed to fix DCO

internal/extproc/translator/openai_gcpanthropic.go
```diff
@@ -673,7 +673,13 @@ func (o *openAIToGCPAnthropicTranslatorV1ChatCompletion) ResponseBody(_ map[stri
 		return nil, nil, tokenUsage, "", fmt.Errorf("failed to unmarshal body: %w", err)
 	}
 
+	responseModel = o.requestModel
+	if anthropicResp.Model != "" {
+		responseModel = string(anthropicResp.Model)
+	}
+
 	openAIResp := &openai.ChatCompletionResponse{
+		Model:   responseModel,
```
yuzisun, line 682: also set the created for completion

yuzisun: should we also set the ID from cloud provider ?

mathetake:
assigning @yuzisun 👍

internal/extproc/translator/openai_gcpanthropic.go
xiaolin593: +1 to set ID from cloud provider.

sukumargaonkar: makes sense
will add `created` and `ID` in future PRs
will keep this PR small

PR #1376, opened 2025-10-16
extproc: preserve removed headers locally for metrics while still stripping them upstream

AyushSawant18588:
**Description**
Metrics labels were missing when a backend configured removeHeaders for headers that are also listed in metricsRequestHeaderAttributes. Those headers were being deleted from the local requestHeaders map before metrics were recorded, so they never appeared as labels. This change ensures those headers are retained locally for metrics while still being stripped before the request reaches the upstream.
- Keep removed headers in the local requestHeaders so metrics can read them.
- Still instruct Envoy to remove those headers upstream via HeaderMutation.RemoveHeaders.


mathetake:
Is this really happening on the main branch? https://github.com/envoyproxy/ai-gateway/commit/1760221ddd0371e48db9c5c443c39addb8ebb21d thought this fixes it

mathetake (approved):

Issue #1191, opened 2025-09-13
Migrate `manifests/envoy-gateway-config` into EG's helm values instead of raw Yamls


mathetake:
*Description*:

Especially, `envoy-gateway-config.yaml` is generated by EG's helm chart, so maintaining the raw yaml here as well could be troublesome in some cases. I believe everything in the dir can be migrated into EG's helm value input, especially envoy-gateway-config.yaml can be under 

https://github.com/envoyproxy/gateway/blob/4b1827909115e2cb509609f773ed58fbd3f47a5e/charts/gateway-helm/values.tmpl.yaml#L98

For redis, I think it can be moved into examples/toke_ratelimit.yaml

mathetake:
well RBAC might be tricky. I couldn't find an option to add additional one from a quick look

saileshd1402:
Hi @mathetake, can I assign this to myself and raise a PR?

mathetake:
sure

kfox1111:
This should be in the right ballpark to remove the need for the config.yaml

along with:
helm upgrade --install valkey oci://registry-1.docker.io/bitnamicharts/valkey -n envoy-gateway-system

```
config:
  envoyGateway:
    extensionApis:
      enableEnvoyPatchPolicy: true
      enableBackend: true
    extensionManager:
      hooks:
        xdsTranslator:
          translation:
            listener:
              includeAll: true
            route:
              includeAll: true
            cluster:
              includeAll: true
            secret:
              includeAll: true
          post:
            - Translation
            - Cluster
            - Route
      service:
        fqdn:
          hostname: ai-gateway-controller.envoy-ai-gateway-system.svc.cluster.local
          port: 1063
    provider:
      kubernetes:
        deploy:
          type: GatewayNamespace
        rateLimitDeployment:
          patch:
            type: StrategicMerge
            value:
              spec:
                template:
                  spec:
                    containers:
                    - imagePullPolicy: IfNotPresent
                      name: envoy-ratelimit
                      image: docker.io/envoyproxy/ratelimit:60d8e81b
                      env:
                      - name: REDIS_AUTH
                        valueFrom:
                          secretKeyRef:
                            name: valkey
                            key: valkey-password
    rateLimit:
      backend:
        type: Redis
        redis:
          url: valkey-primary.envoy-gateway-system.svc.cluster.local:6379
```

kfox1111:
Could the rbac bits be added to the ai gateway chart?

[...]

PR #1377, opened 2025-10-16
refactor: move RBAC for EG into helm chart and drop unnecessary rules

mathetake:
**Description**

This moves the RBAC yaml necessary to be attached to EG controller into our helm chart. This removes one additional step during the installation. 

This also removes the unnecessary/unused roles from the RBAC. The remaining thing is only for the inference pool.

Part of #1191


manifests/charts/ai-gateway-helm/values.yaml
```diff
@@ -178,3 +178,10 @@ controller:
   #
   # This value should be set to a secure random string in production environments instead of the default below.
   mcpSessionEncryptionSeed: "default-insecure-seed"
+
+# Configuration for the Envoy Gateway component that AI Gateway relies on to program Envoy.
+envoyGateway:
+  # The namespace where the Envoy Gateway controller is installed.
+  namespace : envoy-gateway-system
+  # The name of the service account used by the Envoy Gateway controller.
+  serviceAccountName: envoy-gateway
```
saileshd1402: I think the name is hardcoded in envoy gateway helm charts, we should probably also hardcode it:
https://github.com/envoyproxy/gateway/blob/70af785fba094929bc6044a57470d429205c4d5e/charts/gateway-helm/templates/envoy-gateway-serviceaccount.yaml#L4

mathetake: good catch

saileshd1402:
LGTM!

mathetake:
cool e2e tests passing

PR #1382, opened 2025-10-16
fix(metrics): correctly instantiate per-request scope metrics

mathetake:
**Description**

Previously, the metrics struct instance, for example, ChatCompletionMetrics interface implementation was shared across all requests, and only instantiated once at the extproc startup time. However, that implementation has request-scope fields such as model name, request start, etc. In other words, there was a serious race condition in them. This fixes it by instantiating these metrics interface implementations per-request.


aabchoo (approved):
that's crazy 👍 lgtm

mathetake:
i think i need to add a proper race test with parallel requests...

codefromthecrypt:
I have an idea failing test.. I think the easiest way is to make a request that takes longer than another to complete (using sync barriers in a fake httptest server). then it should be simple as the issue doesn't need concurrency to show iiuc

Issue #1083, opened 2025-08-15
Refactor processors code to share common logic


mathetake:
As you see in #1022, currently adding a new processor involves "a lot" of copy paste of the existing infra. chat_completions, embeddings, and messages processors are all sharing majority of request flow handling which is compatible with the control plane's configuration.

We need to make it much more easier since we are looking to add support more processors both for passthrough and translations.

mathetake:
let's see copilot is strong enough to be useful here

PR #1383, opened 2025-10-17
fix: do not modify original request mody

mathetake:
**Description**

OpenAI /chat/completions passthrough translator has unnecessarily modified the original request body. This removes it and fixes the bug like the case where after the fallback, original model name was no longer maintained in the struct, which in turn caused the metrics label corruption and the fallback model name incorrectness.

Supersedes part of #1376


internal/extproc/translator/openai_openai.go
```diff
@@ -62,7 +62,6 @@ func (o *openAIToOpenAITranslatorV1ChatCompletion) RequestBody(original []byte,
 		}
 		// Make everything coherent.
 		o.requestModel = o.modelNameOverride
-		req.Model = o.modelNameOverride
```
mathetake, line 65: Note: This parsed struct is in reality read only, and this was the only place where we accidentally modified -- I checked all the reference of the .Model field in the codebase. Sad that Go doesn't have an immutable keyword at the compiler level.

AyushSawant18588 (approved):
lgtm

internal/extproc/translator/openai_openai.go
aabchoo: May need to update this line: https://github.com/envoyproxy/ai-gateway/blob/main/internal/extproc/chatcompletion_processor.go#L238

```
reqModel := cmp.Or(c.requestHeaders[internalapi.ModelNameHeaderKeyDefault], c.modelNameOverride, c.originalRequestBody.Model)
```

aabchoo: jk this change gets called after the line I linked

edit: might still need to update it.. lemme think

mathetake: i think the code should be fine (the processor code has become so complex and refactoring is the pressing issue... #1083...)

PR #1385, opened 2025-10-17
chore: removes envoy installation requirement for running extproc tests

mathetake:
**Description**
This removes one of the test requirement which is Envoy by leveraging func-e via go tool. Since Envoy v1.36 was released, so this also upgrades the envoy version used in the test to v1.36


CONTRIBUTING.md
```diff
@@ -43,12 +43,6 @@ For example,
 
 - The latest `kubectl` binary for running `make test-e2e`.
```
mathetake, line 44: cc @anuraaga do you have any good idea how to remove kubectl requirement in tests? I think kubectl seems like it's possible to make it "tool" like ignorepaths script where we can leverage exposed main https://github.com/kubernetes/kubernetes/blob/master/cmd/kubectl/kubectl.go

mathetake: This makes the project completely dependency free & only Go is the requirement

mathetake: *except for `make` ..

.github/workflows/build_and_test.yaml
```diff
@@ -140,16 +140,11 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        # Note: we cannot run the latest Envoy version on macOS due to https://github.com/tetratelabs/archive-envoy/issues/12.
-        # Once it's supported, the following "binary installation" steps below can be just removed and
-        # we can simply exec.Cmd with "go tool -modfile=tools/go.mod func-e run" with the envoy version configured via ENVOY_VERSION env var.
```
mathetake, line 145: we no longer need to test on Envoy latest, which made this PR possible

CONTRIBUTING.md
anuraaga: AFAIK it's not possible since they use `replace`

https://github.com/kubernetes/kubernetes/blob/master/go.mod#L225

mathetake: damn

tests/extproc/extproc_test.go
```diff
@@ -59,9 +59,7 @@ var (
 		Region:      "gcp-region",
 		ProjectName: "gcp-project-name",
 	}}}
-	// This always failing backend is configured to have AWS Bedrock schema so that
-	// we can test that the extproc can fallback to the different schema. E.g. Primary AWS and then OpenAI.
-	alwaysFailingBackend = filterapi.Backend{Name: "always-failing-backend", Schema: awsBedrockSchema}
+	alwaysFailingBackend = filterapi.Backend{Name: "always-failing-backend", Schema: openAISchema}
```
mathetake, line 62: This is very hard to explain but this was accidentally working until v1.36 which makes it possible for extproc return the immediate response correctly, so this makes the embedding endpoint test not working: AWS schema doesn't have support for embedding -> try returning immediate response -> previously it was retry-able (due to timeout) vs not retry-able due to extpric immediate response.

In practice, this schema diff retry test case is already covered in the e2e test so no harm

.envoy-version
mathetake, line 1: note: This file is what func-e honors as you can see in the help message by `go tool -modfile=tools/go.mod func-e help run`

.github/workflows/build_and_test.yaml
mathetake: also i realized that latest installation had been broken since one-two months ago due to the upstream repo change, so it was useless anyways

mathetake:
ready for YOLO

PR #1387, opened 2025-10-17
fix: reset previously set/removed values in per-backend header mutation

mathetake:
**Description**

This fixes a bug in the per-backend header mutator where the previous mutation in retry case causes incorrect header presence, absence or incorrect header values.


tests/extproc/testupstream_test.go
```diff
@@ -61,10 +61,6 @@ func TestWithTestUpstream(t *testing.T) {
 			testUpstreamAzureBackend,
 			testUpstreamGCPVertexAIBackend,
 			testUpstreamGCPAnthropicAIBackend,
-			// TODO: this shouldn't be needed. The previous per-backend headers shouldn't affect the subsequent retries.
```
mathetake, line 64: This TODO summarizes what was happening with the bug being fixed here

tests/internal/testenvironment/test_environment.go
```diff
@@ -262,8 +262,8 @@ func requireEnvoy(t testing.TB,
 		"--concurrency", strconv.Itoa(max(runtime.NumCPU(), 2)),
 		// This allows multiple Envoy instances to run in parallel.
 		"--base-id", strconv.Itoa(time.Now().Nanosecond()),
-		// Add debug logging for extproc.
-		"--component-log-level", "ext_proc:trace,http:debug,connection:debug",
+		// Add debug logging for http.
+		"--component-log-level", "http:debug",
```
mathetake, line 266: ext_proc trace is just a noise as well as connection stuff as well

internal/extproc/embeddings_processor_test.go
```diff
@@ -557,86 +556,11 @@ func TestEmbeddingsProcessorUpstreamFilter_ProcessRequestHeaders_WithHeaderMutat
 		require.Equal(t, "custom-value", headers["x-custom"])
 	})
 
-	t.Run("header mutations restored on retry", func(t *testing.T) {
```
mathetake, line 560: this simply re-tests the headermutator internal so it's overlapping with the tests there, hence removed it.

internal/extproc/messages_processor_test.go
```diff
@@ -769,100 +769,6 @@ func TestMessagesProcessorUpstreamFilter_ProcessRequestHeaders_WithHeaderMutatio
 		require.Equal(t, "custom-value", headers["x-custom"])
 	})
 
-	t.Run("header mutations restored on retry", func(t *testing.T) {
```
mathetake, line 772: same

mathetake:
ok ready to go!

internal/extproc/headermutator/header_mutator.go
```diff
@@ -12,10 +12,11 @@ import (
 	extprocv3 "github.com/envoyproxy/go-control-plane/envoy/service/ext_proc/v3"
 
 	"github.com/envoyproxy/ai-gateway/internal/filterapi"
+	"github.com/envoyproxy/ai-gateway/internal/internalapi"
 )
 
 type HeaderMutator struct {
-	// getOrignalHeaders Callback to get removed sensitive headers from the router filter.
+	// getOriginalHeaders Callback to get removed sensitive headers from the router filter.
```
aabchoo: this comment seems inaccurate

internal/extproc/headermutator/header_mutator.go
```diff
@@ -62,21 +69,65 @@ func (h *HeaderMutator) Mutate(headers map[string]string, onRetry bool) *extproc
...
 			}
 		}
+		// 1. Remove any headers that were added in the previous attempt (not part of original headers and not being set now).
+		// 2. Restore any original headers that were modified in the previous attempt (and not being set now).
+		for key := range headers {
+			key = strings.ToLower(key)
+			if shouldIgnoreHeader(key) {
+				continue
+			}
+			if _, set := setHeadersSet[key]; set {
```
aabchoo: nit: keep consistent
```suggestion
			if _, isSet := setHeadersSet[key]; isSet {
```

internal/extproc/headermutator/header_mutator.go
```diff
@@ -62,21 +69,65 @@ func (h *HeaderMutator) Mutate(headers map[string]string, onRetry bool) *extproc
...
+			}
+			originalValue, exists := h.originalHeaders[key]
+			if !exists {
+				delete(headers, key)
+				headerMutation.RemoveHeaders = append(headerMutation.RemoveHeaders, key)
+			} else {
+				// Restore original value.
+				headers[key] = originalValue
+				headerMutation.SetHeaders = append(headerMutation.SetHeaders, &corev3.HeaderValueOption{
+					Header: &corev3.HeaderValue{Key: key, RawValue: []byte(originalValue)},
```
aabchoo, line 105: In previous `SetHeaders`, you used the original key and not the lowercased key. ie h instead of strings.ToLower(h). 

Is there a reason why you decided to do it differently here?

internal/extproc/messages_processor_test.go
```diff
@@ -1043,6 +949,6 @@ func TestMessagesProcessorUpstreamFilter_SetBackend_WithHeaderMutations(t *testi
 		require.Equal(t, []byte("original-value"), restoredHeader.Header.RawValue)
 		require.Equal(t, "original-value", testHeaders["x-custom"])
 		// x-existing should not be restored because it already exists.
```
aabchoo: nit: comment

aabchoo (approved):
overall looks good to me

johnugeorge:
Cc @siddharth1036

internal/extproc/headermutator/header_mutator.go
```diff
@@ -62,21 +69,65 @@ func (h *HeaderMutator) Mutate(headers map[string]string, onRetry bool) *extproc
...
 	return headerMutation
 }
+
+// shouldIgnoreHeader returns true if the header key should be ignored for mutation.
+//
+// Skip Envoy AI Gateway headers since some of them are populated after the originalHeaders are captured.
+// This should be safe since these headers are managed by Envoy AI Gateway itself, not expected to be
+// modified by users via header mutation API.
+//
+// Also, skip Envoy pseudo-headers beginning with ':'.
```
siddharth1036: Should we add example of pseudo headers in comments? I am assuming these are :path, etc

internal/extproc/headermutator/header_mutator.go
```diff
@@ -62,21 +69,65 @@ func (h *HeaderMutator) Mutate(headers map[string]string, onRetry bool) *extproc
...
+func shouldIgnoreHeader(key string) bool {
+	// Ignore Envoy pseudo-headers beginning with ':'.
+	if strings.HasPrefix(key, ":") {
+		return true
+	}
+	// Ignore internal headers beginning with Envoy AI Gateway prefix.
+	if strings.HasPrefix(key, internalapi.EnvoyAIGatewayHeaderPrefix) {
```
siddharth1036: Just wondering if case would be a problem here, as this might be changed to lower case on retries

internal/extproc/headermutator/header_mutator.go
```diff
@@ -62,21 +69,65 @@ func (h *HeaderMutator) Mutate(headers map[string]string, onRetry bool) *extproc
...
+		return true
+	}
+	return false
+}
```
siddharth1036: nit
```suggestion
	// Ignore Envoy pseudo-headers beginning with ':'.
	// Ignore internal headers beginning with Envoy AI Gateway prefix.
	return strings.HasPrefix(key, ":") || strings.HasPrefix(key, internalapi.EnvoyAIGatewayHeaderPrefix) 
}
```

internal/extproc/headermutator/header_mutator.go
```diff
@@ -62,21 +69,65 @@ func (h *HeaderMutator) Mutate(headers map[string]string, onRetry bool) *extproc
...
+				setHeadersSet[key] = struct{}{}
 				headerMutation.SetHeaders = append(headerMutation.SetHeaders, &corev3.HeaderValueOption{
 					Header: &corev3.HeaderValue{Key: h, RawValue: []byte(v)},
 				})
 			}
 		}
+		// 1. Remove any headers that were added in the previous attempt (not part of original headers and not being set now).
+		// 2. Restore any original headers that were modified in the previous attempt (and not being set now).
+		for key := range headers {
+			key = strings.ToLower(key)
```
siddharth1036: Just TOL, should we merge this loop and the loop over original headers? If I understand this loop (if no removals or additions) will make headers equal to orignalHeaders. Should we start with headers = originalHeaders and go ahead with remaining mutations? 
Initially I skipped this to avoid any header that was already added and was different from original header before entering mutation but as we are making them same now just wondering if extra loop is required

internal/extproc/headermutator/header_mutator.go
mathetake: case doesn't matter here as it's internal header

internal/extproc/headermutator/header_mutator.go
mathetake: Let's do it in a follow up later

Issue #568, opened 2025-04-09
IAM roles for service accounts integration with AWS Bedrock auth


shivam-dubey-1:
*Description*:
>What issue is being seen? Describe what should be happening instead of
The AI Gateway's BackendSecurityPolicy CRD doesn't properly support IRSA (IAM Roles for Service Accounts) for AWS authentication. When attempting to use a service account with an associated IAM role for AWS Bedrock access, the configuration fails with validation errors.

Current behavior: The BackendSecurityPolicy CRD validation rejects IRSA configuration.
Expected behavior: The BackendSecurityPolicy should support IRSA configuration for AWS authentication.

*Repro steps*:
>
eksctl create iamserviceaccount \
  --name bedrock-sa \
  --namespace default \
  --cluster cluster_name \
  --region us-east-1 \
  --attach-policy-arn arn:aws:iam::aws:policy/AmazonBedrockFullAccess \
  --approve

apiVersion: aigateway.envoyproxy.io/v1alpha1
kind: BackendSecurityPolicy
metadata:
  name: envoy-ai-gateway-basic-aws-credentials
  namespace: default
spec:
  type: AWSCredentials
  awsCredentials:
    region: us-east-1
    serviceAccount:
      name: bedrock-sa

Error from server (BadRequest): error when creating "bedrock.yaml": BackendSecurityPolicy in version "v1alpha1" cannot be handled as a BackendSecurityPolicy: strict decoding error: unknown field "spec.awsCredentials.serviceAccount"

shivam-dubey-1:
Is there any workaround for not using secrets(aws credentials) for bedrock?

mathetake:
Unfortunately no currently. But you are welcome to contribute the necessary change if that's your priority

PR #1394, opened 2025-10-20
feat: aws auth sdk default chain

seluard:
**Description**

This PR updates the AWS authentication handler to support the AWS SDK’s default credential chain, enabling native integration with modern AWS identity providers like IRSA (IAM Roles for Service Accounts) and EKS Pod Identity.

Previously, credentials had to be explicitly provided via config literals. Now, if no credential file is set, the handler automatically falls back to config.LoadDefaultConfig, which supports:

- IRSA (EKS service accounts)
- EKS Pod Identity
- EC2 Instance Profiles
- Environment variables (AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY)
- Shared credentials files (~/.aws/credentials)

This simplifies AWS deployments and aligns with best practices by avoiding static credentials.
This makes it much easier to deploy the AI Gateway on AWS infrastructure while
following AWS security best practices by avoiding static credentials.

The PR includes:

- Refactored AWS authentication handler to use the default credential chain when no credentials file is provided
- Added example configs for IRSA and EKS Pod Identity
- Updated documentation with setup guides for both methods
- Extended tests to cover the new credential resolution logic
- Update API documentation

Closes https://github.com/envoyproxy/ai-gateway/issues/568

Special notes for reviewers:

The implementation maintains backward compatibility - existing configurations
using credential file literals will continue to work exactly as before. The
default credential chain is only used when CredentialFileLiteral is not provided
(empty/nil).

I've test that Pod Identity association works, but I've not clusters setup with IRSA in place.

When run test locally, seems like some in `cmd/aigw` are e2e, which leeds to failing, i've disable those locally.

### DIFF
--- a/cmd/aigw/testdata/translate_basic.out.yaml
+++ b/cmd/aigw/testdata/translate_basic.out.yaml
@@ -182,8 +182,7 @@ metadata:
 spec:
   directResponse:
     body:
-      inline: No matching route found. It is likely that the model specified your
-        request is not configured in the Gateway.
+      inline: No matching route found. It is likely because the model specified in your request is not configured in the Gateway.
       type: null
     statusCode: 404
 ---
--- a/internal/controller/ai_gateway_route.go
+++ b/internal/controller/ai_gateway_route.go
@@ -146,7 +146,7 @@ func generateHTTPRouteFilters(aiGatewayRoute *aigv1a1.AIGatewayRoute) []*egv1a1.
 					Body: &egv1a1.CustomResponseBody{
 						Inline: ptr.To(
 							// "Likely" since the matching rule can be arbitrary, not necessarily matching on the model name.
-							`No matching route found. It is likely that the model specified your request is not configured in the Gateway.`,
+							`No matching route found. It is likely because the model specified in your request is not configured in the Gateway.`,
 						),
 					},
 				},
--- a/tests/e2e/testupstream_test.go
+++ b/tests/e2e/testupstream_test.go
@@ -83,7 +83,7 @@ func TestWithTestUpstream(t *testing.T) {
 				name:            "openai",
 				modelName:       "non-existent-model",
 				expStatus:       404,
-				expResponseBody: `No matching route found. It is likely that the model specified your request is not configured in the Gateway.`,
+				expResponseBody: `No matching route found. It is likely because the model specified in your request is not configured in the Gateway.`,
 			},
 			{
 				name:               "openai-header-mutation",


mathetake:
this is nice!

internal/controller/backend_security_policy.go
```diff
@@ -86,9 +86,24 @@ func (c *BackendSecurityPolicyController) reconcile(ctx context.Context, bsp *ai
...
-		bsp.Spec.Type != aigv1a1.BackendSecurityPolicyTypeAnthropicAPIKey {
+		bsp.Spec.Type != aigv1a1.BackendSecurityPolicyTypeAnthropicAPIKey
+
+	// Skip rotation for AWS when neither credentials file nor OIDC exchange is configured
+	// This allows IRSA/Pod Identity to work via the default credential chain
+	if bsp.Spec.Type == aigv1a1.BackendSecurityPolicyTypeAWSCredentials {
+		if bsp.Spec.AWSCredentials != nil &&
+			bsp.Spec.AWSCredentials.CredentialsFile == nil &&
+			bsp.Spec.AWSCredentials.OIDCExchangeToken == nil {
+			c.logger.Info("Using default AWS credential chain (IRSA/Pod Identity), skipping rotation",
```
secustor, line 100: Should be probably debug at least. 

This will probably printed for each request

seluard: This it's happening at controller level, not at dataplane.

seluard:
As I did not put on review yet, I've just force push with rebase before doing it.

mathetake (commented):
https://github.com/envoyproxy/ai-gateway/blob/7b141fdd230c703fdb1dc9e4603540b12e7bef09/api/v1alpha1/backendsecurity_policy.go#L271 

could you write up a nice documentation comment here on the API as well?

internal/controller/gateway.go
```diff
@@ -380,8 +380,21 @@ func (c *GatewayController) bspToFilterAPIBackendAuth(ctx context.Context, backe
 		}
 		return &filterapi.BackendAuth{AnthropicAPIKey: &filterapi.AnthropicAPIKeyAuth{Key: apiKey}}, nil
 	case aigv1a1.BackendSecurityPolicyTypeAWSCredentials:
+		awsCred := backendSecurityPolicy.Spec.AWSCredentials
+
+		// If no credentials file or OIDC token is configured, use default credential chain
+		// This allows IRSA/Pod Identity to work automatically
+		if awsCred.CredentialsFile == nil && awsCred.OIDCExchangeToken == nil {
```
mathetake, line 387: `awsCred.CredentialsFile == nil && awsCred.OIDCExchangeToken == nil` this statement is used many times so how about having it as a method in a file in `api/v1alpha1/backendsecurity_policy_helper.go`

internal/controller/gateway.go
```diff
@@ -380,8 +380,21 @@ func (c *GatewayController) bspToFilterAPIBackendAuth(ctx context.Context, backe
...
+			return &filterapi.BackendAuth{
```
mathetake, line 388: can you add a unit test case that reaches this

internal/extproc/backendauth/aws.go
```diff
@@ -34,43 +34,55 @@ type awsHandler struct {
...
+			config.WithSharedCredentialsFiles([]string{name}),
+			config.WithRegion(awsAuth.Region),
+		)
+		if err != nil {
+			return nil, fmt.Errorf("cannot load from credentials file: %w", err)
 		}
 	} else {
-		return nil, fmt.Errorf("aws auth configuration is required")
+		// Use default credential chain (supports IRSA, EKS Pod Identity, etc.)
+		cfg, err = config.LoadDefaultConfig(
```
mathetake, line 69: one question: does this mean that we only load the identity once at this phase and after this point, it won't reload? Or, at signing time the SDK will reload credentials if the aws session token is updated?

seluard: 🤔 .. you're totally right, seems like the credentials will be loaded just one from the provider. Let me update this to  load the provider once and save at the `awsHandler` struct, then at `Do` phase , retrieve credentials from that provider will use the right ones and SDK will handle it automatically.

internal/controller/gateway.go
seluard: > `awsCred.CredentialsFile == nil && awsCred.OIDCExchangeToken == nil` this statement is used many times so how about having it as a method in a file in `api/v1alpha1/backendsecurity_policy_helper.go`

I've just found this occurrency. Do you think it's worth to create a helper?

mathetake: sorry nvm

internal/extproc/backendauth/aws.go
```diff
@@ -115,7 +122,12 @@ func (a *awsHandler) Do(ctx context.Context, requestHeaders map[string]string, h
 	// https://github.com/envoyproxy/envoy/blob/60b2b5187cf99db79ecfc54675354997af4765ea/source/extensions/filters/http/ext_proc/processor_state.cc#L180-L183
 	req.ContentLength = -1
 
-	err = a.signer.SignHTTP(ctx, a.credentials, req,
+	credentials, err := a.credentialsProvider.Retrieve(ctx)
```
mathetake, line 125: how costly is this `credentialsProvider.Retrieve` is my next question.. Previously, aws secrets are rotated on OIDC credentials rotation on the controller side and this newHandler was called on the update.

One another way to resolve this is to return "RequeueAfter" in this case as well 
https://github.com/envoyproxy/ai-gateway/blob/8ee7ae12c3a4534a615d704779e00ef0fb70990f/internal/controller/backend_security_policy.go#L273

which eventually propagates to the ExtProc's secret update hence the newHandler will be called. Maybe we can requeue after 1 minute or so assuming any pod will have more than 1 minute ttl on the session token. 

wdyt?

mathetake: ~yeah `Credentials.Retrieve` seems not goroutine safe~  It seems safe

mathetake: @seluard i am fine with the current change as-is as long as we can be sure that there's almost zero impact on the existing static credential case.

Could you do a quick bench mark on this Do method to compare the results (in line Retrieve or one time Retrieve for static credentials)

seluard: I've just add a bench test, those are the results I got : 

```
❯ go test -bench=BenchmarkAWSHandler_Do -benchmem -tags=benchmark ./internal/extproc/backendauth/
goos: darwin
goarch: arm64
pkg: github.com/envoyproxy/ai-gateway/internal/extproc/backendauth
cpu: Apple M1 Pro
BenchmarkAWSHandler_Do/file_credentials-8                 292096              3909 ns/op            4770 B/op         56 allocs/op
BenchmarkAWSHandler_Do/default_chain_env_credentials-8            261584              3966 ns/op            4759 B/op         56 allocs/op
BenchmarkAWSHandler_Do/concurrent-8                               429165              2577 ns/op            5474 B/op         67 allocs/op
BenchmarkAWSHandler_Do/just_credential_retrieve-8               49032210                24.35 ns/op            0 B/op          0 allocs/op
PASS
ok      github.com/envoyproxy/ai-gateway/internal/extproc/backendauth   11.375s
```

Performance analysis shows that both file-based and environment-based credentials have
nearly identical performance (~3.9-4.0 μs per signing operation) because:

1. The AWS SDK v2 caches credentials internally after the first retrieval
2. Calling Retrieve() on cached credentials is extremely fast (24.35 ns, 0 allocations)
3. Credential retrieval overhead is <0.7% of total signing time (24ns / 3900ns)

So seems to be safe to call Retrieve() on every request - the SDK's caching makes it
essentially free. Both credential sources benefit equally from caching mechanism.

mathetake: Thank you so much!!

internal/extproc/backendauth/aws_bench_test.go
```diff
@@ -0,0 +1,109 @@
+// Copyright Envoy AI Gateway Authors
+// SPDX-License-Identifier: Apache-2.0
+// The full text of the Apache license is available in the LICENSE file at
+// the root of the repo.
+
+//go:build benchmark
```
mathetake: I don't think we need build tag
```suggestion
```

PR #1395, opened 2025-10-20
feat: optionally enable pprof endpoint in extproc

mathetake:
**Description**

This optionally enables pprof endpoint in order to troubleshoot and debug the running application easily, including lock contentions.


cmd/extproc/main.go
```diff
@@ -17,6 +19,16 @@ import (
 )
 
 func main() {
+	// Run the pprof server if the ENABLE_PPROF environment variable is non-empty.
+	if os.Getenv("ENABLE_PPROF") != "" {
```
mathetake: i think it should be fine to enable by default as the impact should be negligible. wdyt all

mathetake: cc @envoyproxy/ai-gateway-maintainers

yuzisun: Are we able to configure this extproc env via the helm chart ?

mathetake: Yes as you see the comment addition in the helm chart values in this PR

wengyao04:
Hi @mathetake do you think if we can track activeStream like this and log or publish the activeStream metrics, I am not quite sure if there is goroutine leak

```
var activeStreams int64

func (s *Server) Process(stream extprocv3.ExternalProcessor_ProcessServer) error {
    atomic.AddInt64(&activeStreams, 1)
    defer atomic.AddInt64(&activeStreams, -1)
    // ...
}
```

mathetake:
It should be already tracked in envoy I guess? In anyways let's discuss outside this PR as it's irrelevant here

mathetake:
Enabling it by default

cmd/extproc/main.go
```diff
@@ -17,6 +20,27 @@ import (
...
+	// The impact should be negligible when the actual pprof endpoints are not being accessed.
+	if _, ok := os.LookupEnv("DISABLE_PPROF"); !ok {
+		go func() {
+			const pprofPort = "6060" // The same default port as in th Go pprof documentation.
+			mux := http.NewServeMux()
+			mux.HandleFunc("/debug/pprof/", pprof.Index)
+			mux.HandleFunc("/debug/pprof/cmdline", pprof.Cmdline)
+			mux.HandleFunc("/debug/pprof/profile", pprof.Profile)
+			mux.HandleFunc("/debug/pprof/symbol", pprof.Symbol)
+			mux.HandleFunc("/debug/pprof/trace", pprof.Trace)
```
nacx: Does it make sense to add something like this, to enable profiles like `heap`, `goroutines`, etc?

```go
mux.HandleFunc("/debug/pprof/{profile}", func(w http.ResponseWriter, r *http.Request) {
    pprof.Handler(r.PathValue("profile")).ServeHTTP(w, r)
})
```

cmd/extproc/main.go
```diff
@@ -17,6 +20,27 @@ import (
...
+			server := &http.Server{Addr: ":" + pprofPort, Handler: mux, ReadHeaderTimeout: 5 * time.Second}
+			log.Printf("starting pprof server on port %s", pprofPort)
+			if err := server.ListenAndServe(); err != nil && errors.Is(err, http.ErrServerClosed) {
+				log.Printf("pprof server stopped: %v", err)
+			}
```
nacx, line 40: Can we move this pprof server to `internal/pprof.go` so that it can be reused? Thi can be useful for the controller as well.

yuzisun: Hi @nacx I am going to merge this as we have a urgent issue to use this for debugging, will open an PR to move this later

PR #1396, opened 2025-10-20
feat: gcp anthropic cache support

alexagriffith:
**Description**

This PR adds support for Anthropic's cache_control feature to the OpenAI to GCP Anthropic translator. Users can now enable caching on a per-message basis by adding a cache field to their OpenAI API requests.

### DIFF
--- a/cmd/extproc/mainlib/main.go
+++ b/cmd/extproc/mainlib/main.go
@@ -26,6 +26,7 @@ import (
 	"google.golang.org/grpc/health/grpc_health_v1"
 
 	"github.com/envoyproxy/ai-gateway/internal/extproc"
+	"github.com/envoyproxy/ai-gateway/internal/filterapi"
 	"github.com/envoyproxy/ai-gateway/internal/internalapi"
 	"github.com/envoyproxy/ai-gateway/internal/mcpproxy"
 	"github.com/envoyproxy/ai-gateway/internal/metrics"
@@ -275,12 +276,12 @@ func Main(ctx context.Context, args []string, stderr io.Writer) (err error) {
 	server.Register(path.Join(flags.rootPrefix, *endpointPrefixes.OpenAI, "/v1/models"), extproc.NewModelsProcessor)
 	server.Register(path.Join(flags.rootPrefix, *endpointPrefixes.Anthropic, "/v1/messages"), extproc.MessagesProcessorFactory(messagesMetrics))
 
-	if watchErr := extproc.StartConfigWatcher(ctx, flags.configPath, server, l, time.Second*5); watchErr != nil {
+	if watchErr := filterapi.StartConfigWatcher(ctx, flags.configPath, server, l, time.Second*5); watchErr != nil {
 		return fmt.Errorf("failed to start config watcher: %w", watchErr)
 	}
 
 	// Create and register gRPC server with ExternalProcessorServer (the service Envoy calls).
-	if err = extproc.StartConfigWatcher(ctx, flags.configPath, server, l, time.Second*5); err != nil {
+	if err = filterapi.StartConfigWatcher(ctx, flags.configPath, server, l, time.Second*5); err != nil {
 		return fmt.Errorf("failed to start config watcher: %w", err)
 	}
 
@@ -295,7 +296,7 @@ func Main(ctx context.Context, args []string, stderr io.Writer) (err error) {
 		if err != nil {
 			return fmt.Errorf("failed to create MCP proxy: %w", err)
 		}
-		if err = extproc.StartConfigWatcher(ctx, flags.configPath, mcpProxyConfig, l, time.Second*5); err != nil {
+		if err = filterapi.StartConfigWatcher(ctx, flags.configPath, mcpProxyConfig, l, time.Second*5); err != nil {
 			return fmt.Errorf("failed to start config watcher: %w", err)
 		}
 
--- a/internal/controller/mcp_route_security_policy.go
+++ b/internal/controller/mcp_route_security_policy.go
@@ -471,7 +471,7 @@ func (c *MCPRouteController) buildOAuthAuthServerMetadataJSON(oauth *aigv1a1.MCP
 
 	// Try to fetch metadata from the well-known endpoint first.
 	if authServer != "" {
-		fetchedMetadata, err := fetchOAuthAuthServerMetadata(authServer)
+		fetchedMetadata, err := fetchOAuthAuthServerMetadata(authServer, maxRetryElapsedTime)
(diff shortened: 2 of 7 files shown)


mathetake:
@xiaolin593  @sukumargaonkar could you take a look

internal/apischema/openai/openai.go
```diff
@@ -431,7 +436,8 @@ type ChatCompletionUserMessageParam struct {
 	Role string `json:"role"`
 	// An optional name for the participant. Provides the model information to
 	// differentiate between participants of the same role.
-	Name string `json:"name,omitempty"`
+	Name  string      `json:"name,omitempty"`
+	Cache *CacheValue `json:"cache,omitempty"`
```
yuzisun: Let's a comment for this field as it is anthropic specific

alexagriffith: fixed

internal/apischema/openai/openai.go
```diff
@@ -432,7 +437,8 @@ type ChatCompletionUserMessageParam struct {
 	Role string `json:"role"`
 	// An optional name for the participant. Provides the model information to
 	// differentiate between participants of the same role.
-	Name string `json:"name,omitempty"`
+	Name                    string `json:"name,omitempty"`
+	*AnthropicMessageFields `json:",inline,omitempty"`
```
yuzisun: from the example, looks like the cache_control is at the individual content level.
https://docs.claude.com/en/docs/build-with-claude/prompt-caching#prompt-caching-examples
also see litellm https://docs.litellm.ai/docs/providers/anthropic#caching---large-context-caching

internal/apischema/openai/openai.go
```diff
@@ -473,7 +481,8 @@ type ChatCompletionToolMessageParam struct {
 	// The role of the messages author, in this case `tool`.
 	Role string `json:"role"`
 	// Tool call that this message is responding to.
-	ToolCallID string `json:"tool_call_id"`
+	ToolCallID              string `json:"tool_call_id"`
+	*AnthropicMessageFields `json:",inline,omitempty"`
```
yuzisun: I did not find doc saying tool message cache is supported. The tool definition `tools` is supported. 
https://docs.claude.com/en/docs/build-with-claude/prompt-caching#prompt-caching-examples
also see litelllm: https://docs.litellm.ai/docs/providers/anthropic#caching---tools-definitions

alexagriffith: the anthropic go sdk has cache option in OfTool

// The properties ToolUseID, Type are required.
type ToolResultBlockParam struct {
	ToolUseID string          `json:"tool_use_id,required"`
	IsError   param.Opt[bool] `json:"is_error,omitzero"`
	// Create a cache control breakpoint at this content block.
	CacheControl CacheControlEphemeralParam         `json:"cache_control,omitzero"`
	Content      []ToolResultBlockParamContentUnion `json:"content,omitzero"`
	// This field can be elided, and will marshal its zero value as "tool_result".
	Type constant.ToolResult `json:"type,required"`
	paramObj
}

alexagriffith: which is where we convert here right?

https://github.com/envoyproxy/ai-gateway/pull/1396/files/e30dc5f4190e4721e9ef8aa45feaa356a6a18e02

internal/apischema/openai/openai.go
alexagriffith: I see what you mean, there are two cases where it can be a list from the openai message

alexagriffith: but our Content isn't structured - its like

```

type StringOrUserRoleContentUnion struct {
	Value any
}
```

alexagriffith: fixed

internal/apischema/openai/openai.go
alexagriffith: oftool should be supported im thinking - tool messages are treated as general message types. added in updated code. based on the type we translate to, that anthropic type always has cachecontrol in it, so I think we need it in the request type as well.

alexagriffith: https://docs.claude.com/en/docs/build-with-claude/prompt-caching#what-can-be-cached

internal/translator/openai_gcpanthropic.go
```diff
@@ -693,28 +717,6 @@ func (o *openAIToGCPAnthropicTranslatorV1ChatCompletion) ResponseError(respHeade
 	return headerMutation, bodyMutation, nil
 }
 
-// anthropicToolUseToOpenAICalls converts Anthropic tool_use content blocks to OpenAI tool calls.
```
alexagriffith, line 636: I just moved it

mathetake:
@yuzisun could you take another review? also @sukumargaonkar mind taking a look ?

hustxiayang:
where is the usage info for cache? I guess we have such information?

hustxiayang:
Just noticed that the cache usage is set even before this PR. Never mind on my comment.

internal/translator/openai_gcpanthropic.go
```diff
@@ -700,18 +771,19 @@ func (o *openAIToGCPAnthropicTranslatorV1ChatCompletion) ResponseBody(_ map[stri
...
+	promptTokens := anthropicResp.Usage.InputTokens + anthropicResp.Usage.CacheReadInputTokens + anthropicResp.Usage.CacheCreationInputTokens
 	tokenUsage = LLMTokenUsage{
-		InputTokens:       uint32(anthropicResp.Usage.InputTokens),                                    //nolint:gosec
-		OutputTokens:      uint32(anthropicResp.Usage.OutputTokens),                                   //nolint:gosec
-		TotalTokens:       uint32(anthropicResp.Usage.InputTokens + anthropicResp.Usage.OutputTokens), //nolint:gosec
-		CachedInputTokens: uint32(anthropicResp.Usage.CacheReadInputTokens),                           //nolint:gosec
+		InputTokens:       uint32(promptTokens),                                                                            //nolint:gosec
+		OutputTokens:      uint32(anthropicResp.Usage.OutputTokens),                                                        //nolint:gosec
+		TotalTokens:       uint32(promptTokens + anthropicResp.Usage.OutputTokens),                                         //nolint:gosec
+		CachedInputTokens: uint32(anthropicResp.Usage.CacheReadInputTokens + anthropicResp.Usage.CacheCreationInputTokens), //nolint:gosec
```
yuzisun, line 770: I think on the openai token usage it is only for the cache read input tokens.

alexagriffith: discussed offline - should be there

yuzisun (approved):

PR #1397, opened 2025-10-21
docs: metrics enrichment with custom headers

ehrnst:
**Description**

https://github.com/envoyproxy/ai-gateway/pull/981/files

After inspecting this PR I found out how to add custom headers as metric labels. This is a small update to the documentation on how to pass the configuration during install.

Related PR: #981

### DIFF
--- a/site/docs/capabilities/observability/metrics.md
+++ b/site/docs/capabilities/observability/metrics.md
@@ -40,9 +40,13 @@ Each metric comes with some default attributes such as:
 - `gen_ai.request.model` - The model name requested (may be overridden)
 - `gen_ai.response.model` - The model name returned in the response
 - `gen_ai.provider.name` - The provider name (e.g., `openai`, `anthropic`)
+
 :::tip
+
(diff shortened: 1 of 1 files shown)


site/docs/capabilities/observability/metrics.md
mathetake:
```suggestion
:::tip
You can enrich the metrics with custom labels extracted from HTTP request headers. It can be configured via the `controller.metricsRequestHeaderAttributes` helm installation value. Please refer to [values.yaml](https://github.com/envoyproxy/ai-gateway/blob/main/manifests/charts/ai-gateway-helm/values.yaml) for more details including other configurations.
:::
```

site/docs/capabilities/observability/metrics.md
```diff
@@ -107,3 +107,38 @@ and then you would get the response like this, assuming you have made some reque
...
+kind: HelmRelease
+metadata:
+  name: envoy-ai-gateway
+spec:
+  values:
+    controller:
+      # Map HTTP request headers to Prometheus metric labels
+      # Format: "header1:label1,header2:label2"
+      metricsRequestHeaderLabels: "x-customer-id:customer_id,x-tenant-id:tenant_id"
+```
```
mathetake: I would rather have it as a single note above rather than a dedicated section like this since we do not want to maintain the whole another installation command here and elsewhere as well as teaching users how to use helm seems redundant as we assume at this point users have installed AIGW once in the installation doc. Plus, we already have the documentation in values.yaml

```suggestion
```

mathetake (approved):

PR #1400, opened 2025-10-21
feat: support response_format for gemini models other than 2.5 via translation from json schema to genai.Schema

hustxiayang:
**Description**
Current implementation of response_format for gemini models uses `responseJsonSchema`. However, in the official documentation, it's only supported for gemini2.5: https://ai.google.dev/gemini-api/docs/structured-output#json-schema. Thus, for other versions of gemini models, we need to use ``responseSchema`. Because the supported fields of `responseSchema` are different than standard json schema, we  also need to add  translations from json schema to `responseSchema`

https://github.com/envoyproxy/ai-gateway/issues/1403.

In this pr, we only do translations for response_format.  `Parameters` in `genai.FunctionDeclaration` is also a `genai.Schema`, but it's typically not complex and we would add translations for that late

### DIFF
--- a/internal/extproc/translator/gemini_helper_test.go
+++ b/internal/extproc/translator/gemini_helper_test.go
@@ -1532,7 +1532,7 @@ func TestOpenAIReqToGeminiGenerationConfigWithJsonSchemaToGemini(t *testing.T) {
 					},
 				},
 			},
-			requestModel: "gemini-pro", // older model that doesn't support native JSON schema
+			requestModel: "gemini-2.0-flash", // older model that doesn't support native JSON schema
 			expectedGenerationConfig: &genai.GenerationConfig{
 				ResponseMIMEType: "application/json",
 				ResponseSchema: &genai.Schema{
@@ -1586,7 +1586,7 @@ func TestOpenAIReqToGeminiGenerationConfigWithJsonSchemaToGemini(t *testing.T) {
 					},
 				},
 			},
-			requestModel: "gemini-pro", // older model
+			requestModel: "gemini-2.0-flash", // older model
 			expectedGenerationConfig: &genai.GenerationConfig{
 				ResponseMIMEType: "application/json",
 				ResponseSchema: &genai.Schema{
@@ -1643,7 +1643,7 @@ func TestOpenAIReqToGeminiGenerationConfigWithJsonSchemaToGemini(t *testing.T) {
 					},
 				},
 			},
-			requestModel: "gemini-pro",
+			requestModel: "gemini-2.0-flash",
 			expectedGenerationConfig: &genai.GenerationConfig{
 				ResponseMIMEType: "application/json",
 				ResponseSchema: &genai.Schema{
@@ -1682,7 +1682,7 @@ func TestOpenAIReqToGeminiGenerationConfigWithJsonSchemaToGemini(t *testing.T) {
 					},
 				},
 			},
-			requestModel: "gemini-pro",
+			requestModel: "gemini-2.0-flash",
 			expectedGenerationConfig: &genai.GenerationConfig{
 				ResponseMIMEType: "application/json",
 				ResponseSchema: &genai.Schema{
@@ -1725,7 +1725,7 @@ func TestOpenAIReqToGeminiGenerationConfigWithJsonSchemaToGemini(t *testing.T) {
 					},
 				},
 			},
-			requestModel: "gemini-pro",
+			requestModel: "gemini-2.0-flash",
 			expectedGenerationConfig: &genai.GenerationConfig{
 				ResponseMIMEType: "application/json",
 				ResponseSchema: &genai.Schema{
@@ -1760,7 +1760,7 @@ func TestOpenAIReqToGeminiGenerationConfigWithJsonSchemaToGemini(t *testing.T) {
 					},
 				},
 			},
-			requestModel:   "gemini-pro",
+			requestModel:   "gemini-2.0-flash",
 			expectedErrMsg: "invalid JSON schema",
 		},
 		{


go.mod
```diff
@@ -13,6 +13,7 @@ require (
 	github.com/aws/aws-sdk-go-v2/aws/protocol/eventstream v1.7.2
 	github.com/aws/aws-sdk-go-v2/config v1.31.13
 	github.com/aws/aws-sdk-go-v2/service/sts v1.38.7
+	github.com/barkimedes/go-deepcopy v0.0.0-20220514131651-17c30cfc62df
```
mathetake: drive-by: I would not accept a random direct dependency like this that is not well maintained or very popular either

internal/extproc/translator/jsonschema_helper.go
```diff
@@ -0,0 +1,388 @@
...
+	"required":    {},
+	"type":        {},
+	"nullable":    {},
+	"allOf":       {},
+	"anyOf":       {},
+	"not":         {},
+}
+
+// GcpConversionError defines a custom error type for consistent error handling.
+type GcpConversionError struct {
```
mathetake: let's not export unnecessarily

internal/extproc/translator/jsonschema_helper.go
```diff
@@ -0,0 +1,388 @@
...
+			keys = append(keys, nestedKeys...)
+		}
+	}
+
+	return keys, nil
+}
+
+// adapted from https://github.com/langchain-ai/langchain/blob/fce8caca16121024547fb0e8eb2d289c8f96396a/libs/core/langchain_core/utils/json_schema.py#L95
+// DereferenceRefsOptions holds optional parameters for the DereferenceRefs function.
+type DereferenceRefsOptions struct {
```
mathetake: same comment: unexport

internal/extproc/translator/jsonschema_helper.go
mathetake, line 1: can you put "jsonSchema" portion in the names of **every** jsonschema specific struct or function in this file? Go is not file scoped but directory scoped so namespace can be easily contaminated

hustxiayang:
@mathetake This PR is ready now. The above comments should be resolved and the suggestions should be followed. Thanks a lot!

internal/extproc/translator/gemini_helper.go
```diff
@@ -414,13 +414,22 @@ func openAIReqToGeminiGenerationConfig(openAIReq *openai.ChatCompletionRequest)
 		case openAIReq.ResponseFormat.OfJSONObject != nil:
 			gc.ResponseMIMEType = mimeTypeApplicationJSON
 		case openAIReq.ResponseFormat.OfJSONSchema != nil:
+			gc.ResponseMIMEType = mimeTypeApplicationJSON
 			var schemaMap map[string]any
 			if err := json.Unmarshal([]byte(openAIReq.ResponseFormat.OfJSONSchema.JSONSchema.Schema), &schemaMap); err != nil {
 				return nil, fmt.Errorf("invalid JSON schema: %w", err)
 			}
+			// it only works with gemini2.5 according to https://ai.google.dev/gemini-api/docs/structured-output#json-schema
+			if strings.Contains(openAIReq.Model, "gemini") && strings.Contains(openAIReq.Model, "2.5") { // TODO: users might be able to change the mapping?
```
yuzisun: let's make sure the model name here is the one after the mutation using modelNameOverride field as the original model alias may not have the version name.

internal/extproc/translator/jsonschema_helper.go
```diff
@@ -0,0 +1,408 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("invalid JSON schema: %w", err)
+	}
+
+	// Call the recursive helper function to perform the dereferencing.
+	processedRefs := make(map[string]struct{})
+	return jsonSchemaDereferenceHelper(schemaObj, schemaObj, skipKeys, &processedRefs)
+}
+
+// jsonSchemaToGapic formats a JSON schema for a gapic request, which is modified from https://github.com/langchain-ai/langchain-google/blob/06a5857841675461ae282648a155e1cd90d1e5d5/libs/vertexai/langchain_google_vertexai/functions_utils.py#L109
```
yuzisun: for my knowledge what does gapic stand for?

hustxiayang: > for my knowledge what does gapic stand for?

I think "google-api-python-client"

internal/extproc/translator/gemini_helper_test.go
```diff
@@ -872,6 +876,253 @@ func TestOpenAIReqToGeminiGenerationConfig(t *testing.T) {
...
+}`),
+						},
+					},
+				},
+			},
+			expectedGenerationConfig: &genai.GenerationConfig{
+				ResponseMIMEType: "application/json",
+				ResponseSchema: &genai.Schema{
+					Properties: map[string]*genai.Schema{
+						"final_answer": {Type: "string"},
```
sukumargaonkar: is this missing a `required` sections?

internal/extproc/translator/jsonschema_helper.go
```diff
@@ -0,0 +1,408 @@
...
+	"strings"
+
+	"google.golang.org/genai"
+)
+
+// get the supported field names
+func jsonSchemaGetAllowedFields() map[string]struct{} {
+	fieldSet := make(map[string]struct{})
+	var schema *genai.Schema
+	t := reflect.TypeOf(schema).Elem()
```
sukumargaonkar: i think the list of allowed fields is static, can we hardcode it instead
wanna avoid reflection.
field list: https://cloud.google.com/vertex-ai/generative-ai/docs/reference/rest/v1/Schema

also, i dont expect the list to change that often

internal/extproc/translator/jsonschema_helper.go
sukumargaonkar: GAPIC (Generated API Client)
https://github.com/googleapis/gapic-generator

internal/extproc/translator/jsonschema_helper.go
```diff
@@ -0,0 +1,408 @@
...
+				continue
+			}
+
+			// Handle reference key "$ref".
+			if k == "$ref" {
+				refPath, isString := v.(string)
+				if !isString {
+					return nil, fmt.Errorf("'$ref' value must be a string")
+				}
+				if _, ok := (*processedRefs)[refPath]; ok {
```
sukumargaonkar: i believe this is added to avoid recursive path reference
for now this silently ignore skips part of the schema when recursive paths exists.
lts better to throw error instead

internal/extproc/translator/jsonschema_helper.go
```diff
@@ -0,0 +1,408 @@
...
+				for pkey, pvalue := range properties {
+					if pSubSchema, ok := pvalue.(map[string]any); ok {
+						var err error
+						convertedSchema["properties"].(map[string]any)[pkey], err = jsonSchemaToGapic(pSubSchema, allowedSchemaFieldsSet)
+						if err != nil {
+							return nil, err
+						}
+					}
+				}
+			}
```
sukumargaonkar, line 397: throw error if properties is not a map to avoid silent failures

internal/extproc/translator/jsonschema_helper.go
```diff
@@ -0,0 +1,408 @@
...
+		case "$defs":
+			continue
+		case "items":
+			if subSchema, ok := value.(map[string]any); ok {
+				var err error
+				convertedSchema["items"], err = jsonSchemaToGapic(subSchema, allowedSchemaFieldsSet)
+				if err != nil {
+					return nil, err
+				}
+			}
```
sukumargaonkar, line 376: throw error if `!ok`

internal/extproc/translator/jsonschema_helper.go
```diff
@@ -0,0 +1,408 @@
...
+import (
+	"encoding/json"
+	"fmt"
+	"reflect"
+	"strings"
+
+	"google.golang.org/genai"
+)
+
+// get the supported field names
```
mathetake: please write complete sentences for any function documentation and what is "supported fields"? could you clarify?

```suggestion
// jsonSchemaGetAllowedFields gets ... foo bar.
```

internal/extproc/translator/jsonschema_helper.go
```diff
@@ -0,0 +1,408 @@
...
+func jsonSchemaGetAllowedFields() map[string]struct{} {
```
mathetake: and more importantly why do you need this? can we have a static global without calling this multiple times? also 99% of the times almost all use of reflect package is wrong and unnecessary. can we remove it?

internal/extproc/translator/jsonschema_helper.go
```diff
@@ -0,0 +1,408 @@
+// Copyright Envoy AI Gateway Authors
+// SPDX-License-Identifier: Apache-2.0
```
mathetake, line 2: could you add *_helper_test.go file and write unit tests for all of the functions in this file

mathetake:
> As described in https://github.com/envoyproxy/ai-gateway/issues/1403. Some other the implementation referred https://github.com/langchain-ai/langchain-google

@hustxiayang i hate to say this multiple times but i would appreciate it if you could avoid just "deribed in #12345" but briefly summarize your changes there in your words for the non trivial change like this. Imagine someone blaming the commit history. It's ok to refer to issue but just putting it would be usually not enough.

also can you clean up "TODO" comments?

internal/extproc/translator/jsonschema_helper.go
```diff
@@ -0,0 +1,408 @@
...
+						var err error
+						convertedSchema["properties"].(map[string]any)[pkey], err = jsonSchemaToGapic(pSubSchema, allowedSchemaFieldsSet)
+						if err != nil {
+							return nil, err
+						}
+					}
+				}
+			}
+		case "type":
+			if typeList, ok := value.([]any); ok {
```
sukumargaonkar: nit: switch case on the type might suite better
also that way if the type of `value` is not list or string you will throw an error.

```go
switch typeList := value.(type) {
case []any:
    if len(typeList) != 2 {
        msg := fmt.Sprintf("If the value of type is a list, the length of the list must be 2. Got %d.", len(typeList))
        return nil, fmt.Errorf("invalid JSON schema: %s", msg)
    }
    // logic for fetching hasNull ....
    if hasNull && nonNullType != nil {
        msg := "If type is a list, it must contain one non-null type and 'null'."
        return nil, fmt.Errorf("invalid JSON schema: %s", msg)
    }
    // rest of the logic for handling list types
case string:
    convertedSchema["type"] = fmt.Sprintf("%s", typeList)
default:
    // Throw error
}
```

sukumargaonkar:
we found that gcp recently added `defs` and `ref` field to Schema object
https://github.com/googleapis/python-genai/pull/820/files#diff-af6bee559d65a77d542f74d9cdcc4a3fd7fa1b82e9474b5f270873f070d565c9R1327-R1338

if they work as advertised can skip de-referencing logic

internal/extproc/translator/gemini_helper.go
```diff
@@ -392,8 +393,13 @@ func openAIToolChoiceToGeminiToolConfig(toolChoice *openai.ChatCompletionToolCho
 	}
 }
 
+// it only works with gemini2.5 according to https://ai.google.dev/gemini-api/docs/structured-output#json-schema, separate it as a small function to make it easir to maintain
```
xiaolin593:
```suggestion
// it only works with gemini2.5 according to https://ai.google.dev/gemini-api/docs/structured-output#json-schema, separate it as a small function to make it easier to maintain
```

internal/extproc/translator/gemini_helper.go
```diff
@@ -392,8 +393,13 @@ func openAIToolChoiceToGeminiToolConfig(toolChoice *openai.ChatCompletionToolCho
...
+func responseJSONSchemaAvailable(requestModel internalapi.RequestModel) bool {
```
xiaolin593, line 397: name it `isGeminiModel`?

All the gemini model name convention seem to have `gemini-2.5-` as prefix. https://cloud.google.com/vertex-ai/generative-ai/docs/learn/model-versions.

Also, the check doesn't handle future Gemini version.

yuzisun: can we do the other around to check flash 2.0 instead ? this is the only model that does not support ref

hustxiayang: @yuzisun @xiaolin593 Hi, this is to follow the doc: "It only works with Gemini 2.5." https://ai.google.dev/gemini-api/docs/structured-output#json-schema. I am not against to make it something to exclude "gemini2.0" and "gemini-1.5" though. I expect this check is just temporary and would make it `true` when we deprecate `gemini-2.0`

internal/extproc/translator/jsonschema_helper.go
hustxiayang: @mathetake followed the suggestion.

internal/extproc/translator/jsonschema_helper.go
hustxiayang: I make it a variable. This might introduce some extra maintain overhead though

internal/extproc/translator/jsonschema_helper.go
hustxiayang: made it a variable

hustxiayang:
@mathetake Updated the description and state the `TODO` as just a note

hustxiayang:
> we found that gcp recently added `defs` and `ref` field to Schema object https://github.com/googleapis/python-genai/pull/820/files#diff-af6bee559d65a77d542f74d9cdcc4a3fd7fa1b82e9474b5f270873f070d565c9R1327-R1338
> 
> if they work as advertised can skip de-referencing logic

Verified that the lasted version of genai is v1.32.0, but it does not include `ref` and `def` yet 
https://github.com/googleapis/go-genai/blob/v1.32.0/types.go#L1214. Thus, we can add this as a future work.

internal/extproc/translator/jsonschema_helper.go
hustxiayang: if `ok` it would continue and if `not ok` we would add the key into the map. I think this covers both cases and we do not need to throw as we prevented recursive path reference?

internal/extproc/translator/jsonschema_helper.go
```diff
@@ -0,0 +1,434 @@
...
+}
+
+// jsonSchemaDereferenceHelper recursively dereferences JSON schema references.
+// Note: `processedRefs` is a pointer to a set (map[string]struct{})
+// to ensure state is shared across recursive calls.
+func jsonSchemaDereferenceHelper(
+	obj any,
+	fullSchema map[string]any,
+	skipKeys []string,
+	processedRefs *map[string]struct{},
```
mathetake: i don't believe a pointer to a map is necessary

hustxiayang: hi, sorry, what do you mean? Do you have suggestions?

mathetake: map is always passed by a reference in Go, so drop `*` from this line

> You don't need to use a pointer with a map.
> Map types are reference types, like pointers or slices[1]
> If you needed to change the Session you could use a pointer:
>
> [1] https://go.dev/blog/maps

internal/extproc/translator/jsonschema_helper.go
mathetake, line 1: are the functions here really worth the complexity? How much gains we get from this manual deep copy vs json unmarshal and marshal (or i think protomarshal or something like that could be used)? I would rather avoid the premature optimization if this is not that serious

mathetake: oh ok majority is not about the deep copy hah

mathetake:
you can browser the coverage overview by clicking here, so I would request for the comprehensive coverage on the helper especially... this would be extremely hard to debug when something goes wrong without proper tests
image

internal/extproc/translator/jsonschema_helper.go
hustxiayang: followed the suggestion

internal/extproc/translator/jsonschema_helper.go
hustxiayang: followed the suggestion

yuzisun:
> you can browser the coverage overview by clicking here, so I would request for the comprehensive coverage on the helper especially... this would be extremely hard to debug when something goes wrong without proper tests image

@mathetake Improved the overall code readability and increased the coverage, this code is in its own isolated files and Gemini flash 2.0 is going to be EOL Feb, 2026, by then we can remove this code but currently it is still required running on production.

https://cloud.google.com/vertex-ai/generative-ai/docs/learn/model-versions

hustxiayang:
@codecov recheck

internal/extproc/translator/jsonschema_helper.go
```diff
@@ -0,0 +1,623 @@
...
+
+// Constants for safety limits
+const (
+	maxRecursionDepth = 100
+)
+
+// Errors for safety violations
+var (
+	ErrMaxRecursionDepthExceeded = fmt.Errorf("maximum recursion depth exceeded")
+	ErrInvalidSchema             = fmt.Errorf("invalid JSON schema")
```
mathetake: unexport

internal/extproc/translator/jsonschema_helper.go
```diff
@@ -0,0 +1,623 @@
...
+			if _, allowed := allowedSchemaFieldsSet[key]; allowed {
+				convertedSchema[key] = value
+			}
+		}
+	}
+	return convertedSchema, nil
+}
+
+// processTypeField handles the "type" field conversion with improved type safety
+func processTypeField(value any) (map[string]any, error) {
```
mathetake:
```suggestion
func processJSONSchemaTypeField(value any) (map[string]any, error) {
```

internal/extproc/translator/jsonschema_helper.go
```diff
@@ -0,0 +1,623 @@
...
+
+	default:
+		return nil, fmt.Errorf("%w: 'type' must be a list or string, got %T", ErrInvalidSchema, value)
+	}
+
+	return result, nil
+}
+
+// processAllOfField handles the "allOf" field conversion
+func processAllOfField(value any, allowedSchemaFieldsSet map[string]struct{}) (map[string]any, error) {
```
mathetake: this needs same jsonschema portion in the name too

internal/extproc/translator/jsonschema_helper.go
```diff
@@ -0,0 +1,623 @@
...
+	subSchema, ok := allOfList[0].(map[string]any)
+	if !ok {
+		return nil, fmt.Errorf("%w: item in 'allOf' must be an object, got %T", ErrInvalidSchema, allOfList[0])
+	}
+
+	return jsonSchemaToGapic(subSchema, allowedSchemaFieldsSet)
+}
+
+// processAnyOfField handles the "anyOf" field conversion
+func processAnyOfField(value any, allowedSchemaFieldsSet map[string]struct{}) (map[string]any, error) {
```
mathetake: this too

internal/extproc/translator/jsonschema_helper.go
mathetake: and jsonschema portion in names

internal/extproc/translator/jsonschema_helper.go
```diff
@@ -0,0 +1,623 @@
...
+	"encoding/json"
+	"fmt"
+	"strings"
+
+	"google.golang.org/genai"
+)
+
+// Constants for safety limits
+const (
+	maxRecursionDepth = 100
```
mathetake:
```suggestion
const (
	jsonSchemaMaxRecursionDepth = 100
```

internal/extproc/translator/jsonschema_helper_test.go
```diff
@@ -0,0 +1,979 @@
...
+						},
+					},
+				},
+				Type:     "object",
+				Required: []string{"item"},
+			},
+		},
+
+		{
+			name: "anyof null for ResponseSchema",
```
hustxiayang, line 893: anyOf

internal/extproc/translator/gemini_helper_test.go
```diff
@@ -1489,3 +1503,340 @@ func TestExtractTextFromGeminiParts(t *testing.T) {
...
+										}
+									}
+								},
+								"required": ["user", "items"]
+							}`),
+						},
+					},
+				},
+			},
+			requestModel: "gemini-pro", // older model
```
hustxiayang: probably better to use a real name gemini-2.0-flash

internal/extproc/translator/gemini_helper_test.go
```diff
@@ -1489,3 +1503,340 @@ func TestExtractTextFromGeminiParts(t *testing.T) {
...
+								"type": "object",
+								"properties": {
+									"invalid_ref": {"$ref": "#/nonexistent/path"}
+								}
+							}`),
+						},
+					},
+				},
+			},
+			requestModel:   "gemini-pro",
```
hustxiayang: probably better to use a real name gemini-2.0-flash

internal/extproc/translator/gemini_helper_test.go
```diff
@@ -835,10 +843,12 @@ func TestOpenAIReqToGeminiGenerationConfig(t *testing.T) {
 				ResponseJsonSchema: map[string]any{"type": "string"},
 			},
 			expectedResponseMode: responseModeJSON,
+			requestModel:         "gemini-2.5-flash",
 		},
 		{
 			name: "json schema (invalid string)",
 			input: &openai.ChatCompletionRequest{
+				Model: "gemini-2.5",
```
hustxiayang, line 851: not necessary and conflict with requestModel?

internal/extproc/translator/gemini_helper_test.go
```diff
@@ -817,10 +823,12 @@ func TestOpenAIReqToGeminiGenerationConfig(t *testing.T) {
 				ResponseJsonSchema: map[string]any{"type": "string"},
 			},
 			expectedResponseMode: responseModeJSON,
+			requestModel:         "gemini-2.5-flash",
 		},
 		{
 			name: "json schema (string)",
 			input: &openai.ChatCompletionRequest{
+				Model: "gemini-2.5",
```
hustxiayang, line 831: not necessary and conflict with requestModel?

internal/extproc/translator/gemini_helper_test.go
```diff
@@ -1489,3 +1503,340 @@ func TestExtractTextFromGeminiParts(t *testing.T) {
...
+									"name": {"type": "string"},
+									"age": {"type": "number"}
+								},
+								"required": ["name", "age"]
+							}`),
+						},
+					},
+				},
+			},
+			requestModel: "gemini-pro", // older model that doesn't support native JSON schema
```
hustxiayang: probably better to use a real name gemini-2.0-flash

internal/extproc/translator/gemini_helper_test.go
```diff
@@ -1489,3 +1503,340 @@ func TestExtractTextFromGeminiParts(t *testing.T) {
...
+											{"type": "number"}
+										]
+									}
+								}
+							}`),
+						},
+					},
+				},
+			},
+			requestModel: "gemini-pro",
```
hustxiayang: probably better to use a real name gemini-2.0-flash

internal/extproc/translator/gemini_helper_test.go
```diff
@@ -1489,3 +1503,340 @@ func TestExtractTextFromGeminiParts(t *testing.T) {
...
+											{"type": "null"}
+										]
+									}
+								}
+							}`),
+						},
+					},
+				},
+			},
+			requestModel: "gemini-pro",
```
hustxiayang: probably better to use a real name gemini-2.0-flash

internal/extproc/translator/gemini_helper_test.go
```diff
@@ -1489,3 +1503,340 @@ func TestExtractTextFromGeminiParts(t *testing.T) {
...
+											"age": {"type": "number"}
+										}
+									}
+								}
+							}`),
+						},
+					},
+				},
+			},
+			requestModel: "gemini-pro",
```
hustxiayang: probably better to use a real name gemini-2.0-flash

internal/extproc/translator/gemini_helper_test.go
yuzisun: requestModel can be the actual provider model name which is configured via `AIGatewayRoute` while model field on original request can be the model alias.

PR #1402, opened 2025-10-21
fix: remove wrapping double-quotes from GCP guided_regex

sukumargaonkar:
**Description**
This PR removes the wrapping double-quotes returned by GCP when using guided_regex

GCP doesn't natively support REGEX response modes, so we instead express them as json schema.
This causes the response to be wrapped in double-quotes.
E.g. `"positive"` (the double-quotes at the start and end are unwanted)


internal/extproc/translator/gemini_helper.go
```diff
@@ -410,39 +422,47 @@ func openAIReqToGeminiGenerationConfig(openAIReq *openai.ChatCompletionRequest)
...
 			}
 
+			responseMode = ResponseModeJSON
+
 			gc.ResponseMIMEType = mimeTypeApplicationJSON
 			gc.ResponseJsonSchema = schemaMap
 		}
 	}
 
 	if openAIReq.GuidedChoice != nil {
```
aabchoo, line 446: response mode may be overridden -- add check to confirm only one of GuidedRegex, Guided Choice,  Responsefromat etc is set

internal/extproc/translator/gemini_helper.go
```diff
@@ -545,10 +565,19 @@ func geminiFinishReasonToOpenAI(reason genai.FinishReason) openai.ChatCompletion
...
+func extractTextFromGeminiParts(parts []*genai.Part, responseMode ResponseMode) string {
 	var text string
 	for _, part := range parts {
 		if part != nil && part.Text != "" {
+			if responseMode == ResponseModeRegex {
+				// GCP doesn't natively support REGEX response modes, so we instead express them as json schema.
+				// This causes the response to be wrapped in double-quotes.
+				// E.g. `"positive"` (the double-quotes at the start and end are unwanted)
+				// Here we remove the wrapping double-quotes.
+				if len(part.Text) > 2 && part.Text[0] == '"' && part.Text[len(part.Text)-1] == '"' {
```
tiswanso: I'm curious whether there's a reason to not use strings.Trim() here: https://pkg.go.dev/strings#Trim

sukumargaonkar: nice catch,
mainly being defensive for scenarios where first and last words are quoted but not connected
`"ERROR" Unable to connect to database "DatabaseModule"`
trim would remove more than intended
added a test case for this
thoughts?

internal/extproc/translator/gemini_helper.go
sukumargaonkar: good point,
Addressed in https://github.com/envoyproxy/ai-gateway/pull/1402/commits/9bc5b8eee3a2cb994fb1253fb30c3564e26e1ca0

sukumargaonkar:
The ubuntu based CI failed but macOS based passed

error in ubuntu based CI
https://github.com/envoyproxy/ai-gateway/actions/runs/18718793788/job/53391903589?pr=1402
```
--- FAIL: TestRun_enabled (0.00s)
    pprof_test.go:33: 
        	Error Trace:	/home/runner/work/ai-gateway/ai-gateway/internal/pprof/pprof_test.go:33
        	Error:      	Received unexpected error:
        	            	Get "http://localhost:6060/debug/pprof/cmdline": dial tcp [::1]:6060: connect: connection refused
        	Test:       	TestRun_enabled
```

seems unrelated to this PR. maybe @mathetake knows more

mathetake:
yeah sorry you can ignore the flake

internal/extproc/translator/gemini_helper.go
tiswanso: I see... makes sense.  TrimPrefix(), TrimSuffix() can be used to trim exact strings so it'd work to limit to the single char.  https://pkg.go.dev/strings#TrimPrefix

internal/extproc/translator/gemini_helper.go
```diff
@@ -29,6 +29,17 @@ const (
...
 
+// ResponseMode represents the type of response mode for Gemini requests
+type ResponseMode string
+
+const (
+	ResponseModeNone  ResponseMode = "NONE"
+	ResponseModeText  ResponseMode = "TEXT"
+	ResponseModeJSON  ResponseMode = "JSON"
+	ResponseModeEnum  ResponseMode = "ENUM"
+	ResponseModeRegex ResponseMode = "REGEX"
```
mathetake: unexport

mathetake: also prefix gemini*

internal/extproc/translator/gemini_helper.go
```diff
@@ -29,6 +29,17 @@ const (
 	httpHeaderKeyContentLength     = "Content-Length"
 )
 
+// ResponseMode represents the type of response mode for Gemini requests
+type ResponseMode string
```
mathetake:
```suggestion
type geminiResponseMode string
```

internal/extproc/translator/gemini_helper.go
mathetake: +1 to strings.Trim*

internal/extproc/translator/gemini_helper.go
sukumargaonkar: Addressed in https://github.com/envoyproxy/ai-gateway/pull/1402/commits/f871ff7d7c0a42b8e4129ad461aef5726d45ce28

internal/extproc/translator/gemini_helper.go
sukumargaonkar: addressed in https://github.com/envoyproxy/ai-gateway/pull/1402/commits/f871ff7d7c0a42b8e4129ad461aef5726d45ce28

internal/extproc/translator/gemini_helper.go
mathetake: @sukumargaonkar i think this is the only pending unresolved comment

sukumargaonkar: Addressed in https://github.com/envoyproxy/ai-gateway/pull/1402/commits/c73f6aa01cacc0405808ca927a8dc3c8736c824d

mathetake: i don't think this if conditions is necessary as strings.Trim* is noop if not applicable

> TrimPrefix returns s without the provided leading prefix string. If s doesn't start with prefix, s is returned unchanged.

internal/extproc/translator/gemini_helper.go
```diff
@@ -407,46 +419,66 @@ func openAIReqToGeminiGenerationConfig(openAIReq *openai.ChatCompletionRequest)
...
 		}
+		responseMode = responseModeJSON
+
 		gc.ResponseMIMEType = mimeTypeApplicationJSON
 		gc.ResponseJsonSchema = openAIReq.GuidedJSON
 	}
 
+	// ResponseFormat and guidedJSON/guidedChoice/guidedRegex are mutually exclusive.
+	// Verify only one is specified.
+	if formatSpecifiedCount > 1 {
```
aabchoo, line 478: What happens when formatSpecifiedCount == 0

sukumargaonkar: nothing really
this is only to validate more than one format specifiers are not specified

internal/extproc/translator/gemini_helper.go
sukumargaonkar: addressed in https://github.com/envoyproxy/ai-gateway/pull/1402/commits/f2d8620e8e7f5244446e5bd8563bf5fd7a6dd6f4

yuzisun (approved):

Issue #1398, opened 2025-10-21
MCP: SSE parser doesn't handle CRLF line endings properly


Eric84626:
Problem The SSE parser in sse.go currently only handles Unix-style line endings (\n) but doesn't properly support Windows-style line endings (\r\n). This can cause parsing issues when receiving SSE streams from servers that use CRLF line endings.
Current Behavior The SSE parser uses hardcoded separators that only work with \n:
When parsing SSE data with \r\n line endings, the parser may fail to correctly identify field boundaries and event separators. And we can see below error from log.
`MCP message is not a response: <nil>`

Expected Behavior The sse parser should handle both Unix (\n) and Windows (\r\n) line endings transparently.

mathetake:
> Lines must be separated by either a U+000D CARRIAGE RETURN U+000A LINE FEED (CRLF) character pair, a single U+000A LINE FEED (LF) character, or a single U+000D CARRIAGE RETURN (CR) character.


oh good catch!

PR #1425, opened 2025-10-23
mcp: gracefully handle CRLF and CR separators in SSE events

nacx:
**Description**

Gracefully handle CRLF and CR separators in SSE events.

Fixes https://github.com/envoyproxy/ai-gateway/issues/1398

@Eric84626 could you give this branch a try?

### DIFF
--- a/internal/mcpproxy/sse.go
+++ b/internal/mcpproxy/sse.go
@@ -6,8 +6,8 @@
 package mcpproxy
 
 import (
-	"bufio"
 	"bytes"
+	"errors"
 	"fmt"
 	"io"
 	"net/http"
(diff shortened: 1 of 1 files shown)


internal/mcpproxy/sse.go
```diff
@@ -72,31 +81,90 @@ func (p *sseEventParser) next() (*sseEvent, error) {
...
 	}
 	for _, msg := range e.messages {
 		_, _ = w.Write(sseDataPrefix)
 		data, _ := jsonrpc.EncodeMessage(msg)
 		_, _ = w.Write(data)
-		_, _ = w.Write(sseFieldSeparator)
+		_, _ = w.Write(e.separator.field())
 	}
-	_, _ = w.Write(sseEventSeparator)
+	_, _ = w.Write(e.separator.event())
```
mathetake: i don't think we need to preserve separators in the events we write since regardless of the host env the clients running, the clients are supposed to be able to handle any separator, so I would remove this repetive call to an interface method and instead simply always use '\n'.

internal/mcpproxy/sse.go
```diff
@@ -72,31 +81,90 @@ func (p *sseEventParser) next() (*sseEvent, error) {
...
 	}
-	p.buf = p.buf[idx+2:]
+	p.buf = p.buf[idx+sepLen:]
 	return ret, nil
 }
 
+// findSeparator finds the index of the next event separator in the buffer.
+// Lines must be separated by either a U+000D CARRIAGE RETURN U+000A LINE FEED (CRLF) character pair,
+// a single U+000A LINE FEED (LF) character, or a single U+000D CARRIAGE RETURN (CR) character.
+func (p *sseEventParser) findSeparator() int {
```
mathetake: this doesn't work as the spec says any separator can be used in a data stream and no guarantee in the consistency in its use. sseSeparatorLF and sseSeparatorCRLF can be mixed in a single stream for example.

nacx: Fixed

internal/mcpproxy/sse.go
nacx: Done

Eric84626:
> **Description**
> 
> Gracefully handle CRLF and CR separators in SSE events.
> 
> **Related Issues/PRs (if applicable)**
> 
> Fixes #1398
> 
> **Special notes for reviewers (if applicable)**
> 
> @Eric84626 could you give this branch a try?

sure, I will try in my local, after that, i will give you the feedback, thanks

Eric84626:
> > **Description**
> > Gracefully handle CRLF and CR separators in SSE events.
> > **Related Issues/PRs (if applicable)**
> > Fixes #1398
> > **Special notes for reviewers (if applicable)**
> > @Eric84626 could you give this branch a try?
> 
> sure, I will try in my local, after that, i will give you the feedback, thanks

hello @nacx, I have tested your code changes in my local, it works as expected. thanks

nacx:
Oh, I read your comment after merge. I'll do in a quick PR

Issue #1186, opened 2025-09-12
Restructure Helm Charts – CRD Installation Refactor


saileshd1402:
## *Description*

We currently face conflicts when installing ai-gateway-crd-helm and ai-gateway-helm as two subcharts under a parent Helm chart. This proposal suggests refactoring CRD handling to avoid duplicate ownership issues when used as subcharts and follow the Helm standards.

### Setup / Problem
- We have two Helm subcharts under a parent chart: **ai-gateway-crd-helm** and **ai-gateway-helm**. Both include the same CRDs, causing ownership / conflict issues when installing the parent chart.  
- While installing **ai-gateway-helm** through Helm CLI, we can use `--skip-crds` to avoid CRD installation, but **ai-gateway-helm** doesn’t have a corresponding setting in `values.yaml` to skip its CRDs when added as a subchart.  
- The current workaround is to install the parent chart with the `--skip-crds` flag, so only **ai-gateway-crd-helm** installs the CRDs and **ai-gateway-helm**’s CRDs (in its `crds/` directory) are skipped. But this feels unintuitive and fragile.

---

### 1. Remove CRDs from `ai-gateway-helm` Chart

- Remove all CRDs from **ai‑gateway‑helm**, so that **ai‑gateway‑crd‑helm** is the single chart responsible for defining and managing CRDs.  
- The CRD chart owns the lifecycle of CRDs — installation, upgrades (as appropriate), etc., while **ai‑gateway‑helm** handles the rest of the resources.  
- Having a single owner of CRDs avoids the conflict/ownership issues when installing multiple charts under a parent.  
- [kserve](https://github.com/kserve/kserve) follows this approach where they have completely separate charts for CRDs ([kserve-crd](https://github.com/kserve/kserve/tree/master/charts/kserve-crd)) and resources ([kserve-resources](https://github.com/kserve/kserve/tree/master/charts/kserve-resources))

### 2. Toggle + Move CRDs into `templates/` in `ai-gateway-helm` Chart

- Move CRDs from **ai‑gateway‑helm**’s `crds/` directory into `templates/`.  
- Add a Boolean toggle in **ai‑gateway‑helm**’s `values.yaml` (e.g. `crds.enabled`) that controls whether the CRD templates are rendered and installed. 
[...]

PR #1431, opened 2025-10-24
refactor: remove /crds folder from ai-gateway-helm chart

saileshd1402:
**Description**

As discussed in this slack thread: https://envoyproxy.slack.com/archives/C07Q4N24VAA/p1761235915409479, we want to separate the crd chart and the resources chart. (option 1 of the proposed approaches in https://github.com/envoyproxy/ai-gateway/issues/1186)

https://github.com/envoyproxy/ai-gateway/issues/1186

Added an extra note on upgrade when done from a previous install of ai-gateway-helm helm chart alone. it uses the `--take-ownership` flag while upgrading: https://github.com/helm/helm/pull/13439, https://helm.sh/docs/helm/helm_upgrade/#options

Open Question:
- Should we rename `ai-gateway-helm` to `ai-gateway-resources-helm`? It would be more consistent but upgrade is slightly trickier

cc: @mathetake @johnugeorge @nagar-ajay


nacx (approved):
Thanks! LGTM!

Issue #957, opened 2025-07-28
add extension of embedding chat request and rerank api


hustxiayang:
*Description*:
In current implementation, embedding api is strictly following OpenAI's definition: https://platform.openai.com/docs/api-reference/embeddings/create. I wanted to extend the functionality with `EmbeddingChatRequest` like vllm" https://github.com/vllm-project/vllm/blob/34a20c49b3f81f64133428b3a0d62309db1256f9/vllm/entrypoints/openai/protocol.py#L1341C6-L1341C75. It's not only for feature support, but also affecting the accuracy of new embedding models like qwen3 embedding.

Also want to add the re-rank api, which is used for a typical RAG application.

Since there are no such definitions in OpenAI, I suggested to add these definitions in internal/apischema/embeddingchat and internal/apischema/rerank.

mathetake:
@hustxiayang from next time, let's not conflate two completely separate things in one issue. Rerank API is already added so I will go ahead and close this one. if it's necessary, feel free to open a new issue focusing on "extensions of embeddings"

hustxiayang:
@mathetake Thanks a lot! I would open a pr to extend embedding support. Btw, the purpose of  including these two was actually to ask about/discuss  "breaking strict openai api schema", because at that time it was not using unified api, so felt very nervous to break the "strict openai schema".

PR #1433, opened 2025-10-24
feat: add passthrough cohere v2 rerank endpoint

AyushSawant18588:
**Description**
This PR adds complete support for Cohere's rerank endpoint (/v2/rerank) across the Envoy AI Gateway. It introduces apischema, processor, translator, metrics instrumentation and updates Docmentation to include rerank endpoint.

https://github.com/envoyproxy/ai-gateway/issues/957

- Cannot add support for rerank in Envoy AI Gateway CLI quickstart using ollama cause there is no cohere v2 rerank endpoint support in Ollama.
- Have not added the following things in this PR cause wanted to confirm the base implementation, will take the following in a separate PR:
   - tracing for rerank
   - integration tests using vcr and cassettes

### DIFF
--- a/internal/extproc/translator/cohere_rerank_v2.go
+++ b/internal/extproc/translator/cohere_rerank_v2.go
@@ -30,14 +30,18 @@ func NewRerankCohereToCohereTranslator(apiVersion string, modelNameOverride inte
 // https://docs.cohere.com/reference/rerank
 type cohereToCohereTranslatorV2Rerank struct {
 	modelNameOverride internalapi.ModelNameOverride
+	// requestModel stores the effective model for this request (override or provided)
+	requestModel internalapi.RequestModel
 	// The path of the rerank endpoint to be used for the request. It is prefixed with the API path prefix.
 	path string
 }
 
 // RequestBody implements [CohereRerankTranslator.RequestBody].
-func (t *cohereToCohereTranslatorV2Rerank) RequestBody(original []byte, _ *cohereschema.RerankV2Request, onRetry bool) (
+func (t *cohereToCohereTranslatorV2Rerank) RequestBody(original []byte, req *cohereschema.RerankV2Request, onRetry bool) (
 	headerMutation *extprocv3.HeaderMutation, bodyMutation *extprocv3.BodyMutation, err error,
 ) {
+	// Store the request model to use as fallback for response model
+	t.requestModel = req.Model
 	var newBody []byte
 	if t.modelNameOverride != "" {
 		// Override the model if configured.
@@ -45,6 +49,8 @@ func (t *cohereToCohereTranslatorV2Rerank) RequestBody(original []byte, _ *coher
 		if err != nil {
 			return nil, nil, fmt.Errorf("failed to set model name: %w", err)
 		}
+		// Make everything coherent.
+		t.requestModel = t.modelNameOverride
 	}
 
 	// Always set the path header to the rerank endpoint so that the request is routed correctly.
@@ -80,7 +86,7 @@ func (t *cohereToCohereTranslatorV2Rerank) ResponseBody(_ map[string]string, bod
 ) {
 	var resp cohereschema.RerankV2Response
 	if err := json.NewDecoder(body).Decode(&resp); err != nil {
-		return nil, nil, tokenUsage, "", fmt.Errorf("failed to unmarshal body: %w", err)
+		return nil, nil, tokenUsage, t.requestModel, fmt.Errorf("failed to unmarshal body: %w", err)
 	}
 
 	// Token accounting: rerank only has input tokens; output tokens do not apply.
@@ -96,8 +102,8 @@ func (t *cohereToCohereTranslatorV2Rerank) ResponseBody(_ map[string]string, bod
 		}
 	}
 
-	// Cohere rerank responses do not echo model; report empty responseModel.
-	responseModel = ""
+	// Cohere rerank responses do not echo model; report the effective request model if known.
+	responseModel = t.requestModel
 	return
 }
 
--- a/site/docs/capabilities/llm-integrations/supported-endpoints.md
+++ b/site/docs/capabilities/llm-integrations/supported-endpoints.md
@@ -209,6 +209,7 @@ curl -H "Content-Type: application/json" \
 **Supported Providers:**
 
 - Cohere
+- Any Cohere-compatible provider that supports rerank, including vLLM.
 
 **Example:**
 
--- a/examples/basic/README.md
+++ b/examples/basic/README.md
@@ -11,6 +11,7 @@ traffic for various AI providers.
 - `azure_openai.yaml` - Azure OpenAI integration
 - `gcp_vertex.yaml` - GCP Vertex AI integration
 - `tars.yaml` - TARS integration
+- `cohere.yaml` - Cohere integration
 
 For AWS Bedrock, we recommend using either `aws-pod-identity.yaml` (EKS 1.24+) or
 `aws-irsa.yaml` (all EKS versions) for production deployments instead of static credentials. [Docs](https://docs.aws.amazon.com/eks/latest/best-practices/identity-and-access-management.html#_identities_and_credentials_for_eks_pods)
--- a/internal/extproc/rerank_processor.go
+++ b/internal/extproc/rerank_processor.go
@@ -317,9 +317,8 @@ func (r *rerankProcessorUpstreamFilter) ProcessResponseBody(ctx context.Context,
 	r.costs.OutputTokens += tokenUsage.OutputTokens
 	r.costs.TotalTokens += tokenUsage.TotalTokens
 
-	// Set the response model for metrics from the model in response body or model name from request headers.
-	resModel := cmp.Or(responseModel, r.requestHeaders[internalapi.ModelNameHeaderKeyDefault])
-	r.metrics.SetResponseModel(resModel)
+	// Set the response model for metrics
+	r.metrics.SetResponseModel(responseModel)
 
 	// Update metrics with token usage (rerank records only input tokens in metrics package).
 	r.metrics.RecordTokenUsage(ctx, tokenUsage.InputTokens, r.requestHeaders)
--- a/internal/extproc/translator/cohere_rerank_v2_test.go
+++ b/internal/extproc/translator/cohere_rerank_v2_test.go
@@ -160,6 +160,7 @@ func TestCohereToCohereTranslatorV2Rerank_ResponseBody(t *testing.T) {
 	} {
 		t.Run(tc.name, func(t *testing.T) {
 			translator := NewRerankCohereToCohereTranslator("v2", "")
+			translator.(*cohereToCohereTranslatorV2Rerank).requestModel = "rerank-english-v3"
 			headerMutation, bodyMutation, tokenUsage, responseModel, err := translator.ResponseBody(
 				map[string]string{contentTypeHeaderName: jsonContentType},
 				strings.NewReader(tc.responseBody),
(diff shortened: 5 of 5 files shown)


internal/apischema/cohere/rerank_v2.go
```diff
@@ -0,0 +1,98 @@
...
+	TopN *int `json:"top_n,omitempty"`
+	// Optional: truncate long documents to this many tokens. Default: 4096.
+	MaxTokensPerDoc *int `json:"max_tokens_per_doc,omitempty"`
+}
+
+// RerankV2Response represents the response from Cohere Rerank API v2.
+// Docs: https://docs.cohere.com/reference/rerank
+type RerankV2Response struct {
+	// Ordered list of ranked documents with scores.
+	Results []RerankV2Result `json:"results"`
```
gavrissh: Beneficial to use cohere pkg for these?
https://github.com/cohere-ai/cohere-go/blob/main/v_2.go#L5717

cmd/extproc/mainlib/main.go
```diff
@@ -250,6 +251,7 @@ func Main(ctx context.Context, args []string, stderr io.Writer) (err error) {
 	server.Register(path.Join(flags.rootPrefix, "/v1/completions"), extproc.CompletionsProcessorFactory(completionMetrics))
 	server.Register(path.Join(flags.rootPrefix, "/v1/embeddings"), extproc.EmbeddingsProcessorFactory(embeddingsMetrics))
 	server.Register(path.Join(flags.rootPrefix, "/v1/images/generations"), extproc.ImageGenerationProcessorFactory(imageGenerationMetrics))
+	server.Register(path.Join(flags.rootPrefix, "/v2/rerank"), extproc.RerankProcessorFactory(rerankMetrics))
```
mathetake: Non OpenAI endpoints should be prefixed (though this might change but for now let's be consistent on this)

```suggestion
	server.Register(path.Join(flags.rootPrefix, "/cohere/v2/rerank"), extproc.RerankProcessorFactory(rerankMetrics))
```

AyushSawant18588: Since OpenAI does not have a rerank endpoint and cohere is a widely used rerank endpoint and a standard followed by other providers, should we consider adopting Cohere's API as the ingress interface for Rerank like OpenAI is for other endpoints?

mathetake: yes, that is a good point and it also applies to anthropic's /messages too. We are discussing it offline right now with other maintainers

AyushSawant18588: yes, so should I add /cohere prefix in this PR?

mathetake: https://github.com/envoyproxy/ai-gateway/issues/1457

internal/extproc/translator/cohere_rerank_v2.go
```diff
@@ -0,0 +1,127 @@
...
+package translator
+
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+	"path"
+	"strconv"
+
+	cohereapi "github.com/cohere-ai/cohere-go/v2"
```
nacx: There was a conscious decision not to rely on libraries such as the OpenAI library, etc.
Given that for rerank, we're using a really small data model (request and response are quite simple objects), can we just define them, like with the other providers, and avoid bringing a new dependency?

mathetake: +1

AyushSawant18588: I was not aware of that, earlier I had followed that approach only of defining the request, response and error objects but as can be seen in this commit: https://github.com/envoyproxy/ai-gateway/blob/ef01b31c3b8f5585b5aa39516ca0dd731e40cf1d/internal/apischema/cohere/rerank_v2.go, the structs needed are a bit complex. 
Cause of this thought we can use the cohere sdk directly to simplify maintenance and ensure alignment with the official API and also for anthropic I saw sdk is used https://github.com/envoyproxy/ai-gateway/blob/f547bad61e1711fe413c6bdb79d2199674ba0025/internal/extproc/translator/anthropic_anthropic.go#L16, hence I updated to use Cohere sdk. This way, we won’t have to manually maintain or update the structs as the API evolves. Let me know your thoughts on this approach, happy to revert if we’d prefer to stick with explicit struct definitions

mathetake: sorry we won't repeat the same discussion over and over again but you can follow https://github.com/envoyproxy/ai-gateway/pull/1147 here for example. Anthropic SDK use will be removed soon, which i am working on. so could you go ahead and avoid using corehere SDK (ok to use it only in tests)

AyushSawant18588: Okay thanks I will go through this and revert to using explicit struct definitions

mathetake: yep, thank you. key points are: 
* Cross provider compatibility we want is not a provider's concern. E.g. They do not care about edge cases we have like https://github.com/openai/openai-go/issues/484
* Maintaining a struct for a few endpoint is not that hard
* Usually AI providers auto generate SDKs from an ambiguous OpenAPI spec (at least openai/anthropic), which has weird performance cost as well as making it hard for us to add vendor specific fields

nacx: @mathetake can you add the points in the referenced PR and these ones to a document in the docs directory so we have this decision recorded in a proper file we can better reference in the future?

AyushSawant18588:
> Thanks! Just some minor comments.
> 
> Can you add an e2e test for this API, guarded by the presence of the credentials, so we have a way of validating it live?

Added the e2e test for cohere rerank API. Will have to add Cohere API key in GitHub Secrets for the test to run in CI otherwise it will be skipped like for other providers. Also I noticed e2e tests for OpenAI, AWS and Anthropic are being skipped in CI cause their credentials are not in GitHub Secrets.

cmd/extproc/mainlib/main.go
AyushSawant18588: Have added the cohere prefix

tests/e2e/basic_test.go
```diff
@@ -63,6 +66,11 @@ func Test_Examples_Basic(t *testing.T) {
 		require.NoError(t, err)
 		require.NoError(t, e2elib.KubectlApplyManifestStdin(t.Context(), strings.ReplaceAll(string(anthropicManifest), "ANTHROPIC_API_KEY", cc.AnthropicAPIKey)))
 
+		// Apply Cohere resources if credentials are set
+		cohereManifest, err := os.ReadFile(manifestDir + "/cohere.yaml")
+		require.NoError(t, err)
+		require.NoError(t, e2elib.KubectlApplyManifestStdin(t.Context(), strings.ReplaceAll(string(cohereManifest), "COHERE_API_KEY", cc.CohereAPIKey)))
```
mathetake, line 72: just to confirm: you did test locally with the API key right?

internal/extproc/translator/cohere_rerank_v2.go
mathetake: https://github.com/envoyproxy/ai-gateway/blob/a23a002678b39f06df31583fac3d2f558cc88b88/docs/why-not-official-sdk.md added the doc in #1464

tests/e2e/basic_test.go
AyushSawant18588: yup tested with API key

examples/basic/cohere.yaml
```diff
@@ -0,0 +1,87 @@
+# Copyright Envoy AI Gateway Authors
```
mathetake, line 1: can you add a line to https://github.com/envoyproxy/ai-gateway/blob/main/examples/basic/README.md to mention this file?

internal/extproc/translator/cohere_rerank_v2.go
```diff
@@ -0,0 +1,129 @@
...
+			tokenUsage.TotalTokens = tokenUsage.InputTokens
+		}
+		if resp.Meta.Tokens.OutputTokens != nil {
+			tokenUsage.OutputTokens = uint32(*resp.Meta.Tokens.OutputTokens) //nolint:gosec
+			tokenUsage.TotalTokens += tokenUsage.OutputTokens
+		}
+	}
+
+	// Cohere rerank responses do not echo model; report empty responseModel.
+	responseModel = ""
```
mathetake: instead of return empty, you should return the request model (either original or overridden one) like other places.

internal/extproc/translator/translator.go
```diff
@@ -125,6 +126,49 @@ type OpenAIEmbeddingTranslator interface {
...
+		err error,
+	)
+
+	// ResponseHeaders translates the response headers.
+	// 	- `headers` is the response headers.
+	//	- This returns `headerMutation` that can be nil to indicate no mutation.
+	ResponseHeaders(headers map[string]string) (
+		headerMutation *extprocv3.HeaderMutation,
+		err error,
+	)
```
mathetake, line 151: i think we can remove this until we need

internal/metrics/rerank_metrics.go
```diff
@@ -0,0 +1,72 @@
...
+		return &rerank{
+			baseMetrics: b.newBaseMetrics(genaiOperationRerank),
+		}
+	}
+}
+
+// RecordTokenUsage implements [RerankMetrics.RecordTokenUsage].
+func (r *rerank) RecordTokenUsage(ctx context.Context, inputTokens uint32, requestHeaders map[string]string) {
+	// Some rerank responses omit token usage information; skip recording when zero.
+	if inputTokens == 0 {
```
mathetake, line 62: i mean maybe it's useful to know input token is zero indicated in metrics?

site/docs/capabilities/llm-integrations/supported-endpoints.md
```diff
@@ -191,6 +191,40 @@ curl -H "Content-Type: application/json" \
...
+**Features:**
+
+- ✅ Single-query document reranking
+- ✅ Model selection via request body or `x-ai-eg-model` header
+- ✅ Token usage tracking and cost calculation
+- ✅ Provider fallback and load balancing
+
+**Supported Providers:**
+
+- Cohere
```
mathetake, line 211: could you mention the Cohere compatible providers here like vllm like embeddings endpoint

internal/metrics/rerank_metrics.go
AyushSawant18588: I saw not all models of cohere return token usage in the response and there is no field in request to force token usage in response like openai. Given this even when input token is non zero it will incorrectly be recorded in metrics as zero. I did this to prevent it, so is this fine?

mathetake: Cool thank you for clarifying

internal/extproc/translator/translator.go
AyushSawant18588: Its there for all translators so added, should I keep for consistency or remove as its not needed?

mathetake: Yeah it's alright to keep it as is on second thought. I'm planning to do the huge refactoring anyways

examples/basic/cohere.yaml
AyushSawant18588: done

internal/extproc/translator/cohere_rerank_v2.go
AyushSawant18588: done

site/docs/capabilities/llm-integrations/supported-endpoints.md
AyushSawant18588: done

mathetake (approved):

Issue #1412, opened 2025-10-22
Data race seen in test-coverage


aabchoo:
_Description_:

Test-coverage failed because of the following:

https://github.com/envoyproxy/ai-gateway/actions/runs/18721861080/job/53396124126?pr=1411#step:5:120

```
WARNING: DATA RACE
Write at 0x00c0020055f0 by goroutine 247:
  runtime.mapaccess2()
      /Users/runner/hostedtoolcache/go/1.25.3/arm64/src/internal/runtime/maps/runtime_swiss.go:114 +0x2ec
  sigs.k8s.io/kubectl-validate/pkg/validator.(*Validator).infoForGVK()
      /Users/runner/go/pkg/mod/sigs.k8s.io/kubectl-validate@v0.0.5-0.20250915070809-d2f2d68fba09/pkg/validator/validator.go:307 +0xe28
  sigs.k8s.io/kubectl-validate/pkg/validator.(*Validator).Parse()
      /Users/runner/go/pkg/mod/sigs.k8s.io/kubectl-validate@v0.0.5-0.20250915070809-d2f2d68fba09/pkg/validator/validator.go:62 +0x318
  sigs.k8s.io/kubectl-validate/pkg/cmd.ValidateDocument()
      /Users/runner/go/pkg/mod/sigs.k8s.io/kubectl-validate@v0.0.5-0.20250915070809-d2f2d68fba09/pkg/cmd/validate.go:288 +0x48
  github.com/envoyproxy/gateway/internal/gatewayapi/resource.Validator.Validate()
      /Users/runner/go/pkg/mod/github.com/envoyproxy/gateway@v0.5.0-rc.1.0.20251020052851-16aa8100498d/internal/gatewayapi/resource/validator.go:41 +0x338
  github.com/envoyproxy/gateway/internal/gatewayapi/resource.loadKubernetesYAMLToResources.func1()
      /Users/runner/go/pkg/mod/github.com/envoyproxy/gateway@v0.5.0-rc.1.0.20251020052851-16aa8100498d/internal/gatewayapi/resource/load.go:83 +0x31c
  github.com/envoyproxy/gateway/internal/gatewayapi/resource.IterYAMLBytes()
      /Users/runner/go/pkg/mod/github.com/envoyproxy/gateway@v0.5.0-rc.1.0.20251020052851-16aa8100498d/internal/gatewayapi/resource/load.go:675 +0x244
[...]

PR #1445, opened 2025-10-27
deps: resolves EG false positive CVEs issue

mathetake:
**Description**

This pins the EG to the "correct" version that doesn't cause false positive CVEs.

Reference: https://github.com/envoyproxy/gateway/issues/7248
Closes: #1412


mathetake:
hah

mathetake:
oh ok i see what's wrong now

mathetake:
https://github.com/envoyproxy/gateway/pull/7351 this is the blocker

.trivyignore
mathetake, line 1: 🔥

PR #1449, opened 2025-10-28
mcp: allow configuring the header for mcp backend api auth

nacx:
**Description**

This allows configuring the header where api credentials will be injected when talking to the upstream MCP servers. Previously, we could only inject Bearer token credentials, but some MCP servers require them to be on a different header. With this change users can configure the header like in the following example:

```yaml
kind: MCPRoute
metadata:
  name: mcp-route
  namespace: default
spec:
  parentRefs:
    - name: aigw-run
      kind: Gateway
  path: "/mcp"
  backendRefs:
    - name: mcp-backend
      kind: Backend
      group: gateway.envoyproxy.io
      securityPolicy:
        apiKey:
          header: "X-Api-Key"
          secretRef:
            name: apikey-secret

```

N/A

N/A

### DIFF
--- a/api/v1alpha1/mcp_route.go
+++ b/api/v1alpha1/mcp_route.go
@@ -143,7 +143,7 @@ type MCPToolFilter struct {
 
 // MCPBackendSecurityPolicy defines the security policy for a sp
 type MCPBackendSecurityPolicy struct {
-	// APIKey is a mechanism to access a backend. The API key will be injected into the Authorization header.
+	// APIKey is a mechanism to access a backend. The API key will be injected into the request headers.
 	// +optional
 	APIKey *MCPBackendAPIKey `json:"apiKey,omitempty"`
 }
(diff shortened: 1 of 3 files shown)


api/v1alpha1/mcp_route.go
mathetake: so i think we should change this documentation too

nacx: I'll keep it generic as "will be injected in the request headers"

mathetake: sg

api/v1alpha1/mcp_route.go
```diff
@@ -161,6 +161,14 @@ type MCPBackendAPIKey struct {
 	//
 	// +optional
 	Inline *string `json:"inline,omitempty"`
+
+	// Header is the HTTP header to inject the API key into.
+	// If not specified, the default is "Authorization".
```
mathetake: let's document that when this is "authorization" the value is prefixed with "bearer"

PR #1183, opened 2025-09-12
feat: add Azure Managed Identity token provider support

MattMencel:
**Description**

This commit adds comprehensive support for Azure Managed Identity authentication for AI Gateway backends through the BackendSecurityPolicy.

The new AzureManagedIdentityTokenProvider enables seamless authentication with Azure OpenAI and other Azure AI services using both system-assigned and user-assigned managed identities. This eliminates the need to manage API keys or secrets for Azure-hosted workloads, improving security and operational simplicity.

Key features:
  - System-assigned and user-assigned managed identity support
  - OIDC token exchange for cross-cloud scenarios
  - Fallback to Kubernetes secret-based authentication
  - Comprehensive test coverage including integration tests
  - Example configurations demonstrating various authentication scenarios
  - Updated API documentation and CRD schemas with proper validation
  - Improved .gitignore to exclude Go test binaries

**Related Issues/PRs (if applicable)**

#1180 

**Special notes for reviewers (if applicable)**

This PR includes extensive test coverage with both unit tests and integration tests using envtest. The implementation follows the existing token provider pattern and integrates cleanly with the current BackendSecurityPolicy architecture.

PR #1451, opened 2025-10-28
feat: azure/aks workload identity support

ehrnst:
**Description**

The PR is a small re-work of #1183 which I could not get to work since it was using `azidentity.NewManagedIdentityCredential(managedIDOptions)` which would leverage the underlying AKS node's identity. I added  `azidentity.NewWorkloadIdentityCredential(workloadIDOptions)` that allows pods to run with a specific Azure identity, federated with the kubernetes cluster.

Related to PR #1183 (Azure Managed Identity support by @mattmancel)

- CRD Changes: The CRD file was manually updated to match API changes due to controller-gen access issues on Windows. you may want to regenerate the CRD to verify
- I tested this from all angles I could think of on a clean AKS cluster. Installing AI Gateway with the latest helm chart, patching the deployments with my local build. Then i tested annotating the helm install like this 👇  before making requests against the GW.

```helm upgrade aieg oci://registry-1.docker.io/envoyproxy/ai-gateway-helm \
  --version 0.0.0-2861c1b5b182b278841eec609d0f2b497b150ce9 \
  --set controller.serviceAccount.annotations."azure\.workload\.identity/client-id"="d245ed76-8227-" \
  --set controller.serviceAccount.annotations."azure\.workload\.identity/tenant-id"="e5806cb5-f8f4-" \
  --set controller.image.repository="ehrnst.azurecr.io/ai-gateway-controller" \
  --set controller.image.tag="azure-managed-identity-v3"
  ```
I dont know if we should change to specifically creating a service account, or if we should continue with the patch approach during install or after.

- I noticed that the controller will create a secret (witht the bearer token) which is used by the extproc container during calls to Azure OpenAI. I did not inspect how this rotation work, but i did remove my identity and delete the secret to verify authentication agains azure stopped working.


mathetake (changes requested):
Thanks for the work! However, I have a suggestion that would drastically change the current code.

Right now, the impl is using the SA mounted by Envoy AI Gateway controller, and it distributes/copies that identity into the secret that is eventually mounted by extproc. That I think is a bad practice from security perspective since you have no idea to tell which pod is actually accessing the service since this makes potentially tens of envoy poids will access Azure sevice with the same identity.

So, I would suggest to do the same thing as https://github.com/envoyproxy/ai-gateway/pull/1394 like by using Azure SDK to get the token dynamically inside the extproc using the SA mounted by the Envoy pod, and insert the token in the header. That would mean you have to write more code in https://github.com/envoyproxy/ai-gateway/tree/main/internal/extproc/backendauth than controller package right now

ehrnst:
> Thanks for the work! However, I have a suggestion that would drastically change the current code.
> 
> Right now, the impl is using the SA mounted by Envoy AI Gateway controller, and it distributes/copies that identity into the secret that is eventually mounted by extproc. That I think is a bad practice from security perspective since you have no idea to tell which pod is actually accessing the service since this makes potentially tens of envoy poids will access Azure sevice with the same identity.
> 
> So, I would suggest to do the same thing as #1394 like by using Azure SDK to get the token dynamically inside the extproc using the SA mounted by the Envoy pod, and insert the token in the header. That would mean you have to write more code in https://github.com/envoyproxy/ai-gateway/tree/main/internal/extproc/backendauth than controller package right now

I agree, I was just trying to get the origina #1183 to work. But I guess this will affect the current Azure authentication implementation as well?

I can see if i can make something work with the help of some collegues and various agents.

mathetake:
> I agree, I was just trying to get the origina https://github.com/envoyproxy/ai-gateway/pull/1183 to work. But I guess this will affect the current Azure authentication implementation as well?

no, we can make it work without affecting it.

ehrnst:
> > I agree, I was just trying to get the origina #1183 to work. But I guess this will affect the current Azure authentication implementation as well?
> 
> no, we can make it work without affecting it.

I see 0.4 is released now, congrats! i will try to make this PR ready for next release.

mathetake:
@ehrnst let me know if you intend to continue the work or not -- otherwise i would like to close this one for now to clarify which PRs are alive

ehrnst:
Hi. Sorry, I’ve been busy with work lately, like moving our Ai gateway to production .

We want this feature, so I will find some time to look at it again

Sent from Outlook for iOS<https://aka.ms/o0ukef>
________________________________

mathetake:
let me mark this as draft until you pass all tests and address my suggestion!

ehrnst:
Just jumping in here to say sorry for the delay. but I also have to admit that i'm no developer and need to rely on LLMs to help me out with the task, but i'm afraid that the result will break existing functionality, as i don't understand the entire code base. During my local development i see a lot of AI slop being added due to my little knowledge of the existing code. Which makes me uncomfortable.
If anyone is willing to help with it that would be great. If not i will find some time again to revisit this. Been way to busy with work lately.

PR #1452, opened 2025-10-28
test: make rate limit e2e less flaky

mathetake:
**Description**
Since #1410, the deployment created by EG for the rate limit server is in crashing loop until redis server is up and running in the rate limit e2e test setup, which caused a bit of flake in the test. This fixes it by restarting the deployment just in case.

This also removes the only remaining use of t.Context in test cleanup function in e2e tests.


mathetake:
4 times in a row without flake so far

mathetake:
cool 10 runs without flake, good to go

PR #1455, opened 2025-10-28
docs: updates RELEASES.md

mathetake:
**Description**

This updates RELEASES.md, notably making the following change:
* Clarify that we might release outside the normal release cycle depending on the community needs
* Clarify that we do not guarantee Go package level compatibility since that's not the intended use case for this project, except for api package. In fact, we have already broken internal/filterapi which used to be a public package.

### DIFF
--- a/cmd/aigw/run_test.go
+++ b/cmd/aigw/run_test.go
@@ -22,43 +22,22 @@ import (
 	"github.com/tetratelabs/func-e/api"
 	gwapiv1 "sigs.k8s.io/gateway-api/apis/v1"
 
-	"github.com/envoyproxy/ai-gateway/cmd/extproc/mainlib"
 	"github.com/envoyproxy/ai-gateway/internal/filterapi"
-	internaltesting "github.com/envoyproxy/ai-gateway/internal/testing"
 )
 
 // TestRun verifies that the main run function starts up correctly without making any actual requests.
 //
 // The real e2e tests are in tests/e2e-aigw.
 func TestRun(t *testing.T) {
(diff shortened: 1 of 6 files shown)


RELEASES.md
```diff
@@ -2,12 +2,15 @@
...
 ## Release Cycles
 
-Since Envoy AI Gateway depends on the Envoy Gateway and Envoy Proxy, we will follow the release cycle of the Envoy Gateway.
-
+Since Envoy AI Gateway depends on the Envoy Gateway and Envoy Proxy, we will normally follow the release cycle of the Envoy Gateway.
 In other words, we aim to cut the release of the Envoy AI Gateway a few days or a week after the new version of the Envoy Gateway is released. Therefore, the release cycle of the Envoy AI Gateway will be approximately every 2-3 months.
 
-We increment the major version number when we have a major architectural change or a major feature addition.
+However, we might release the Envoy AI Gateway out of the regular release cycle if the community requests it, depending on the significance of the change.
+Usually, the main branch of Envoy AI Gateway is tested against both the latest stable version of Envoy Gateway and its main branch, so we can release the Envoy AI Gateway anytime if needed.
```
nacx:
```suggestion
However, given the rapidly changing nature of the AI ecosystem, we aim to release frequently, as soon as features are available.
The main branch of Envoy AI Gateway is tested against both the latest stable version of Envoy Gateway and its main branch, allowing us to confidently release the Envoy AI Gateway at any time if needed.
```

nacx (approved):

PR #1461, opened 2025-10-29
feat: provide unified thinking config among anthropic and gemini models

hustxiayang:
**Description**
`thinking_config` among anthropic and gemini models are similar, for example: `thinking` and `thinking_config`; `budget_tokens` and `thinkingBudget`.  Our users do not need to set up different thinking configs for different models as we can provide a unified interface among different providers.

Related to https://github.com/envoyproxy/ai-gateway/issues/1463


docs/proposals/004-vendor-specific-fields/proposal.md
```diff
@@ -36,23 +36,14 @@ type ChatCompletionRequest struct {
...
-	ThinkingConfig *genai.GenerationConfigThinkingConfig `json:"thinkingConfig,omitempty"`
-}
-
-// AnthropicVendorFields contains GCP Anthropic-specific fields.
-type AnthropicVendorFields struct {
-	Thinking *anthropic.ThinkingConfigParamUnion `json:"thinking,omitzero"`
+	// SafetySettings: Safety settings in the request to block unsafe content in the response.
+	//
+	// https://cloud.google.com/vertex-ai/docs/reference/rest/v1/SafetySetting
+	SafetySettings []*genai.SafetySetting `json:"safetySettings,omitzero"`
```
gavrissh, line 52: is this unrelated to the `thinking` field change and a new field?

mathetake:
so if i understand correctly, this goes completely beyond the original @sukumargaonkar 's proposal https://github.com/envoyproxy/ai-gateway/blob/main/docs/proposals/004-vendor-specific-fields/proposal.md 

This is trying to define "AI Gateway Schema", and the scope and impact is huge. Could you write up a design doc and have more comments from community here? 

At the **very least**, I would request for very detailed documents and future planning on this direction as well as this kind of thing requires user facing documentation in the website, not just code. sg?

mathetake:
Please note that breaking change regarding this is **extremely impactful** than the control plane API. Once this is exposed, you cannot assume that clients will migrate properly, so we have to be very careful

hustxiayang:
@mathetake Sure, thanks a lot for the comments! One key motivation is that we do not want to break the existing codes internally. Would prepare a doc also. Thanks!

yuzisun:
> so if i understand correctly, this goes completely beyond the original @sukumargaonkar 's proposal https://github.com/envoyproxy/ai-gateway/blob/main/docs/proposals/004-vendor-specific-fields/proposal.md
> 
> This is trying to define "AI Gateway Schema", and the scope and impact is huge. Could you write up a design doc and have more comments from community here?
> 
> At the **very least**, I would request for very detailed documents and future planning on this direction as well as this kind of thing requires user facing documentation in the website, not just code. sg?

@mathetake It is not against @sukumargaonkar's original proposal for vendor specific fields. However for thinking we found out that different providers have very similar looking definitions, so it is worth unifying at the gateway level. @hustxiayang will provider a detailed writeup.

yuzisun:
> Please note that breaking change regarding this is **extremely impactful** than the control plane API. Once this is exposed, you cannot assume that clients will migrate properly, so we have to be very careful

agreed, I think we can still keep the GCP thinking vendor field but recommend the new unified thinking API shape.

mathetake:
ah sorry i should clarify that my comment about breaking change is about this new "unified api portion", not about vendor stuff already in place, and that's why I am saying we should not take this addition lightly without any very careful consideration.

mathetake:
basically API added here in this PRs will *never* be reverted without very long migration period (think about openai/anthropic, or any other public API's deprecation window)

yuzisun:
> basically API added here in this PRs will _never_ be reverted without very long migration period (think about openai/anthropic, or any other public API's deprecation window)

Adding an agenda tomorrow to discuss the deprecation policy.

nacx:
Premature abstractions can come at a significant maintenance cost, so we need proper, informed decisions before pushing for them.

>thinking_config among different providers are similar, for example: thinking and thinking_config; budget_tokens and thinkingBudget. Our users do not need to set up different thinking configs for different models as we can provide a unified interface among different providers.

This PR description is quite vague and doesn't give a good understanding on what will be covered in reality and the real user benefits. Please edit the description and add a detailed breakdown of the providers that are supported today and how their APIs look like. Let's include all [all the currently supported providers](https://aigateway.envoyproxy.io/docs/capabilities/llm-integrations/supported-providers), so we can get a good understanding on what this change would really cover (because not all of them support the same features through the OpenAI comapt APIs. For example, Cohere does not support the `reasoning_effort` field).

That would help. lot understand what this really covers and what it's optimizing for.

docs/proposals/004-vendor-specific-fields/proposal.md
hustxiayang: Hi, this is just to remove the `AnthropicVendorFields` and `GCPVertexAIGenerationConfig` in the proposals?

mathetake:
yeah so i won't be able to make it to the meeting tomorrow (sorry 🙏) but as long as the deprecation policy is agreed upon in a sane way as well as the API design is reviewed by multiple people given the prior art research (not limited to litellm), then all good to me.

of course the user facing documentation is a must but that's after the agreement we can do that

hustxiayang:
@mathetake @yuzisun @nacx It's been a while after the discussion. Any plans/concerns to merge this PR?

mathetake:
@yuzisun  @hustxiayang this needs the end user documentation change https://aigateway.envoyproxy.io/docs/capabilities/llm-integrations/vendor-specific-fields

yuzisun:
> @yuzisun @hustxiayang this needs the end user documentation change https://aigateway.envoyproxy.io/docs/capabilities/llm-integrations/vendor-specific-fields

Updated https://github.com/envoyproxy/ai-gateway/pull/1590

PR #1462, opened 2025-10-29
test: force deletion on cleanup

mathetake:
**Description**

In e2e, we cleanup resources for faster iterations when the resource is constrained such as on GHA. To make it faster and less flaky, this commit adds "force=true" flag to the deletion, which is safe anyways since it's only used in tests.


mathetake:
I believe this resolves very rare flake like this https://github.com/envoyproxy/ai-gateway/actions/runs/18912938907/job/53988466850 where Delete hung up

Issue #1389, opened 2025-10-17
Implement tracing for Anthropic /messages


mathetake:
_Description_:

Currently, the tracer implementation for /messages is missing

PR #1464, opened 2025-10-29
refactor: scaffolds anthropic messages structs for tracing

mathetake:
**Description**
This scaffolds the bare minimum data structures for Anthropic messages, which previously was not defined correctly. The sets the stage for tracing work, which will add more fields in addition to the structs defined here.

Preparation for #1389


internal/extproc/messages_processor_test.go
mathetake, line 1: deleted here were not testing anything int the first place so deleted

internal/extproc/translator/anthropic_gcpanthropic_test.go
mathetake, line 1: same comment: not testing anything so deleted

internal/extproc/messages_processor.go
```diff
@@ -273,6 +275,14 @@ func (c *messagesProcessorUpstreamFilter) ProcessResponseBody(ctx context.Contex
 			c.metrics.RecordRequestCompletion(ctx, true, c.requestHeaders)
 		}
 	}()
+	if code, _ := strconv.Atoi(c.responseHeaders[":status"]); !isGoodStatusCode(code) {
+		// For now, simply pass through error responses without modification.
+		// TODO: do the error conversion like other processors, to be able to capture any error in the proper
```
mathetake, line 280: this is a serious TODO, which basically means that this processor has been very fragile until we solve this

PR #1468, opened 2025-10-30
fix: set index of tool calls in stream mode

hustxiayang:
**Description**

There is an `index`  field (https://github.com/openai/openai-python/blob/4e8856576211064b09c0cc4a1ed35b82b169abe2/src/openai/types/chat/chat_completion_chunk.py#L48) in the tool calls definition in stream mode:  and it's not optional.  Third party libs like opentelemetry(https://github.com/open-telemetry/opentelemetry-python-contrib/blob/447aac2b0fbe88998e970941c1160147f355ec73/instrumentation-genai/opentelemetry-instrumentation-openai-v2/src/opentelemetry/instrumentation/openai_v2/patch.py#L285) is implemented based on this assumption.  The index field should be set according the number of tool calls in the streaming events which can be tracked by checking message start/stop event types.

- aws bedrock: track the tool index by contentBlockStart/Stop events
- gcp anthropic: track the tool index by content_block_start/stop events

**Discuss/suggestion**: 

- `ChatCompletionMessageToolCallParam` should not be reused in the response type.
- For both converse api and anthropic api, the stream chunks have strong dependencies. We should use complete responses to write unit tests.

### DIFF
--- a/tests/internal/testupstreamlib/testupstream/main_test.go
+++ b/tests/internal/testupstreamlib/testupstream/main_test.go
@@ -338,7 +338,7 @@ func Test_main(t *testing.T) {
 			var message eventstream.Message
 			message, err = decoder.Decode(response.Body, nil)
 			require.NoError(t, err)
-			require.Equal(t, "contentBlockDelta", message.Headers.Get("event-type").String())
+			require.Equal(t, "contentBlockDelta", message.Headers.Get(":event-type").String())
 			require.JSONEq(t, fmt.Sprintf("{\"contentBlockIndex\": 0, \"delta\":{\"text\":\"%d\"}}", i+1), string(message.Payload))
 
 			// Ensure that the server sends the response line every second.


internal/extproc/translator/gemini_helper.go
```diff
@@ -647,6 +647,44 @@ func extractToolCallsFromGeminiParts(parts []*genai.Part) ([]openai.ChatCompleti
...
+		toolCallID := uuid.New().String()
+
+		toolCall := openai.ChatCompletionChunkChoiceDeltaToolCall{
+			ID:   &toolCallID,
+			Type: openai.ChatCompletionMessageToolCallTypeFunction,
+			Function: openai.ChatCompletionMessageToolCallFunctionParam{
+				Name:      part.FunctionCall.Name,
+				Arguments: string(args),
+			},
+			Index: 0,
```
yuzisun: here should handle multiple tool calls to ensure that multiple tool calls in a single streaming response receive proper incremental indices

hustxiayang:
@yuzisun we discussed that tool call index in stream mode for gemini models is based on the assumption that all tool calls are in the same chunk. This is not verified.

PR #1470, opened 2025-10-31
aigw: removes the manual certs gen

mathetake:
**Description**

This removes the no-longer-necessary certs gen code path in `aigw run` command. This leverages the change in EG's upstream: envoyproxy/gateway/pull/7362


mathetake:
fyi @codefromthecrypt

codefromthecrypt:
Thanks so much!

PR #1471, opened 2025-11-01
test: use EG v1.6.0 rc1

mathetake:
**Description**
Envoy Gateway v1.6.0-rc.1 has been released a few days ago, so this commit uses it for our testing to prepare for our next release: https://github.com/envoyproxy/gateway/releases/tag/v1.6.0-rc.1

### DIFF
--- a/docs/why-not-official-sdk.md
+++ b/docs/why-not-official-sdk.md
@@ -1,0 +1,22 @@
+## Why not use the "official SDKs"?
+
+In [`internal/apischema`](../internal/apischema), we have defined our own data structures for various providers like OpenAI, Azure OpenAI, Anthropic, etc., for our translation logic as well as for observability purposes.
+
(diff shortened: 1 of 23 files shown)


.github/workflows/build_and_test.yaml
```diff
@@ -184,6 +184,8 @@ jobs:
         include:
           - name: latest
             envoy_gateway_version: v0.0.0-latest
+          - name: v1.6.0-rc.1
+            envoy_gateway_version: v1.6.0-rc.1
           - name: v1.5.0
```
zhaohuabing, line 189: If we want to use a new feature in v1.6.0, should we disable the related tests of that feature for v1.5.0? Or we just use whatever that are only available in v1.5.0?

For example: https://github.com/envoyproxy/ai-gateway/pull/1473

mathetake: After v1.6 is released, yes we will remove but not happening yet

PR #1473, opened 2025-11-03
chore: use header for mcp www-authentication as it's already supported in EG CustomResponse API

zhaohuabing:
**Description**
The current MCP implementation used the body field in `CustomResponse` to carry the `www-authentication` value and then copied it to the header in the extension server.

This PR switches to using header directly, as header has already been supported in CustomResponse API.

### DIFF
--- a/internal/extensionserver/mcproute.go
+++ b/internal/extensionserver/mcproute.go
@@ -7,7 +7,6 @@ package extensionserver
 
 import (
 	"fmt"
-	"os"
 	"strings"
 
 	egextension "github.com/envoyproxy/gateway/proto/extension"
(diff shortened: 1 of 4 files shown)


internal/extensionserver/mcproute.go
```diff
@@ -58,12 +55,6 @@ func (s *Server) maybeGenerateResourcesForMCPGateway(req *egextension.PostTransl
...
 	s.modifyMCPGatewayGeneratedCluster(req.Clusters)
-
-	// Modify OAuth custom response filters to add WWW-Authenticate headers.
-	// TODO: remove this step once Envoy Gateway supports this natively in the BackendTrafficPolicy ResponseOverride.
-	// https://github.com/envoyproxy/gateway/pull/6308
-	s.modifyMCPOAuthCustomResponseFilters(req.Listeners)
-
 	// TODO: remove this step once Envoy Gateway supports this natively in the BackendTrafficPolicy ResponseOverride.
 	// https://github.com/envoyproxy/gateway/pull/6308
 	s.modifyMCPOAuthCustomResponseRoute(req.Routes)
```
zhaohuabing, line 60: This will be handled in a follow-up PR.

zhaohuabing:
we need to wait until EG v1.6 is released and use 1.6 for E2E tests.

Issue #1459, opened 2025-10-29
mcp: fine-grained access control on tools


zhaohuabing:
_Description_:
Enforce access restrictions on tools based on JWT claims in the access token, ensuring clients can only access authorized tools. This is useful when tools contain sensitive data or when fine-grained access control is needed.

Example:
```yaml
spec:
  ...
  securityPolicy:
     oauth:
       issuer: "https://auth-server.example.com"
       protectedResourceMetadata:
         resource: "https://api.example.com/mcp"
    authorization:
      rules:
        - Allow
          tools: 
            - service: mcp-server1
              tools: [tool1, tool2, tool3]
           Principal:
             jwt: 
               claims:
                 - name: username
                   values: [john, alice]
```

Update the example based on the latest [MCP authorization spec](https://modelcontextprotocol.io/specification/draft/basic/authorization#scope-challenge-handling):

```yaml
spec:
  ...
  securityPolicy:
    oauth:
      issuer: "https://auth-server.example.com"
      protectedResourceMetadata:
        resource: "https://api.example.com/mcp"
    authorization:
      rules:
      - tools: [mcp-backend1_tool1, mcp-backend1_tool2, mcp-backend2_tool1]
        scopes: [foo, bar]
      - tools: [mcp-backend3_tool1]
        scopes: [baz]  
```

nacx:
Implemented in https://github.com/envoyproxy/ai-gateway/pull/1482

PR #1482, opened 2025-11-04
mcp:  JWT scope based authorization

zhaohuabing:
**Description**

This PR introduces MCP spec compatible scope based authorization for MCPRoutes.

According to the 2025-11-25 version of the MCP spec, the MCP Gateway should enforce scope-based authorization on behalf of the backend MCP server, and include the required scopes in the `WWW-Authenticate` header of the 403 response if authoriation fails.

https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization#runtime-insufficient-scope-errors

> When a client makes a request with an access token with insufficient scope during runtime operations, the server SHOULD respond with:
HTTP 403 Forbidden status code 
WWW-Authenticate header with the Bearer scheme and additional parameters:
error="insufficient_scope" - indicating the specific type of authorization failure
scope="required_scope1 required_scope2" - specifying the minimum scopes needed for the operation
resource_metadata - the URI of the Protected Resource Metadata document (for consistency with 401 responses)
error_description (optional) - human-readable description of the error

Example:

```yaml
spec:
  ...
  securityPolicy:
    oauth:
      ...
    authorization:
      rules:
        - source:
            jwt:
              scopes:
                - echo
          target:
            tools:
              - backend: mcp-backend-authorization
                toolName: echo
                when: args.text.matches("^Hello, .*!$")
        - source:
            jwt:
              scopes:
                - sum
          target:
            tools:
              - backend: mcp-backend-authorization
                toolName: sum
```

Reference: https://modelcontextprotocol.io/specification/draft/basic/authorization#scope-challenge-handling

Implements #1459


api/v1alpha1/mcp_route.go
```diff
@@ -222,6 +227,43 @@ type MCPRouteOAuth struct {
...
+	// If not specified, the default action is Deny.
+	// +optional
+	DefaultAction *egv1a1.AuthorizationAction `json:"defaultAction"`
+}
+
+// MCPRouteAuthorizationRule defines an authorization rule for MCPRoute based on the MCP authorization spec.
+// Reference: https://modelcontextprotocol.io/specification/draft/basic/authorization#scope-challenge-handling
+type MCPRouteAuthorizationRule struct {
+	// Tools defines the list of tool names this rule applies to. The name must be a fully qualified tool name including the backend name.
+	// For example, "mcp-backend-name__tool-name".
```
zhaohuabing: Or we could introduce an intermediate `MCPToolName` type:
```yaml
type MCPToolName struct{
  Backend string,
  Tool string,
}
```

I prefer the full qualified name for simplicity.

api/v1alpha1/mcp_route.go
```diff
@@ -222,6 +227,43 @@ type MCPRouteOAuth struct {
...
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:MaxItems=16
+	Tools []string `json:"tools"`
+
+	// Scopes defines the list of JWT scopes required for the rule.
+	// If multiple scopes are specified, all scopes must be present in the JWT for the rule to match.
+	//
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:MaxItems=16
+	Scopes []egv1a1.JWTScope `json:"scopes"`
```
zhaohuabing, line 307: Right now we only support the JWT scope base authz required by the MCP spec.

We can add other identifiers later if required, for example, the client id from API Key auth.

internal/mcpproxy/authorization.go
```diff
@@ -0,0 +1,124 @@
...
+	token, err := bearerToken(headers.Get("Authorization"))
+	if err != nil {
+		m.l.Info("missing or invalid bearer token", slog.String("error", err.Error()))
+		return false
+	}
+
+	claims := jwt.MapClaims{}
+	// JWT verification is performed by Envoy before reaching here. So we only need to parse the token without verification.
+	parser := jwt.NewParser(jwt.WithoutClaimsValidation())
+	if _, _, err := parser.ParseUnverified(token, claims); err != nil {
```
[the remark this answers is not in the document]
zhaohuabing: JWT verification is performed by Envoy before reaching here. So we only need to parse the token without verification.

internal/mcpproxy/authorization.go
```diff
@@ -0,0 +1,169 @@
...
+	// If the rules are defined, a valid bearer token is required.
+	token, err := bearerToken(headers.Get("Authorization"))
+	if err != nil {
+		m.l.Info("missing or invalid bearer token", slog.String("error", err.Error()))
+		return false
+	}
+
+	claims := jwt.MapClaims{}
+	// JWT verification is performed by Envoy before reaching here. So we only need to parse the token without verification.
+	if _, _, err := jwt.NewParser().ParseUnverified(token, claims); err != nil {
```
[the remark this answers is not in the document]
zhaohuabing: This is a false positive - AIGW controller generates an EG `SecuritPolicy` with a JWT authn filter, which verifies the JWT token before it reaches the MCP proxy.

internal/mcpproxy/handlers.go
```diff
@@ -540,6 +540,14 @@ func (m *MCPProxy) handleToolCallRequest(ctx context.Context, s *session, w http
 		return fmt.Errorf("%w: %s", errInvalidToolName, toolName)
 	}
 
+	// Enforce authentication if required by the route.
+	if route.authorization != nil {
+		if !m.authorizeRequest(route.authorization, headers, backendName, toolName, p.Arguments) {
+			onErrorResponse(w, http.StatusForbidden, "authorization failed")
```
zhaohuabing: According to the new MCP spec, the MCP Gateway should include the missing scopes in the WWW-Authenticate header of the 403 response when authorization fails.

However, adding theses scopes is tricky because our authorization model is more flexible than the one defined in the MCP spec. For example, there are cases where we simply don't have meaningful scopes to return:
* No rules with matching tools are found for the target
* The rule is based on other claims in the source (not scopes), so there's no corresponding scope we can include in the response.

zhaohuabing: Simplified the authorization API by removing the DefaultAction and Action within the rule. We can add it back later if required.

api/v1alpha1/mcp_route.go
```diff
@@ -227,6 +234,87 @@ type MCPRouteOAuth struct {
...
+	ToolName string `json:"toolName"`
+
+	// Arguments is a CEL expression that must evaluate to true for the rule to match.
+	// The expression is evaluated with a single variable "args" bound to the tool call arguments as a dynamic object.
+	// Guard against missing fields with null checks (e.g., args["foo"] != null && args["foo"]["bar"] == "val").
+	//
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:MaxLength=4096
+	// +optional
+	Arguments *string `json:"arguments,omitempty"`
```
zhaohuabing: After discussing this with @nacx offline, we chose CEL for argument evaluation.

CEL is easier to write and understand than regex evaluation when the arguments are complex object types.

For example:

To check against this argument:

```json
{
  "key-level1": {
    "key-level2": {
      "key-level3": "value"
    }
  }
}
```

The regex match:
```
/"key-level1"\s*:\s*\{\s*"key-level2"\s*:\s*\{\s*"key-level3"\s*:\s*"value"\s*\}\s*\}/gm
```

The CEL match:
```
args["key-level1"]["key-level2"]["key-level3"] == "value"
```

api/v1alpha1/mcp_route.go
```diff
@@ -227,6 +234,87 @@ type MCPRouteOAuth struct {
...
+	Target MCPAuthorizationTarget `json:"target"`
+}
+
+// MCPAuthorizationTarget defines the target of an authorization rule.
+type MCPAuthorizationTarget struct {
+	// Tools defines the list of tools this rule applies to.
+	//
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:MaxItems=16
```
nacx, line 285: Is there a reason for this limitation?

api/v1alpha1/mcp_route.go
```diff
@@ -227,6 +234,87 @@ type MCPRouteOAuth struct {
...
+	Tools []ToolCall `json:"tools"`
+}
+
+// MCPAuthorizationSource defines the source of an authorization rule.
+type MCPAuthorizationSource struct {
+	// JWTSource defines the JWT scopes required for this rule to match.
+	//
+	// +kubebuilder:validation:Required
+	JWTSource JWTSource `json:"jwtSource"`
```
nacx: I'd just call it `jwt` to make the YAML API less verbose.
```suggestion
	JWT JWTSource `json:"jwt"`
```

api/v1alpha1/mcp_route.go
nacx: And it would be easy enough to just add a `claims` field to support the rest of claims in a JWT token.

api/v1alpha1/mcp_route.go
```diff
@@ -227,6 +234,87 @@ type MCPRouteOAuth struct {
...
+
+	// TODO : we can add more fields in the future, e.g., audiences, claims, etc.
+}
+
+// ToolCall represents a tool call in the MCP authorization target.
+type ToolCall struct {
+	// BackendName is the name of the backend this tool belongs to.
+	//
+	// +kubebuilder:validation:Required
+	BackendName string `json:"backendName"`
```
nacx: Simplify the names for a less verbose and easier to write/read YAML API?
```suggestion
	Backend string `json:"backend"`
```

internal/filterapi/mcpconfig.go
```diff
@@ -27,6 +27,9 @@ type MCPRoute struct {
 
 	// Backends is the list of backends that this route can route to.
 	Backends []MCPBackend `json:"backends"`
+
+	// Authorization is the authorization configuration for this route.
+	Authorization *MCPRouteAuthorization `json:"authorization,omitempty"`
```
nacx, line 32: Not for this PR, but for the future:
This config is stored on a k8s secret, and in scenarios where there are many MCPRoutes and many Authorization rules, the contents may exceed the max sicze of a k8s object.
We may need to come up with a different way to store this to avoid reaching these limits. We'll create a follow-up issue to track this.

internal/filterapi/mcpconfig.go
```diff
@@ -58,3 +61,54 @@ type MCPToolSelector struct {
...
+}
+
+type MCPAuthorizationTarget struct {
+	// Tools defines the list of tools this rule applies to.
+	Tools []ToolCall `json:"tools"`
+}
+
+type MCPAuthorizationSource struct {
+	// JWTSource defines the JWT scopes required for this rule to match.
+	JWTSource JWTSource `json:"jwtSource,omitempty"`
```
nacx: Consider simplifying the names of the field in this internal api as well

internal/mcpproxy/handlers.go
```diff
@@ -540,6 +540,22 @@ func (m *MCPProxy) handleToolCallRequest(ctx context.Context, s *session, w http
 		return fmt.Errorf("%w: %s", errInvalidToolName, toolName)
 	}
 
+	// Enforce authentication if required by the route.
+	if route.authorization != nil {
+		allowed, requiredScopes := m.authorizeRequest(route.authorization, headers, backendName, toolName, p.Arguments)
```
nacx, line 554: I think it is worth having an interface to allow better authorization implementations when we need them. I was thinking something like:

```go
type Authorization[P mcp.Params] interface {
	Authorize(polixy *filterapi.MCPRouteAuthorization, principal Principal, bakend string, params P) error
    // We can also have an AuthResult interface with "OAuth" implementations that provide the required scopes, etc, so that we can have a common auth framework letting each MCP method implement auth the right way.
}
type Principal interface {} // Add whatever methods we consider appropriate here
```

Where `Principal` is the representation of the authenticated principal (source). We can have a `JWTPrincipal` implementation, etc, and we can have a `ToolCallAuthorization = Authorization[*mcp.CallToolParams]`.

This allows us to:

* Decouple the Authorization process from the HTTP request. We can have a "principal extractor" or something similar to extract the authenticated principal from the incoming request, then use it in the authorization call.
* Make things more test-friendly as everything is interface-driven and we can have mock or test implementations for the tests.
* Have different auth implementations for the different methods under the same clean abstraction.
* Support more features of the tool call request (like metadata based policies) without changing the signature.

api/v1alpha1/mcp_route.go
nacx: WDYT about renaming this to `condition`? If we do this, we could also easily support `meta` in addition to `args` to have policies based on the contents of the metadata.

internal/mcpproxy/handlers.go
```diff
@@ -540,6 +540,22 @@ func (m *MCPProxy) handleToolCallRequest(ctx context.Context, s *session, w http
...
+		if !allowed {
+			// Specify the minimum required scopes in the WWW-Authenticate header.
+			// Reference: https://mcp.mintlify.app/specification/2025-11-25/basic/authorization#runtime-insufficient-scope-errors
+			if len(requiredScopes) > 0 {
+				if challenge := buildInsufficientScopeHeader(requiredScopes, route.authorization.ResourceMetadataURL); challenge != "" {
+					w.Header().Set("WWW-Authenticate", challenge)
+				}
+			}
+			onErrorResponse(w, http.StatusForbidden, "authorization failed")
```
nacx: "access denied" is a more common message
```suggestion
			onErrorResponse(w, http.StatusForbidden, "access denied")
```

tests/e2e/mcp_route_authorization_test.go
nacx, line 1: Nice! :)

internal/mcpproxy/authorization.go
```diff
@@ -0,0 +1,239 @@
...
+		return true, nil
+	}
+
+	// If no rules are defined, deny all requests.
+	if len(authorization.Rules) == 0 {
+		return false, nil
+	}
+
+	// If the rules are defined, a valid bearer token is required.
+	token, err := bearerToken(headers.Get("Authorization"))
```
nacx, line 108: I think we should decouple the "authenticated principal extraction" from the authorization code (see my other comments around interfaces, etc).

internal/mcpproxy/authorization.go
```diff
@@ -0,0 +1,239 @@
...
+		m.l.Info("failed to parse JWT token", slog.String("error", err.Error()))
+		return false, nil
+	}
+
+	scopeSet := sets.New(extractScopes(claims)...)
+	var requiredScopesForChallenge []string
+
+	for _, rule := range authorization.Rules {
+		if !m.toolMatches(backendName, toolName, rule.Target, arguments) {
+			continue
```
nacx, line 128: If I understand correctly, with the current logic, if the MCPRoute has, say 20 tools, and I just want to limit access so that for one particular tool I can not use a particular argument (e.g. you can't call the filesystem tool passing "/root" as an argument), then I could have to create 20 auth rules to allow all tools, and then the rule with the argument filter for that particular rule, right?

Is this the UX we want?
Should we consider bringing back the default action? Having a default allow and explicit denies can also help when MCP servers add more tools, etc (and we can consider that with the existing "tool filtering" we're only exposing the tools that we want to be called... so having to always list all allowed tools if you have any rule could not be the best UX.

WDYT?

nacx: Or perhaps we should also support wildcards in tool names? We need to come up with a way to facilitate users doing what they need without being excessively verbose on the API.

api/v1alpha1/mcp_route.go
zhaohuabing: Should `meta` be inside the `ToolCall`, or `MCPAuthorizationTarget`?

api/v1alpha1/mcp_route.go
zhaohuabing: Just want to set a sane limitation for the size of the tool list for one rule. We can adjust it as needed in the future, but I think 16 should be OK for now.

internal/filterapi/mcpconfig.go
zhaohuabing: Tracked in https://github.com/envoyproxy/ai-gateway/issues/1613

internal/mcpproxy/authorization.go
zhaohuabing: That's a good point. I think we should bring the default action back. Supporting wildcard is not enough here - the tool names may not share a similar name pattern.

zhaohuabing: The use case in your example should work now, with a `Deny` rule for restricted folders and a `Allow` rule with the required sopes.
```yaml
authorization:
  defaultAction: Deny
  rules:
    - action: Deny
      target:
        tools:
          - backend: backend1
            tool: listFiles
            arguments: 'args.folder == "restricted"'

    - action: Allow
      source:
        jwt:
          scopes:
            - read
      target:
        tools:
          - backend: backend1
            tool: listFiles
```

internal/mcpproxy/handlers.go
zhaohuabing: I agree that we could eventually introduce a nicer abstraction for different principal types, but maybe we can wait until we actually have another one to support. The current code might be a bit naive and not the cleanest, but it does work for now, and we do need that special handling to gather the required scopes from the JWT principal across rules.

Since it’s functioning as-is, my thought is to hold off on refactoring until we have a clearer need. But I might be missing something—what do you think?

api/v1alpha1/mcp_route.go
zhaohuabing: Oh meta is part of the CallToolParams - then condition is a good.

zhaohuabing: Thinking about it again, condition feels a bit vague and also used in the status field, some alternatives:

* When
* Predicate
* Matcher

When reads naturally in the context of authorization rules: 

```yaml
tools:
  - backend: backend1
    tool: listFiles
    when: args.folder == "restricted"
```

zhaohuabing:
> WDYT about renaming this to condition? If we do this, we could also easily support meta in addition to args to have policies based on the contents of the metadata.

https://github.com/envoyproxy/ai-gateway/pull/1482/files#r2587966887

Thinking about it again, condition feels a bit vague and also used in the status field, some alternatives:

* When
* Predicate
* Matcher

When reads naturally in the context of authorization rules: 

```yaml
tools:
  - backend: backend1
    tool: listFiles
    when: args.folder == "restricted"
```

api/v1alpha1/mcp_route.go
nacx: I like the `when` better than the other two options. It is also used for additional conditions in Istio AuthorizationPolicies, so it's not a completely foreign name.

api/v1alpha1/mcp_route.go
```diff
@@ -177,6 +177,8 @@ type MCPBackendAPIKey struct {
 }
 
 // MCPRouteSecurityPolicy defines the security policy for a MCPRoute.
+//
+// +kubebuilder:validation:XValidation:rule="!has(self.authorization) || has(self.oauth)",message="oauth must be configured when authorization is set"
```
nacx: It hink we should relax this validation and require that _any_ authN method has been set (apikey, extauthz). In the end it is about making sure there is an identity without caring about _who_ populated that identity.

nacx: We could also remove this validation completely, because the `source` in rules is also optional. In this case, we need to properly define the API:

* If we explicitly define `authorization`, do we always require an identity? Or do we allow anonymous requests?
* Based on what we decide on this, the "empty" source in the AuthZ rules would mean "allow any authenticated principal", or "allow anyone, even unauthenticated ones".

We need to pick an approach and align the validations with it.

zhaohuabing: When a JWT source is configured in the Authorization, the OAuth2 is required for both acquiring the JWT token and the verification of the JWT token. Reading scopes and claims from an unverified token would be insecure.

The validation has been updated to "oauth must be configured when any authorization rule uses a jwt source"

nacx: Nice! Validation LGTM. Is there any reason you moved the PR back to draft? It LGTM

zhaohuabing: Trying to clean up some code and add more tests to cover the changes.

zhaohuabing:
The CI failure is not related to this PR:

We'll need to create a PAT to access the Github MCP server.
```
time=2025-12-16T23:48:41.239Z level=ERROR msg="failed to create MCP session" component=mcp-proxy backend=github error="MCP initialize request failed with status code 400 and body=bad request: GitHub App Server-To-Server Tokens are not supported for this endpoint\n"
```

nacx:
Yeah, looks like there are changes on the GH MCP: https://github.com/github/github-mcp-server/issues/1610

internal/mcpproxy/authorization.go
```diff
@@ -0,0 +1,255 @@
...
+	scopeSet := sets.New[string]()
+	token, err := bearerToken(headers.Get("Authorization"))
+	// This is just a sanity check. The actual JWT verification is performed by Envoy before reaching here, and the token
+	// should always be present and valid.
+	if err != nil {
+		m.l.Info("missing or invalid bearer token", slog.String("error", err.Error()))
+	} else {
+		claims := jwt.MapClaims{}
+		// JWT verification is performed by Envoy before reaching here. So we only need to parse the token without verification.
+		if _, _, err := jwt.NewParser().ParseUnverified(token, claims); err != nil {
```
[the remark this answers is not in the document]
nacx: Explained in the comment why verification is not needed here

Issue #1478, opened 2025-11-03
Flaky test: TestRerank_RecordRequestCompletion


mathetake:
https://github.com/envoyproxy/ai-gateway/actions/runs/19041831219/job/54380851574
```
--- FAIL: TestRerank_RecordRequestCompletion (0.01s)
    rerank_metrics_test.go:180: 
        	Error Trace:	/Users/runner/work/ai-gateway/ai-gateway/internal/metrics/rerank_metrics_test.go:180
        	Error:      	Max difference between 0.02 and 0.027829541 allowed is 0.005, but difference was -0.007829540999999999
        	Test:       	TestRerank_RecordRequestCompletion
```

mathetake:
@AyushSawant18588 could you take a look and make sure this passes with `go test -race -count=1000`?

PR #1487, opened 2025-11-04
fix: index of choice in stream mode was always 0

hustxiayang:
**Description**

Index of choice was always set as 0. In vertex api, it would return several parts in one chunk. The response id should be the same across all streaming chunks.

### DIFF
--- a/tests/e2e-aigw/examples_mcp_test.go
+++ b/tests/e2e-aigw/examples_mcp_test.go
@@ -107,6 +107,7 @@ func TestMCP_standalone(t *testing.T) {
 				toolName: "context7__get-library-docs",
 				params: map[string]any{
 					"context7CompatibleLibraryID": "/mongodb/docs",
+					"page":                        1,
 				},
 			},
 			{


hustxiayang:
https://github.com/envoyproxy/ai-gateway/issues/1478

nacx:
Could you please add unit tests so that we don't break this by mistake in the future?

yuzisun (approved):

yuzisun:
will merge after adding the test

hustxiayang:
@nacx Sure, would add it. Thanks a lot for the comment!

hustxiayang:
@nacx @yuzisun I added tests for "n==2" in non-stream mode. In stream mode, it never succeed and found it's not supported https://docs.cloud.google.com/vertex-ai/generative-ai/docs/model-reference/inference#sample-requests-text-stream-response

internal/translator/gemini_helper.go
```diff
@@ -774,10 +774,10 @@ func buildGCPModelPathSuffix(publisher, model, gcpMethod string, queryParams ...
 func geminiCandidatesToOpenAIStreamingChoices(candidates []*genai.Candidate, responseMode geminiResponseMode) ([]openai.ChatCompletionResponseChunkChoice, error) {
 	choices := make([]openai.ChatCompletionResponseChunkChoice, 0, len(candidates))
 
-	for _, candidate := range candidates {
-		if candidate == nil {
-			continue
-		}
+	// candidateCount is only supported in non-stream mode
+	// https://docs.cloud.google.com/vertex-ai/generative-ai/docs/model-reference/inference#sample-requests-text-stream-response
+	if len(candidates) == 1 && candidates[0] != nil {
```
yuzisun: I think we do not need explicitly check the length even if streaming only support one candidate, the original code should still work.

Issue #1485, opened 2025-11-04
AIGatewayRoute with model-based matching fails to route /v1/chat/completions requests despite /v1/models listing the model


googs1025:
_Description_:

Issue: AIGatewayRoute with model-based matching fails to route /v1/chat/completions requests despite /v1/models listing the model

After deploying the InferencePool example [README](https://github.com/envoyproxy/ai-gateway/tree/main/examples/inference-pool), the /v1/models endpoint correctly returns a list of available models (e.g., meta-llama/Llama-3.1-8B-Instruct, mistral:latest). However, when sending a standard OpenAI-compatible request to /v1/chat/completions with a model field in the JSON body, Envoy AI Gateway returns:

```
No matching route found. It is likely because the model specified in your request is not configured in the Gateway.
```
This suggests that the gateway fails to match the model field from the request body to the routes defined in the AIGatewayRoute, even though the model is listed and the route appears correctly configured.

Expected behavior:

The request should be routed to the appropriate backend (e.g., vllm-llama3-8b-instruct InferencePool) based on the model value in the JSON payload, without requiring custom headers like x-ai-eg-model.

Repro Steps
Deploy the example manifests:
```
kubectl apply -f base.yaml
kubectl apply -f aigwroute.yaml
```
Forward the gateway service port:
```
kubectl port-forward svc/envoy-default-inference-pool-with-aigwroute-d416582c 8081:80 -n envoy-gateway-system
```
Verify models are listed:
```
curl http://localhost:8081/v1/models
```

Send a chat completion request:
```
curl -X POST http://localhost:8081/v1/chat/completions \
  -H "Content-Type: application/json" \
  -d '{
        "model": "meta-llama/Llama-3.1-8B-Instruct",
        "messages": [{"role": "user", "content": "Say this is a test"}]
      }'
```
Observe error:
```
No matching route found. It is likely because the model specified in your request is not configured in the Gateway.
```

```yaml
root@iZ6wed7dv05mxum2042djiZ:~/ai-gateway/examples/inference-pool# kubectl apply -f base.yaml
service/mistral-upstream created
deployment.apps/mistral-upstream created
inferencepool.inference.networking.k8s.io/mistral created
inferenceobjective.inference.networking.x-k8s.io/mistral created
serviceaccount/mistral-epp created
service/mistral-epp created
deployment.apps/mistral-epp created
configmap/plugins-config created
role.rbac.authorization.k8s.io/pod-read created
[...]

PR #1488, opened 2025-11-04
docs: fix example in examples/inference-pool

googs1025:
**Description**
fix: https://github.com/envoyproxy/ai-gateway/issues/1485

### DIFF
--- a/internal/extproc/translator/openai_gcpanthropic.go
+++ b/internal/extproc/translator/openai_gcpanthropic.go
@@ -125,7 +125,7 @@ func translateAnthropicToolChoice(openAIToolChoice *openai.ChatCompletionToolCho
 		if choice.Type == openai.ToolTypeFunction && choice.Function.Name != "" {
 			toolChoice = anthropic.ToolChoiceUnionParam{
 				OfTool: &anthropic.ToolChoiceToolParam{
-					Type:                   constant.Tool(choice.Type),
+					Type:                   constant.Tool("tool"),
 					Name:                   choice.Function.Name,
 					DisableParallelToolUse: disableParallelToolUse,
 				},
--- a/internal/extproc/translator/openai_gcpanthropic_test.go
+++ b/internal/extproc/translator/openai_gcpanthropic_test.go
@@ -973,7 +973,7 @@ func TestTranslateOpenAItoAnthropicTools(t *testing.T) {
 			},
 			expectedTools: anthropicTestTool,
 			expectedToolChoice: anthropic.ToolChoiceUnionParam{
-				OfTool: &anthropic.ToolChoiceToolParam{Type: "function", Name: "my_func"},
+				OfTool: &anthropic.ToolChoiceToolParam{Type: "tool", Name: "my_func"},
 			},
 		},
 		{
--- a/manifests/charts/ai-gateway-helm/templates/admission_webhook.yaml
+++ b/manifests/charts/ai-gateway-helm/templates/admission_webhook.yaml
@@ -30,7 +30,7 @@ webhooks:
     sideEffects: None
     admissionReviewVersions: ["v1"]
     timeoutSeconds: 10
-    failurePolicy: Ignore
+    failurePolicy: Fail
 ---
 {{- if .Values.controller.mutatingWebhook.certManager.enable }}
 apiVersion: cert-manager.io/v1


examples/inference-pool/aigwroute.yaml
```diff
@@ -38,16 +38,6 @@ spec:
...
-            - type: Exact # This demonstrates how to configure complex header matching rule such as API key auth for a particular model.
-              name: Authorization
-              value: sk-abcdefghijklmnopqrstuvwxyz
-        - headers:
-            - type: Exact
-              name: x-ai-eg-model
-              value: meta-llama/Llama-3.1-8B-Instruct
-            - type: Exact
-              name: Authorization
-              value: sk-zyxwvutsrqponmlkjihgfedcba
```
nacx, line 50: Can you keep these examples and update the README instead?

We have [e2e tests that use this configuration](https://github.com/envoyproxy/ai-gateway/blob/main/tests/e2e-inference-extension/inference_pool_test.go#L55-L87) that are failing because of this change.

We want to make sure the examples work, and that's the reason many of them are used in e2e tests. Let's just fix the README and keep the current configuration instead?

examples/inference-pool/README.md
```diff
@@ -2,11 +2,19 @@
...
+- One standard `Backend` as a fallback for non-InferencePool traffic
+
+Routing is controlled by the `x-ai-eg-model` HTTP header.
+
 ## Files in This Directory
 
 - **`envoy-gateway-values-addon.yaml`**: Envoy Gateway values addon for InferencePool support. Combine with `../../manifests/envoy-gateway-values.yaml`.
 - **`base.yaml`**: Complete example that includes Gateway, AIServiceBackend, InferencePool CRDs, and a sample application deployment.
 - **`aigwroute.yaml`**: Example AIGatewayRoute that uses InferencePool as a backend.
+- **`vllm-llama3-8b-instruct.yaml`**: Example vllm-llama3-8b-instruct that uses InferencePool as a backend.
```
nacx: Instead of adding this new file, we could document here the same we have documented in the docs site: https://aigateway.envoyproxy.io/docs/capabilities/inference/aigatewayroute-inferencepool#step-3-deploy-inference-backends (which are BTW the same ones [applied in the e2e setup](https://github.com/envoyproxy/ai-gateway/blob/main/tests/internal/e2elib/e2elib.go#L367-L386)).

nacx (commented):
Left a couple of comments:

* We use the examples in e2e tests to make sure they stay current and work.
* Can you update the README to document what's missing to be able to run the examples manually, instead of changing the examples and adding the new file?

Thank you!

googs1025:
This makes reading easier: https://github.com/googs1025/ai-gateway/tree/reading/examples/inference-pool

examples/inference-pool/README.md
googs1025: removed. and add some info in README.md

examples/inference-pool/aigwroute.yaml
```diff
@@ -41,13 +41,6 @@ spec:
             - type: Exact # This demonstrates how to configure complex header matching rule such as API key auth for a particular model.
               name: Authorization
               value: sk-abcdefghijklmnopqrstuvwxyz
-        - headers:
```
googs1025, line 44: I deleted the duplicate parts.

examples/inference-pool/aigwroute.yaml
googs1025: In my actual testing, the request only succeeded when I deleted this part. I'm not sure what the problem is.

```go
            - type: Exact # This demonstrates how to configure complex header matching rule such as API key auth for a particular model.
              name: Authorization
              value: sk-abcdefghijklmnopqrstuvwxyz
```

googs1025: like this: 

```bash
root@iZ6wec7mrotnckp5tv4y8yZ:~# curl -v http://localhost:8080/v1/chat/completions   -H "Content-Type: application/json"   -H "x-ai-eg-model: meta-llama/Llama-3.1-8B-Instruct"   -H "Authorization: Bearer sk-abcdefghijklmnopqrstuvwxyz"   -d '{
    "model": "meta-llama/Llama-3.1-8B-Instruct",
    "messages": [{"role": "user", "content": "Explain quantum computing."}]
  }'
* Host localhost:8080 was resolved.
* IPv6: ::1
* IPv4: 127.0.0.1
*   Trying [::1]:8080...
* Connected to localhost (::1) port 8080
> POST /v1/chat/completions HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/8.5.0
> Accept: */*
> Content-Type: application/json
> x-ai-eg-model: meta-llama/Llama-3.1-8B-Instruct
> Authorization: Bearer sk-abcdefghijklmnopqrstuvwxyz
> Content-Length: 130
>
< HTTP/1.1 404 Not Found
< content-length: 115
< content-type: text/plain
< date: Tue, 04 Nov 2025 14:59:23 GMT
<
* Connection #0 to host localhost left intact
No matching route found. It is likely because the model specified in your request is not configured in the Gateway.root@iZ6wec7mrotnckp5tv4y8yZ:~#

```

examples/inference-pool/aigwroute.yaml
nacx: It's not a duplicate. It uses a different API key. You can either remove this _and_ remove the test case that uses this key, or keep it.

googs1025: thanks for reminding this, I keep this in yaml file

examples/inference-pool/README.md
```diff
@@ -2,10 +2,20 @@
...
+Routing is controlled by the `x-ai-eg-model` HTTP header.
+
 ## Files in This Directory
 
 - **`envoy-gateway-values-addon.yaml`**: Envoy Gateway values addon for InferencePool support. Combine with `../../manifests/envoy-gateway-values.yaml`.
-- **`base.yaml`**: Complete example that includes Gateway, AIServiceBackend, InferencePool CRDs, and a sample application deployment.
+- **`base.yaml`**: Deploys all inference backends and supporting resources using the **standard approach documented in the official guide**. This includes:
+  - A `mistral` backend with custom Endpoint Picker configuration
+  - A standard fallback backend (`envoy-ai-gateway-basic-testupstream`) for non-InferencePool routing
+  - The `vllm-llama3-8b-instruct` inference backend (deployed exactly as described in [Step 3: Deploy Inference Backends](https://aigateway.envoyproxy.io/docs/capabilities/inference/aigatewayroute-inferencepool#step-3-deploy-inference-backends))
```
nacx: This third point is not included in the `base.yaml` and could be confusing as-is. I'd just add a note outside the bullet points, or simply remove it as you already mention that in teh Quick start section below.

examples/inference-pool/README.md
```diff
@@ -27,16 +37,61 @@ This example demonstrates how to use AI Gateway with the InferencePool feature,
...
+```bash
+GATEWAY_HOST=$(kubectl get gateway/inference-pool-with-aigwroute -n default -o jsonpath='{.status.addresses[0].value}')
+echo "Gateway available at: http://${GATEWAY_HOST}"
+```
+
+Then send a request:
+
+```bash
+curl -X POST "http://${GATEWAY_HOST}/v1/chat/completions" \
+  -H "x-ai-eg-model: meta-llama/Llama-3.1-8B-Instruct" \
```
nacx, line 61: Requests to this model need the additional Authorization header, right? (here and in the other examples that use this model).

nacx (approved):
Thanks!

examples/inference-pool/aigwroute.yaml
googs1025: oh. this is my bad.

`-H "Authorization: Bearer sk-abcdefghijklmnopqrstuvwxyz"` -> `-H "Authorization: sk-abcdefghijklmnopqrstuvwxyz"`

googs1025: then this can work as expected

PR #1489, opened 2025-11-04
fix: mcp route to gateway cross namespace reference

### LINKED ISSUE #1483: Fix MCP Route & Gateway Reconciliation for cross namespace reference
Follows #1185 

Description: 
When MCPRoute is created in a different namespace than the gateway, the MCPRoute changes are not propagated to the extproc.

Reproducing steps:
Create/Edit an MCPRoute in different namespace than Gateway resource. When the controller trys to sync gateway, it fails with error `Gateway not found: {namespace: <> , name: <>`

Hritik003:
**Description**
Previously when MCPRoute being reconciled, the controller assumed that all the parentRefs are local, which was a bug for cross-namespace use cases. This fixes the bug which existed in the syncGateawy function of MCPRoute controller.

**Related Issues / PRs**

Fixes #1483

### DIFF
--- a/cmd/aigw/translate.go
+++ b/cmd/aigw/translate.go
@@ -30,7 +30,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	gwapiv1 "sigs.k8s.io/gateway-api/apis/v1"
-	gwapiv1a3 "sigs.k8s.io/gateway-api/apis/v1alpha3"
 	kyaml "sigs.k8s.io/yaml"
 
 	aigv1a1 "github.com/envoyproxy/ai-gateway/api/v1alpha1"
@@ -110,7 +109,7 @@ func collectObjects(yamlInput string, out io.Writer, logger *slog.Logger) (
 	mcpRoutes []*aigv1a1.MCPRoute,
 	aigwBackends []*aigv1a1.AIServiceBackend,
 	backendSecurityPolicies []*aigv1a1.BackendSecurityPolicy,
-	backendTLSConfigs []*gwapiv1a3.BackendTLSPolicy,
+	backendTLSConfigs []*gwapiv1.BackendTLSPolicy,
 	gws []*gwapiv1.Gateway,
 	secrets []*corev1.Secret,
 	envoyProxies []*egv1a1.EnvoyProxy,
@@ -184,7 +183,7 @@ func translateCustomResourceObjects(
 	mcpRoutes []*aigv1a1.MCPRoute,
 	aigwBackends []*aigv1a1.AIServiceBackend,
 	backendSecurityPolicies []*aigv1a1.BackendSecurityPolicy,
-	backendTLSPolicies []*gwapiv1a3.BackendTLSPolicy,
+	backendTLSPolicies []*gwapiv1.BackendTLSPolicy,
 	gws []*gwapiv1.Gateway,
 	usedDefinedSecrets []*corev1.Secret,
 	logger *slog.Logger,
--- a/manifests/charts/ai-gateway-helm/templates/admission_webhook.yaml
+++ b/manifests/charts/ai-gateway-helm/templates/admission_webhook.yaml
@@ -54,84 +54,39 @@ metadata:
   namespace: {{ .Release.Namespace }}
 spec:
   selfSigned: {}
-{{- else }}
 ---
+{{- else }}
+{{- $caCrt := "" }}
+{{- $tlsCrt := "" }}
+{{- $tlsKey := "" }}
+{{/* Check fi the secret exists to avoid regenerating the certificate on upgrades */}}
+{{- $existing := lookup "v1" "Secret" .Release.Namespace .Values.controller.mutatingWebhook.tlsCertSecretName }}
+{{- if $existing }}
+  {{- $caCrt = index $existing.data .Values.controller.mutatingWebhook.caBundleName }}
+  {{- $tlsCrt = index $existing.data .Values.controller.mutatingWebhook.tlsCertName }}
+  {{- $tlsKey = index $existing.data .Values.controller.mutatingWebhook.tlsKeyName }}
+{{- else }}
+  {{- $serviceName := include "ai-gateway-helm.controller.fullname" . }}
+  {{- $ca := genCA (printf "%s-ca" $serviceName) 3650 }}
+  {{- $dnsNames := list
+        $serviceName
+        (printf "%s.%s" $serviceName .Release.Namespace)
+        (printf "%s.%s.svc" $serviceName .Release.Namespace)
+        (printf "%s.%s.svc.cluster.local" $serviceName .Release.Namespace)
+  -}}
+  {{- $cert := genSignedCert (printf "%s.%s.svc" $serviceName .Release.Namespace) nil $dnsNames 365 $ca }}
+  {{- $caCrt = $ca.Cert | b64enc }}
+  {{- $tlsCrt = $cert.Cert | b64enc }}
+  {{- $tlsKey = $cert.Key | b64enc }}
+{{- end }}
 apiVersion: v1
 kind: Secret
 metadata:
   name: {{ .Values.controller.mutatingWebhook.tlsCertSecretName }}
   namespace: {{ .Release.Namespace }}
-stringData:
-  ca.crt: |
-    -----BEGIN CERTIFICATE-----
-    MIIDOzCCAiOgAwIBAgIUU+g1Upp1Qtfpk87zY5H2/EY55QUwDQYJKoZIhvcNAQEL
-    BQAwLTELMAkGA1UEBhMCQVUxHjAcBgNVBAMMFWFpLWdhdGV3YXktY29udHJvbGxl
-    cjAeFw0yNTA1MjAxNjQzNTJaFw0zNTA1MTgxNjQzNTJaMC0xCzAJBgNVBAYTAkFV
-    MR4wHAYDVQQDDBVhaS1nYXRld2F5LWNvbnRyb2xsZXIwggEiMA0GCSqGSIb3DQEB
-    AQUAA4IBDwAwggEKAoIBAQDKN5YmMh7TgGqNpedC0DWBWdn2pMiHtCeRlTkluDjK
-    l+ZeleiR7rooNUXc6gE02RAaRCEaNMSZL3m6BkZ1Xoo92Mvabu+ORkwApO+OTIvj
-    NsYb3/blsST1qHXApm7n886Ed80CG3Jczi7AioXsAhTv+SoJeQJsoKLeVYV5m5l/
-    j4xoJl9fY+lzpmgdcALBm7FDrAbsEgjKwmFEQAxTNxWowZDiARW21io45saC411S
-    m/ZhthSxDQpqSzPwYcXwR04syZxGUewYrpIE54hRsM8KwpqNEZVnjlaKBssiEgG8
-    97sx9wDb3HLzep7FShKz4LslePAc8DmvdYjnooZaxzsfAgMBAAGjUzBRMB0GA1Ud
-    DgQWBBS9puJ0i+zKW4Y3FY2NvRKAb0ONYzAfBgNVHSMEGDAWgBS9puJ0i+zKW4Y3
-    FY2NvRKAb0ONYzAPBgNVHRMBAf8EBTADAQH/MA0GCSqGSIb3DQEBCwUAA4IBAQCZ
-    ZgPCQnaXt/xSJ7oUFBMba5TLeqvzIKPeDvS0ii64tZeQ75R7nQvSVt2QnGDjpyJ+
-    00ERja7jWpjL3IWijmE199vv40ZY4pajhBAL3y8wPf6vRh8d7TO4XmN7hser2tci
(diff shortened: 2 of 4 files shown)


nacx:
Thanks!

Looking at the existing e2e tests for MCP routes, we already have one test that uses an MCP route that has a cross-namespace reference to the gateway: https://github.com/envoyproxy/ai-gateway/blob/main/tests/e2e/testdata/mcp_route.yaml#L96-L106

And that e2e test is already passing.

I haven't looked deep into the details yet, but where is the real gap in testing, if the current tests that exercise routing already worked, but the feature was not fully working as expected?

mathetake:
@nacx previously, the yaml was applied altogether so sync bug being fixed here doesn't matter, and the tests still passing because other local-MCP route in anyways triggers the ext proc config creation. This is about fixing the case where only the cross-namspaced MCPRoute is applied/edited and the sync not happening.

Hritik003:
> Thanks!
> 
> Looking at the existing e2e tests for MCP routes, we already have one test that uses an MCP route that has a cross-namespace reference to the gateway: https://github.com/envoyproxy/ai-gateway/blob/main/tests/e2e/testdata/mcp_route.yaml#L96-L106
> 
> And that e2e test is already passing.
> 
> I haven't looked deep into the details yet, but where is the real gap in testing, if the current tests that exercise routing already worked, but the feature was not fully working as expected?

the [existing test ](https://github.com/envoyproxy/ai-gateway/blob/main/tests/e2e/testdata/mcp_route.yaml#L96-L106) didnt have a namespace in its parent ref, hence it is always defaulted to mcp-route's ns.

tests/e2e/testdata/cross_namespace_mcproute.yaml
mathetake, line 1: can you instead of adding this new yaml, could you edit the existing test case's MCPRoute and see the propagation happens like https://github.com/envoyproxy/ai-gateway/blob/2fa6b27fe9bfdb0be4f108f9dec595735050cd1d/tests/e2e/testupstream_test.go#L256

internal/controller/mcp_route_test.go
```diff
@@ -372,3 +372,50 @@ func TestMCPRouteController_ensureMCPBackendRefHTTPFilter(t *testing.T) {
...
+	gateway2 := &gwapiv1.Gateway{
+		ObjectMeta: metav1.ObjectMeta{Name: "gateway2", Namespace: "other-ns"},
+	}
+
+	err := c.Create(t.Context(), gateway1)
+	require.NoError(t, err)
+	err = c.Create(t.Context(), gateway2)
+	require.NoError(t, err)
+
+	//Create Controller
```
mathetake: unnecessary comment

internal/controller/mcp_route_test.go
```diff
@@ -372,3 +372,50 @@ func TestMCPRouteController_ensureMCPBackendRefHTTPFilter(t *testing.T) {
...
+	mcpRoute := &aigv1a1.MCPRoute{
+		ObjectMeta: metav1.ObjectMeta{Name: "test-route", Namespace: "default"},
+		Spec: aigv1a1.MCPRouteSpec{
+			ParentRefs: []gwapiv1.ParentReference{
+				{Name: gwapiv1.ObjectName("gateway1"), Namespace: ptr.To(gwapiv1.Namespace("other-ns"))},
+				{Name: gwapiv1.ObjectName("gateway2"), Namespace: ptr.To(gwapiv1.Namespace("default"))},
+			},
+		},
+	}
+	//Sync Gateways
```
mathetake: same

tests/e2e/cross_namespace_test.go
```diff
@@ -69,3 +69,44 @@ func TestCrossNamespace(t *testing.T) {
...
 }
+
+// TestCrossNamespaceMCPRoute tests MCPRoute with cross-namespace references.
+// This test validates that:
+//  1. A Gateway in one namespace (gateway-ns) can be referenced by an MCPRoute in another namespace (mcp-route-ns)
+//  2. An MCP Backend in one namespace (mcp-backend-ns) can be referenced by an MCPRoute in another namespace (mcp-route-ns)
+//     when a valid ReferenceGrant exists
+//  3. The generated HTTPRoutes and other resources work correctly across namespaces
+//  4. Traffic routing works end-to-end through the cross-namespace MCP setup.
+func TestCrossNamespaceMCPRoute(t *testing.T) {
```
mathetake, line 78: let's delete as per the comment

mathetake:
ah existing tests are not cross name space in the first place heh

tests/e2e/testdata/cross_namespace_mcproute.yaml
mathetake: nvm

mathetake:
@Hritik003 could you check the ci and follow the same style here

mathetake:
also you don't need to explain every change in your PR description. it's a commit message per CONTRIBUTING.md

Hritik003:
> also you don't need to explain every change in your PR description. it's a commit message per CONTRIBUTING.md

mb ;-;

mathetake:
@Hritik003 could you makes sure `make precommit` is executed 🙏

Hritik003:
> @Hritik003 on second though, the test case added here is not necessary. thanks @nacx
> 
> https://github.com/envoyproxy/ai-gateway/blob/2fa6b27fe9bfdb0be4f108f9dec595735050cd1d/tests/e2e/testdata/mcp_route.yaml#L113
> 
> it's already doing a cross name stuff. could you
> 
> * delete the newly added e2e test
> * Address my original comment [fix: fix mcp route to gateway cross namespace reference #1489 (comment)](https://github.com/envoyproxy/ai-gateway/pull/1489#discussion_r2491258210)

the cross namespace is for gateway right ? which is present in the parent ref? and not the backend ref?

mathetake:
It's a crossnamespace reference 
https://github.com/envoyproxy/ai-gateway/blob/2fa6b27fe9bfdb0be4f108f9dec595735050cd1d/tests/e2e/testdata/mcp_route.yaml#L100

nacx:
>the [existing test ](https://github.com/envoyproxy/ai-gateway/blob/main/tests/e2e/testdata/mcp_route.yaml#L96-L106) didnt have a namespace in its parent ref, hence it is always defaulted to mcp-route's ns.

Yes, there was a namespace reference in the parent reference.
Looks like @mathetake explanation could be the cause of test passing: there are other MCPRoutes being reconciled that will trigger the Gateway reconcile (even if the cross-namespace route fails) making the existing test to "accidentally work".

In order to be sure about the root cause, can you:

1. In your cross-namespace test, add a comment to the MCPRoute explaining that in that test ,we are only using ONE MCPRoute to make sure there are no side effects from the reconciliation of other routes (this will keep the test future-proof, and we will know we shouldn't add more routes to it).
2. Run this e2e test you added **without the code fixes**, and post here the results, to prove the theory is OK? If the theory is OK this new e2e should always fail without the fix.

mathetake:
+1

tests/e2e/cross_namespace_test.go
```diff
@@ -69,3 +71,40 @@ func TestCrossNamespace(t *testing.T) {
...
 }
+
+// TestCrossNamespaceMCPRoute tests MCPRoute with cross-namespace references.
+// This test validates that:
+//  1. A Gateway in one namespace (mcp-gw) can be referenced by an MCPRoute in another namespace (mcp-tenant)
+func TestCrossNamespaceMCPRoute(t *testing.T) {
+	const manifest = "testdata/cross_namespace_mcproute.yaml"
+	require.NoError(t, e2elib.KubectlApplyManifest(t.Context(), manifest))
+	t.Cleanup(func() {
+		_ = e2elib.KubectlDeleteManifest(t.Context(), manifest)
```
mathetake: t.Context is already canceled at the time when cleanup functions are called 
```suggestion
		_ = e2elib.KubectlDeleteManifest(context.Background(), manifest)
```

tests/e2e/cross_namespace_test.go
```diff
@@ -69,3 +71,40 @@ func TestCrossNamespace(t *testing.T) {
...
+	})
+
+	const egSelector = "gateway.envoyproxy.io/owning-gateway-name=mcp-gateway"
+	e2elib.RequireWaitForGatewayPodReady(t, egSelector)
+	fwd := e2elib.RequireNewHTTPPortForwarder(t, e2elib.EnvoyGatewayNamespace, egSelector, e2elib.EnvoyGatewayDefaultServicePort)
+	defer fwd.Kill()
+	client := mcp.NewClient(&mcp.Implementation{Name: "demo-http-client", Version: "0.1.0"}, nil)
+	t.Run("mcp-request-succeeds with cross-namespace gateway", func(t *testing.T) {
```
mathetake: i don't think this subtest is necessary

Hritik003: how do we want to test the `testdata/cross_namespace_mcproute.yaml` in that case?

mathetake: i meant delete this exact line ... `t.Run("mcp-request-succeeds with cross-namespace gateway", func(t *testing.T) {`

mathetake: what's the purpose of this nested subtest in other words??

tests/e2e/cross_namespace_test.go
```diff
@@ -69,3 +71,39 @@ func TestCrossNamespace(t *testing.T) {
...
+		_ = e2elib.KubectlDeleteManifest(context.Background(), manifest)
+	})
+
+	const egSelector = "gateway.envoyproxy.io/owning-gateway-name=mcp-gateway"
+	e2elib.RequireWaitForGatewayPodReady(t, egSelector)
+	fwd := e2elib.RequireNewHTTPPortForwarder(t, e2elib.EnvoyGatewayNamespace, egSelector, e2elib.EnvoyGatewayDefaultServicePort)
+	defer fwd.Kill()
+	client := mcp.NewClient(&mcp.Implementation{Name: "demo-http-client", Version: "0.1.0"}, nil)
+
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
```
mathetake: this should be inside of eventually closure below

tests/e2e/testdata/cross_namespace_mcproute.yaml
```diff
@@ -0,0 +1,113 @@
...
+    matchLabels:
+      app: mcp-backend-tenant
+  template:
+    metadata:
+      labels:
+        app: mcp-backend-tenant
+    spec:
+      containers:
+        - name: mcp-backend-tenant
+          image: docker.io/hritikraj003/ai-gateway-testmcpserver:latest
```
mathetake: can you fix this?

tests/e2e/cross_namespace_test.go
```diff
@@ -69,3 +71,39 @@ func TestCrossNamespace(t *testing.T) {
...
+	var err error
+	var sess *mcp.ClientSession
```
mathetake: i mean if you are not using session outside, then shall we just make it local in the closure?

Hritik003: session is kept outside so that it is accessible to `t.Cleanup(func() { _ = sess.Close() })` or do we even need it?

mathetake: for what you want to close it in t.Cleanup? Can you elaborate on the rationale?

tests/e2e/cross_namespace_test.go
```diff
@@ -69,3 +71,39 @@ func TestCrossNamespace(t *testing.T) {
...
+	})
+
+	const egSelector = "gateway.envoyproxy.io/owning-gateway-name=mcp-gateway"
+	e2elib.RequireWaitForGatewayPodReady(t, egSelector)
+	fwd := e2elib.RequireNewHTTPPortForwarder(t, e2elib.EnvoyGatewayNamespace, egSelector, e2elib.EnvoyGatewayDefaultServicePort)
+	defer fwd.Kill()
+	client := mcp.NewClient(&mcp.Implementation{Name: "demo-http-client", Version: "0.1.0"}, nil)
+
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	var err error
```
mathetake: also no reason to declare here

tests/e2e/cross_namespace_test.go
Hritik003: thinking behind was to ensure the session is closed when the test finishes, regardless of whether it passes or fails.

mathetake: no reason to escape outside eventually closure anyways..

mathetake: > thinking behind was to ensure the session is closed when the test finishes, regardless of whether it passes or fails.

what you are saying is "why we close session", and not answering my question about "why do you want to close the session outside the closure by sacrificing the readability by closure capturing the variable outside"

Hritik003: You're right - the original design with the cleanup outside was poor,
the closure captures a variable from outer scope that gets assigned inside the `require.Eventually` loop
and the resource allocation (inside Eventually) and cleanup (outside Eventually) are separated, hurting readability

mathetake:
@Hritik003 the new test failing

Hritik003:
> @Hritik003 the new test failing

Worked locally though ;-;
Will check

mathetake:
@Hritik003 thanks for the multiple iterations! I appreciate it!

PR #1492, opened 2025-11-04
feat: route-level body mutation for top-level field

### LINKED ISSUE #1481: Allow route level body mutation
_Description_:

Similar to route level header mutation, route level body mutation will allow user to specify what field to modify/override/replace. 

For instance, when have multiple backends OpenAI and Azure stack together to serve GPT-4.1 traffic. route level body mutation allow user to specify service_tier value only for OpenAI but Azure which doesn't support service_tier.

Similar override will apply to Anthropic's service_tier override.

xiaolin593:
**Description**

Support route-level body mutation for top-level field. route level body mutation is useful when have heterogeneous cloud-provider backends support same model but with slightly different request body fields to override esp  OpenAI's `service_tier` field

Close #1481

### DIFF
--- a/internal/extproc/translator/openai_gcpvertexai_test.go
+++ b/internal/extproc/translator/openai_gcpvertexai_test.go
@@ -1653,7 +1653,7 @@ func getChatCompletionResponseChunk(body []byte) []openai.ChatCompletionResponse
 func TestOpenAIToGCPVertexAITranslatorV1ChatCompletion_StreamingParallelToolIndex(t *testing.T) {
 	translator := NewChatCompletionOpenAIToGCPVertexAITranslator("gemini-2.0-flash-001").(*openAIToGCPVertexAITranslatorV1ChatCompletion)
 	// Mock multiple GCP streaming response with parallel tool calls
-	gcpToolCallsChunk := `{
+	gcpToolCallsChunk := `data: {
     "candidates": [
         {
             "content": {
@@ -1670,8 +1670,9 @@ func TestOpenAIToGCPVertexAITranslatorV1ChatCompletion_StreamingParallelToolInde
                 "role": "model"
             }
         }
-    ],
-	"candidates": [
+]}
+
+data: {"candidates": [
         {
             "content": {
                 "parts": [
@@ -1687,8 +1688,7 @@ func TestOpenAIToGCPVertexAITranslatorV1ChatCompletion_StreamingParallelToolInde
                 "role": "model"
             }
         }
-    ],
-}`
+]}`
 
 	expectedChatCompletionChunks := []openai.ChatCompletionResponseChunk{
 		{
@@ -1702,7 +1702,7 @@ func TestOpenAIToGCPVertexAITranslatorV1ChatCompletion_StreamingParallelToolInde
 								Index: int64(0),
 								ID:    ptr.To("123"),
 								Function: openai.ChatCompletionMessageToolCallFunctionParam{
-									Arguments: `{"location": "New York City"}`,
+									Arguments: `{"location":"New York City"}`,
 									Name:      "get_weather",
 								},
 								Type: "function",
@@ -1724,7 +1724,7 @@ func TestOpenAIToGCPVertexAITranslatorV1ChatCompletion_StreamingParallelToolInde
 								Index: int64(1),
 								ID:    ptr.To("123"),
 								Function: openai.ChatCompletionMessageToolCallFunctionParam{
-									Arguments: `{"location": "Shang Hai}`,
+									Arguments: `{"location":"Shang Hai"}`,
 									Name:      "get_weather",
 								},
 								Type: "function",
@@ -1750,6 +1750,7 @@ func TestOpenAIToGCPVertexAITranslatorV1ChatCompletion_StreamingParallelToolInde
 
 	body := bodyMut.Mutation.(*extprocv3.BodyMutation_Body).Body
 	chatCompletionChunks := getChatCompletionResponseChunk(body)
+	require.Len(t, chatCompletionChunks, 2)
 
 	for idx, chunk := range chatCompletionChunks {
 		chunk.Choices[0].Delta.ToolCalls[0].ID = ptr.To("123")


xiaolin593:
> I think this is not bad and looks a good start. The API still leaves a room for the future addition of more complex jsonpatch style!
> 
> cc @tiswanso do you have any comments?

Looks like top-level field can solve the use-case cited at here https://github.com/envoyproxy/ai-gateway/discussions/1386#discussioncomment-14714421 already?

What is the specific need for jsonpatch deep level field mutation? Wondering if we ever encounter such complex use case for ChatCompletion, Completion or other AI workload APIs.

mathetake:
at least vendor specific fields forced at gateway would be the use case, i.e. any field in https://aigateway.envoyproxy.io/docs/capabilities/llm-integrations/vendor-specific-fields/  so the path might be nested at least not necessarily require jsonpatch though

xiaolin593:
had to force-push due to one of commits wasn't signed.

internal/controller/gateway.go
```diff
@@ -178,6 +178,11 @@ func headerMutationToFilterAPI(m *aigv1a1.HTTPHeaderMutation) *filterapi.HTTPHea
 	return ret
 }
 
+// TODO
+func bodyMutationToFilterAPI(m *aigv1a1.HTTPBodyMutation) *filterapi.HTTPBodyMutation {
```
aabchoo, line 182: Finish todo?

tiswanso:
>> What is the specific need for jsonpatch deep level field mutation? Wondering if we ever encounter such complex use case for ChatCompletion, Completion or other AI workload APIs.

> at least vendor specific fields forced at gateway would be the use case, i.e. any field in https://aigateway.envoyproxy.io/docs/capabilities/llm-integrations/vendor-specific-fields/ so the path might be nested at least not necessarily require jsonpatch though

Right... the nested args is similar to the use-case I actually have with an internally supported provider gateway/proxy that requires setting of `user.appkey` in the body.  

Only allowing setting the top-level arg would still allow my use-case to work but with the caveat that it'd clobber any existing sub-args set under the top-level arg in the payload.  JsonPath would allow for supporting more semantics there.

internal/extproc/messages_processor.go
```diff
@@ -205,6 +207,18 @@ func (c *messagesProcessorUpstreamFilter) ProcessRequestHeaders(ctx context.Cont
 		}
 	}
 
+	// Apply body mutations from the route and also restore original body on retry.
+	if b := c.bodyMutator; b != nil {
+		if bodyMutation != nil && bodyMutation.GetBody() != nil && len(bodyMutation.GetBody()) > 0 {
```
mathetake: so this means that only the translation case can apply body mutation? can you fix it?

internal/extproc/bodymutator/body_mutator.go
```diff
@@ -0,0 +1,78 @@
...
+
+func NewBodyMutator(bodyMutations *filterapi.HTTPBodyMutation, originalBody []byte) *BodyMutator {
+	return &BodyMutator{
+		originalBody:  originalBody,
+		bodyMutations: bodyMutations,
+	}
+}
+
+// Mutate mutates the request body based on the body mutations and restores original body if mutated previously.
+func (b *BodyMutator) Mutate(requestBody []byte, onRetry bool) ([]byte, error) {
```
mathetake, line 77: would it be possible to use sjson? this function copy the entire body twice (unmarshal & marshal)

mathetake (commented):
can you also add test cases in tests/extproc/testupstream_test.go so we can verify end-to-end

internal/extproc/chatcompletion_processor.go
```diff
@@ -261,6 +263,18 @@ func (c *chatCompletionProcessorUpstreamFilter) ProcessRequestHeaders(ctx contex
 		}
 	}
 
+	// Apply body mutations from the route and also restore original body on retry.
+	if b := c.bodyMutator; b != nil {
+		if bodyMutation != nil && bodyMutation.GetBody() != nil && len(bodyMutation.GetBody()) > 0 {
+			mutatedBody, mutationErr := c.bodyMutator.Mutate(bodyMutation.GetBody(), c.onRetry)
+			if mutationErr != nil {
+				c.logger.Warn("failed to apply body mutation", "error", mutationErr)
```
yuzisun: this should be error

internal/extproc/bodymutator/body_mutator.go
```diff
@@ -0,0 +1,78 @@
...
+	if onRetry && b.originalBody != nil {
+		// On retry, restore the original body first
+		requestBody = b.originalBody
+	}
+
+	// Parse the request body as JSON
+	var bodyData map[string]interface{}
+	if err := json.Unmarshal(requestBody, &bodyData); err != nil {
+		return nil, fmt.Errorf("failed to parse request body as JSON: %w", err)
+	}
```
yuzisun: Let's use sjson, unmarshall is slow.

internal/extproc/messages_processor.go
xiaolin593: not sure I follow. can you elaborate?

mathetake: bodyMutation is nil == translation didn't happen. So you are not applying body mutator API to the cases, for example, openai backends for chat/completions or anthropic for /messages. I don't think we want to do such partial logic for this API. can you apply unconditionally regardless of bodyMutation != nil or not

xiaolin593:
> Another question on API: what if we want to do this mutation conditionally? E.g. i want fieldA in embeddings endpoint but not in chat/completions etc?

Do you have a concrete example for embedding mutation? I'd imagine not much mutation on embedding API as it's very simpler than chat/completion.

mathetake:
i mean for sure we wouldn't want to add service_tier to embeddings requests? But in that case it can be set at route level where embeddings vs chat completions differ in its mutation config yeah so maybe no issue

xiaolin593:
> i mean for sure we wouldn't want to add service_tier to embeddings requests? But in that case it can be set at route level where embeddings vs chat completions differ in its mutation config yeah so maybe no issue

embedding(e.g. text-embedding-3-small)  and chat/completion model(e.g. gpt-4.1) are different architecture-based models (vector based vs transformer/neural network). 

I think the problem you describe only happen *iff* there is a model supports both /v1/embedding and /v1/chat/completions. Otherwise, most of time, I'd expect user to set separate routes for embedding and chat/completion and choose body mutation when needed, so there will be no overlap even hypothetically embedding and chat/completion share same API schema i.e. same field "service_tier".

Back to "mutate conditionally", one valid assumption is user wants to have different service_tier/max_output_token for Response and ChatCompletion APIs for same backend, this one is tricky and I'd defer until the use case rises.

internal/extproc/bodymutator/body_mutator.go
```diff
@@ -0,0 +1,122 @@
...
+	if len(b.bodyMutations.Set) > 0 {
+		for _, field := range b.bodyMutations.Set {
+			if field.Path != "" {
+				// Check value type to determine appropriate sjson method
+				if isJSONValue(field.Value) {
+					// Use SetRawBytes for JSON values (quoted strings, numbers, booleans, objects, arrays)
+					mutatedBody, err = sjson.SetRawBytes(mutatedBody, field.Path, []byte(field.Value))
+				} else {
+					// Use SetBytes for plain string values
+					mutatedBody, err = sjson.SetBytes(mutatedBody, field.Path, field.Value)
```
siddharth1036: Is there a reason we didn't use ReplaceInPlace:true and Optimistic:true for sjson updates?

internal/extproc/chatcompletion_processor.go
```diff
@@ -268,6 +270,32 @@ func (c *chatCompletionProcessorUpstreamFilter) ProcessRequestHeaders(ctx contex
...
+			if mutationErr != nil {
+				c.logger.Error("failed to apply body mutation", "error", mutationErr)
+			} else {
+				bodyMutation.Mutation = &extprocv3.BodyMutation_Body{Body: mutatedBody}
+			}
+		}
+	}
+
+	// Ensure bodyMutation is not nil for subsequent processing
+	if bodyMutation == nil {
```
siddharth1036, line 277: Just wondering when would the bodyMutation be nil?

siddharth1036:
Hi @xiaolin593 , Great Work! Minor comments. One doubt that I have is should we add shouldIgnoreField() for fields such as model being updated via mutations? as such mutations will create inconsistencies in metrics

internal/extproc/chatcompletion_processor.go
xiaolin593: when there is no translation and no body mutation at route.

internal/extproc/bodymutator/body_mutator.go
xiaolin593:
```
// Optimistic is a hint that the value likely exists which
// allows for the sjson to perform a fast-track search and replace.
```
the value may not exist when use `set`, intention was to allow add operation via body mutation.

tiswanso:
Nice work!  I tested locally a bit (commit 21c8ffb) and it's looking good to me and will work for my current scenario.

Test:

1.  install with local helm CRDs & local built version of extproc and controller
   
    ```
     helm upgrade -i aieg-crd ./manifests/charts/ai-gateway-crds-helm \
        --namespace envoy-ai-gateway-system --create-namespace
 
     helm upgrade -i aieg ./manifests/charts/ai-gateway-helm \
        --namespace envoy-ai-gateway-system \
        --set extProc.image.repository=localhost:5001/ai-gateway-extproc --set controller.image.repository=localhost:5001/ai-gateway-controller \
        --set extProc.image.tag=pr1492 --set controller.image.tag=pr1492 \
        --create-namespace
    ```

2.  Setup basic example: 

     ```
     kubectl apply -f https://raw.githubusercontent.com/envoyproxy/ai-gateway/main/examples/basic/basic.yaml
     ```

3.  Edit the `AIServiceBackend` to add a `BodyMutation` (I also did a `HeaderMutation` below):

    ```
    kind: AIServiceBackend
    metadata:
      name: envoy-ai-gateway-basic-testupstream
      namespace: default
    spec:
      backendRef:
        group: gateway.envoyproxy.io
        kind: Backend
        name: envoy-ai-gateway-basic-testupstream
      bodyMutation:
        set:
        - path: user
          value: '{{"appkey": "<redacted>"}}'
      headerMutation:
        set:
        - name: user
          value: '{{"appkey": "<redacted>"}}'
      schema:
        name: OpenAI
    ```

4. curl to the gateway for the mock inference model

    ```
    curl -H "Content-Type: application/json" \
      -d '{
            "model": "some-cool-self-hosted-model",
            "messages": [
                {
                    "role": "system",
                    "content": "Hi."
                }
            ]
        }' \
      $GATEWAY_URL/v1/chat/completions
     ```

5.  See the `BodyMuation` in the logs of the mock model (`envoy-ai-gateway-basic-testupstream-...`)

    ```
    │ [testupstream] header "X-Ai-Eg-Original-Path": [/v1/chat/completions]                                                                       │
    │ [testupstream] header "X-Forwarded-Host": [localhost:8080]                                                                                  │
    │ [testupstream] header "Accept": [*/*]                                                                                                       │
    │ [testupstream] header "X-Forwarded-For": [10.244.0.14]                                                                                      │
    │ [testupstream] header "X-Forwarded-Proto": [http]                                                                                           │
    │ [testupstream] header "X-Envoy-External-Address": [127.0.0.1]                                                                               │
    │ [testupstream] header "X-Ai-Eg-Model": [some-cool-self-hosted-model]                                                                        │
    │ [testupstream] header "User": [{{"appkey": "<redacted>"}}]                                              │
    │ [testupstream] header "Content-Length": [252]                                                                                               │
    │ [testupstream] header "User-Agent": [curl/8.12.1]                                                                                           │
    │ [testupstream] header "Content-Type": [application/json]                                                                                    │
    │ [testupstream] header "X-Request-Id": [623754e9-0014-4f13-b75d-8596e7fe04b3]                                                                │
    │ [testupstream] header "X-Ai-Eg-Internal-Req-Id": [623754e9-0014-4f13-b75d-8596e7fe04b3-3b015948-3cc9-4f44-a9c7-5dc52eeaa704]                │
    │ [testupstream] no expected host: got envoy-ai-gateway-basic-testupstream.default.svc.cluster.local                                          │
    │ [testupstream] no expected headers                                                                                                          │
    │ [testupstream] no non-expected headers in the request                                                                                       │
    │ [testupstream] no expected testupstream-id                                                                                                  │
    │ [testupstream] Request body (252 bytes): {                                                                                                  │
    │         "model": "some-cool-self-hosted-model",                                                                                             │
    │         "messages": [                                                                                                                       │
    │             {                                                                                                                               │
    │                 "role": "system",                                                                                                           │
    │                 "content": "Hi."                                                                                                            │
    │             }                                                                                                                               │
    │         ]                                                                                                                                   │
    │     ,"user":{{"appkey": "<redacted>"}}}                                                                 │
    │ [testupstream] no expected request body                                                                                                     │
    │ [testupstream] no response headers                                                                                                          │
    │ [testupstream] detected model in the request: some-cool-self-hosted-model                                                                   │
    │ [testupstream] response sent: {"choices":[{"message":{"role":"assistant", "content":"I feel the need - the need for speed."}}]}
    ```

internal/extproc/messages_processor.go
```diff
@@ -212,6 +214,32 @@ func (c *messagesProcessorUpstreamFilter) ProcessRequestHeaders(ctx context.Cont
...
+				}
+			}
+		} else if bodyMutation.GetBody() != nil && len(bodyMutation.GetBody()) > 0 {
+			mutatedBody, mutationErr := c.bodyMutator.Mutate(bodyMutation.GetBody(), c.onRetry)
+			if mutationErr != nil {
+				c.logger.Error("failed to apply body mutation", "error", mutationErr)
+			} else {
+				bodyMutation.Mutation = &extprocv3.BodyMutation_Body{Body: mutatedBody}
+			}
+		}
```
mathetake: can you share this code in internal/extproc/util.go + add the same change to all endpoints?

internal/extproc/bodymutator/body_mutator.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+	if strings.HasPrefix(value, "[") && strings.HasSuffix(value, "]") {
+		return true
+	}
+
+	// Default to plain string
+	return false
+}
+
+// Mutate mutates the request body based on the body mutations and restores original body if mutated previously.
```
mathetake, line 76: so sjson has an "option" to indicate {Optimistic: bool, ReplaceInPlace: bool} that can be passed to sjson.SetOptions function (Delete are alias to Set with nil bytes). Ideally, we should use ReplaceInPlace=false for the first mutation and then after the first mutation, we should use ReplaceInPlace=true for all operations to avoid copy as much as possible.

Not sure why sjson has no API for passing options to Delete variant API but Delete APIs are always using default options, which is ReplaceInPlace=false (if i read their impl correctly). So, maybe switching to Set for all Delete API usage to be able to explicitly pass the option

yuzisun: @mathetake Can we optimize in a separate PR ?

internal/extproc/messages_processor.go
xiaolin593: @mathetake applied same change to other endpoints. can you take another look?

internal/extproc/bodymutator/body_mutator.go
mathetake: i mean it should be a few lines of change and can be done very easily, but if you insist i am fine

xiaolin593: I can add it in this PR.

xiaolin593: Passed `ReplaceInPlace` per onRetry value to SetRawByteOptions and SetBytesOptions. 

For Delete, I think using `DeleteBytes` API is semantically intuitive and readable despite sjson internally forwards to `SetBytes` operation.

internal/extproc/bodymutator/body_mutator.go
```diff
@@ -0,0 +1,123 @@
...
+		}
+	}
+
+	// Apply sets
+	replaceInPlace := onRetry
+	if len(b.bodyMutations.Set) > 0 {
+		for _, field := range b.bodyMutations.Set {
+			if field.Path != "" {
+				// Check value type to determine appropriate sjson method
+				if isJSONValue(field.Value) {
```
mathetake, line 110: so this can be called in non request path but instead configuration loading phrase right? can you at least put a TODO comment on that?

internal/extproc/bodymutator/body_mutator.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+
+	mutatedBody := requestBody
+	var err error
+
+	// Apply removals first
+	if len(b.bodyMutations.Remove) > 0 {
+		for _, fieldName := range b.bodyMutations.Remove {
+			if fieldName != "" {
+				mutatedBody, err = sjson.DeleteBytes(mutatedBody, fieldName)
```
mathetake, line 95: so DeleteBytes by default copies the the underlying buffer, so let's put a TODO comment here too saying we should optimize by using SetBytesOptions with nil value + replaceInPlace tracking like below

PR #1479, opened 2025-11-03
mcp: proxy stdio servers in standalone mode

nacx:
**Description**

Opening as a draft to get early feedback.

When running in standalone mode, it allows running and proxying `stdio` MCP servers. This PR only supports defining the servers in the `servers.json` format, as for now, we don't want to introduce new nouns for a feature that does not apply to Kubernetes.

With this change, the CLI can be started with a file like the following:

```
aigw run --mcp-json '{
  "mcpServers": {
    "everything": {
      "command": "npx",
      "args": ["-y","@modelcontextprotocol/server-everything"]
    },
    "filesystem": {
      "command": "npx",
      "args": ["-y","@modelcontextprotocol/server-filesystem"]
    }
  }
}'
```

**Related Issues/PRs (if applicable)**

Depends on https://github.com/envoyproxy/gateway/pull/7427 for a clean implementation.

**Special notes for reviewers (if applicable)**

N/A

Issue #950, opened 2025-07-26
Backend with static ip is not working


Xunzhuo:
*Description*:

currently we patch upstream level aigw extproc when cluster LoadAssignment is not nil.

```
		if cluster.LoadAssignment == nil {
			s.log.Info("LoadAssignment is nil", "cluster_name", cluster.Name)
			return
		}
		if len(cluster.LoadAssignment.Endpoints) != len(httpRouteRule.BackendRefs) {
			s.log.Info("LoadAssignment endpoints length does not match backend refs length",
				"cluster_name", cluster.Name, "endpoints_length", len(cluster.LoadAssignment.Endpoints), "backend_refs_length", len(httpRouteRule.BackendRefs))
			return
		}
```

This is not working with when we configure static IP for inference remove service:

```
apiVersion: aigateway.envoyproxy.io/v1alpha1
kind: AIServiceBackend
metadata:
  name: envoy-ai-gateway-tke-qwen3-coder
  namespace: default
spec:
  schema:
    name: OpenAI
  backendRef:
    name: envoy-ai-gateway-tke-qwen3-coder
    kind: Backend
    group: gateway.envoyproxy.io
---
apiVersion: gateway.envoyproxy.io/v1alpha1
kind: Backend
metadata:
  name: envoy-ai-gateway-tke-qwen3-coder
spec:
  endpoints:
  - ip:
      address: x.x.x.x
      port: xxxx
  type: Endpoints
```

Envoy Gateway built EDS type cluster for this, so LoadAssignment is nil, and the upstream level extproc is never getting work.

mathetake:
https://github.com/envoyproxy/ai-gateway/issues/902 yeah it's the same underlying issue as this issue

PR #1496, opened 2025-11-05
feat: allow using ip addresses in auto-config backends

nacx:
**Description**

The auto-config template configures backends automatically, always using `hostname`. If the input config uses IP addresses, the generated configuration would be invalid. This fixes the config so that IP addresses are properly rendered.

The need for this will be more evident when https://github.com/envoyproxy/ai-gateway/pull/1479 and https://github.com/envoyproxy/gateway/pull/7427 are merged, to allow cleanly proxying MCP servers running locally.

N/A

### DIFF
--- a/manifests/charts/ai-gateway-helm/templates/admission_webhook.yaml
+++ b/manifests/charts/ai-gateway-helm/templates/admission_webhook.yaml
@@ -30,7 +30,7 @@ webhooks:
     sideEffects: None
     admissionReviewVersions: ["v1"]
     timeoutSeconds: 10
-    failurePolicy: Ignore
+    failurePolicy: Fail
 ---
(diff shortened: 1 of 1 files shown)


mathetake:
so i am skeptical if the local ips are working properly for AIGatewayRoutes due to the reason described in #950 -- basically extension servers are not able to set LocalityLbEndpoints metadata. If it's working, then EG made some magical progress in its implementation detail. Did you check the translation, auth, etcs are workign?

nacx:
Had a chat with @mathetake capturing the background on the limitation. This works for MCP because it's not going via extproc, but won't work for LLM traccos, so I'm closing it for now.
Thanks for the context @mathetake!

PR #1501, opened 2025-11-06
chore: remove nip.io now that EG supports localhost backends in cli

nacx:
**Description**

Now that EG has merged the patch that allows using `localhost` in Backends when running in standalone mode, we can remove the uses of `.nip.io` we were using to overcome the limitation when proxying local Ollama LLMs or local MCP servers.

I've verified all Docker guides we had as well as all ollama configs and everything works as expected (cc @codefromthecrypt)

https://github.com/envoyproxy/gateway/pull/7427

This requires upgrading to the latest EG from main (the change hasn't yet been cherry-picked to the 1.6 release branch). Opening as a draft to avoid bumping the EG dependency beyond 1.6 until we release.

### DIFF
--- a/cmd/aigw/README.md
+++ b/cmd/aigw/README.md
@@ -2,7 +2,7 @@
 
 ## Quick Start
 
-[docker-compose.yml](docker-compose.yaml) builds and runs `aigw`, targeting
+[docker-compose.yaml](docker-compose.yaml) builds and runs `aigw`, targeting
 [Ollama][ollama] for OpenAI chat completion requests on port 1975.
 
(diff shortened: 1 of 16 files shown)


codefromthecrypt (approved):
great and this doesn't just solve the auto-config (ENV driven) mode, but also config in general. The first time I tried aigw, I lost many hours trying not knowing there were tricks needed just to use localhost. That time was before we had auto-config so it was just a really obscure pitfall in the normal yaml.

nacx:
The change in EG is released and this is ready for review and merge

codefromthecrypt:
thanks I am really happy you did this

PR #1507, opened 2025-11-07
refactor: decouples translators from extproc

mathetake:
**Description**
The translators shouldn't have to be tied with ext_proc APIs since it's none of their concerns. This commit decouples the translator package from them so that we can easily use them in dynamic modules too.

Preparation for #90


mathetake:
won't merge until v0.4

mathetake:
Sorry bad merge

internal/extproc/translator/openai_gcpvertexai_test.go
```diff
@@ -1504,7 +1504,7 @@ func TestOpenAIToGCPVertexAITranslatorV1ChatCompletion_StreamingParallelToolInde
                 "role": "model"
             }
         }
-    ],
+    ]
```
hustxiayang: really sorry, the `gcpToolCallsChunk` should have "\n\n" between the chunks, otherwise the test does not make sense.  I put it in https://github.com/envoyproxy/ai-gateway/pull/1537. Sorry for this mistake!

mathetake: don't be sorry! let's make sure all tests are working properly carefully from next time :)

PR #1513, opened 2025-11-07
fix(anthropic): marshalling errors for anthropic system messages 

secustor:
**Description**

Currently if a client sends system messages to the Anthropic API endpoint of the Gateway JSON parsing fails. 

That field is an union type of `string | []object`. https://docs.claude.com/en/api/messages#body-system

When it is a string: 
```
[2025-11-07 08:56:49.901][24][warning][ext_proc] [source/extensions/filters/http/ext_proc/ext_proc.cc:1689] [Tags: "ConnectionId":"12015","StreamId":"14675843249488310277"] Received gRPC error on stream: 2, message error processing request message: cannot process request body: /v1/messages endpoint requires Anthropic format: failed to unmarshal Anthropic Messages body: json: cannot unmarshal string into Go struct field MessagesRequest.system of type []*anthropic.SystemPrompt{via_upstream}
```
or in the array format too: 

```
[2025-11-06 17:54:18.763][24][warning][ext_proc] [source/extensions/filters/http/ext_proc/ext_proc.cc:1689] [Tags: "ConnectionId":"6492","StreamId":"8941995427420900364"] Received gRPC error on stream: 2, message error processing request message: cannot process request body: /v1/messages endpoint requires Anthropic format: failed to unmarshal Anthropic Messages body: json: cannot unmarshal array into Go struct field MessagesRequest.system of type anthropic.SystemPrompt{via_upstream}
```

After this system prompts are correctly handled: 

```
  curl -v -H "Content-Type: application/json" \
    -d '{
      "model": "eu.anthropic.claude-haiku-4-5-20251001-v1:0",
      "messages": [
        {
          "role": "user",
          "content": "Hello, how are you?"
         }
     ],
   "system": "You are Albert Einstein",
   "max_tokens": 100
 }' \
 $GATEWAY_URL/anthropic/v1/messages?beta=true
```

```
{
  "model": "claude-haiku-4-5-20251001",
  "id": "msg_bdrk_01Wm5C56XatnufSKseRs3qeH",
  "type": "message",
  "role": "assistant",
  "content": [
    {
      "type": "text",
      "text": "# Guten Tag!\n\nI'm doing quite well, thank you for asking. I find myself in good spirits today. There's something about a genuine greeting that reminds one of the simple pleasures in life—a moment of human connection, unpretentious and warm.\n\nI've been pondering some problems this morning, as one does. The mind, I've found, works best when allowed to wander freely between ideas, making unexpected connections. It's rather like how"
    }
  ],
  "stop_reason": "max_tokens",
  "stop_sequence": null,
  "usage": {
    "input_tokens": 17,
    "cache_creation_input_tokens": 0,
    "cache_read_input_tokens": 0,
    "output_tokens": 100
  }
}
```

### DIFF
--- a/cmd/aigw/translate.go
+++ b/cmd/aigw/translate.go
@@ -30,7 +30,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	gwapiv1 "sigs.k8s.io/gateway-api/apis/v1"
-	gwapiv1a3 "sigs.k8s.io/gateway-api/apis/v1alpha3"
 	kyaml "sigs.k8s.io/yaml"
 
 	aigv1a1 "github.com/envoyproxy/ai-gateway/api/v1alpha1"
@@ -110,7 +109,7 @@ func collectObjects(yamlInput string, out io.Writer, logger *slog.Logger) (
 	mcpRoutes []*aigv1a1.MCPRoute,
 	aigwBackends []*aigv1a1.AIServiceBackend,
 	backendSecurityPolicies []*aigv1a1.BackendSecurityPolicy,
-	backendTLSConfigs []*gwapiv1a3.BackendTLSPolicy,
+	backendTLSConfigs []*gwapiv1.BackendTLSPolicy,
 	gws []*gwapiv1.Gateway,
 	secrets []*corev1.Secret,
 	envoyProxies []*egv1a1.EnvoyProxy,
@@ -184,7 +183,7 @@ func translateCustomResourceObjects(
 	mcpRoutes []*aigv1a1.MCPRoute,
 	aigwBackends []*aigv1a1.AIServiceBackend,
 	backendSecurityPolicies []*aigv1a1.BackendSecurityPolicy,
-	backendTLSPolicies []*gwapiv1a3.BackendTLSPolicy,
+	backendTLSPolicies []*gwapiv1.BackendTLSPolicy,
 	gws []*gwapiv1.Gateway,
 	usedDefinedSecrets []*corev1.Secret,
 	logger *slog.Logger,
--- a/internal/controller/mcp_route.go
+++ b/internal/controller/mcp_route.go
@@ -432,7 +432,11 @@ func (c *MCPRouteController) newPerBackendRefHTTPRoute(ctx context.Context, dst 
 // syncGateways synchronizes the gateways referenced by the MCPRoute by sending events to the gateway controller.
 func (c *MCPRouteController) syncGateways(ctx context.Context, mcpRoute *aigv1a1.MCPRoute) error {
 	for _, p := range mcpRoute.Spec.ParentRefs {
-		c.syncGateway(ctx, mcpRoute.Namespace, string(p.Name))
+		gwNamespace := mcpRoute.Namespace
+		if p.Namespace != nil {
+			gwNamespace = string(*p.Namespace)
+		}
+		c.syncGateway(ctx, gwNamespace, string(p.Name))
 	}
 	return nil
 }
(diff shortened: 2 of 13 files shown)


internal/apischema/anthropic/anthropic.go
```diff
@@ -168,7 +168,7 @@ type Thinking struct{} // TODO when we need it for observability, etc.
 
 // SystemPrompt represents a system prompt to guide the model's behavior.
 // https://docs.claude.com/en/api/messages#body-system
-type SystemPrompt struct{} // TODO when we need it for observability, etc.
+type SystemPrompt interface{} // TODO when we need it for observability, etc.
```
yuzisun, line 171: Would this be a problem for other fields as well ?

yuzisun:
@secustor thanks for the fix, did you test the tool calls which my have similar issues.

nacx:
I think this will affect all fields, and that we should change all occurences of empty `struct{}` with `interface{}`.

An example to illustrte the issue:

```
type Foo struct{}
type Bar interface{}

func TestFoo(t *testing.T) {
	raw := []byte(`{"key":"value"}`)

	var foo Foo
	var bar Bar
	require.NoError(t, json.Unmarshal(raw, &foo))
	require.NoError(t, json.Unmarshal(raw, &bar))

	t.Logf("foo: %+v\nbar: %+v", foo, bar)
}
```
```
=== RUN   TestFoo
        foo: {}
        bar: map[key:value]
--- PASS: TestFoo (0.00s)
PASS
```

So I think we should:

* Change all `struct{}` to `interface{}` in the Anthropic model.
* Given how close we are to the release and how impactful this change is (breaking existing Anthropic support), it is probably safer to just revert the commits that refactor the Anthropic model. We already made the decision of releasing frequently after 0.4, so it would be easy to release the Anthropic model refactoring after 0.4.

Thoughts?

I'm leaning towards reverting those PRs now for the release, and keep working on the Anthropic model and release it shortly after v0.4, when we have the typed model plus metrics & tracing.

yuzisun:
> I think this will affect all fields, and that we should change all occurences of empty `struct{}` with `interface{}`.
> 
> An example to illustrte the issue:
> 
> ```
> type Foo struct{}
> type Bar interface{}
> 
> func TestFoo(t *testing.T) {
> 	raw := []byte(`{"key":"value"}`)
> 
> 	var foo Foo
> 	var bar Bar
> 	require.NoError(t, json.Unmarshal(raw, &foo))
> 	require.NoError(t, json.Unmarshal(raw, &bar))
> 
> 	t.Logf("foo: %+v\nbar: %+v", foo, bar)
> }
> ```
> 
>  ```
> === RUN   TestFoo
>         foo: {}
>         bar: map[key:value]
> --- PASS: TestFoo (0.00s)
> PASS
> ```
> 
> So I think we should:
> 
> * Change all `struct{}` to `interface{}` in the Anthropic model.
> * Given how close we are to the release and how impactful this change is (breaking existing Anthropic support), it is probably safer to just revert the commits that refactor the Anthropic model. We already made the decision of releasing frequently after 0.4, so it would be easy to release the Anthropic model refactoring after 0.4.
> 
> Thoughts?

agreed, let’s revert the previous PR https://github.com/envoyproxy/ai-gateway/pull/1513 considering the scope of impact of this and broke the major use case like Claude Code. We need to test it thoroughly.

internal/apischema/anthropic/anthropic.go
secustor: After your comment I have checked it and yeah other are effected too.

secustor:
No, we tested our setup with ClaudeCode when we ran into this. After that everything worked fine again. 

> @secustor thanks for the fix, did you test the tool calls which my have similar issues.

No, only a simple hello world.  As commented above other fields have also fall backs to other types.

yuzisun:
> No, we tested our setup with ClaudeCode when we ran into this. After that everything worked fine again.
> 
> > @secustor thanks for the fix, did you test the tool calls which my have similar issues.
> 
> No, only a simple hello world. As commented above other fields have also fall backs to other types.

Right tool calls and thinking will stop working as these fields are not passed along.

mathetake:
these structs are read only so as long as this unmarshal is resolved, there should be zero impact though, I'm ok reverting it since it's near v0.4 release

mathetake:
https://github.com/envoyproxy/ai-gateway/pull/1516 reverting

Issue #1457, opened 2025-10-28
Revisit endpoint prefixes


mathetake:
_Background_:

Currently, [we have a static prefixes for AI endpoints depending on whether or not the original provider is OpenAI](https://github.com/envoyproxy/ai-gateway/blob/7ba9f32ef27783d25dea99bf363745b13c210caf/cmd/extproc/mainlib/main.go#L249-L254). For exmaples,

* OpenAI format's /chat/completions, /embeddings, etc, are simply exposed at ${GATEWAY_ADDRESS}/v1/that/endpoint
* Anthropic's /messages endpoint is exposed at ${GATEWAY_ADDRESS}/anthropic/v1/messages where we have "anthropic" endpoint.
* Corehe's /rerank API _will be_ exposed at ${GATEWAY_ADDRESS}/cohere/v2/rerank https://github.com/envoyproxy/ai-gateway/pull/1433

Historically, we were only planning to support for OpenAI, so there was no problem to this. However, it turned out that it is inevitable to add support for more endpoints outside the OpenAI.

The original rational on having no prefix for openai vs /anthropic prefix was that both providers has two identical endpoints /models. Without using additional attributes, we cannot _easily_ distinguish the requests to /models endpoints, and aigw cannot determin which format it should return the response in to these requests. *Luckily*, so far, /models is the only example in the wild where providers have conflicts.

_What other projects do_:

* vLLM: They have flat structures  https://docs.vllm.ai/en/latest/serving/openai_compatible_server.html#supported-apis 
    * not only openai endpoints, but they also have rerank API from cohere https://docs.vllm.ai/en/latest/serving/openai_compatible_server.html#re-rank-api
* litellm: Inconsistency across the project. For example,
    * /rerank API in cohere is directly at the top level (even without /v2) https://docs.litellm.ai/docs/rerank
    * /ocr API in mistral is exposed with /v1/ocr. https://docs.litellm.ai/docs/ocr
    * AWS Bedrock's /converse endpoint is exposed with /aws prefix: https://docs.litellm.ai/docs/bedrock_converse

_What we want to achieve_:

* Consistency in the convention and not special casing OpenAI
* Less friction in the configuration on the client/downstream applications 
    * Do not want to impose "AI Gateway specific" caveats like "putting this prefix when you want to access aigw". For example, /anthtropic prefix is one example 
* Have a future-proof configuration 
[...]

PR #1514, opened 2025-11-07
feat: make prefix of endpoints configurable

AyushSawant18588:
**Description**
Introduces configurable provider endpoint prefixes via global configurations.

Compatibility
Non-breaking: defaults preserve current behavior for OpenAI (/), Cohere (/cohere), Anthropic (/anthropic) if not set.

https://github.com/envoyproxy/ai-gateway/issues/1457


siddharth1036:
@AyushSawant18588 can you check if e2e tests are required with custom prefixes?

manifests/charts/ai-gateway-helm/values.yaml
```diff
@@ -14,6 +14,10 @@ endpointConfig:
   #
   # This can be used for providing a separation between AIGatewayRoutes and normal HTTPRoutes when the top level "/v1/" is not desired.
   rootPrefix: "/"
+  # Comma-separated key-value pairs for endpoint prefixes.
+  # Format: "openaiPrefix:/prefix1,anthropicPrefix:/prefix2"
+  # Example: "openaiPrefix:/,coherePrefix:/cohere,anthropicPrefix:/anthropic"
+  endpointPrefixes: ""
```
mathetake: i would rather have an explicit exhaustive list here to avoid manual validation at runtime
```suggestion
  openai: ""
  anthropic: "/anthropic"
  cohere: "/cohere"
```

mathetake: also could you add detailed comments here by showing exact paths examples. At least one for each provider

AyushSawant18588: The idea behind using key-value pairs for endpoint prefixes was to make the design more scalable. As we add support for more providers in the future, maintaining an explicit list would require updating multiple files (helm values, controller deployment, controller, extproc, gateway_mutator, internalapi, etc.) each time. With the current approach, any new provider addition would only require changes within the EndpointPrefixes struct and related functions in the internalapi package.
What are your thoughts on this trade-off?

mathetake: I would prefer my suggestion for now, not trying to solve non existing problem

mathetake: btw you can concatenate strings in helm functions so your concern on multiple changes doesn't apply here i believe

mathetake: my suggestion is "let's not expose the error prone string format to users". I am not saying we shouldn't use the concatenated ones in arguments in extproc/controller, it should be totally fine

AyushSawant18588: Oh thanks, got your point so expose explicit prefixes via values but concatenate and  pass it as a string in endpointPrefixes arg to controller. So this structure is fine for string after concatenation right?
```
--endpointPrefixes='endpointConfig.endpointPrefixes=openaiPrefix:/,coherePrefix:/cohere,anthropicPrefix:/anthropic'
```

mathetake: yes

mathetake:
For e2e you can reuse the existing e2e-namespaced for non default config https://github.com/envoyproxy/ai-gateway/tree/main/tests/e2e-namespaced @AyushSawant18588

AyushSawant18588:
> For e2e you can reuse the existing e2e-namespaced for non default config https://github.com/envoyproxy/ai-gateway/tree/main/tests/e2e-namespaced @AyushSawant18588

can you tell what e2e test are you suggesting to add?

mathetake:
i would say pass your new helm value with non-default prefix for openai in the e2e setup, and then access the gateway

tests/e2e-namespaced/namespaced_test.go
```diff
@@ -67,3 +73,125 @@ func TestNamespaced(t *testing.T) {
...
+	thisDir := filepath.Dir(thisFile)
+	helmChartPath := filepath.Join(thisDir, "..", "..", "manifests", "charts", "ai-gateway-helm")
+
+	// Upgrade AI Gateway with the OpenAI endpoint prefix in namespace "test".
+	t.Log("Upgrading AI Gateway with endpointConfig.openai=/openai in namespace 'test'")
+	helm := testsinternal.GoToolCmdContext(ctx, "helm", "upgrade", "ai-eg", "--force",
+		helmChartPath,
+		"--set", "endpointConfig.openai=/openai",
+		"--set", "controller.watch.namespaces={gateway-ns,route2-ns}",
+		"-n", "envoy-ai-gateway-e2e", "--create-namespace")
```
mathetake: so instead of reconfiguring the EG here, can you pass the value in TestMain function (like namespace config)

```
func TestMain(m *testing.M) {
	e2elib.TestMain(m, e2elib.AIGatewayHelmOption{
		Namespace: "envoy-ai-gateway-e2e", // Also install AI Gateway on a different namespace
		AdditionalArgs: []string{
			// Configure the controller to only watch certain namespaces
			// By skipping the "route1-ns" the models defined in that namespace routes
			// won't be returned in the ListModels response.
			"--set", "controller.watch.namespaces={gateway-ns,route2-ns}",
		},
	}, false, true,
	)
}
```

Otherwise, the tests in this package result becomes the order-dependent.

AyushSawant18588: As the tests are under one namespaced folder there can be only one main function right and the first test test does not need this prefix setting, so need to reconfigure ai gateway for this test, as these test are not running in parallel will this be fine? I have followed the same logic for e2e tests as here https://github.com/envoyproxy/ai-gateway/blob/main/tests/e2e/otel_tracing_test.go. Let me know if this is fine or I should configure prefix for all tests in namespaced folder?

mathetake: Otel stuff feels a tech debt (very slow and flaky) so I would rather have it configured and make all tests order independent as I suggested.

AyushSawant18588: Okay so will add prefix in Main function itself and test it in TestNamespaced then no need for a new test function, sounds good? Just need to add the openai prefix here: https://github.com/envoyproxy/ai-gateway/blob/bbb6ccaad899b957b174db21d764cd034bc776c2/tests/e2e-namespaced/namespaced_test.go#L61

mathetake: Yea sounds good

cmd/controller/main.go
```diff
@@ -154,6 +155,11 @@ func parseAndValidateFlags(args []string) (flags, error) {
 		"",
 		"Comma-separated key-value pairs for mapping HTTP request headers to otel span attributes. Format: x-session-id:session.id,x-user-id:user.id.",
 	)
+	endpointPrefixes := fs.String(
+		"endpointPrefixes",
+		"",
+		"Comma-separated key-value pairs for endpoint prefixes. Format: openaiPrefix:/,coherePrefix:/cohere,anthropicPrefix:/anthropic.",
```
mathetake: "Prefix" suffix feels redundant as the context is already prefix 
```suggestion
		"Comma-separated key-value pairs for endpoint prefixes. Format: openai:/,cohere:/cohere,anthropic:/anthropic.",
```

AyushSawant18588: cool, I have made the change

PR #1515, opened 2025-11-07
fix: set the remaining extra fields for inputsShema

hustxiayang:
**Description**

In the original codes, it drops every thing other than `properties`. Thus, other data like `defs` are dropped. it should be stored into extra filelds.


internal/extproc/translator/openai_gcpanthropic.go
```diff
@@ -194,6 +184,28 @@ func translateOpenAItoAnthropicTools(openAITools []openai.Tool, openAIToolChoice
 					inputSchema.Required = requiredSlice
 				}
 
+				// Keys to skip
+				keysToSkip := map[string]bool{
```
mathetake: * Define this a global map (don't forget to add proper prefix in its variable name) to avoid allocations here
* People don't use `map[string]bool` for set in Go, but use `map[string]struct{}` instead.

hustxiayang: updated, thanks a lot for your comments

mathetake:
looks good to me! deferring to @yuzisun

hustxiayang:
@mathetake thanks for your free golang tutorial!

nacx:
@yuzisun good to merge?

mathetake:
ping @yuzisun

hustxiayang:
@yuzisun Please help to review and approve it, thanks!

PR #1519, opened 2025-11-09
chore: align otel embeddings with final openinference spec

codefromthecrypt:
**Description**

This aligns the OpenInference embeddings data capture with the final specification I worked upstream to complete.

https://github.com/Arize-ai/openinference/tree/main/spec

**Related Issues/PRs (if applicable)**

https://github.com/Arize-ai/openinference/pull/2238

**Special notes for reviewers (if applicable)**

Thanks very much @axiomofjoy from Arize who spent a lot of think time on this on the other side, helping to balance the desire to have things documented (for external implementations like this) with the practical realities of Eval UX I'm less familiar with.

PR #1520, opened 2025-11-09
fix: updates image generation request tracing to match latest OpenInference

codefromthecrypt:
**Description**

updates image generation request tracing to match latest OpenInference
- Span name changed to ImagesResponse
- Add output.mime_type and output.value attributes
- Remove llm.model_name from request attributes (now in invocation_parameters)
- Fix missing RecordResponse call in translator

https://github.com/envoyproxy/ai-gateway/pull/1519


codefromthecrypt:
will work on this more when online again (long flight ahead)

codefromthecrypt:
note I also polished test style, since codecov is angry anyway

nacx (approved):

PR #1521, opened 2025-11-09
feat: parse out thought summary from gemini models' response when include thought is true

hustxiayang:
**Description**

When `includeThought` is true, Gemini would also generate summary of thinking process. We need to parse out this kind of data to users. Otherwise, we would return thought process together with output to users.

Depends/base or replace https://github.com/envoyproxy/ai-gateway/pull/1461

### DIFF
--- a/internal/translator/gemini_helper.go
+++ b/internal/translator/gemini_helper.go
@@ -645,7 +645,7 @@ func geminiCandidatesToOpenAIChoices(candidates []*genai.Candidate, responseMode
 			thoughtSummary, content := extractTextAndThoughtSummaryFromGeminiParts(candidate.Content.Parts, responseMode)
 			if thoughtSummary != "" {
 				message.ReasoningContent = &openai.ReasoningContentUnion{
-					Value: &openai.AWSBedrockReasoningContent{
+					Value: &openai.ReasoningContent{
 						ReasoningContent: &awsbedrock.ReasoningContentBlock{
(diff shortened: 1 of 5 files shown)


internal/translator/gemini_helper.go
```diff
@@ -537,9 +538,22 @@ func geminiCandidatesToOpenAIChoices(candidates []*genai.Candidate, responseMode
...
 				Role: openai.ChatMessageRoleAssistant,
 			}
-			// Extract text from parts.
-			content := extractTextFromGeminiParts(candidate.Content.Parts, responseMode)
-			message.Content = &content
+			// Extract thought summary and text from parts.
+			thoughtSummary, content := extractTextAndThoughtSummaryFromGeminiParts(candidate.Content.Parts, responseMode)
+			if thoughtSummary != "" {
+				message.ReasoningContent = &openai.ReasoningContentUnion{
+					Value: &openai.AWSBedrockReasoningContent{
```
yuzisun: since reasoning content is also used by gemini now, we should remove the AWSBedrock naming prefix.

yuzisun:
@hustxiayang can you help resolve the conflicts here

internal/translator/gemini_helper.go
hustxiayang: @yuzisun I removed the `AWSBedrock` prefix in stream and non-stream

PR #1524, opened 2025-11-10
fix: fix tool call index of stream chunks for gemini models when the tool calls are not all in the same chunk

hustxiayang:
**Description**

In https://github.com/envoyproxy/ai-gateway/pull/1468,  tool call index in stream mode for gemini models is based on the assumption that all tool calls are in the same chunk. This is not true. Thus, we still need a global index to track.


internal/extproc/translator/gemini_helper.go
```diff
@@ -682,9 +682,7 @@ func extractToolCallsFromGeminiParts(toolCalls []openai.ChatCompletionMessageToo
 // extractToolCallsFromGeminiPartsStream extracts tool calls from Gemini parts for streaming responses.
 // Each tool call is assigned an incremental index starting from 0, matching OpenAI's streaming protocol.
 // Returns ChatCompletionChunkChoiceDeltaToolCall types suitable for streaming responses, or nil if no tool calls are found.
-func extractToolCallsFromGeminiPartsStream(toolCalls []openai.ChatCompletionChunkChoiceDeltaToolCall, parts []*genai.Part) ([]openai.ChatCompletionChunkChoiceDeltaToolCall, error) {
-	toolCallIndex := int64(0)
-
+func (o *openAIToGCPVertexAITranslatorV1ChatCompletion) extractToolCallsFromGeminiPartsStream(toolCalls []openai.ChatCompletionChunkChoiceDeltaToolCall, parts []*genai.Part) ([]openai.ChatCompletionChunkChoiceDeltaToolCall, error) {
```
nacx: This is no longer a generic helper method. Worth moving to `openai_gcpvertexai.go`? (same for the other method modified in this PR).

internal/extproc/translator/gemini_helper_test.go
```diff
@@ -1901,7 +1901,10 @@ func TestExtractToolCallsFromGeminiPartsStream(t *testing.T) {
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			calls, err := extractToolCallsFromGeminiPartsStream(toolCalls, tt.input)
+			o := openAIToGCPVertexAITranslatorV1ChatCompletion{
+				toolCallIndex: int64(-1),
```
nacx: Initializing this with a concrete value just because the implementation uses an increment looks error-prone. Can we change this and use the onctructor instead? (This would hide these implementation details). We often read tests to understand how certain types work or should be used, and not exposing this would make it easier to maintain.

internal/extproc/translator/openai_gcpvertexai.go
```diff
@@ -41,7 +41,7 @@ type gcpVertexAIErrorDetails struct {
 // NewChatCompletionOpenAIToGCPVertexAITranslator implements [Factory] for OpenAI to GCP Gemini translation.
 // This translator converts OpenAI ChatCompletion API requests to GCP Gemini API format.
 func NewChatCompletionOpenAIToGCPVertexAITranslator(modelNameOverride internalapi.ModelNameOverride) OpenAIChatCompletionTranslator {
-	return &openAIToGCPVertexAITranslatorV1ChatCompletion{modelNameOverride: modelNameOverride}
+	return &openAIToGCPVertexAITranslatorV1ChatCompletion{modelNameOverride: modelNameOverride, toolCallIndex: int64(-1)}
```
nacx: Having to set the `-1` in the constructor just because of how implementation increments its value looks a bit error-prone. If possible, I would try to make the zero values "just work fine".

hustxiayang: @nacx Hi, Thanks a lot for the comment! This is to follow openai's guidance. They removed the guides with chat completion though, but vllm's test  use use it: https://github.com/vllm-project/vllm/blob/4673e465ff1140592431b82272b52ee08a7faf57/tests/tool_use/test_parallel_tool_calls.py#L87C1-L87C28. I think it's quite safe because if there is no tool use block, the index is not used or updated?

nacx: I agree it should be safe as this is a private type and it is created via the constructor, and that it works.

My point is more about general readability and maintainability best practices. If I, as a maintainer, had to instantiate that type directly in the future for whatever reason, it is very easy to miss initializing that to `-1` (and I would have to remember that there is some method that relies on that...). If instead, we make our types "just work" with the zero values, which is a Go best practice, we're making the code more robust and maintainable.

hustxiayang: I see. I thought this could be captured by tests. But sure, to make it look maintainable, I would follow your comments! Thanks!

nacx: Thanks!  🙇

internal/extproc/translator/gemini_helper.go
hustxiayang: I moved the two functions to openai_gcpvertexai.go and the corresponding test files are also moved. Thanks!

internal/extproc/translator/gemini_helper_test.go
hustxiayang: Changed to use constructor instead.

internal/extproc/translator/openai_gcpvertexai.go
hustxiayang: initialize the toolcall index as 0 instead and update after every tool call.

hustxiayang:
@nacx I think I've resolved all the comments. Do you mind to review again? Thanks!

hustxiayang:
This is still on assumption that a single chunk can contain at least one tool call. Currently verified it for our models, this might be because that gemini would validate the arguments to be valid json, but it may change later.

internal/extproc/translator/openai_gcpvertexai_test.go
```diff
@@ -1255,11 +1255,511 @@ func TestOpenAIToGCPVertexAITranslatorV1ChatCompletion_StreamingResponseBody(t *
...
+
+	require.Nil(t, headerMut)
+	require.NoError(t, err)
+	require.NotNil(t, bodyMut)
+	require.NotNil(t, bodyMut.Mutation)
+
+	body := bodyMut.Mutation.(*extprocv3.BodyMutation_Body).Body
+	chatCompletionChunks := getChatCompletionResponseChunk(body)
+
+	for idx, chunk := range chatCompletionChunks {
```
mathetake, line 1754: this test is broken as the inline json above is invalid hence this assertions are not running. i am fixing this in #1507

hustxiayang: ah, sorry for that! Seems quite wrong.

PR #1533, opened 2025-11-11
fix: validate responses from aws, gcp anthropic and vertex ai

hustxiayang:
**Description**
Currently, the responses of aws, gcp anthropic and gcp vertex ai missed several fields, like `Model`, `ID`, `Created`. These fields are not optional in openai api. Thus, users complained it can not pass pydantic velidation

Also refactor a bit to make serialization for openai chunks common for all chat completion requests

### DIFF
--- a/internal/extproc/messages_processor.go
+++ b/internal/extproc/messages_processor.go
@@ -330,6 +330,7 @@ func (c *messagesProcessorUpstreamFilter) ProcessResponseBody(ctx context.Contex
 	c.costs.InputTokens += tokenUsage.InputTokens
 	c.costs.OutputTokens += tokenUsage.OutputTokens
 	c.costs.TotalTokens += tokenUsage.TotalTokens
+	c.costs.CachedInputTokens += tokenUsage.CachedInputTokens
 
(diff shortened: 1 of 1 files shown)


hustxiayang:
@mathetake @nacx Do you mind to review this pr? Thanks a lot!

internal/translator/openai_awsbedrock.go
```diff
@@ -644,7 +642,9 @@ func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) ResponseBody(_ map[string
 		// We use request model as response model since bedrock does not return the modelName in the response.
 		Model:   o.requestModel,
 		Object:  "chat.completion",
+		Created: openai.JSONUNIXTime(time.Now()),
 		Choices: make([]openai.ChatCompletionResponseChoice, 0),
+		ID:      uuid.New().String(),
```
yuzisun, line 638: should we use the provider generated ID instead ?

yuzisun: for bedrock the request id is stored in `x-amzn-requestid`

Issue #1530, opened 2025-11-11
MCP Router failed to call resources/read.


Hyzhou:
Envoy AIGateway Version：0.4.0-rc1
Envoy Gateway Versin：1.5.4

Using the MCP Inspector (v0.17.2) to test the gateway MCPRoute. test proxy to mcp sample server. (https://github.com/Joooook/12306-mcp)

resources/list was ok
```
Request: 
{
	method: "resources/list"
	params: {}
}

Response: 
{
	resources: [
		0: {
			name: "modelscope__stations"
			uri: "data://all-stations"
		}
	]
}
```

resources/read was crashed
```
Request: 
{
	method: "resources/read"
	params: {
		uri: "data://all-stations"
	}
}

Response: 
{
	error: "MCP error -32001: Error POSTing to endpoint (HTTP 400): invalid resource name data://all-stations: invalid resource name: data://all-stations"
}
```

Without gateway, using MCP Inspector direct access mcp server, resources list works well.

resources/read was ok
```
Request: 
{
	method: "resources/read"
	params: {
		uri: "data://all-stations"
	}
}

Response:
{
	uri: "data://all-stations"
	text: "{"VAP":{"station_id":"@bjb","station_name":"北京北","station_code":"VAP","station_pinyin":"beijingbei",..."
}
```

PR #1535, opened 2025-11-11
fix: encode backend name in MCP resource URIs

nacx:
**Description**

In MCP, resource read requests do not include the name; just the resource URI:

```json
{
  "jsonrpc": "2.0",
  "id": 2,
  "method": "resources/read",
  "params": {
    "uri": "file:///project/src/main.rs"
  }
}
```

When returning the list of available resources, we were returning the backend encoded in the resource names, like:

```json
{
  "jsonrpc": "2.0",
  "id": 1,
  "result": {
    "resources": [
      {
        "uri": "file:///project/src/main.rs",
        "name": "backend__main.rs",
        "title": "Rust Software Application Main File",
        "description": "Primary application entry point",
        "mimeType": "text/x-rust"
      }
    ],
    "nextCursor": "next-page-cursor"
  }
}
```

However, when MCP clients perform the `resources/read` operation, they only send the URL, and there was not enough information there to decide what backend the request should be routed to.

In this PR, I'm proposing to encode the backend in the resource URI scheme part, like:

```json
"uri": "backend+file:///project/src/main.rs",
```

The spec allows for custom schemas [1].

I've considered other approaches, like generating URIs like: `aigw://backend/<encoded original URI>`, but an important thing to take into account is that **we can't do that for resource template URIs**, as they'll be parameterized like `file://files/{id}.txt`, and clients are expected to be able to simply perform variable substitution on those URLs. This means that we need to cleanly preserve the original URI somehow, and encoding the backend in the scheme looks like a simple and clean solution.

Fixes https://github.com/envoyproxy/ai-gateway/issues/1530

~~Opening the PR as a draft for early feedback. I still need to run some more tests.~~

Something important I noticed while developing this is that our MCP extproc tests were hardcoding some of the conventions in the calls we made to the server, preventing the detection of this issue. In this PR I also change the tests to use the outputs from the server, so that the entire interactions are consistent and don't make any assumptions.

1: https://modelcontextprotocol.io/specification/2025-06-18/server/resources#custom-uri-schemes

### DIFF
--- a/internal/mcpproxy/handlers.go
+++ b/internal/mcpproxy/handlers.go
@@ -650,6 +650,10 @@ func (m *MCPProxy) proxyResponseBody(ctx context.Context, s *session, w http.Res
 				case *jsonrpc.Response:
 					// Correct the ID to match the original request if possible.
 					if req != nil {
+						if err = m.maybeResponseModify(ctx, req, msg, backend.Name); err != nil {
+							m.l.Error("failed to modify response", slog.String("error", err.Error()))
+							continue
+						}
 						msg.ID = req.ID
 					}
 					m.recordResponse(ctx, msg)
--- a/tests/extproc/mcp/mcp_test.go
+++ b/tests/extproc/mcp/mcp_test.go
@@ -411,7 +411,7 @@ func testReadResource(t *testing.T, m *mcpEnv) {
 	})
 	require.NoError(t, err)
 	require.Len(t, r.Contents, 1)
-	require.Equal(t, testmcp.DummyResource.URI, r.Contents[0].URI)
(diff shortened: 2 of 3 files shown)


nacx:
@Hyzhou can you give this PR a try?

nacx:
I completed the tests and verified that resources, templates and completions work fine.
Extproc tests will now catch the misalignments that were not catched before.

Hyzhou:
> @Hyzhou can you give this PR a try?

The problem has been fixed.

Response:
```
{
  contents: [
    0: {
      uri: "modelscope+data://all-stations"
      text: "{"VAP":{"station_id":"@bjb","station_name":"北京北","station_code":"VAP","station_pinyin":"beijingbei",..."
    }
  ]
}
```

PR #1536, opened 2025-11-13
feat: support more delimiter for SSE events in GCP gemini

sukumargaonkar:
**Description**
The SSE event spec supports 3 different delimiters (pair of CRLF / LF / CR). This PR updates the stream processing translator for GCP VertexAI to support all 3 delimiters.

[SSE docs] about delimiters
> **Note:** The docs define an event as `event = *( comment / field ) end-of-line`
`end-of-line` is defined as a single CRLF or LF or CR (not a pair)
But the `comment` and `field` definitions also end in `end-of-line` implying that an event always ends in a pair of CRLR / CR / LF

[SSE docs]: https://html.spec.whatwg.org/multipage/server-sent-events.html#parsing-an-event-stream

### DIFF
--- a/internal/translator/anthropic_anthropic.go
+++ b/internal/translator/anthropic_anthropic.go
@@ -94,12 +94,7 @@ func (a *anthropicToAnthropicTranslator) ResponseBody(_ map[string]string, body 
 	if err := json.NewDecoder(body).Decode(anthropicResp); err != nil {
 		return nil, nil, tokenUsage, responseModel, fmt.Errorf("failed to unmarshal body: %w", err)
 	}
-	tokenUsage = LLMTokenUsage{
-		InputTokens:       uint32(anthropicResp.Usage.InputTokens),                                    //nolint:gosec
-		OutputTokens:      uint32(anthropicResp.Usage.OutputTokens),                                   //nolint:gosec
-		TotalTokens:       uint32(anthropicResp.Usage.InputTokens + anthropicResp.Usage.OutputTokens), //nolint:gosec
-		CachedInputTokens: uint32(anthropicResp.Usage.CacheReadInputTokens),                           //nolint:gosec
-	}
+	tokenUsage = ExtractLLMTokenUsageFromUsage(anthropicResp.Usage)
 	responseModel = cmp.Or(internalapi.ResponseModel(anthropicResp.Model), a.requestModel)
 	return nil, nil, tokenUsage, responseModel, nil
 }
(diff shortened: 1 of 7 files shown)


internal/translator/openai_gcpvertexai.go
```diff
@@ -218,27 +249,22 @@ func (o *openAIToGCPVertexAITranslatorV1ChatCompletion) parseGCPStreamingChunks(
 	var chunks []genai.GenerateContentResponse
 
 	// Read buffered body and new input, then split into individual chunks.
-	bodyBytes, err := io.ReadAll(io.MultiReader(bytes.NewReader(o.bufferedBody), body))
-	if err != nil {
-		return nil, err
-	}
-	lines := bytes.Split(bodyBytes, []byte("\n\n"))
+	bodyReader := io.MultiReader(bytes.NewReader(o.bufferedBody), body)
+	scanner := bufio.NewScanner(bodyReader)
```
nacx: When implementing the [mcp sse parser](https://github.com/envoyproxy/ai-gateway/blob/main/internal/mcpproxy/sse.go), I got issues with an initial implementation that used the Scanner; for some servers, I got long lines and the parsing failed with `bufio.ErrTooLong` (bufio.Scanner: token too long). I only caught this when using some real MCP servers.
I don't know if this can be the case as well here, but just a data point to consider not using the scanner if long lines could be expected.

sukumargaonkar: good point.
as of go 1.25.3 the default for `MaxScanTokenSize` seems to be 64KB

Since each streaming chunk represents couple of generated tokens, the size of each chunk should not exceed this limit. though i couldn't find docs that guarantee that the chunk will be smaller than a specific size

nacx: What comes from the LLM is inherently unpredictable, and I don't know if we can make that statement (for example, there are some tests that use chunks that have tool calls, and the length of the arguments could be unpredictable...

I agree it is _unlikely_ to happen, but my main concern is that we are introducing a limitation we didn't have before, and I don't want this to hit us in the form of a bug.

sukumargaonkar: two approaches i can think of
1. make the `MaxScanTokenSize` configurable
2. follow the [mcp sse parser](https://github.com/envoyproxy/ai-gateway/blob/main/internal/mcpproxy/sse.go) logic to explicitly parse the chunks

I personally prefer option 1 as relying on scanner to do the parsing aspect is simpler and easier maintain 

what do you think @nacx ?

sukumargaonkar:
unsure about test failures

`github.com/envoyproxy/ai-gateway/tests/e2e-upgrade`
```
timed out waiting for the condition on pods/envoy-default-upgrade-test-bd44e180-cd5d859c6-2rj22
    upgrade_test.go:146: 
        	Error Trace:	/home/runner/work/ai-gateway/ai-gateway/tests/e2e-upgrade/upgrade_test.go:146
        	Error:      	Received unexpected error:
        	            	[before upgrade (requests made: 0, current pods: envoy-default-upgrade-test-bd44e180-6d66b8f5d5-7q68j(Running), envoy-default-upgrade-test-bd44e180-cd5d859c6-2rj22(Running))] unexpected status code 500: body=
        	Test:       	TestUpgrade/rolling_out_pods
```

`github.com/envoyproxy/ai-gateway/tests/e2e-inference-extension`
```
error: error executing jsonpath "'{.items[0].spec.initContainers[*].name} {.items[0].spec.containers[*].name}'": Error executing template: array index out of bounds: index 0, length 0. Printing more information for debugging the template:
	template was:
		'{.items[0].spec.initContainers[*].name} {.items[0].spec.containers[*].name}'
	object given to jsonpath engine was:
		map[string]interface {}{"apiVersion":"v1", "items":[]interface {}{}, "kind":"List", "metadata":map[string]interface {}{"resourceVersion":""}}
```

internal/translator/openai_gcpvertexai.go
nacx: What about a third option: in the previous implementation, we're loading everything in memory and doing a `bytes.Split`. Why don't just keep that, but use the right function to split? I mean, test for the presence of the delimiters to figure out the "split characters", then keep the original code just splitting using those.

This way we don't change the behaviour, keep it backwards-compatible, and we don't introduce a new config option that users won't know what value to set confidently and how to set it. Also, if we find loading everything in memory to be an issue (it hasn't been till now), then we can address the stream reader approach, but probably it's not needed for this PR.

sukumargaonkar: updated implementation in https://github.com/envoyproxy/ai-gateway/pull/1536/commits/52781e5c167bfd1aaf6332873730bfd5d7fe8303

another round of review @nacx ?

PR #1537, opened 2025-11-13
fix:  fix stream parallel tool call index tests

hustxiayang:
**Description**
As @mathetake pointed out, the test function in https://github.com/envoyproxy/ai-gateway/pull/1524 is quite wrong. @mathetake fixed some in https://github.com/envoyproxy/ai-gateway/pull/1507, but there is still one bug: I did not insert "\n\n" in the string `gcpToolCallsChunk`, thus, it just parsed out one gcp chunk, which makes the test meaningless (it's to test across chunks).

Really sorry for the mistake in the tests!

### DIFF
--- a/internal/extproc/mocks_test.go
+++ b/internal/extproc/mocks_test.go
@@ -17,6 +17,7 @@ import (
 	"github.com/stretchr/testify/require"
 	"google.golang.org/grpc/metadata"
 
+	cohere "github.com/envoyproxy/ai-gateway/internal/apischema/cohere"
 	"github.com/envoyproxy/ai-gateway/internal/apischema/openai"
 	"github.com/envoyproxy/ai-gateway/internal/backendauth"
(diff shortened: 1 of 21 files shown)


internal/extproc/translator/openai_gcpvertexai_test.go
```diff
@@ -1687,8 +1688,7 @@ func TestOpenAIToGCPVertexAITranslatorV1ChatCompletion_StreamingParallelToolInde
                 "role": "model"
             }
         }
-    ],
-}`
+]}`
```
nacx, line 1691: To all these tests, add an assertion like:
```go
require.Len(t, chatCompletionChunks, <expected length>)
```
Otherwise, we still don't guarantee we're asserting the right things as we don't guarantee in the test that we enter the assertion for loop.

hustxiayang: added `require.Len(t, chatCompletionChunks, 2)`

PR #1539, opened 2025-11-13
feat: implement otel tracing for cohere v2 rerank endpoint

AyushSawant18588:
**Description**

This adds tracing support for cohere v2 rerank endpoint and follows the semantics mentioned for rerank in https://github.com/Arize-ai/openinference/blob/main/spec/semantic_conventions.md

### DIFF
--- a/internal/extproc/mocks_test.go
+++ b/internal/extproc/mocks_test.go
@@ -17,6 +17,7 @@ import (
 	"github.com/stretchr/testify/require"
 	"google.golang.org/grpc/metadata"
 
+	cohere "github.com/envoyproxy/ai-gateway/internal/apischema/cohere"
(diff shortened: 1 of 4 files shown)


internal/tracing/openinference/cohere/rerank.go
```diff
@@ -0,0 +1,185 @@
...
+		if total > 0 {
+			attrs = append(attrs, attribute.Int(openinference.LLMTokenCountTotal, total))
+		}
+	}
+
+	return attrs
+}
+
+// recordResponseError emits exception event and error status on the span.
+func recordResponseError(span trace.Span, statusCode int, body string) {
```
nacx: This method is identical to the one in the `openai` implementation. can you move it out to the `openinference` package and reuse it?

PR #1545, opened 2025-11-14
refactor: move config watcher out of extproc

mathetake:
**Description**

This decouples the filter config watcher from the extproc package as it's already solely operating on the filterapi package.

Preparation for #90

### DIFF
--- a/internal/filterapi/watcher.go
+++ b/internal/filterapi/watcher.go
@@ -13,7 +13,7 @@ import (
 	"time"
 )
 
-// ConfigReceiver is an interface that can receive *filterapi.Config updates.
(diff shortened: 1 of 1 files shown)


internal/filterapi/watcher.go
```diff
@@ -3,23 +3,21 @@
...
 	"context"
 	"fmt"
 	"log/slog"
 	"os"
 	"time"
-
-	"github.com/envoyproxy/ai-gateway/internal/filterapi"
 )
 
 // ConfigReceiver is an interface that can receive *filterapi.Config updates.
```
aabchoo:
```suggestion
// ConfigReceiver is an interface that can receive *Config updates.
```

PR #1546, opened 2025-11-15
refactor: move the runtime filter configuration out of extproc

mathetake:
**Description**

This commit decouples "runtime filter config" that is derived from the filterapi.Config to be able to reuse the code in dynamic modules port #90.

Preparation for #90


internal/extproc/server.go
```diff
@@ -63,39 +61,9 @@ func NewServer(logger *slog.Logger, tracing tracing.Tracing) (*Server, error) {
 
 // LoadConfig updates the configuration of the external processor.
 func (s *Server) LoadConfig(ctx context.Context, config *filterapi.Config) error {
-	backends := make(map[string]*processorConfigBackend, len(config.Backends))
```
mathetake, line 66: note: mainly this PR is moving this logic and the processorConfig* structs into filterapi/runtimefc package. Others are mechanical

internal/filterapi/runtimefc/runtimefc.go
```diff
@@ -0,0 +1,84 @@
+// Copyright Envoy AI Gateway Authors
+// SPDX-License-Identifier: Apache-2.0
+// The full text of the Apache license is available in the LICENSE file at
+// the root of the repo.
+
+// Package runtimefc is the API for the construct of runtime filter configurations parsed from filterapi.Config.
+//
+// Decoupled from filterapi to avoid circular dependencies.
```
mathetake: This is the reason why we cannot have this directly inside the filterapi package

mathetake:
PTAL @nacx

internal/filterapi/runtimefc/runtimefc.go
nacx: Looking at it, it doesn't feel quite right, then, does it?
I mean, we have an "api" package that depends on implementation-specific things like the backend-auth, because we need to call impl-specific constructors from this package...

Perhaps we need to split this differently?

mathetake: ack

mathetake: resolved -- Moved the interface into filterapi and now no cycle imports

nacx (approved):
Thank you!

PR #1550, opened 2025-11-19
fix: udpate anthropic token usage

alexagriffith:
**Description**
update anthropic -> translation code to correctly translate token usage

### DIFF
--- a/internal/translator/anthropic_usage.go
+++ b/internal/translator/anthropic_usage.go
@@ -38,16 +38,6 @@ func ExtractLLMTokenUsageFromUsage(usage anthropic.Usage) LLMTokenUsage {
 	)
 }
 
-// ExtractLLMTokenUsageFromMessageUsage extracts token usage from streaming message_start events.
-func ExtractLLMTokenUsageFromMessageUsage(usage anthropic.Usage) LLMTokenUsage {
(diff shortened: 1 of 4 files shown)


internal/translator/anthropic_usage.go
```diff
@@ -0,0 +1,59 @@
...
+	return LLMTokenUsage{
+		InputTokens:       uint32(totalInputTokens),                //nolint:gosec
+		OutputTokens:      uint32(outputTokens),                    //nolint:gosec
+		TotalTokens:       uint32(totalInputTokens + outputTokens), //nolint:gosec
+		CachedInputTokens: uint32(totalCachedTokens),               //nolint:gosec
+	}
+}
+
+// ExtractLLMTokenUsageFromUsage extracts token usage from anthropic.Usage struct (non-streaming).
+func ExtractLLMTokenUsageFromUsage(usage anthropic.Usage) LLMTokenUsage {
```
aabchoo, line 32: Is there a reason why you have the same function twice (if not counting delta usage) with different names?

What's the difference between MessageDeltaUsage and Usage?

alexagriffith: sorry, was an error when I was refactoring. removed it

PR #1526, opened 2025-11-10
feat: enable enterprise web search for gemini models

hustxiayang:
**Description**

This is to enable feature `Web Grounding for Enterprise`: https://docs.cloud.google.com/vertex-ai/generative-ai/docs/grounding/web-grounding-enterprise

**Related Issues/PRs (if applicable)**
https://github.com/envoyproxy/ai-gateway/issues/1417

PR #1554, opened 2025-11-20
feat: add new features(thinking_level and media_resolution) of gemini3

hustxiayang:
**Description**
Some new features were introduced in gemini3:
**1 thinking_level:**
https://ai.google.dev/gemini-api/docs/gemini-3?thinking=low#thinking_level
This is similar to reasoning_effort of openai, thus, unified them.

**2 media_resolution**
https://ai.google.dev/gemini-api/docs/gemini-3?thinking=low#media_resolution
This is similar to detail in openai, thus, unified them.

The difference is that openai does not provide a global config of media_resolution. Thus, added it as gcp specific, but still use detail to make the name consistent.

**Some related PRs:**
thinking_budget is in
https://github.com/envoyproxy/ai-gateway/pull/1461
thinking_level and thinking_budget are both supported, but can not use them together.

Other features under review:
**1 web search:**
https://github.com/envoyproxy/ai-gateway/pull/1526

**2 parse the thought summary:**
https://github.com/envoyproxy/ai-gateway/pull/1521


internal/apischema/openai/openai.go
```diff
@@ -1580,6 +1581,9 @@ type GCPVertexAIGenerationConfig struct {
 	//
 	// https://cloud.google.com/vertex-ai/docs/reference/rest/v1/GenerationConfig#ThinkingConfig
 	ThinkingConfig *genai.ThinkingConfig `json:"thinkingConfig,omitzero"`
+
+	// Detail is media resolution in gemini models
+	Detail genai.MediaResolution `json:"detail,omitempty"`
```
yuzisun: can you add the reference link?

hustxiayang: added, thanks for the comment

yuzisun: Actually this filed should be added to the image part details

yuzisun: the link is wrong, it points to the caching API

hustxiayang: `this filed should be added to the image part details` is already in openai' spec

hustxiayang: `it points to the caching API` I think the definition is the same

internal/apischema/openai/openai.go
```diff
@@ -1580,6 +1582,9 @@ type GCPVertexAIGenerationConfig struct {
 	//
 	// https://cloud.google.com/vertex-ai/docs/reference/rest/v1/GenerationConfig#ThinkingConfig
 	ThinkingConfig *genai.ThinkingConfig `json:"thinkingConfig,omitzero"`
+
+	// Detail here is to set global media resolution in gemini models: https://ai.google.dev/api/caching#MediaResolution
+	Detail genai.MediaResolution `json:"detail,omitempty"`
```
yuzisun: `details` can be confusing without context at the top level, since it is in the GCP vendor field section we should keep the name `medium_resolution`.

hustxiayang: updated.

yuzisun (approved):

Issue #1552, opened 2025-11-20
How to achive tracing per gateway?


pkalemba:
Hi!﻿

I'm currently testing the ai-gateway, and I found documentation about OTEL tracing, but it describes global tracing. Is there any option to enable it per gateway?﻿

I'd like to configure tracing selectively for specific gateway instances rather than enabling it across the entire setup. Could you point me to the relevant configuration options or documentation?﻿

Thanks in advance!﻿

sqaisar:
This will be really nice if we could add OTEL_EXPORTER_OTLP_ENDPOINT in the CRDs rather than defining it in the ai gateway controller's helm chart configs. 

It'll be a really nice feature. I don't think it exists today.

sqaisar:
@pkalemba  I've opened the above PR https://github.com/envoyproxy/ai-gateway/pull/1561 that adds that support to define environment variables per route giving it the ability to define tracing config per ai gateway deployment.
let's see if that gets approved and merged.

mathetake:
yeah this is a valid request. however, unfortunately, #1561 is not the way to go. Because of the how request-body-based (model name) routing is performed, the filter configuration (extproc config) is gateway-wide side effect, and the env var like this should be a gateway-scoped configuration instead of router level (as you see, the PR has appending multiple env var configs from multiple routes). The "ai gateway filter config" `filterConfig` will be deprecated soon anyways (see below), and the existing cpu/memory config shouldn't exist there too (historically before we could only attach one route to a gateway, hence the resource config is there).

The path forward IMO is introduce "GatewayConfig" CRD in our API where we allow the config like this as well as the resource (cpu/memory) config, and use the annotation to indicate the configuration attached to a Gateway object. Then, when we promote the existing v1alpha1 API to beta, I would suggest to drop the `filterConfig` field from AIGatewayRoute cc @johnugeorge @yuzisun 

Meanwhile, you can already deploy AIGW in any namespace and specify the watch target namespace, and that would allow you to configure the different tracing config for multiple Gateways. I hope this helps

nacx:
Agree that routes are not the right places to configure such things. This is an `infrastructure` concern and, as such, should be configured outside the scope of a route.

[...]

PR #1561, opened 2025-11-24
feat: allow AIGatewayRoute to define environment variables

sqaisar:
**Description**

This PR introduces the ability to define environment variables for the external processor container directly within the `AIGatewayRoute` CRD. This allows for per-gateway configuration of environment variables, which is particularly useful for use cases like configuring tracing (e.g., `OTEL_EXPORTER_OTLP_HEADERS`) or other runtime settings specific to a single gateway instance without affecting others.

Key changes include:
- Added `ExtProcEnv` field to `AIGatewayRouteSpec` to support custom environment variables.
- Updated the gateway mutator to apply these environment variables to the external processor container.
- Added comprehensive tests to ensure environment variables are applied correctly and isolated per gateway.
- Updated documentation to reflect how to use this feature for configuration.

**Example Usage**

You can now specify environment variables for the external processor in your `AIGatewayRoute` configuration under `spec.filterConfig.externalProcessor.env`. These variables are applied specifically to the external processor container associated with this gateway route.

```yaml
kind: AIGatewayRoute
metadata:
  name: my-ai-gateway
spec:
  filterConfig:
    type: ExternalProcessor
    externalProcessor:
      env:
        - name: OTEL_EXPORTER_OTLP_HEADERS
          value: "api-key=secret-token"
        - name: LOG_LEVEL
          value: "debug"
  rules:
    - backendRefs:
        - name: my-ai-backend
```

**How it works**

When the AI Gateway controller reconciles the `AIGatewayRoute`, it checks for the presence of `spec.filterConfig.externalProcessor.env`. If defined, these environment variables are injected into the external processor container of the corresponding Gateway deployment.

- **Isolation**: The environment variables are scoped to the specific Gateway instance managed by this `AIGatewayRoute`. They do not affect other gateways or the global controller configuration.
- **Merging**: These variables are merged with any default environment variables required by the system. If there's a conflict, the values defined in `AIGatewayRoute` take precedence (depending on the specific merging logic implemented, but typically specific overrides general).

https://github.com/envoyproxy/ai-gateway/issues/1552

The implementation ensures that environment variables defined in one `AIGatewayRoute` do not leak into other gateways or the global configuration. Verification tests covering `mutatePod` have been added to confirm this isolation.

### DIFF
--- a/internal/controller/gateway_mutator_per_gateway_env_test.go
+++ b/internal/controller/gateway_mutator_per_gateway_env_test.go
@@ -84,3 +84,102 @@ func TestGatewayMutator_mutatePod_PerGatewayEnvVars(t *testing.T) {
 	}
 	require.Equal(t, expectedEnvVars, extProcContainer.Env)
 }
+
+func TestGatewayMutator_mutatePod_NoSideEffects(t *testing.T) {
+	fakeClient := requireNewFakeClientWithIndexes(t)
+	fakeKube := fake2.NewClientset()
+	// Initialize with global env vars
+	g := newTestGatewayMutator(fakeClient, fakeKube, "", "", "", "GLOBAL_VAR=global-value", "", false)
+
+	// Route 1 with specific env var
+	const gwName1, gwNamespace = "gateway-1", "test-namespace"
+	err := fakeClient.Create(t.Context(), &aigv1a1.AIGatewayRoute{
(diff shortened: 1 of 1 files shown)


mathetake:
https://github.com/envoyproxy/ai-gateway/issues/1552#issuecomment-3572705608

PR #1562, opened 2025-11-24
fix: skip mediaResolution and thinking_level for older versions of gemini

hustxiayang:
**Description**

Skip mediaResolution and thinking_level for older versions of gemini. 

Currently the gemini backend would report errors that those older versions of gemini models do not support these new parameters.

Two options:

1 We can keep it as it so that users know these are not supported;
2 The drawback of keeping it as it is is that it might break existing codes: some users might used these parameters for OpenAI models, previously these parameters are automatically ignored, now it would return an error. Also, some users might still want to use the same setup, if the parameters are not supported, just ignore the parameters.

### DIFF
--- a/internal/translator/gemini_helper.go
+++ b/internal/translator/gemini_helper.go
@@ -457,7 +457,7 @@ func openAIToolChoiceToGeminiToolConfig(toolChoice *openai.ChatCompletionToolCho
 // Check and Update these functions when new Gemini model versions are released
 // and their feature support changes.
 
-// it only works with gemini2.5 according to https://ai.google.dev/gemini-api/docs/structured-output#json-schema, separate it as a small function to make it easier to maintain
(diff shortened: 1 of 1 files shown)


internal/translator/gemini_helper.go
```diff
@@ -447,9 +447,31 @@ func openAIToolChoiceToGeminiToolConfig(toolChoice *openai.ChatCompletionToolCho
...
+// Gemini Version-Specific Feature Availability Functions
+// ------------------------------------------------------------
+//
+// These functions implement version-specific feature gating for Gemini models
+// to prevent potential breaking existing codes when using newer features with older model versions.
+//
+// Check and Update these functions when new Gemini model versions are released
+// and their feature support changes.
+
 // it only works with gemini2.5 according to https://ai.google.dev/gemini-api/docs/structured-output#json-schema, separate it as a small function to make it easier to maintain
```
yuzisun: change the comment as now we have gemini 3

hustxiayang: updated. Thanks for the comment.

PR #1564, opened 2025-11-24
feat: adds dynamic modules based data plane

mathetake:
**Description**

This commit adds the complete implementation of dynamic modules based data plane as an alternative to the existing extproc based implementation. It passes all the existing tests for extproc meaning that it has the exact feature parity with extproc. More precisely, this passes all the tests in `make test-data-plane` without making any changes to the test cases themselves.

Note that this still lacks the control plane implementation to allow users to actually use it. It requires some non trivial plumbing but since this PR comes with the raw Envoy configuration for dynamic modules based data plane, which is used in the data plane end to end tests, it should be not that hard to implement the control plane.

#90


mathetake:
first test case passing!!!!!

mathetake:
making progresses but still many tests failing

mathetake:
finally passed most of the testupstream tests except for three test cases!!!!!!!!!

mathetake:
finally passed all existing testupstream extproc test cases.........

mathetake:
YES the rest is to backfill the unit tests

PR #1567, opened 2025-11-25
refactor: fixes inconsistent span passing in embeddings 

mathetake:
**Description**

This fixes the inconsistency for embeddings in the span usage in translators.

Preparation for #90
Follow up on #1565

### DIFF
--- a/internal/translator/translator.go
+++ b/internal/translator/translator.go
@@ -77,13 +77,13 @@ type Translator[ReqT any, SpanT any] interface {
 type (
 	// OpenAIChatCompletionTranslator translates the OpenAI's /chat/completions endpoint.
 	OpenAIChatCompletionTranslator = Translator[openai.ChatCompletionRequest, tracing.ChatCompletionSpan]
-	// OpenAIEmbeddingTranslator = translates the OpenAI's /embeddings endpoint.
+	// OpenAIEmbeddingTranslator translates the OpenAI's /embeddings endpoint.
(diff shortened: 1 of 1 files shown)


internal/translator/translator.go
```diff
@@ -78,7 +78,7 @@ type (
 	// OpenAIChatCompletionTranslator translates the OpenAI's /chat/completions endpoint.
 	OpenAIChatCompletionTranslator = Translator[openai.ChatCompletionRequest, tracing.ChatCompletionSpan]
 	// OpenAIEmbeddingTranslator = translates the OpenAI's /embeddings endpoint.
-	OpenAIEmbeddingTranslator = Translator[openai.EmbeddingRequest, any]
+	OpenAIEmbeddingTranslator = Translator[openai.EmbeddingRequest, tracing.EmbeddingsSpan]
```
mathetake, line 81: this is the fix

aabchoo:
Will the same need to be done for CohereRerankTranslator, AnthropicMessagesTranslator, and OpenAIImageGenerationTranslator?

mathetake:
> Will the same need to be done for CohereRerankTranslator, AnthropicMessagesTranslator, and OpenAIImageGenerationTranslator?


so the tracers are not implemented for the other things hence any. But good q anyways!

mathetake:
oops aaron you are right we have at least span for image gen...

PR #1572, opened 2025-11-25
refactor: consolidates all metrics into one interface

mathetake:
**Description**

This consolidates all metrics interfaces into one interface as they are almost all copy and pasted from each other.

Preparation for #90


internal/extproc/messages_processor.go
```diff
@@ -332,9 +332,16 @@ func (c *messagesProcessorUpstreamFilter) ProcessResponseBody(ctx context.Contex
...
 	if c.stream {
 		c.metrics.RecordTokenLatency(ctx, tokenUsage.OutputTokens, body.EndOfStream, c.requestHeaders)
+		// Emit usage once at end-of-stream using final totals.
+		if body.EndOfStream {
+			c.metrics.RecordTokenUsage(ctx,
+				metrics.OptUint32(c.costs.InputTokens), metrics.OptUint32(c.costs.CachedInputTokens), metrics.OptUint32(c.costs.OutputTokens), c.requestHeaders)
+		}
+	} else {
+		c.metrics.RecordTokenUsage(ctx,
+			metrics.OptUint32(tokenUsage.InputTokens), metrics.OptUint32(tokenUsage.CachedInputTokens), metrics.OptUint32(tokenUsage.OutputTokens), c.requestHeaders)
```
mathetake, line 344: note: this was a bug in Anthropic impl

internal/metrics/metrics.go
```diff
@@ -89,3 +92,53 @@ func NewMetricsFromEnv(ctx context.Context, stdout io.Writer, promReader sdkmetr
 	mp := sdkmetric.NewMeterProvider(options...)
 	return mp.Meter("envoyproxy/ai-gateway"), mp.Shutdown, nil
 }
+
+// OptUint32 is an optional uint32 type represented as an uint64 to allow for a None value.
+type OptUint32 uint64
+
+// OptUint32None represents the None value for OptUint32.
+const OptUint32None = OptUint32(0xffffffff_00000000)
```
mathetake, line 100: this will be deleted in a follow up when moving LLMCost into this package where we can have the optional stuff as an implementation detail contained here vs being exported from here. I was thinking to do that in this PR but the blast radius will be much larger (as it will involve translator package).

mathetake:
tomorrow i will backfill the unit tests for the metrics impl

PR #1583, opened 2025-11-26
refactor: move Token Usage definition to metrics package from translator

mathetake:
**Description**

This moves translator.LLMTokenUsage struct into metrics package from translator package to make sure that all call site of metrics interface uses the same signature by simply passing the cost info from translator. This will help unify all the per-endpoint implementation detail.

Preparation for #90

### DIFF
--- a/internal/translator/anthropic_anthropic_test.go
+++ b/internal/translator/anthropic_anthropic_test.go
@@ -94,8 +94,7 @@ func TestAnthropicToAnthropic_ResponseBody_non_streaming(t *testing.T) {
 	require.NoError(t, err)
 	require.Nil(t, headerMutation)
 	require.Nil(t, bodyMutation)
-	expected := tokenUsageFrom(9, 16, 25)
-	expected.SetCachedInputTokens(0)
+	expected := tokenUsageFrom(9, 0, 16, 25)
 	require.Equal(t, expected, tokenUsage)
 	require.Equal(t, "claude-sonnet-4-5-20250929", responseModel)
 }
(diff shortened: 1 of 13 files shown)


internal/metrics/metrics.go
```diff
@@ -120,7 +114,7 @@ type Metrics interface {
 	// RecordTokenUsage records token usage metrics.
 	//
 	// Depending on the endpoint, some token types are not available and should be passed as OptUint32None.
-	RecordTokenUsage(ctx context.Context, inputTokens, cachedInputTokens, outputTokens OptUint32, requestHeaders map[string]string)
+	RecordTokenUsage(ctx context.Context, usage TokenUsage, requestHeaders map[string]string)
```
mathetake, line 117: this and the type addition below are the main change and others are mechanical

internal/metrics/metrics.go
```diff
@@ -142,3 +136,108 @@ type Factory interface {
 func NewMetricsFactory(meter metric.Meter, requestHeaderLabelMapping map[string]string, operation GenAIOperation) Factory {
 	return &metricsImplFactory{metrics: newGenAI(meter), requestHeaderAttributeMapping: requestHeaderLabelMapping, operation: string(operation)}
 }
+
+// TokenUsage represents the token usage reported usually by the backend API in the response body.
+//
+// Fields are not exported to control the optionality of each field via the accompanying boolean flags.
```
mathetake, line 142: important

mathetake:
OK this is the last refactoring piece I wanted for dynamic modules... as a good side effect, now we can also consolidate all the processor code into one generic type for ext_proc too, which will essentially resolve #1083

nacx (approved):

PR #1590, opened 2025-11-29
docs: update doc for extension fields 

yuzisun:
**Description**
Explained how the unified thinking field works across providers and documented automatic translation: GCP Vertex AI (Gemini) → generationConfig.thinkingConfig, while GCP Anthropic and AWS Bedrock use it directly. Update extension field usage for both vendor specific and unified fields.


hustxiayang:
Can we also add the mediaResolution and websearch in this doc too?

PR #1584, opened 2025-11-27
feat: add audio endpoint support /v1/audio/speech

tanthcstt:
**Description**

This PR adds support for the OpenAI-compatible `/v1/audio/speech` text-to-speech endpoint with translation support for GCP Vertex AI backends.

**Implementation Files**:

- `internal/extproc/audiospeech_processor.go` - Core processor logic
- `internal/extproc/audiospeech_processor_test.go` - Comprehensive test coverage
- `internal/translator/audio_openai_gcpvertexai.go` - Backend translation logic
- `internal/metrics/audio_speech_metrics.go` - Metrics and observability

**Key Features**:

- JSON request body parsing for TTS requests
- Voice mapping between OpenAI and Gemini voice models
- Binary audio response handling
- Streaming audio generation support
- Cost tracking and usage metrics

**Voice Mapping**:
| OpenAI | Gemini | Description |
|--------|--------|-------------|
| alloy | Zephyr | Neutral, balanced |
| echo | Puck | Male voice |
| fable | Aoede | Expressive |
| onyx | Fenrir | Deep male |
| nova | Kore | Female voice |
| shimmer | Thetis | Soft female |

**Translation Logic**:

1. Converts OpenAI TTS request to Gemini's audio generation format
2. Maps OpenAI voice names to Gemini prebuilt voices
3. Sets response modality to `AUDIO` in generation config
4. Configures speech settings (voice, temperature)
5. Extracts binary audio data from streaming response chunks
6. Returns raw audio data in OpenAI-compatible format

**Metrics**:

- Request tracking with voice and model details
- Character count as input tokens
- Audio generation duration tracking
- Backend performance monitoring
- Custom header label mapping support

**Related Issues/PRs (if applicable)**
#1499

Issue #1499, opened 2025-11-06
Support for OpenAI Audio / Transcription Endpoints


mathetake:
### Discussed in https://github.com/envoyproxy/ai-gateway/discussions/1498

<div type='discussions-op-text'>

<sup>Originally posted by **tanthcstt** November  5, 2025</sup>
Does Envoy Gateway Support for OpenAI Audio / Transcription Endpoints</div>

tanthcstt:
I’m working on it. I’d like to do it.

tanthcstt:
I have no experience with Kubernetes; it was a little bit hard for me to deploy and test, but finally it works. I will create a PR soon. I’m trying to understand what I did in this repo first :) AI has helped me a lot.

tanthcstt:
@mathetake
/v1/audio/transcriptions and /v1/audio/speech work with the OpenAI provider. Do I need to create a translator for Gemini audio understanding? It’s a little different, but both can be used for audio transcription. Gemini uses the generateContent endpoint with a prompt, while OpenAI has a separate endpoint (/audio/transcriptions).

johnugeorge:
@tanthcstt  You may start without translation(passthrough for Openai provider). Translators can be added later. Take a look at image generation api.

tanthcstt:
Do all tests need to pass before creating a PR?
@johnugeorge  
<img width="443" height="729" alt="Image" src="https://github.com/user-attachments/assets/b8d239d8-0104-4c1f-9b94-5c8f66160167" />

tanthcstt:
@mathetake #1592 #1584

PR #1592, opened 2025-12-01
feat: add support for audio transcription endpoint /v1/audio/transcriptions

tanthcstt:
**Description**

This PR adds support for the OpenAI-compatible `/v1/audio/transcriptions` speech-to-text endpoint with translation support for GCP Vertex AI  backends.

## Key features
- Streaming (SSE) and non-streaming response handling
- SSE streaming chunk parsing and accumulation of transcription text
- Token usage tracking (input/output/total) and metrics emission

Additional notes: clients may request server-sent-event (SSE) style streaming by passing an explicit `stream` boolean parameter (see table below). When `stream=true`, the gateway will select the streaming endpoint and translate SSE chunks into the OpenAI-compatible streaming behavior.

## Supported request parameters

| Parameter | Type | Description |
|---|---:|---|
| `model` | string | Model identifier (e.g., `whisper-1`) |
| `language` | string | Language of the audio (optional) |
| `prompt` | string | Custom transcription prompt (optional) |
| `response_format` | string | Output format (optional) |
| `temperature` | float | Sampling temperature (optional) |
| `stream` | bool | Request streaming (SSE) responses when `true` (optional) |
| `timestamp_granularities` | `[]string` | Timestamp detail level (optional) |

## Backend support

- **OpenAI**: passthrough translator that forwards request bodies with optional model override.
- **GCP Vertex AI / Gemini**: translator builds Gemini requests, handles streaming and non-streaming responses, and converts them back to OpenAI-compatible responses.

## Related issues / PRs

- Closes / relates to: #1499
- it can be conflict with PR https://github.com/envoyproxy/ai-gateway/pull/1584

### DIFF
--- a/cmd/extproc/mainlib/main.go
+++ b/cmd/extproc/mainlib/main.go
@@ -255,6 +255,7 @@ func Main(ctx context.Context, args []string, stderr io.Writer) (err error) {
 	embeddingsMetricsFactory := metrics.NewMetricsFactory(meter, metricsRequestHeaderAttributes, metrics.GenAIOperationEmbedding)
 	imageGenerationMetricsFactory := metrics.NewMetricsFactory(meter, metricsRequestHeaderAttributes, metrics.GenAIOperationImageGeneration)
 	rerankMetricsFactory := metrics.NewMetricsFactory(meter, metricsRequestHeaderAttributes, metrics.GenAIOperationRerank)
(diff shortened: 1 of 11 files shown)


mathetake:
Per https://github.com/envoyproxy/ai-gateway/pull/1584#pullrequestreview-3535877047, let's work on a feature step by step. The first endpoint impl is done including all the user facing doc, then let's come back here and do the same thing with the reviewed context/style, etc rather than doing the exactly same back-and-forth in parallel in multiple PRs

PR #1593, opened 2025-12-01
mcp: setup benchmark test for MCPProxy

zirain:
**Description**

This PR create a benchmark test for AIGW withe diffrent `mcp-session-encryption-iterations`

Run this benchmark with following:

```console
go test -timeout=15m -run='^$$' -bench=. -benchmem -benchtime=1x ./tests/bench/...
2025/12/03 13:50:12 starting DUMB MCP Streamable-HTTP server on :8080 at /mcp
goos: darwin
goarch: arm64
pkg: github.com/envoyproxy/ai-gateway/tests/bench
cpu: Apple M1 Pro
BenchmarkMCP/BaseLine-10                       1          31305583 ns/op           49736 B/op        720 allocs/op
BenchmarkMCP/Iterations_100-10                 1           1394125 ns/op           22640 B/op        300 allocs/op
PASS
ok      github.com/envoyproxy/ai-gateway/tests/bench    5.394s
```

~~This result that the default `SessionCrypto` cost a lot, we'd better improve it or change the default setting.~~

cc @nacx


internal/mcpproxy/handlers.go
```diff
@@ -231,10 +231,16 @@ func (m *MCPProxy) servePOST(w http.ResponseWriter, r *http.Request) {
 			// The MCP proxy uses it to locate the route’s backends and initialize sessions with them.
 			route := r.Header.Get(internalapi.MCPRouteHeader)
 			if route == "" {
-				errType = metrics.MCPErrorInternal
-				m.l.Error("cannot find route header in the downstream request")
-				onErrorResponse(w, http.StatusInternalServerError, "missing route header")
-				return
+				if len(m.routes) == 1 {
```
zirain: this's a hack for running MCPProxy without front envoy.

nacx:
Thanks!

Now that we've seen that running the MCPServer "outside" Envoy has a negligible impact, could you refactor the benchmarks as:

* Focus on calling an externally running proxy:
  * The test cases could have a name, port, and the command to start the proxy.
  * Each test case would run the "start proxy" command and wait for it to be available.
  * Now that https://github.com/envoyproxy/ai-gateway/pull/1604 is merged, we can have several tests configured with different commands to set different values for the iterations.
  * Having commands to launch the proxy in the test cases makes it trivial for anyone to run the benchmarks with their own proxy, to more easily compare.
* Running the `aigw` with `go run` has worse performance than just running a compiled binary, so for more real-world benchmarks, I would not use `go run`. We can think about smart ways of deciding what binary to run in the default benchmarks (from the `out` directory), or make it configurable like reading from an env var, etc.

Not having to run the "out of Envoy" allows us to remove the hacks. We can also remove the `NoopSessionCrypto` and just focus on real-world examples.

WDYT?

cmd/aigw/run.go
```diff
@@ -138,14 +138,15 @@ func run(ctx context.Context, c cmdRun, o *runOpts, stdout, stderr io.Writer) er
...
-		extProcLauncher:          o.extProcLauncher,
+		isDebug:                        c.Debug,
+		envoyGatewayResourcesOut:       resourcesBuf,
+		stderrLogger:                   debugLogger,
+		stderr:                         stderr,
+		tmpdir:                         filepath.Dir(o.logPath), // runDir
+		udsPath:                        o.extprocUDSPath,
+		adminPort:                      c.AdminPort,
+		extProcLauncher:                o.extProcLauncher,
+		mcpSessionEncryptionIterations: c.MCPSessionEncryptionIterations,
```
zirain, line 149: this's missed in https://github.com/envoyproxy/ai-gateway/pull/1604

tests/internal/testmcp/server.go
```diff
@@ -130,6 +131,9 @@ func NewServer(opts *Options) *http.Server {
 		WriteTimeout: opts.WriteTimeout,
 		Handler:      handler,
 		ConnState: func(conn net.Conn, state http.ConnState) {
+			if opts.DisableLog {
```
zirain, line 134: without this, too much noise during bench test.

tests/bench/bench_test.go
```diff
@@ -0,0 +1,188 @@
...
+		DumbEchoServer:    true,
+		WriteTimeout:      writeTimeout,
+		DisableLog:        true,
+	})
+	b.Cleanup(func() {
+		_ = mcpServer.Close()
+	})
+
+	// Reset the timer to exclude setup time from the results
+	b.ResetTimer()
```
nacx: nit: I'd move this to the benchmark test, and remove it from the helper function, so all the control on the benchmark test itself in in hte main test function

tests/bench/aigw.yaml
```diff
@@ -0,0 +1,123 @@
...
+###################################################################################
+apiVersion: gateway.envoyproxy.io/v1alpha1
+kind: Backend
+metadata:
+  name: test
+  namespace: default
+spec:
+  endpoints:
+    - ip:
+        address: 0.0.0.0
```
nacx: nit: we could better change this to `127.0.0.1`? In the recent versions of EG and EAIGW we allow localhost addresses when running in standalone mode.

tests/bench/bench_test.go
```diff
@@ -0,0 +1,188 @@
...
+				if txt.Text != "dumb echo: hello MCP" {
+					b.Fatalf("unexpected text: %q", txt.Text)
+				}
+			}
+		})
+	}
+}
+
+func killProcessListeningOn(port int) error {
+	cmd := exec.Command("lsof", "-nP", "-sTCP:LISTEN", "-i", fmt.Sprintf("TCP:%d", port), "-t") // nolint: gosec
```
nacx: Instead of doing this, I think it would be better to:

1. Create the command like this:

```go
cmd := exec.Command("your-command")
cmd.SysProcAttr = &syscall.SysProcAttr{
    Setpgid: true, // put into new process group so we can kill the entire process tree (and children)
}
```

2. In the test teardown, kill the process and children:

```go
// Kill the whole process group:
syscall.Kill(-cmd.Process.Pid, syscall.SIGKILL)
```

This way we can ensure that test execution does not leak processes, and we don't need this hacky kill based on listening ports.

WDYT?

tests/bench/bench_test.go
nacx: Actually we cna jsut remove it as we're already doing the reset before the tool call.

internal/mcpproxy/handlers.go
```diff
@@ -231,10 +231,16 @@ func (m *MCPProxy) servePOST(w http.ResponseWriter, r *http.Request) {
...
+					for mcpRouteName := range m.routes {
+						route = mcpRouteName
+					}
+				} else {
+					errType = metrics.MCPErrorInternal
+					m.l.Error("cannot find route header in the downstream request")
+					onErrorResponse(w, http.StatusInternalServerError, "missing route header")
+					return
+				}
```
nacx: We can remove the changes to this file now that we're always executing the CLI, right?

tests/bench/bench_test.go
zirain: TIL

internal/mcpproxy/handlers.go
zirain: missed this one, reverted

PR #1598, opened 2025-12-01
mcp: allow configuring the session encryption iterations

nacx:
**Description**

Allow configuring the number of iterations to be executed for the MCP session encryption.
MCP sessions are encrypted using  PBKDF2 for key derivation and AES-GCM for encryption. This provides some flexibility to allow users to balance between security and performance.

The default implementation defaults to 100.000 iterations, which should be a reasonable tradeoff between security and speed. This PR introduces a new Helm value to allow users change this value to make the MCP proxy faster if needed.

The following benchmarks show the behaviour for different values:

```
Running tool: /opt/homebrew/opt/go@1.24/bin/go test -benchmem -run=^$ -bench ^BenchmarkPBKDF2AesGcmSessionCrypto$ github.com/envoyproxy/ai-gateway/internal/mcpproxy

goos: darwin
goarch: arm64
pkg: github.com/envoyproxy/ai-gateway/internal/mcpproxy
cpu: Apple M1
BenchmarkPBKDF2AesGcmSessionCrypto/encrypt_100-8         	   85375	     14220 ns/op	    2560 B/op	      20 allocs/op
BenchmarkPBKDF2AesGcmSessionCrypto/decrypt_100-8         	   87002	     14356 ns/op	    2260 B/op	      16 allocs/op
BenchmarkPBKDF2AesGcmSessionCrypto/encrypt_1000-8        	    8824	    126737 ns/op	    2560 B/op	      20 allocs/op
BenchmarkPBKDF2AesGcmSessionCrypto/decrypt_1000-8        	    9558	    116771 ns/op	    2260 B/op	      16 allocs/op
BenchmarkPBKDF2AesGcmSessionCrypto/encrypt_10000-8       	    1056	   1512627 ns/op	    2560 B/op	      20 allocs/op
BenchmarkPBKDF2AesGcmSessionCrypto/decrypt_10000-8       	     799	   1350919 ns/op	    2260 B/op	      16 allocs/op
BenchmarkPBKDF2AesGcmSessionCrypto/encrypt_50000-8       	     211	   5655061 ns/op	    2560 B/op	      20 allocs/op
BenchmarkPBKDF2AesGcmSessionCrypto/decrypt_50000-8       	     208	   5703031 ns/op	    2260 B/op	      16 allocs/op
BenchmarkPBKDF2AesGcmSessionCrypto/encrypt_100000-8      	     100	  11378411 ns/op	    2560 B/op	      20 allocs/op
BenchmarkPBKDF2AesGcmSessionCrypto/decrypt_100000-8      	     100	  11343512 ns/op	    2260 B/op	      16 allocs/op
BenchmarkPBKDF2AesGcmSessionCrypto/encrypt_200000-8      	      52	  23629280 ns/op	    2560 B/op	      20 allocs/op
BenchmarkPBKDF2AesGcmSessionCrypto/decrypt_200000-8      	      50	  26281746 ns/op	    2260 B/op	      16 allocs/op
PASS
ok  	github.com/envoyproxy/ai-gateway/internal/mcpproxy	15.551s
```

N/A

N/A

### DIFF
--- a/internal/translator/cohere_rerank_v2.go
+++ b/internal/translator/cohere_rerank_v2.go
@@ -11,6 +11,7 @@ import (
 	"io"
 	"path"
 	"strconv"
+	"strings"
 
 	"github.com/tidwall/sjson"
 
@@ -111,7 +112,7 @@ func (t *cohereToCohereTranslatorV2Rerank) ResponseBody(_ map[string]string, bod
 func (t *cohereToCohereTranslatorV2Rerank) ResponseError(respHeaders map[string]string, body io.Reader) (
 	newHeaders []internalapi.Header, newBody []byte, err error,
 ) {
-	if v, ok := respHeaders[contentTypeHeaderName]; ok && v != jsonContentType {
+	if v, ok := respHeaders[contentTypeHeaderName]; ok && !strings.Contains(v, jsonContentType) {
 		buf, err := io.ReadAll(body)
 		if err != nil {
 			return nil, nil, fmt.Errorf("failed to read error body: %w", err)
@@ -125,10 +126,10 @@ func (t *cohereToCohereTranslatorV2Rerank) ResponseError(respHeaders map[string]
 		if err != nil {
 			return nil, nil, fmt.Errorf("failed to marshal error body: %w", err)
 		}
-		newHeaders = []internalapi.Header{
-			{pathHeaderName, t.path},
-			{contentTypeHeaderName, jsonContentType},
-		}
+		newHeaders = append(newHeaders,
+			internalapi.Header{contentTypeHeaderName, jsonContentType},
+			internalapi.Header{contentLengthHeaderName, strconv.Itoa(len(newBody))},
+		)
 	}
 	return
 }
--- a/internal/translator/openai_awsbedrock.go
+++ b/internal/translator/openai_awsbedrock.go
@@ -562,7 +562,7 @@ func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) ResponseError(respHeaders
 ) {
 	statusCode := respHeaders[statusHeaderName]
 	var openaiError openai.Error
-	if v, ok := respHeaders[contentTypeHeaderName]; ok && v == jsonContentType {
+	if v, ok := respHeaders[contentTypeHeaderName]; ok && strings.Contains(v, jsonContentType) {
 		var bedrockError awsbedrock.BedrockException
 		if err = json.NewDecoder(body).Decode(&bedrockError); err != nil {
 			return nil, nil, fmt.Errorf("failed to unmarshal error body: %w", err)
--- a/internal/translator/openai_embeddings.go
+++ b/internal/translator/openai_embeddings.go
@@ -11,6 +11,7 @@ import (
 	"io"
 	"path"
 	"strconv"
+	"strings"
 
 	"github.com/tidwall/sjson"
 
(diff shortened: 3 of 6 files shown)


manifests/charts/ai-gateway-helm/values.yaml
```diff
@@ -196,12 +196,18 @@ controller:
   # resources (configuration) is big. Defaults to 4MB.
   # maxRecvMsgSize: "4194304"
 
-  # mcpSessionEncryptionSeed is an arbitrary string seed used to derive the MCP session encryption key.
-  #	Do not include commas as they are used as separators. You can optionally pass "fallback" seed after the first one to allow for key rotation.
-  #	For example: "new-seed,old-seed-for-fallback". The fallback seed is only used for decryption.
-  #
-  # This value should be set to a secure random string in production environments instead of the default below.
-  mcpSessionEncryptionSeed: "default-insecure-seed"
```
nacx, line 204: This could be a breaking change, but we're in version `0.4` and the current Deployment template was not reading this value (!) anyway, so I think it is OK to just "move" this setting.

internal/mcpproxy/crypto_test.go
```diff
@@ -6,78 +6,145 @@
...
-func TestEncryptionIsSalted(t *testing.T) {
-	sc := DefaultSessionCrypto("test", "")
-
-	enc1, err := sc.Encrypt("plaintext")
-	require.NoError(t, err)
-	enc2, err := sc.Encrypt("plaintext")
-	require.NoError(t, err)
-
-	require.NotEqual(t, enc1, enc2)
+func TestEncryptSession(t *testing.T) {
```
nacx, line 22: The tests in this file haven't changed but I've refactored them a bit to make them trivial to use with additional crypto implementations, if find convenient to have other impls.

internal/mcpproxy/crypto.go
```diff
@@ -73,24 +73,12 @@ func (p pbkdf2AesGcm) Encrypt(plaintext string) (string, error) {
 		return "", err
 	}
 
-	block, err := aes.NewCipher(key)
```
nacx, line 76: No real changes to the code from this point. Just a small refactoring to make the AES encrypt/decrypt method reusable, should we want to explore having other crypto impls.

manifests/charts/ai-gateway-helm/values.yaml
```diff
@@ -196,12 +196,18 @@ controller:
...
+    # sessionEncryptionSeed is an arbitrary string seed used to derive the MCP session encryption key.
+    #	Do not include commas as they are used as separators. You can optionally pass "fallback" seed after the first one to allow for key rotation.
+    #	For example: "new-seed,old-seed-for-fallback". The fallback seed is only used for decryption.
+    #
+    # This value should be set to a secure random string in production environments instead of the default below.
+    sessionEncryptionSeed: "default-insecure-seed"
+    # Number of PBKDF2 iterations to use for deriving the MCP session encryption key.
+    # A higher number increases security but also increases CPU usage and time consumed during key derivation.
+    # The default value of 100,000 is a reasonable balance for most production environments.
+    sessionEncryptionIterations: 100000
```
mathetake: should this also be fallback-able? when a user wants to change, currently no downtime migration is not possible. What do you think about 

```
mcp:
  encryption:
  default:
    seed: "xyz"
    iterations: 10
  fallback:
    seed: "default-insecure-seed""
    iterations: 100000 
```

? Putting `100000` in the fallback allows us to put the faster version in the default encryption (which i think is what we want?)

nacx: Fallback is mostly for rotation. not to fallback to "slow", right? I agree on adding the fallback settings as well (I'll do) for the rotation use case.

mathetake: "fallback" is a bad word but this is 100% about rotation. The configuration in the "fallback" is only used in the decryption to be able to decode the session-id encoded with a previous setting (that's the current logic for seed fallback)

mathetake (approved):

PR #1599, opened 2025-12-01
refactor(tracing): rearrange type parameters & avoid nested generic type parameter 

mathetake:
**Description**

This stops the nested generics type embedding and switches to have explicit type parameters for generic tracing interfaces.

Follow up on #1582


internal/tracing/api/api.go
```diff
@@ -38,7 +38,7 @@ type (
 		Shutdown(context.Context) error
 	}
 	// RequestTracer standardizes tracer implementations for non-MCP requests.
-	RequestTracer[ReqT any, SpanT any] interface {
+	RequestTracer[ReqT any, RespT any, RespChunkT any] interface {
```
mathetake, line 41: this is the gist of the change. SpanT was generic too hence nested, which resulted in cryptic type declaration. So this switches to explicitly pass only the necessary concrete type params

PR #1594, opened 2025-12-01
feat: add Realtime Client Secrets API support

tanthcstt:
**Description**

This PR adds support for the OpenAI-compatible `/v1/realtime/client_secrets` endpoint with translation support for GCP Vertex AI backends. This endpoint enables WebSocket authentication token generation for realtime API connections.

**Provider Support**:

| Provider | Endpoint | Description |
|----------|----------|-------------|
| OpenAI | `/v1/realtime/sessions` | Pass-through to native endpoint |
| GCP Vertex AI | `/v1alpha/auth_tokens` | Translated to GCP auth token format |

**Related Issues/PRs (if applicable)**
#1566

PR #1429, opened 2025-10-24
feat: initial implementation of openai responses api

sivanantha321:
**Description**

**Related Issues/PRs (if applicable)**

**Special notes for reviewers (if applicable)**

PR #1600, opened 2025-12-01
refactor: consolidates all processors with one generic type

### LINKED ISSUE #1083: Refactor processors code to share common logic

mathetake:
**Description**
This consolidates all the copy-pasted processors that existed per endpoint we support into one generic processor. This was made possible thanks to the series of refactoring that we landed in the past few weeks primarily for dynamic module work #90. 

Notably, now in order to add an endpoint, majority of the new code will be in translator (where also have shared generic interface) as well as the type definition. No longer it requires the huge copy paste of processors.

Resolves #1083 
Blocker for #1429 #1584 #1592 #1594

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -28,7 +28,7 @@ require (
 	github.com/google/go-cmp v0.7.0
 	github.com/google/jsonschema-go v0.3.0
 	github.com/google/uuid v1.6.0
-	github.com/modelcontextprotocol/go-sdk v1.0.0
+	github.com/modelcontextprotocol/go-sdk v1.1.0
 	github.com/openai/openai-go v1.12.0
 	github.com/openai/openai-go/v2 v2.7.1
 	github.com/prometheus/client_golang v1.23.2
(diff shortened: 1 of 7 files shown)


mathetake:
wow cool all tests passing

mathetake:
Tomorrow i will backfill the tests and then good to go

internal/extproc/processor_impl.go
mathetake, line 1: this is where I generalized the ChatCompletionsProcessor with a few type params, almost all the logic is the same

internal/endpointspec/endpointspec.go
mathetake, line 1: all new endpoints will only touch this file by declaring one type implementing one small interface and that's it

mathetake:
Cool i came up with some cool trick while eating dinner and now we no longer need to touch internal/extproc AT ALL for adding endpoints!!!!!!!

internal/extproc/processor_impl.go
```diff
@@ -77,9 +43,10 @@ func RerankProcessorFactory(f metrics.Factory, tracer tracing.RerankTracer) Proc
 //
 // Returns:
 // * ProcessorFactory: A factory function to create processors based on the configuration.
-func newFactory[ReqT any, RespT any, RespChunkT any, EndpointSpecT endpointspec.Spec[ReqT, RespT, RespChunkT]](
+func NewFactory[ReqT any, RespT any, RespChunkT any, EndpointSpecT endpointspec.Spec[ReqT, RespT, RespChunkT]](
 	f metrics.Factory,
 	tracer tracing.RequestTracer[ReqT, RespT, RespChunkT],
+	_ EndpointSpecT, // This is a type marker to bind EndpointSpecT without specifying ReqT, RespT, RespChunkT explicitly.
```
mathetake, line 49: This is so cool

PR #1607, opened 2025-12-03
fix: add reasoning content in request and response for gcp anthropic

hustxiayang:
**Description**
This PR is to fix the following issues:
1 Add reasoning content in the request is also missing in the gcp anthropic
2 The reasoning output of gcp anthropic is not parsed out

In this way, reasoning claude models can have an unified interface.

Other issues:
1 The assistant message of gcp anthropic did not cover the case of array. Fixed.


internal/apischema/openai/openai.go
```diff
@@ -515,8 +516,8 @@ type ChatCompletionAssistantMessageParamContent struct {
 	Text *string `json:"text,omitempty"`
 
 	// The signature for a thinking block.
-	Signature               *string `json:"signature,omitempty"`
-	RedactedContent         []byte  `json:"redactedContent,omitempty"`
+	Signature               *string               `json:"signature,omitempty"`
```
yuzisun, line 519: Extra spaces ?

internal/apischema/openai/openai.go
```diff
@@ -1583,6 +1584,43 @@ func (e EmbeddingUnion) MarshalJSON() ([]byte, error) {
...
+	Value any
+}
+
+// UnmarshalJSON implements json.Unmarshaler to handle both []byte and string formats.
+func (r *RedactedContentUnion) UnmarshalJSON(data []byte) error {
+	// Try to unmarshal as []byte first (base64 encoded).
+	var str string
+	if err := json.Unmarshal(data, &str); err == nil {
+		// Try to decode as base64 first (this would be []byte encoded as base64)
+		if decoded, err := base64.StdEncoding.DecodeString(str); err == nil {
```
yuzisun, line 1604: Why do we need to decode in gateway ?

internal/apischema/openai/openai.go
hustxiayang: Hi, this is formatted by "make precommit"

internal/apischema/openai/openai.go
hustxiayang: This is to check whether it's `[]byte`, because in awsbedrock, the type is  `[]byte` but in gcp anthropic, the type is string

PR #1610, opened 2025-12-03
controller: removes all panics from extension server

mathetake:
**Description**

This removes all panics from extension server code paths, mostly for making it resilient against any bad input from EG

### DIFF
--- a/internal/extensionserver/extensionserver_test.go
+++ b/internal/extensionserver/extensionserver_test.go
@@ -44,11 +44,9 @@ import (
 )
 
 // mustToAny marshals the provided message to an Any message.
-func mustToAny(msg proto.Message) *anypb.Any {
+func mustToAny(t *testing.T, msg proto.Message) *anypb.Any {
 	b, err := proto.Marshal(msg)
-	if err != nil {
-		panic(fmt.Sprintf("BUG: failed to marshal message: %v", err))
-	}
+	require.NoError(t, err)
 	const envoyAPIPrefix = "type.googleapis.com/"
 	return &anypb.Any{
 		TypeUrl: envoyAPIPrefix + string(msg.ProtoReflect().Descriptor().FullName()),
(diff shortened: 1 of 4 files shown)


mathetake:
@nacx this is the one i think i forgot to include when upstreaming MCP...

internal/extensionserver/extensionserver_test.go
```diff
@@ -42,6 +43,19 @@ import (
 	"github.com/envoyproxy/ai-gateway/internal/internalapi"
 )
 
+// mustToAny marshals the provided message to an Any message.
+func mustToAny(msg proto.Message) *anypb.Any {
```
nacx: WDYT about adding the `testing.T` as a parameter to make sure we're never tempted to copy this method in non-test code? Also a comment saying we shouldn't use this in proper code would be good to have.
```suggestion
func mustToAny(t *testing.T, msg proto.Message) *anypb.Any {
```

internal/extensionserver/inferencepool.go
```diff
@@ -184,25 +183,32 @@ func buildEPPMetadata(metadata *corev3.Metadata, inferencePool *gwaiev1.Inferenc
...
 // for each InferencePool's endpoint picker service.
-func buildClustersForInferencePoolEndpointPickers(clusters []*clusterv3.Cluster) []*clusterv3.Cluster {
+func buildClustersForInferencePoolEndpointPickers(clusters []*clusterv3.Cluster) ([]*clusterv3.Cluster, error) {
 	result := make([]*clusterv3.Cluster, 0, len(clusters))
 	for _, cluster := range clusters {
 		if pool := getInferencePoolByMetadata(cluster.Metadata); pool != nil {
-			result = append(result, buildExtProcClusterForInferencePoolEndpointPicker(pool))
+			c, err := buildExtProcClusterForInferencePoolEndpointPicker(pool)
+			if err != nil {
+				continue
```
nacx: We should probably log the errors when this happens.

nacx (approved):
Overall LGTM. Just a couple of comments about making sure the methods that panic are not used in normal code, and about making sure we log errors where we panicked before, if we're not returning the errors.

internal/extensionserver/inferencepool.go
nacx: Here, and also in other methods that now return an error, if the code continues, we need to make sure we log the error at least.

PR #1611, opened 2025-12-03
feat: initial tracing for Anthropic /messages API

### LINKED ISSUE #1389: Implement tracing for Anthropic /messages

mathetake:
**Description**

This adds the initial implementation of distributed tracing for Anthropic's /messages endpoint. To do that, this defines the concrete API schema instead of generic map[string]any that had been used previously.

* Redo of #1464 which was reverted in #1516
* Closes #1389 

___

Note: The implementation is passing all tests as well as the manual integration test with Claude Code locally. The TODO is that i haven't implemented all the input content block param type, so that and VCR tests will be covered in the followup.

### DIFF
--- a/internal/tracing/openinference/anthropic/messages.go
+++ b/internal/tracing/openinference/anthropic/messages.go
@@ -9,12 +9,12 @@ import (
 	"encoding/json"
 	"fmt"
 
-	"github.com/envoyproxy/ai-gateway/internal/metrics"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/codes"
 	"go.opentelemetry.io/otel/trace"
 
 	"github.com/envoyproxy/ai-gateway/internal/apischema/anthropic"
+	"github.com/envoyproxy/ai-gateway/internal/metrics"
 	tracing "github.com/envoyproxy/ai-gateway/internal/tracing/api"
 	"github.com/envoyproxy/ai-gateway/internal/tracing/openinference"
 )
--- a/internal/translator/anthropic_usage_test.go
+++ b/internal/translator/anthropic_usage_test.go
@@ -9,8 +9,9 @@ import (
 	"testing"
 
 	"github.com/anthropics/anthropic-sdk-go"
-	"github.com/envoyproxy/ai-gateway/internal/metrics"
 	"github.com/stretchr/testify/assert"
+
+	"github.com/envoyproxy/ai-gateway/internal/metrics"
 )
 
 func TestExtractLLMTokenUsage(t *testing.T) {


mathetake:
sorry not ready yet

mathetake:
i think i will go ahead and implement the tracing

mathetake:
ok tracing working
image

mathetake:
ok verified that it's working fine with Claude Code too!

mathetake:
image

internal/apischema/anthropic/anthropic.go
```diff
@@ -3,46 +3,579 @@
...
+// as well as for observability purposes on a best-effort basis.
+//
+// Notably, round trip idempotency is not guaranteed when using this struct.
+type MessagesRequest struct {
+	// Model is the model to use for the request.
+	Model string `json:"model,omitempty"`
+
+	// Messages is the list of messages in the conversation.
+	// https://docs.claude.com/en/api/messages#body-messages
+	Messages []MessageParam `json:"messages"`
```
aabchoo, line 31: Why is `MessageParam` the only non-omit empty field?

Message is an already existing class from MessageParam. Why did you decide to name this field Messages?

internal/apischema/anthropic/anthropic.go
```diff
@@ -3,46 +3,579 @@
...
+
+const (
+	MessageRoleUser      MessageRole = "user"
+	MessageRoleAssistant MessageRole = "assistant"
+)
+
+// MessageContent represents the content of a message in the Anthropic Messages API.
+// https://docs.claude.com/en/api/messages#body-messages-content
+type MessageContent struct {
+	Text  string              // Non-empty iif this is not array content.
```
aabchoo:
```suggestion
	Text  string              // Non-empty if this is not array content.
```

internal/apischema/anthropic/anthropic.go
```diff
@@ -3,46 +3,579 @@
...
+	Array []ContentBlockParam // Non-empty iif this is array content.
```
aabchoo:
```suggestion
	Array []ContentBlockParam // Non-empty if this is array content.
```

internal/apischema/anthropic/anthropic.go
```diff
@@ -3,46 +3,579 @@
...
-	if model, ok := m["model"].(string); ok {
-		return model
+	// Try to unmarshal as array of MessageContentArrayElement.
+	var array []ContentBlockParam
+	if err := json.Unmarshal(data, &array); err == nil {
+		m.Array = array
+		return nil
 	}
-	return ""
+	return fmt.Errorf("message content must be either string or array")
```
aabchoo: Keep consistent with line 140. Either `text` or `string` but not both

internal/apischema/anthropic/anthropic.go
```diff
@@ -3,46 +3,579 @@
...
+	case "text":
+		var textBlock TextBlockParam
+		if err := json.Unmarshal(data, &textBlock); err != nil {
+			return fmt.Errorf("failed to unmarshal text block: %w", err)
+		}
+		m.Text = &textBlock
+		return nil
+	default:
+		// TODO add others when we need it for observability, etc.
+		// Fow now, we ignore undefined types.
```
aabchoo, line 174: Why not raise an error for unimplemented type?

internal/apischema/anthropic/anthropic.go
```diff
@@ -3,46 +3,579 @@
...
+	switch typ.String() {
+	case "custom":
+		var tool Tool
+		if err := json.Unmarshal(data, &tool); err != nil {
+			return fmt.Errorf("failed to unmarshal tool: %w", err)
+		}
+		t.Tool = &tool
+		return nil
+	default:
+		// TODO add others when we need it for observability, etc.
```
aabchoo, line 244: similar question to above

internal/apischema/anthropic/anthropic.go
```diff
@@ -3,46 +3,579 @@
...
+		// Fow now, we ignore undefined types.
+		return nil
+	}
+}
+
```
aabchoo, line 249: Missing marshal func?

internal/apischema/anthropic/anthropic.go
```diff
@@ -3,46 +3,579 @@
...
+	OutputTokens float64 `json:"output_tokens"`
+}
+
+// MessagesStreamEvent represents a single event in the streaming response from the Anthropic Messages API.
+// https://docs.claude.com/en/docs/build-with-claude/streaming
+type MessagesStreamEvent struct {
+	// The type of the streaming event.
+	Type MessagesStreamEventType `json:"type"`
+	// MessageStart is present if the event type is "message_start" or "message_delta".
+	MessageStart *MessagesStreamEventMessageStart
```
aabchoo: missing json?

aabchoo (commented):
will continue to reviewing the rest

internal/apischema/anthropic/anthropic.go
```diff
@@ -3,46 +3,579 @@
...
+	// The number of input tokens read from the cache.
+	CacheReadInputTokens float64 `json:"cache_read_input_tokens"`
+	// The number of input tokens which were used.
+	InputTokens float64 `json:"input_tokens"`
+	// The number of output tokens which were used.
+	OutputTokens float64 `json:"output_tokens"`
+}
+
+// MessagesStreamEvent represents a single event in the streaming response from the Anthropic Messages API.
+// https://docs.claude.com/en/docs/build-with-claude/streaming
```
aabchoo, line 449: Reason for calling it events instead of chunk? In tracing it's considered a chunk.

internal/tracing/openinference/anthropic/messages.go
```diff
@@ -0,0 +1,304 @@
...
+	Messages []anthropic.MessageParam `json:"messages,omitempty"`
+	Tools    []anthropic.Tool         `json:"tools,omitempty"`
+}
+
+// buildRequestAttributes builds OpenInference attributes from the request.
+func buildRequestAttributes(req *anthropic.MessagesRequest, body string, config *openinference.TraceConfig) []attribute.KeyValue {
+	attrs := []attribute.KeyValue{
+		attribute.String(openinference.SpanKind, openinference.SpanKindLLM),
+		attribute.String(openinference.LLMSystem, openinference.LLMSystemAnthropic),
+		attribute.String(openinference.LLMModelName, req.Model),
```
aabchoo, line 110: can req be nil? we should check if req + config is nil before using it

internal/tracing/openinference/anthropic/messages.go
```diff
@@ -0,0 +1,304 @@
...
+					maybeRedacted = openinference.RedactedValue
+				}
+				attrs = append(attrs, attribute.String(openinference.InputMessageAttribute(i, openinference.MessageContent), maybeRedacted))
+			case content.Array != nil:
+				for j, param := range content.Array {
+					switch {
+					case param.Text != nil:
+						text := param.Text.Text
+						if config.HideInputText {
+							text = openinference.RedactedValue
```
aabchoo: maybeRedacted instead of text to be consistent

internal/tracing/openinference/anthropic/messages.go
```diff
@@ -0,0 +1,304 @@
...
+	role := resp.Role
+	if !config.HideOutputs && !config.HideOutputMessages {
+		for i, content := range resp.Content {
+			attrs = append(attrs, attribute.String(openinference.OutputMessageAttribute(i, openinference.MessageRole), string(role)))
+
+			switch {
+			case content.Text != nil:
+				txt := content.Text.Text
+				if config.HideOutputText {
+					txt = openinference.RedactedValue
```
aabchoo, line 189: same as above

internal/tracing/openinference/anthropic/messages.go
```diff
@@ -0,0 +1,304 @@
...
+					attrs = append(attrs,
+						attribute.String(openinference.OutputMessageToolCallAttribute(i, 0, openinference.ToolCallFunctionArguments), string(inputStr)),
+					)
+				}
+			}
+		}
+	}
+
+	// Token counts are considered metadata and are still included even when output content is hidden.
+	u := resp.Usage
```
aabchoo, line 209: Use extractTokenUsageFromAnthropic?

internal/tracing/span.go
```diff
@@ -80,6 +81,7 @@ type (
 	embeddingsSpan      = responseSpan[openai.EmbeddingResponse, struct{}]
 	imageGenerationSpan = responseSpan[openaisdk.ImagesResponse, struct{}]
 	rerankSpan          = responseSpan[cohereschema.RerankV2Response, struct{}]
+	messageSpan         = responseSpan[anthropicschema.MessagesResponse, anthropicschema.MessagesStreamEvent]
 )
```
aabchoo: Do you know why the below sets up vars for spans below? I assume you need to add the anthropic message span there as well

internal/translator/anthropic_gcpanthropic.go
```diff
@@ -33,42 +34,31 @@ type anthropicToGCPAnthropicTranslator struct {
 
 // RequestBody implements [AnthropicMessagesTranslator.RequestBody] for Anthropic to GCP Anthropic translation.
 // This handles the transformation from native Anthropic format to GCP Anthropic format.
-func (a *anthropicToGCPAnthropicTranslator) RequestBody(_ []byte, body *anthropicschema.MessagesRequest, _ bool) (
+func (a *anthropicToGCPAnthropicTranslator) RequestBody(raw []byte, req *anthropicschema.MessagesRequest, _ bool) (
 	newHeaders []internalapi.Header, newBody []byte, err error,
 ) {
 	// Extract model name for GCP endpoint from the parsed request.
```
aabchoo: move comment below line 41

internal/translator/anthropic_gcpanthropic.go
```diff
@@ -33,42 +34,31 @@ type anthropicToGCPAnthropicTranslator struct {
...
 
 	// Remove the model field since GCP doesn't want it in the body.
-	delete(anthropicReq, "model")
+	// Note: Do not operate on raw here, as that would mutate the original request body.
+	// Hence, we do the SetBytesOptions above to create mutatedBody first.
+	newBody, _ = sjson.DeleteBytes(mutatedBody, "model")
 
 	// Add GCP-specific anthropic_version field (required by GCP Vertex AI).
 	// Uses backend config version (e.g., "vertex-2023-10-16" for GCP Vertex AI).
 	if a.apiVersion == "" {
```
aabchoo: Should move this up before getting the mutated body? Unnecessary to do that logic if we are going to return an error

aabchoo (commented):
last bit of comments -- thanks for all the hard work!

internal/apischema/anthropic/anthropic.go
mathetake: fixed the omit empty stuff to align the official doc.

as for the field name, it aligned with the official doc where the field name is messages where the type is named MessageParam

image

internal/apischema/anthropic/anthropic.go
mathetake: so i removed the "type" above. The thing is that we don't need to have json tag here since we have a custom Unmarshal function to handle the union

internal/apischema/anthropic/anthropic.go
mathetake: so the thing is that there's bunch of fields we are not interested in. If we raise the error, then these requests will fail and you'll never know, say, what kind of request condent Claude Code will send. Note that this is only used for observability, not translation.

internal/apischema/anthropic/anthropic.go
mathetake: Marshal is only needed for some existing tests, and not used in the main code path because this is only used for observability atm. and this toolunion is not used in the tests, so i omitted marshal

internal/apischema/anthropic/anthropic.go
mathetake: yeah good catch. renamed to chunk

internal/tracing/openinference/anthropic/messages.go
mathetake: req will never be nil

internal/apischema/anthropic/anthropic.go
aabchoo: Sounds good to me. Btw I couldn't find that when looking at the link above https://docs.claude.com/en/api/messages#body-messages

image

internal/tracing/span.go
mathetake: will be cleaned up in https://github.com/envoyproxy/ai-gateway/pull/1622

internal/tracing/openinference/anthropic/messages.go
mathetake: good point -- done

mathetake:
@aabchoo thanks for the look!!!!!!!!!!

aabchoo (approved):

PR #1614, opened 2025-12-04
mcp: fix mCP to MCP in the struct

Hritik003:
**Description**

fix `mCP` naming to `MCP` in the MCPRoute struct


Hritik003:
This is something, which is generated by the client-gen itself and ehe file should be mCPRoutes (lowercase) because it's an unexported/private implementation type. Hence closing the PR ;-;

PR #1618, opened 2025-12-05
feat: indexing for referencegrant

### LINKED ISSUE #1375: Add Indexing Support for reference grant resource
Currently all resources in ai gateway has indexing support, need to add this support to referencegrant resource as well. This will be taken up as a followup to : 
https://github.com/envoyproxy/ai-gateway/pull/1361

siddharth1036:
**Description**

This commit adds indexing for referencegrants. The indexing key uses To Kind + namespace in referencegrant for indexing which would help fetch only related referencegrants given aiservicebackend and namespace.

Closes #1375


internal/controller/controller.go
```diff
@@ -408,6 +419,20 @@ func getSecretNameAndNamespace(secretRef *gwapiv1.SecretObjectReference, namespa
 	return fmt.Sprintf("%s.%s", secretRef.Name, namespace)
 }
 
+func getReferenceGrantIndexKey(namespace, kind string) string {
+	return fmt.Sprintf("%s/%s", namespace, kind)
```
mathetake: can you use the same format as others (name.namespace)

PR #1621, opened 2025-12-05
fix: return user understandable response for request parse and validation

hustxiayang:
**Description**
Improve error handling by distinguishing between internal errors and user-facing errors, returning appropriate HTTP status codes and safe error messages to clients.

  Key Changes:                                                                                                                                                                    
  1. New user_facing_errors.go with sentinel errors                                                                                                                               
  2. Updated all translators to wrap errors appropriately                                                                                                                         
  3. Modified request processor to detect and return user-facing errors                                                                                                           
  4. Comprehensive test coverage across the stack


hustxiayang:
@mathetake could you please tell me why it's `internal error`? Suppose the user's input is not valid, and can not be unmarshaled? Why we can not return the error to users? Or you just mean I should change the error message and not to include the error?

yuzisun:
> @mathetake could you please tell me why it's `internal error`? Suppose the user's input is not valid, and can not be unmarshaled? Why we can not return the error to users? Or you just mean I should change the error message?

@hustxiayang we need to separate out the validation errors, process body can return other internal gateway errors which should not return immediate response.

hustxiayang:
@mathetake @yuzisun Thanks a lot for your comments! This is only to return immediate response when the request is not parsable or not able to be processed. I reverted the changes in https://github.com/envoyproxy/ai-gateway/blob/932c4e2ebf444d0d32de76821ee91ae386b123ae/internal/extproc/processor_impl.go#L282, and make this one still return nil as this is not real auth. For other returns, they are still nil as in the description.

internal/internalapi/user_facing_errors.go
```diff
@@ -0,0 +1,65 @@
...
+	// by the target API (e.g., image formats, content types).
+	ErrUnsupportedMessageFormat = errors.New("message format not supported by target API")
+
+	// ErrInvalidParameterValue indicates a parameter value is outside the valid range
+	// for the target API (e.g., temperature, top_p).
+	ErrInvalidParameterValue = errors.New("parameter value out of valid range for target API")
+
+	// ErrUnsupportedFeature indicates the request uses a feature not supported by the target API
+	// (e.g., streaming, function calling, specific response formats).
+	ErrUnsupportedFeature = errors.New("requested feature not supported by target API")
```
mathetake: Let's not define every single one of the error mode like this, which means exposing the implementation details out of the translator. Leave only one or two error like `ErrorInvalidRequest`, etc only return things like `fmt.Errorf("%w: %s", ErrorInvalidRequest, "details like foo field is missing")`, and delete any other error types. Otherwise this GetUserFacingError etc will explode indefinitely and making the callsite unnecessarily check all errors sequentially where they don't need to understand the error detail (all the switch statement)
```suggestion
```

hustxiayang: @mathetake Thanks a lot! Would update it

hustxiayang: @mathetake just for clarification: I think I should delete any other error types and remove all switch statements for sure, but we still want to return to our users some details, like "invalid image data URI" rather just simply ""invalid request body", is that correct?

mathetake: yes. wraping the defined Error type with the detail with fmt.Errorf as i suggested would suffice your goal

hustxiayang: Thanks a lot for the confirmation!

internal/endpointspec/endpointspec.go
```diff
@@ -84,7 +84,8 @@ func (ChatCompletionsEndpointSpec) ParseBody(
 ) (internalapi.OriginalModel, *openai.ChatCompletionRequest, bool, []byte, error) {
 	var req openai.ChatCompletionRequest
 	if err := json.Unmarshal(body, &req); err != nil {
-		return "", nil, false, nil, fmt.Errorf("failed to unmarshal chat completion request: %w", err)
+		// Return safe error type with details about the parsing failure
```
mathetake: could you remove the unnecessary comment like this here and elsewhere too

hustxiayang:
@mathetake I marked it as ready for review, do you mind to review it again? Thanks!

internal/endpointspec/endpointspec.go
hustxiayang: removed! thanks a lot!

hustxiayang:
@mathetake sure, would add. Thanks a lot!

yuzisun:
@mathetake can you help take a look again, added the upstream test.

yuzisun:
@mathetake can you help approve ?

mathetake (approved):
Sorry for the late response!

PR #1623, opened 2025-12-05
feat: introduce GatewayConfig CRD for gateway-scoped configuration

sqaisar:
**Description**

Introduce GatewayConfig CRD for gateway-scoped configuration of the external processor. Adds controller lifecycle handling (including finalizer management), deprecates AIGatewayRoute-level resource configuration in favor of `GatewayConfig`, and updates docs with migration guidance and new options.

### Example

```yaml
kind: GatewayConfig
metadata:
  name: my-gateway-config
  namespace: default
spec:
  extProc:
   kubernetes:
      env:
        - name: OTEL_EXPORTER_OTLP_ENDPOINT
          value: "http://otel-collector:4317"
      resources:
        requests:
          cpu: "100m"
          memory: "128Mi"
        limits:
          cpu: "500m"
          memory: "512Mi"
---
kind: Gateway
metadata:
  name: my-gateway
  namespace: default
  annotations:
    aigateway.envoyproxy.io/gateway-config: my-gateway-config
spec:
  gatewayClassName: envoy-gateway
  listeners:
    - name: http
      protocol: HTTP
      port: 8080
```

Fixes [1]

Docs, controller, and CRD changes included; route-level resources are deprecated—see migration guidance in docs.

1: https://github.com/envoyproxy/ai-gateway/issues/1552


internal/controller/gateway_config.go
```diff
@@ -0,0 +1,228 @@
...
+// syncGatewayConfig is the main logic for reconciling the GatewayConfig resource.
+func (c *GatewayConfigController) syncGatewayConfig(ctx context.Context, gatewayConfig *aigv1a1.GatewayConfig) error {
+	// Find all Gateways that reference this GatewayConfig.
+	referencingGateways, err := c.findReferencingGateways(ctx, gatewayConfig)
+	if err != nil {
+		return fmt.Errorf("failed to find referencing Gateways: %w", err)
+	}
+
+	// Handle finalizer based on whether any Gateways reference this GatewayConfig.
+	if gatewayConfig.DeletionTimestamp != nil {
```
mathetake: i think we can remove the finalizer for now unlike others, this one doesn't create sub resources that needs to be GCed. Let's simplifies the impl for now

internal/controller/gateway_config.go
```diff
@@ -0,0 +1,228 @@
...
+		c.gatewayEventChan <- event.GenericEvent{Object: gw}
+	}
+
+	return nil
+}
+
+// findReferencingGateways finds all Gateways in the same namespace that reference this GatewayConfig.
+func (c *GatewayConfigController) findReferencingGateways(ctx context.Context, gatewayConfig *aigv1a1.GatewayConfig) ([]*gwapiv1.Gateway, error) {
+	var gateways gwapiv1.GatewayList
+	if err := c.client.List(ctx, &gateways, client.InNamespace(gatewayConfig.Namespace)); err != nil {
```
mathetake: can you create an index like other places instead of looking up all gateways?

internal/controller/gateway_config.go
```diff
@@ -0,0 +1,228 @@
...
+			referencingGateways = append(referencingGateways, gw)
+		}
+	}
+
+	return referencingGateways, nil
+}
+
+// MapGatewayToGatewayConfig is a handler function that maps Gateway events to GatewayConfig reconcile requests.
+// This is used by the controller builder to watch Gateway resources.
+func (c *GatewayConfigController) MapGatewayToGatewayConfig(_ context.Context, obj client.Object) []reconcile.Request {
```
mathetake: it looks we don't need this one as it results in infinite loop. When Gateway is reconciled, it will trigger the pod mutation afterwards where the config will be read. I believe it's enough for our purpose

site/docs/capabilities/gateway-config.md
mathetake, line 1: this is not only for observability so shall we put this one level above

internal/controller/gateway_mutator.go
```diff
@@ -389,3 +393,73 @@ func (g *gatewayMutator) mutatePod(ctx context.Context, pod *corev1.Pod, gateway
...
 	return nil
 }
+
+// fetchGatewayAndConfig returns the Gateway and the referenced GatewayConfig (if present).
+func (g *gatewayMutator) fetchGatewayAndConfig(ctx context.Context, gatewayName, gatewayNamespace string) (*gwapiv1.Gateway, *aigv1a1.GatewayConfig) {
+	// Fetch the Gateway object.
+	var gateway gwapiv1.Gateway
+	if err := g.c.Get(ctx, client.ObjectKey{Name: gatewayName, Namespace: gatewayNamespace}, &gateway); err != nil {
+		if apierrors.IsNotFound(err) {
+			g.logger.Info("Gateway not found, using global defaults",
```
mathetake: I read like "default of what?" so let's be explicit about it
```suggestion
			g.logger.Info("Gateway not found, using global default configuration",
```

internal/controller/gateway_mutator.go
```diff
@@ -389,3 +393,73 @@ func (g *gatewayMutator) mutatePod(ctx context.Context, pod *corev1.Pod, gateway
...
+				"gateway_name", gatewayName, "gateway_namespace", gatewayNamespace)
+		} else {
+			g.logger.Error(err, "failed to get Gateway, using global defaults",
```
mathetake:
```suggestion
			g.logger.Error(err, "failed to get Gateway, using global default configuration",
```

internal/controller/gateway_mutator.go
```diff
@@ -389,3 +393,73 @@ func (g *gatewayMutator) mutatePod(ctx context.Context, pod *corev1.Pod, gateway
...
+				"gateway_name", gatewayName, "gateway_namespace", gatewayNamespace)
+		}
+		return nil, nil
+	}
+
+	// Check for GatewayConfig annotation.
+	if gateway.Annotations == nil {
+		return &gateway, nil
+	}
```
mathetake: Read on nil map is safe so it should be fine to delete
```suggestion
```

internal/controller/gateway_mutator.go
```diff
@@ -389,3 +393,73 @@ func (g *gatewayMutator) mutatePod(ctx context.Context, pod *corev1.Pod, gateway
 	}
 	return nil
 }
+
+// fetchGatewayAndConfig returns the Gateway and the referenced GatewayConfig (if present).
+func (g *gatewayMutator) fetchGatewayAndConfig(ctx context.Context, gatewayName, gatewayNamespace string) (*gwapiv1.Gateway, *aigv1a1.GatewayConfig) {
```
mathetake: The Gateway is not used atm so let's simplify 
```suggestion
func (g *gatewayMutator) fetchGatewayConfig(ctx context.Context, gatewayName, gatewayNamespace string) (*aigv1a1.GatewayConfig) {
```

internal/controller/ai_gateway_route.go
```diff
@@ -43,6 +43,12 @@ const (
 	egOwningGatewayNamespaceLabel                      = egAnnotationPrefix + "owning-gateway-namespace"
 	// apiKeyInSecret is the key to store OpenAI API key.
 	apiKeyInSecret = "apiKey"
+	// GatewayConfigAnnotationKey is the annotation key used on Gateway objects to reference a GatewayConfig.
+	// The value should be the name of the GatewayConfig resource in the same namespace as the Gateway.
+	GatewayConfigAnnotationKey = "aigateway.envoyproxy.io/gateway-config"
+	// GatewayConfigFinalizerName is the finalizer added to GatewayConfig resources to prevent deletion
+	// while they are still referenced by Gateway objects.
+	GatewayConfigFinalizerName = "aigateway.envoyproxy.io/gateway-config-protection"
```
mathetake: we don't have anything like this for other resources, so let's remove the finalizer stuff for now, and think about this one later after this PR together withe other resources

api/v1alpha1/gateway_config.go
```diff
@@ -0,0 +1,106 @@
...
+type GatewayConfigExtProc struct {
+	// Env specifies environment variables to be added to the external processor container.
+	// These variables are merged with any global environment variables configured in the
+	// AI Gateway controller. If there are conflicting variable names, the values defined
+	// here take precedence over global values.
+	//
+	// Common use cases include:
+	// - OTEL tracing configuration (e.g., OTEL_EXPORTER_OTLP_HEADERS, OTEL_EXPORTER_OTLP_ENDPOINT)
+	// - Log level overrides
+	// - Feature flags
```
mathetake: These are existent for env var

api/v1alpha1/gateway_config.go
```diff
@@ -0,0 +1,106 @@
...
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []GatewayConfig `json:"items"`
+}
+
+// GatewayConfigSpec defines the configuration for the AI Gateway.
+type GatewayConfigSpec struct {
+	// ExtProc defines the configuration for the external processor container.
+	//
+	// +optional
+	ExtProc *GatewayConfigExtProc `json:"extProc,omitempty"`
```
nacx, line 63: What about using the EnvoyGatway's [KubernetesContainerSpec](https://github.com/envoyproxy/gateway/blob/main/api/v1alpha1/shared_types.go#L211) type instead?
This way we already get all the config fields in the container, and better align with a known API users may be already using to configure the `EnvoyProxy` resource.

internal/controller/gateway_config.go
sqaisar: Sure, then in case user remove a `GatewayConfig` that's referenced in any of the Gateways on it's reconciliation it'll just ignore if that config isn't found.

site/docs/capabilities/gateway-config.md
sqaisar: Make sense, I've moved it up; outside observability

sqaisar:
> Thank you for taking a shot at this!

Thankyou for quickly reviewing the changes. I've resolved all of them. Please take a look.

api/v1alpha1/gateway_config.go
sqaisar: This actually makes more sense to use the KubernetesContainerSpec from EnvoyGatway's . I'll update this. 
And for the above change I'm not sure if I understand this, if this gatewayconfig is only being used in kubernetes or there are plans to create other provider types apart from kubernetes?

sqaisar: I've updated the ExtProc type to use the KubernetesContainerSpec type.

api/v1alpha1/gateway_config.go
```diff
@@ -0,0 +1,79 @@
...
```
mathetake, line 63: as per the suggestion by @nacx, can we have another indirection to be able to handle standalone mode and kubernetes mode like 
```suggestion
	ExtProc *GatewayConfigExtProc `json:"extProc,omitempty"`
}

type GatewayConfigExtProc struct {
	Kubernetes *egv1a1.KubernetesContainerSpec `json:"kubernetes,omitempty"`
}
```

sqaisar: It's done.

sqaisar:
@mathetake  I've updated the type to `GatewayConfig.extProc.kubernetes`
for now all changes looks concluded. Please review.

api/v1alpha1/gateway_config.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+// GatewayConfigExtProc holds runtime-specific configuration for the external processor.
+// The Kubernetes field mirrors Envoy Gateway's KubernetesContainerSpec so we stay aligned
+// with the EnvoyProxy configuration surface and inherit the full set of supported fields.
```
mathetake: i think we can delete these two lines as it's a rationale for the API design vs the comment here is the API documentation for end user

internal/controller/controller.go
```diff
@@ -339,6 +354,18 @@ func mcpRouteToAttachedGatewayIndexFunc(o client.Object) []string {
 	return ret
 }
 
+func gatewayToGatewayConfigIndexFunc(o client.Object) []string {
+	gateway := o.(*gwapiv1.Gateway)
+	if gateway.Annotations == nil {
+		return nil
+	}
```
mathetake: i think we can delete? (reading nil map is safe in Go)

internal/controller/gateway_mutator.go
```diff
@@ -297,25 +311,44 @@ func (g *gatewayMutator) mutatePod(ctx context.Context, pod *corev1.Pod, gateway
...
 
-	// Currently, we have to set the resources for the extproc container at route level.
-	// We choose one of the routes to set the resources for the extproc container.
+	// Prefer GatewayConfig resources; otherwise leave empty.
 	var resources corev1.ResourceRequirements
-	for i := range routes.Items {
-		fc := routes.Items[i].Spec.FilterConfig
-		if fc != nil && fc.ExternalProcessor != nil && fc.ExternalProcessor.Resources != nil {
-			resources = *fc.ExternalProcessor.Resources
-		}
```
mathetake, line 307: we can't delete these lines here until we cut the next version, otherwise this will break the existing cluster.

internal/controller/controller.go
sqaisar: though I'd done that already. 
Updated!

internal/controller/gateway_mutator.go
sqaisar: Done.

sqaisar:
Added more tests to make up for coverage.

internal/controller/gateway_mutator.go
```diff
@@ -297,25 +311,52 @@ func (g *gatewayMutator) mutatePod(ctx context.Context, pod *corev1.Pod, gateway
 		podspec.ImagePullSecrets = append(podspec.ImagePullSecrets, g.extProcImagePullSecrets...)
 	}
 
-	// Currently, we have to set the resources for the extproc container at route level.
-	// We choose one of the routes to set the resources for the extproc container.
+	// TODO: remove after the next release once GatewayConfig adoption is complete.
```
mathetake:
```suggestion
	// TODO: remove after the next release v0.5.
```

sqaisar:
> LGTM! I will wait for other maintainers to take a look until tomorrow otherwise good to go!
> 
> cc @envoyproxy/ai-gateway-maintainers

Thankyou.

Xunzhuo:
why not just use EnvoyProxy CR?

nacx:
> why not just use EnvoyProxy CR

Mostly because the configuration of the extproc container is generated on the fly by EAIGW, and its config can't be preconfigured in the EP resource.

internal/controller/gateway_config.go
```diff
@@ -0,0 +1,155 @@
...
+	// Find all Gateways that reference this GatewayConfig.
+	referencingGateways, err := c.findReferencingGateways(ctx, gatewayConfig)
+	if err != nil {
+		return fmt.Errorf("failed to find referencing Gateways: %w", err)
+	}
+
+	if gatewayConfig.DeletionTimestamp != nil {
+		c.notifyReferencingGateways(gatewayConfig, referencingGateways)
+		return nil
+	}
```
nacx: This seems redundant as it's the same as the code that follows
```suggestion
```

internal/controller/gateway_config_test.go
```diff
@@ -0,0 +1,337 @@
...
+	require.Contains(t, updated.Status.Conditions[0].Message, "failed to find referencing Gateways")
+}
+
+func TestGatewayConfigConditionsNotAccepted(t *testing.T) {
+	conds := gatewayConfigConditions(aigv1a1.ConditionTypeNotAccepted, "nope")
+	require.Len(t, conds, 1)
+	require.Equal(t, aigv1a1.ConditionTypeNotAccepted, conds[0].Type)
+	require.Equal(t, metav1.ConditionFalse, conds[0].Status)
+	require.Equal(t, "nope", conds[0].Message)
+}
```
nacx, line 376: Can you also add tests to verify the behavior when a Gateway references an unexisting gateway config?

internal/controller/gateway_mutator_test.go
```diff
@@ -477,3 +478,152 @@ func TestParseImagePullSecrets(t *testing.T) {
...
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			g := &gatewayMutator{extProcImage: tt.base}
+			require.Equal(t, tt.expected, g.resolveExtProcImage(tt.extProc))
+		})
+	}
+}
```
nacx, line 702: Can you add a test case to the `_mutatePod` test (or into another dedicated test) that verifies the mutated pod when a Gateway config is present for it?

examples/gateway-config/comprehensive.yaml
```diff
@@ -0,0 +1,64 @@
...
+  name: staging-ai-gateway
+  namespace: default
+  annotations:
+    aigateway.envoyproxy.io/gateway-config: comprehensive-config
+spec:
+  gatewayClassName: envoy-gateway
+  listeners:
+    - name: http
+      protocol: HTTP
+      port: 8081
```
nacx, line 78: Thanks for the example!

We aim for having always working examples, and this is one of the reasons we use the examples in our `e2e` tests. Can you add a test to our [e2e otel tracing](https://github.com/envoyproxy/ai-gateway/blob/main/tests/e2e/otel_tracing_test.go) that uses this example `GatewayConfig` ?

sqaisar: makes sense, I've added a test to it with GatewayConfig.

internal/controller/gateway_mutator_test.go
sqaisar: Done.

internal/controller/gateway_config_test.go
sqaisar: Done, Added a dedicated test for non existing gateway config.

sqaisar:
> Thanks @sqaisar! This looks great. I've left a few comments mostly related to testing and I think this will be good to go once done!

I've resolved them. Looks better i think.

internal/controller/gateway_config_test.go
```diff
@@ -0,0 +1,404 @@
...
+	require.Equal(t, metav1.ConditionFalse, updated.Status.Conditions[0].Status)
+	require.Contains(t, updated.Status.Conditions[0].Message, "failed to find referencing Gateways")
+}
+
+func TestGatewayConfigController_GatewayReferencesNonExistingConfig(t *testing.T) {
+	fakeClient := requireNewFakeClientForGatewayConfig(t)
+	eventCh := internaltesting.NewControllerEventChan[*gwapiv1.Gateway]()
+	c := NewGatewayConfigController(fakeClient, ctrl.Log, eventCh.Ch)
+
+	// Create a Gateway that references a GatewayConfig that doesn't exist yet.
```
nacx: This is not what I meant. The thing we're missing in the tests is the behaviour when the referenced config does not exist (and will never exist). For example, users make a typo in the annotation value. What is the expected behaviour then? This is what I think this test needs to validate.

internal/controller/gateway_mutator_test.go
```diff
@@ -252,6 +276,50 @@ func TestGatewayMutator_mutatePod(t *testing.T) {
 						require.NoError(t, err)
 					}
 
+					// Create Gateway and GatewayConfig for GatewayConfig test case
+					if tt.name == "with GatewayConfig" {
```
nacx: Instead of relying on the test name, which makes it harder to evolve these tests, just add a property to the test case: `gatewayConfig *aigv1a1.GatewayConfig`, and if it is set, then create it with the fake client. You can extract this GW config into a package variable to make it easier to reference in the test if most of its information is static.

tests/e2e/otel_tracing_test.go
```diff
@@ -183,3 +183,152 @@ func TestOTELTracingWithConsoleExporter(t *testing.T) {
...
+func TestOTELTracingWithGatewayConfig(t *testing.T) {
+	ctx, cancel := context.WithTimeout(t.Context(), 5*time.Minute)
+	defer cancel()
+
+	// Get the source directory relative to this test file.
+	_, filename, _, ok := runtime.Caller(0)
+	require.True(t, ok, "failed to get source file path")
+	sourceDir := filepath.Dir(filename)
+
+	manifest := filepath.Join(sourceDir, "testdata", "otel_tracing_gateway_config.yaml")
```
nacx: Instead of doing this, delete this custom file and point directly to the file in the `examples` directory. This is what I meant in my previous comment. You already spotted an issue where the example was missing the `kubernetes` keyword, and we want to keep examples always working. Let's do that by using those in the e2e tests, as we do with other examples.

mathetake:
@sqaisar would you mind addressing comments by @nacx? this is almost there to the finish line and we would love to see this land before the next v0.5 release!

sqaisar:
> @sqaisar would you mind addressing comments by @nacx? this is almost there to the finish line and we would love to see this land before the next v0.5 release!

Yes doing these now, I was on holidays past few days.

internal/controller/gateway_config_test.go
sqaisar: Updated test to validate the actual scenario where a Gateway references a non-existent config (e.g., typo). Test now verifies that reconciling returns no error and no events are sent.

tests/e2e/otel_tracing_test.go
sqaisar: Now uses examples/gateway-config/comprehensive.yaml directly (following the same pattern as other example tests). Updated the example to be self-contained with all necessary resources (Namespace, GatewayClass, Backend, AIServiceBackend, AIGatewayRoute). Deleted custom testdata files.

sqaisar:
@nacx @mathetake  Updated the tests according to the suggestions, please review.

mathetake:
@sqaisar thank you for your contribution!!

PR #1624, opened 2025-12-05
feat: add InvokeModel API for gpt models in aws bedrock

hustxiayang:
**Description**
Add InvokeModel API support for gpt models in aws bedrock.  The motivation is to provide consistent services cross providers, https://github.com/envoyproxy/ai-gateway/issues/1644 for more details about the motivation.


internal/translator/openai_awsinvokeopenai.go
```diff
@@ -0,0 +1,325 @@
...
+
+	// URL encode the model name for the path to handle special characters (e.g., ARNs)
+	encodedModelName := url.PathEscape(o.requestModel)
+
+	// Set the path for AWS Bedrock InvokeModel API
+	// https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_InvokeModel.html#API_runtime_InvokeModel_RequestSyntax
+	pathTemplate := "/model/%s/invoke"
+	if req.Stream {
+		// https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_InvokeModelWithResponseStream.html#API_runtime_InvokeModelWithResponseStream_RequestSyntax
+		pathTemplate = "/model/%s/invoke-with-response-stream"
```
yuzisun, line 66: why the path is different from the AWS anthropic invoke model ?
https://github.com/envoyproxy/ai-gateway/blob/main/internal/translator/anthropic_awsanthropic.go#L70

hustxiayang: this is what I was referring to: https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_InvokeModelWithResponseStream.html, but I found it can pass the stream tests.

internal/translator/openai_awsopenai.go
```diff
@@ -0,0 +1,325 @@
...
+		if json.Unmarshal(buf, &existingOpenAIError) == nil && existingOpenAIError.Error.Message != "" {
+			// Already in OpenAI format, return as-is
+			newBody = buf
+		} else {
+			// Try to parse as AWS error and convert to OpenAI format
+			var awsError struct {
+				Type    string `json:"__type,omitempty"`
+				Message string `json:"message"`
+				Code    string `json:"code,omitempty"`
+			}
```
yuzisun: this is already defined here 
https://github.com/envoyproxy/ai-gateway/blob/main/internal/apischema/awsbedrock/awsbedrock.go#L479

internal/translator/translator.go
```diff
@@ -19,15 +19,16 @@ import (
...
+	pathHeaderName             = ":path"
+	statusHeaderName           = ":status"
+	contentTypeHeaderName      = "content-type"
+	contentLengthHeaderName    = "content-length"
+	awsErrorTypeHeaderName     = "x-amzn-errortype"
+	jsonContentType            = "application/json"
+	eventStreamContentType     = "text/event-stream"
+	openAIBackendError         = "OpenAIBackendError"
+	awsBedrockBackendError     = "AWSBedrockBackendError"
+	awsInvokeModelBackendError = "AWSInvokeModelBackendError"
```
gavrissh: `awsInvokeModelBackendError` and `awsBedrockBackendError` should be similar right?

api/v1alpha1/shared_types.go
```diff
@@ -75,6 +75,11 @@ const (
 	// https://aws.amazon.com/bedrock/anthropic/
 	// https://docs.claude.com/en/api/claude-on-amazon-bedrock
 	APISchemaAWSAnthropic APISchema = "AWSAnthropic"
+	// APISchemaAWSOpenAI is the schema for OpenAI models hosted on AWS Bedrock.
+	// Uses the AWS Bedrock InvokeModel API with OpenAI format for requests and responses.
+	//
+	// https://aws.amazon.com/bedrock/
+	APISchemaAWSOpenAI APISchema = "AWSOpenAI"
```
gavrissh: The schema should be named - `AWSInvokeModel` or `AWSBedrockInvoke`?
As AWS Bedrock has an OpenAI-compatible API as well

hustxiayang: Yeah, I would use `AWSInvokeModel`, thanks a lot for your comment!

internal/translator/translator.go
hustxiayang: I wanted to separate them since I thought it's easier to maintain, but I could make them the same. Thanks!

api/v1alpha1/shared_types.go
hustxiayang: I think we should use `AWSInvokeOpenAI`, this is because we also need schema for claude models, which would need translations.

internal/translator/openai_awsopenai.go
hustxiayang: I would reuse it.

internal/translator/translator.go
hustxiayang: updated, thanks!

hustxiayang:
@gavrissh @yuzisun I think I fixed the comment, and tested it with "openai.gpt-oss-20b-1:0" locally. Let me know if you have further concerns. Thanks!

internal/translator/openai_awsinvokeopenai.go
```diff
@@ -0,0 +1,322 @@
...
+		responseModel = o.requestModel
+	} else {
+		// Non-streaming: handle regular JSON response
+		var openAIResp openai.ChatCompletionResponse
+		if err = json.NewDecoder(body).Decode(&openAIResp); err != nil {
+			return nil, nil, metrics.TokenUsage{}, "", fmt.Errorf("failed to decode response: %w", err)
+		}
+
+		// Extract token usage
+		tokenUsage.SetInputTokens(uint32(openAIResp.Usage.PromptTokens))      //nolint:gosec
```
yuzisun, line 145: does prompt token include cached tokens ?

internal/translator/openai_awsinvokeopenai.go
```diff
@@ -0,0 +1,322 @@
...
+			return nil, nil, metrics.TokenUsage{}, "", fmt.Errorf("failed to read streaming body: %w", err)
+		}
+
+		// Convert AWS EventStream to OpenAI SSE format and return it as newBody
+		newBody, tokenUsage, err = o.convertEventStreamToSSE(buf, span)
+		if err != nil {
+			return nil, nil, metrics.TokenUsage{}, "", fmt.Errorf("failed to convert EventStream to SSE: %w", err)
+		}
+
+		if endOfStream && !strings.HasSuffix(string(newBody), "data: [DONE]\n") {
```
yuzisun: standard SSE uses \n\n after the DONE marker

internal/translator/openai_awsinvokeopenai.go
hustxiayang: it contains. you think it's better to reuse function `metrics.ExtractTokenUsageFromExplicitCaching` here?

hustxiayang: it seems that many places did not use this function. Probably we can make it separate from this pr.

internal/translator/openai_awsinvokeopenai.go
hustxiayang: updated

PR #1625, opened 2025-12-05
feat: add gemini embedding translation

hustxiayang:
**Description**
In vertex ai, it does not support `gemini` api, but support `predict` api, as described in : https://docs.cloud.google.com/vertex-ai/generative-ai/docs/model-reference/text-embeddings-api

In predict api, each instance could have a specific task_type:
```
{
  "instances": [
    {
      "task_type": "RETRIEVAL_DOCUMENT",
      "title": "document title",
      "content": "I would like embeddings for this text!"
    },
  ]
}
```
, but do not support token arrays (because gemini does not expose their tokenizers)

Thus, I  extended the `input` in the request body, and here is what is supported:

| Input Format | OpenAI-compatible Backend | GCP Vertex AI Backend |
  |:-------------|:---------------|:----------------------|
  | `string` | ✅ passthrough | ✅ translated |
  | `[]string` | ✅ passthrough | ✅ translated |
  | `[]int64` | ✅ passthrough | ❌ not applicable |
  | `[][]int64` | ✅ passthrough | ❌ not applicable |
  | `EmbeddingInputItem` | ❌ unknown format | ✅ translated |
  | `[]EmbeddingInputItem` | ❌ unknown format | ✅ translated |

Additionally, in gemini api, it  allows users to set a **global** task type for all strings in the embedding requests. This should be convenient for users, so I also add a simple workaround to support this: for each instance, I copy the global task type into the instances.


internal/apischema/openai/openai.go
```diff
@@ -1587,6 +1660,10 @@ type Embedding struct {
 
 	// Index: The index of the embedding in the list of embeddings.
 	Index int `json:"index"`
+
+	// If the input text was truncated due to having a length longer than the allowed maximum input.
+	// https://github.com/googleapis/go-genai/blob/cb486e101dc66794d52125dd22ff43ff4c0e76a6/types.go#L2807
+	Truncated bool `json:"truncated,omitempty"`
```
yuzisun, line 1665: This is gcp specific field right ?

hustxiayang: yes

internal/apischema/gcp/gcp.go
```diff
@@ -36,3 +40,70 @@ type GenerateContentRequest struct {
...
+// https://docs.cloud.google.com/vertex-ai/generative-ai/docs/model-reference/text-embeddings-api#parameter-list
+type Parameters struct {
+	// When set to true, input text will be truncated. When set to false, an error is returned if the input text is longer than the maximum length supported by the model. Defaults to true.
+	AutoTruncate bool `json:"auto_truncate,omitempty"`
+
+	// Used to specify output embedding size. If set, output embeddings will be truncated to the size specified.
+	OutputDimensionality int `json:"outputDimensionality,omitempty"`
+}
+
+// https://github.com/googleapis/python-aiplatform/blob/30e41d01f3fd0ef08da6ad6eb7f83df34476105e/google/cloud/aiplatform_v1/types/prediction_service.py#L63
```
aabchoo, line 67: Did you mean to link https://docs.cloud.google.com/vertex-ai/generative-ai/docs/model-reference/text-embeddings-api#syntax

hustxiayang: Hi, one it source, one is doc. Probably I pasted this one because I pasted // https://docs.cloud.google.com/vertex-ai/generative-ai/docs/model-reference/text-embeddings-api#syntax for several times, but still need to give some source. I could change it to the doc link if you think this is better.

Issue #1619, opened 2025-12-05
MCP proxy should notify clients when there are tools changes


arifsetiawan:
_Description_:

Currently, when there are changes in MCPRoute (add/remove server, enable/disable tools), client needs to reconnect to get updated list of tools. It would be great if AIGW can supports dynamic tool discovery as stated in https://modelcontextprotocol.info/docs/concepts/tools/#tool-discovery-and-updates. Based on standard, AIGW should notify connected clients when there are tools changes by sending `notifications/tools/list_changed`

PR #1630, opened 2025-12-06
mcp: send a tools changed notification when mcp routes change tools

nacx:
**Description**

Sends a tool list changed notification when the MCPRoutes change. We were already notifying when the upstream MCP servers changed the notifications, but now we're also notifying when the routes change or the filtered tools change (without upstream MCP server changes).

Fixes https://github.com/envoyproxy/ai-gateway/issues/1619

N/A

### DIFF
--- a/internal/translator/anthropic_anthropic.go
+++ b/internal/translator/anthropic_anthropic.go
@@ -36,6 +36,7 @@ type anthropicToAnthropicTranslator struct {
 	stream                 bool
 	buffered               []byte
 	streamingResponseModel internalapi.ResponseModel
+	streamingTokenUsage    metrics.TokenUsage
 }
 
 // RequestBody implements [AnthropicMessagesTranslator.RequestBody].
@@ -83,11 +84,17 @@ func (a *anthropicToAnthropicTranslator) ResponseBody(_ map[string]string, body 
 		if err != nil {
 			return nil, nil, tokenUsage, a.requestModel, fmt.Errorf("failed to read body: %w", err)
 		}
+
+		// If this is a fresh start (no buffered data), reset the streaming token usage
+		if len(a.buffered) == 0 {
+			a.streamingTokenUsage = metrics.TokenUsage{}
+		}
+
 		a.buffered = append(a.buffered, buf...)
-		tokenUsage = a.extractUsageFromBufferEvent(span)
+		a.extractUsageFromBufferEvent(span)
 		// Use stored streaming response model, fallback to request model for non-compliant backends
 		responseModel = cmp.Or(a.streamingResponseModel, a.requestModel)
-		return
+		return nil, nil, a.streamingTokenUsage, responseModel, nil
 	}
 
 	// Parse the Anthropic response to extract token usage.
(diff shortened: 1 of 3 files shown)


tests/extproc/mcp/env.go
```diff
@@ -133,12 +138,23 @@ func requireNewMCPEnv(t *testing.T, forceJSONResponse bool, writeTimeout time.Du
...
 
 	m := new(mcpEnv)
+	m.mcp1, m.mcp2 = mcp1, mcp2
 	m.collector = collector
 	m.writeTimeout = writeTimeout
 	m.extProcMetricsURL = fmt.Sprintf("http://localhost:%d/metrics", env.ExtProcAdminPort())
 	m.baseURL = fmt.Sprintf("http://localhost:%d%s", env.EnvoyListenerPort(), path)
 
 	m.client = mcp.NewClient(&mcp.Implementation{Name: "demo-http-client", Version: "0.1.0"}, &mcp.ClientOptions{
+		ToolListChangedHandler: func(_ context.Context, request *mcp.ToolListChangedRequest) {
```
nacx, line 148: From this point the change is not related to the issue. This adds tests that were missing to verify upstream tool change notification propagation.

internal/mcpproxy/session.go
```diff
@@ -228,6 +239,14 @@ func (s *session) streamNotifications(ctx context.Context, w http.ResponseWriter
 			if heartbeatTicker != nil {
 				heartbeatTicker.Reset(heartbeatInterval)
 			}
+		case <-toolsChanged:
```
mathetake: hmm. does this mean that `toolsChanged` shared across many notification streams/sessions but only one item will be sent by the config change? I think it will not work as expected i think when we have N streams

mathetake: i think instead of channel, we can use `sync.Cond` with multiple waiter

nacx: oh, yes, good catch! I'll fix

nacx: Done. PTAL.
I've opted for a silightly different approach:

* I've moved the config-related stuff to a `config.go` (apologies for the noise).
* I've created a changeSignaler interface, that still uses a channel, but instead of sending to it, it clsoes the channel to wake up all watchers. I think it's a bit cleaner than the `sync.Cond` in this case, because this allows us to use the existing `select` loop we're using for upstream events or heartbeats.

internal/mcpproxy/config_test.go
```diff
@@ -0,0 +1,386 @@
...
+								Include: []string{"tool1", "tool2"}, // Different tools
+							},
+						},
+					},
+				},
+			},
+		},
+	}
+
+	// Start watcher goroutines to make sure all of them are notified
```
nacx, line 293: This is a new test added after the review comments to make sure we can notify N watchers

internal/mcpproxy/config.go
```diff
@@ -0,0 +1,195 @@
...
+	defer m.mu.Unlock()
+	return m.notify
+}
+
+// Signal notifies all watchers of a configuration change.
+func (m *multiWatcherSignaler) Signal() {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	close(m.notify)                // Wake everyone
+	m.notify = make(chan struct{}) // Create a new channel for future updates
```
mathetake, line 194: ah nice

PR #1633, opened 2025-12-07
fix: do not reset usage when no remaining streaming buffer

yuzisun:
**Description**
Remove the usage reset when there is no remaining stream buffer as this is expected and fixed the unrealistic test case which masked the issue.


internal/translator/anthropic_anthropic.go
```diff
@@ -85,11 +85,6 @@ func (a *anthropicToAnthropicTranslator) ResponseBody(_ map[string]string, body
 			return nil, nil, tokenUsage, a.requestModel, fmt.Errorf("failed to read body: %w", err)
 		}
 
-		// If this is a fresh start (no buffered data), reset the streaming token usage
```
yuzisun, line 88: This is actually expected between different chunks, hence removing the logic introduced in https://github.com/envoyproxy/ai-gateway/pull/1631

PR #1641, opened 2025-12-09
feat: add support for Google Search grounding as a tool type for Gemini models

jamesbuddrige:
**Description**

Adds support for Google Search grounding [1] as a tool type for Gemini models, complementing the existing enterprise web search support added in #1526. Enterprise search and Google Search are two different tools in Vertex. See [2] for more details.

[1]: https://docs.cloud.google.com/vertex-ai/generative-ai/docs/grounding/grounding-with-google-search
[2]: https://docs.cloud.google.com/vertex-ai/generative-ai/docs/grounding/web-grounding-enterprise#overview

The implementation translates the google_search tool type to Gemini's GoogleSearch with support for all configuration options:
- exclude_domains: Filter out specific domains from search results (Vertex AI only)
- blocking_confidence: Set phishing block threshold (Vertex AI only)
- time_range_filter: Restrict results to a time window (Gemini API only)

Response grounding metadata is already handled by the existing GroundingMetadata field from #1526.

Related PR: #1526

The config structs are prefixed with GCP (e.g., GCPGoogleSearchConfig, GCPTimeRangeFilter) to follow the existing pattern for vendor-specific extensions (GCPVertexAIVendorFields, GCPVertexAIGenerationConfig).

I'm not completely sure this is the right design pattern for tool options going forward. These fields are necessary to expose the full functionality of the underlying API, but adding vendor-specific config structs to the OpenAI schema does muddy the interface a bit. The alternative would be something more generic (e.g., map[string]any), but that loses type safety. Open to feedback on whether this approach makes sense or if we want to establish a different pattern for future tool configurations.


mathetake:
@yuzisun @hustxiayang could you take a look?

yuzisun:
@jamesbuddrige can you help update the doc here to include the GoogleSearch tool
https://aigateway.envoyproxy.io/docs/capabilities/llm-integrations/vendor-specific-fields#gcp-vertex-ai-gemini

internal/apischema/openai/openai.go
```diff
@@ -1075,11 +1075,26 @@ const (
 	ToolTypeFunction            ToolType = "function"
 	ToolTypeImageGeneration     ToolType = "image_generation"
 	ToolTypeEnterpriseWebSearch ToolType = "enterprise_search"
+	ToolTypeGoogleSearch        ToolType = "google_search"
 )
 
+// GCPGoogleSearchConfig contains GCP-specific configuration for Google Search grounding.
```
aabchoo, line 1082: Could you link https://pkg.go.dev/google.golang.org/genai#GoogleSearch or the source of where you got this from?

jamesbuddrige: Added

yuzisun (approved):

mathetake (approved):

PR #1645, opened 2025-12-10
test: fix premature request closure in benchmarks

howardjohn:
**Description**

This fixes an issue in the benchmark where the context is cancelled before the request processing has finished. This causes Go to consider the connection as un-reusable, leading to 0% connection re-use. This only impacts servers that return data across multiple TCP packets (which is 100% legal thing to do), hence why it is not always showing up.

With this change:
```
name                     time/op
MCP/Baseline_NoProxy-32  44.5µs ± 3%
MCP/Agent_Gateway-32     84.4µs ± 1%
MCP/EAIGW_EAIGW_Default  21.3ms ± 0%
MCP/EAIGW_Config_100-32   257µs ± 0%
```

Before this change:
```
name                  time/op
MCP/Agent_Gateway-32   183µs ± 3%
```

Additionally, pulls in https://github.com/modelcontextprotocol/go-sdk/pull/709 which is also required to fix this.

:heart:


howardjohn:
Thanks for the quick approval! I trust that any blogs that were published based on this inaccurate benchmark would be updated :slightly_smiling_face:

nacx:
> I trust that any blogs that were published based on this inaccurate benchmark would be updated 🙂

Sure. With the changes in this code, I've run the same benchmarks, and they show a difference of 0.2 ms between EAIGW and AgentGateway.

I trust as well that the content that has been published on the agent gateway side with unbalanced comparisons on the MCP performance (kubecon presentations and charts, and the Istio proposal that advocates for moving from envoy to agentgateway, showing the incomplete MCP benchmarks) will be updated as well to properly explain the **sub-millisecond difference**. Thank you! 🙂

Issue #1644, opened 2025-12-10
Provide InvokeModel API in AWS Platform


hustxiayang:
_Description_:

**Proposal: Integration of the `InvokeModel` API within the AWS Platform**

We propose the addition of the `InvokeModel` API in our AWS platform integration, driven by the following key motivations: the need for better compatibility and existing issues with the current `Converse` API.

### 1. Enhanced Compatibility and Cross-Provider Services

Our primary goal is to provide seamless, cross-provider services to our users. For instance, both Google Cloud Platform (GCP) and Amazon Web Services (AWS) offer access to models like the Claude series. When a user makes a request to a model such as `claude-sonnet-4`, we want the underlying implementation to be identical across all providers. This approach allows us to fully utilize our resources, offer more resilient and efficient services, and ensure a consistent user experience with minimal configuration concerns.

Currently, the AWS `Converse` API has significant differences compared to the Anthropic Messages API (and the OpenAI API). While we have implemented numerous workarounds to mitigate these discrepancies, some incompatibilities are unavoidable and frequently cause integration issues.

### 2.  `Converse` API Specific Issues

The `Converse` API itself presents several functional limitations and bugs that have led our users to avoid it. Below are two recent cases illustrating these problems:

*   **A. Inability to Force Model to Skip a Tool Call**
    Users cannot explicitly instruct the model *not* to make a tool call after the model has previously initiated one.

    In the OpenAI API, users can set `tool_choice` to `None` or completely remove the `tools` definition to signal to the model that no further tool calls should be made. In the `Converse` API, neither of these options is available.
    *   First, the `Converse` API does not support `None` as a valid `tool_choice` parameter, as documented in the [AWS Bedrock API Reference](docs.aws.amazon.com).
    *   Second, we cannot simply remove the `tools` definition during subsequent interactions. If a user asks a question, the model responds with a tool call, and the user then appends the tool results for the next request, attempting to remove the `tools` definition in that subsequent request results in a validation error:
        ```
        Error code: 400 - {'type': 'error', 'error': {'type': 'ValidationException:internal.amazon.com', 'code': '400', 'message': 'The toolConfig field must be defined when using toolUse and toolResult content blocks.'}}
        ```
        An online example of this issue can be found in the [dify GitHub repository](github.com).

*   **B. Ambiguity of the `tool_choice` Parameter**
    The specification for `tool_choice` in the `Converse` API is ambiguous.

[...]

PR #1648, opened 2025-12-11
feat: add InvokeModel API support for claude models in aws bedrock

hustxiayang:
**Description**
Add InvokeModel API support for claude models in aws bedrock.  The motivation is to provide consistent services cross providers, https://github.com/envoyproxy/ai-gateway/issues/1644 for more details about the motivation.

Other Changes:
I put common codes related to anthropic into `anthropic_helper.go`, so that both aws and gcp can share these codes.

### DIFF
--- a/internal/apischema/openai/openai.go
+++ b/internal/apischema/openai/openai.go
@@ -869,6 +869,7 @@ type WebSearchLocation struct {
 type ThinkingUnion struct {
 	OfEnabled  *ThinkingEnabled  `json:",omitzero,inline"`
 	OfDisabled *ThinkingDisabled `json:",omitzero,inline"`
+	OfAdaptive *ThinkingAdaptive `json:",omitzero,inline"`
 }
 
 type ThinkingEnabled struct {
@@ -887,6 +888,10 @@ type ThinkingDisabled struct {
 	Type string `json:"type,"`
 }
 
+type ThinkingAdaptive struct {
+	Type string `json:"type,"`
+}
+
 // MarshalJSON implements the json.Marshaler interface for ThinkingUnion.
 func (t *ThinkingUnion) MarshalJSON() ([]byte, error) {
 	if t.OfEnabled != nil {
@@ -894,6 +899,9 @@ func (t *ThinkingUnion) MarshalJSON() ([]byte, error) {
 	}
 	if t.OfDisabled != nil {
 		return json.Marshal(t.OfDisabled)
+	}
+	if t.OfAdaptive != nil {
+		return json.Marshal(t.OfAdaptive)
 	}
 	// If both are nil, return an empty object or an error, depending on your desired behavior.
 	return []byte(`{}`), nil
@@ -923,6 +931,12 @@ func (t *ThinkingUnion) UnmarshalJSON(data []byte) error {
 			return err
 		}
 		t.OfDisabled = &disabled
+	case "adaptive":
+		var adaptive ThinkingAdaptive
+		if err := json.Unmarshal(data, &adaptive); err != nil {
+			return err
+		}
+		t.OfAdaptive = &adaptive
 	default:
 		return fmt.Errorf("invalid thinking union type: %s", typeVal)
 	}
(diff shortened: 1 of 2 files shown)


yuzisun:
@hustxiayang can you help resolve the conflicts?

api/v1alpha1/shared_types.go
```diff
@@ -15,7 +15,7 @@ package v1alpha1
 type VersionedAPISchema struct {
 	// Name is the name of the API schema of the AIGatewayRoute or AIServiceBackend.
 	//
-	// +kubebuilder:validation:Enum=OpenAI;Cohere;AWSBedrock;AzureOpenAI;GCPVertexAI;GCPAnthropic;Anthropic;AWSAnthropic
+	// +kubebuilder:validation:Enum=OpenAI;Cohere;AWSBedrock;AzureOpenAI;GCPVertexAI;GCPAnthropic;Anthropic;AWSAnthropic;AWSInvokeAnthropic
```
yuzisun: We should not need to create another schema, `AWSAnthropic` is the schema used for `AIServiceBackend`. Depending on different endpoint paths we select the corresponding translator.

internal/endpointspec/endpointspec.go
```diff
@@ -114,6 +114,8 @@ func (ChatCompletionsEndpointSpec) GetTranslator(schema filterapi.VersionedAPISc
 		return translator.NewChatCompletionOpenAIToOpenAITranslator(schema.OpenAIPrefix(), modelNameOverride), nil
 	case filterapi.APISchemaAWSBedrock:
 		return translator.NewChatCompletionOpenAIToAWSBedrockTranslator(modelNameOverride), nil
+	case filterapi.APISchemaAWSInvokeAnthropic:
```
yuzisun: use `APISchemaAWSAnthropic`

internal/translator/anthropic_helper.go
```diff
@@ -0,0 +1,1130 @@
...
+
+	"github.com/envoyproxy/ai-gateway/internal/apischema/awsbedrock"
+	"github.com/envoyproxy/ai-gateway/internal/apischema/openai"
+	"github.com/envoyproxy/ai-gateway/internal/internalapi"
+	"github.com/envoyproxy/ai-gateway/internal/json"
+	"github.com/envoyproxy/ai-gateway/internal/metrics"
+	"github.com/envoyproxy/ai-gateway/internal/tracing/tracingapi"
+)
+
+// TODO: Double check these
```
yuzisun: what needs to be checked ?

api/v1alpha1/shared_types.go
hustxiayang: updated

internal/endpointspec/endpointspec.go
hustxiayang: updated as well

internal/translator/anthropic_helper.go
hustxiayang: removed the comments, this should be todo during coding.

internal/filterapi/filterconfig.go
```diff
@@ -114,20 +114,21 @@ const (
...
-	// APISchemaAWSBedrock represents the AWS Bedrock API schema.
+	// APISchemaAWSBedrock represents the AWS Bedrock Converse API schema.
 	APISchemaAWSBedrock APISchemaName = "AWSBedrock"
 	// APISchemaAzureOpenAI represents the Azure OpenAI API schema.
 	APISchemaAzureOpenAI APISchemaName = "AzureOpenAI"
 	// APISchemaGCPVertexAI represents the Google Cloud Gemini API schema.
 	// Used for Gemini models hosted on Google Cloud Vertex AI.
 	APISchemaGCPVertexAI APISchemaName = "GCPVertexAI"
-	// APISchemaGCPAnthropic represents the Google Cloud Anthropic API schema.
+	// APISchemaGCPAnthropic represents the schema from OpenAI API to Google Cloud Anthropic API.
```
yuzisun: This is not correct, it is the same schema if you use message API directly not always translated from OpenAI API.

hustxiayang: updated

yuzisun (approved):

PR #1650, opened 2025-12-12
feat: add tokenize endpoint

hustxiayang:
**Description**
This PR is to add the generic `/tokenize` path  to get the number of input tokens for OpenAI's chat completion/completion requests. Since OpenAI does not have an official API of tokenize for chat completion, we use the definitions from vLLM: https://github.com/vllm-project/vllm/blob/d2c919dcc20b1ea77a94fa01e813ebbb31f8a66a/vllm/entrypoints/openai/protocol.py#L1897, but still call it OpenAI schema, as it's to get the input tokens of OpenAI's chat completion/completion schema.

**Coverage**
This PR covers OpenAI-compatible schema (for example, vLLM), gemini's vertex ai, gcp anthropic, aws anthropic, and aws bedrock. For OpenAI models, since it does not provide a dedicated tokenization endpoint for chat completion/completion, users can deploy a tokenization service with tiktoken: https://github.com/openai/tiktoken.


mathetake (changes requested):
* Can you delete the entire examples/tokenize example yamls as they are irrelevant here ie. no specific config to tokenize endpoint. If you think about it, we don't have yamls for /messages /rerank, etc.
* Let's start small for the new feature like this instead of sending a giant PR. E.g. start with only AWS bedrock. That would be helpful for you to land code faster but also helping reviewers a lot. So could you remove the other code like gcp translation etc.

mathetake:
@hustxiayang I see you left TODOs in the description -- so let me mark the PR as draft until if you feel ready for review after resolving them.

hustxiayang:
discussed with @aabchoo offline, and we agreed to split this PR into different providers' PR to make it easier to review.

hustxiayang:
closed because all split PRs have been merged.

PR #1654, opened 2025-12-15
ci: restrict several test jobs to run only within the `envoyproxy/ai-gateway` repository.

### LINKED ISSUE #1655: when i fork the repo, my fork wiil run ci and will fail
Currently these integration tests also run in forked repositories, and because configuration information (such as environment variables) is missing, they may cause errors.

Aias00:
**Description**

Currently these integration tests also run in forked repositories, and because configuration information (such as environment variables) is missing, they may cause errors.

restrict several test jobs to run only within the `envoyproxy/ai-gateway` repository.

fixes #1655 

### DIFF
--- a/.github/workflows/build_and_test.yaml
+++ b/.github/workflows/build_and_test.yaml
@@ -133,7 +133,7 @@ jobs:
 
   test_extproc:
     needs: changes
-    if: ${{ needs.changes.outputs.code == 'true' }}
+    if: ${{ needs.changes.outputs.code == 'true' && github.repository == 'envoyproxy/ai-gateway' }}
     name: External Processor Test (${{ matrix.os }})
     strategy:
       fail-fast: false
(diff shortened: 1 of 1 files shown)


mathetake:
i am sorry what's the issue with these jobs failing in your fork repo? It is your repository so you are free to apply these changes. Also, ignoring the failures of these jobs in your repo vs adding a workaround here is essentially the same. The change here cannot be guaranteed to be maintained here as the conditional is always false, hence drift will be expected to be honest. I would tackle the root cause though if I were you, like to see if we can add a "skip" for missing credentials etc. at the same time, missing credentials are already taken care of and the tests using them are already skipped so not sure what to do on that front though.

Aias00:
> i am sorry what's the issue with these jobs failing in your fork repo? It is your repository so you are free to apply these changes. Also, ignoring the failures of these jobs in your repo vs adding a workaround here is essentially the same. The change here cannot be guaranteed to be maintained here as the conditional is always false, hence drift will be expected to be honest. I would tackle the root cause though if I were you, like to see if we can add a "skip" for missing credentials etc. at the same time, missing credentials are already taken care of and the tests using them are already skipped so not sure what to do on that front though.

thx, got that

PR #1656, opened 2025-12-15
fix: adjust otel metric minimum value assertions to allow zero and reorder verification checks in tests.

### LINKED ISSUE #1420: flaky test: tests/extproc/vcr/TestOtelOpenAIChatCompletions_metrics
```
    --- FAIL: TestOtelOpenAIChatCompletions_metrics/chat-streaming-detailed-usage (1.12s)
        otel_chat_completions_metrics_test.go:94: 
            	Error Trace:	/Users/runner/work/ai-gateway/ai-gateway/tests/extproc/vcr/otel_chat_completions_metrics_test.go:169
            	            				/Users/runner/work/ai-gateway/ai-gateway/tests/extproc/vcr/otel_chat_completions_metrics_test.go:94
            	Error:      	"0" is not greater than "0"
            	Test:       	TestOtelOpenAIChatCompletions_metrics/chat-streaming-detailed-usage
```

https://github.com/envoyproxy/ai-gateway/actions/runs/18754655225/job/53503362239

Aias00:
**Description**

In VCR/mock environments or on fast CI runners, the execution time for streaming responses can be extremely short, resulting in calculated metrics like gen_ai.server.time_per_output_token or gen_ai.server.time_to_first_token being exactly 0.0. The original assertion require.Greater(t, value, 0.0) incorrectly flagged this valid state as a failure. Fix:

Updated verifyTimeToFirstTokenMetrics and verifyTimePerOutputTokenMetricsWithOriginal
to use require.GreaterOrEqual(t, value, 0.0) instead of strict inequality.
Refactored the verification logic to ensure metric attributes and existence are validated before checking their values, ensuring that a value of 0.0 is accepted only when the metric is properly reported.

Fixes #1420


mathetake:
Did you verify that the flake is actually gone? e.g. go test -rune=foo -count=1000?

Aias00:
> Did you verify that the flake is actually gone? e.g. go test -rune=foo -count=1000?

I'll give it a try, and if any problems arise, I'll fix them.

codefromthecrypt (commented):
it seems a code smell to allow zero here. considering metrics should be derived with nanosecond precision... any idea what's causing this?

Aias00:
> Did you verify that the flake is actually gone? e.g. go test -rune=foo -count=1000?

i tried 

`go test -v -run TestOtelOpenAIChatCompletions_metrics ./tests/extproc/vcr/... -count=1000 -timeout=30m`

my local. but this will cause the port resource exhausted
so i run 

`for i in {1..1000}; do
  echo "Run #$i"
  go test -v -run TestOtelOpenAIChatCompletions_metrics ./tests/extproc/vcr/...
  if [ $? -ne 0 ]; then
     echo "Failed at run #$i"
     exit 1
  fi
  sleep 1
done`

and they all passed

mathetake:
I'm sorry but #1657 fixes it at the root cause

Issue #1652, opened 2025-12-15
CI is failing with the latest EG


zhaohuabing:
_Description_:

```
time=2025-12-15T02:04:56.821Z level=INFO msg="starting admin server" address=[::]:1064
panic: proto: syntax error (line 1:1): invalid field name: goo.gle/debugstr

goroutine 102 [running]:
github.com/envoyproxy/ai-gateway/internal/extproc.(*Server).setBackend(0xc0003d0780, {0x225df38, 0xc000253440}, {0x22664e0, 0xc000235d90}, {0xc0004481e0, 0x49}, 0x1, 0x1?)
  /home/ubuntu/ai-gateway/internal/extproc/server.go:316 +0x765
github.com/envoyproxy/ai-gateway/internal/extproc.(*Server).Process(0xc0003d0780, {0x2274658, 0xc0002c1ad0})
  /home/ubuntu/ai-gateway/internal/extproc/server.go:188 +0x4a5
github.com/envoyproxy/go-control-plane/envoy/service/ext_proc/v3._ExternalProcessor_Process_Handler({0x1d99700?, 0xc0003d0780}, {0x226f9a0, 0xc000164200})
  /home/ubuntu/go/pkg/mod/github.com/envoyproxy/go-control-plane/envoy@v1.36.0/service/ext_proc/v3/external_processor_grpc.pb.go:106 +0xd8
google.golang.org/grpc.(*Server).processStreamingRPC(0xc0002fc908, {0x225df38, 0xc0002533e0}, 0xc0007aa4e0, 0xc000276690, 0x33d8800, 0x0)
  /home/ubuntu/go/pkg/mod/google.golang.org/grpc@v1.77.0/server.go:1714 +0x1172
google.golang.org/grpc.(*Server).handleStream(0xc0002fc908, {0x225f490, 0xc0007aa1a0}, 0xc0007aa4e0)
  /home/ubuntu/go/pkg/mod/google.golang.org/grpc@v1.77.0/server.go:1836 +0xd85
google.golang.org/grpc.(*Server).serveStreams.func2.1()
  /home/ubuntu/go/pkg/mod/google.golang.org/grpc@v1.77.0/server.go:1063 +0x7f
created by google.golang.org/grpc.(*Server).serveStreams.func2 in goroutine 100
```

https://github.com/envoyproxy/ai-gateway/actions/runs/20115137470/job/57740847394?pr=1645

mathetake:
[...]

PR #1651, opened 2025-12-15
extproc: adds workaround for protobuf dep breaking change in Envoy

mathetake:
**Description**

This adds a workaround for a breaking change introduced in the latest Envoy version, which in turns is causing a CI failure with Envoy Gateway latest. Note that the workaround here might be removed soon, but the main purpose of this commit is to unblock the CIs/PRs.

**Related Issues/PRs (if applicable)**
Closes #1652

PR #1660, opened 2025-12-16
fix: use full-path attributes to avoid text-proto encoded metadata

mathetake:
**Description**

Previously, we are getting the cluster/endpoint metadata as an attribute sent by envoy. Internally, Envoy serializes it as a protobuf text. However, since the recent change in the upstream protobuf, which in turn Envoy depends on, the breaking change in its format has been introduced. We added a workaround in #1651 but it's fragile fundamentally since it's still the text protobuf.

This switches our implementation to list the full path for the metadata that we want to get in the ExtProc fiter configuration which allows us to avoid encountering the text formatted full metadata. In order not to break the existing deployment, we keep the legacy code path until the next release.

Followup on #1651 and #1652

### DIFF
--- a/internal/extproc/server.go
+++ b/internal/extproc/server.go
@@ -298,55 +298,52 @@ func (s *Server) setBackend(ctx context.Context, p Processor, internalReqID stri
 	if attributes == nil || len(attributes.Fields) == 0 { // coverage-ignore
 		return status.Error(codes.Internal, "missing attributes in request")
 	}
-	var metadataFieldKey string
+	// metadataFieldKey is the key for the entire metadata field in the attributes for backward compatibility.
+	// backendNamePath is the path to the backend name in the metadata.
+	var metadataFieldKey, backendNamePath string
 	if isEndpointPicker {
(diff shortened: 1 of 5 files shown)


internal/extproc/server.go
```diff
@@ -298,63 +298,60 @@ func (s *Server) setBackend(ctx context.Context, p Processor, internalReqID stri
...
+		backendNamePath = internalapi.XDSUpstreamHostMetadataBackendNamePath
 	}
-	// This should contain the endpoint metadata.
-	hostMetadata, ok := attributes.Fields[metadataFieldKey]
-	if !ok {
+
+	var backendName string
+	if b, ok := attributes.Fields[backendNamePath]; ok {
+		backendName = b.GetStringValue()
+	} else if hostMetadata, ok := attributes.Fields[metadataFieldKey]; ok {
```
mathetake, line 315: note: this block is the same as the existing code before #1651

nacx (approved):
Nice! :)

PR #1663, opened 2025-12-16
mcp: add mcp metrics fixes for tools/call

### LINKED ISSUE #1661: mcp: errors on tool/list tools/call do not appear in metrics
_Description_:

1. Whenever there is an error on methods such as  tools/list, tools/call, ping etc... by the mcp client, AIGW doesnt record the Failed statuses in the metrics

Lets look at a example, to know what is causing the inconsistency: 

The tools/call response struct has an `IsError` field that indicates tool-level errors. When this is true, the response is still a valid JSON-RPC response (no protocol error), but the tool execution itself failed. 

> [!NOTE]
> The code currently doesn't check this field, so it incorrectly counts it as a success.

2. When a MCP Route is created with list of two backends, each backend with two tools (say)
b1---> t1, t2
b2 ---> t3,t4
Followed by MCP Route creation, the mcp server say now deleted one of its tool t4, and when MCP Client now calls a tools/call request to t4 AIGW still treats it as a success metric and increments the same.

> [!NOTE]
> The code currently doesn't track this, since it treats it as part of allowed lists map

_Repro steps_:

- Hit any tool with empty args to trigger an error, and you will see it is still an valid rpc response with `isError=true`.
```
{
  "content": [
    {
      "type": "text",
      "text": "missing required parameter: method"
    }
  ],
  "isError": true
}
```
- Create a mcp route with t1(belongs to mcp server), t2(does not belong to mcp server), now hit a tools/call to t2 tool, you should see unknown tool as part of the error.
```
{
"jsonrpc":"2.0",
"id":6,"error":
   {
      "code":-32602,
       "message":"unknown tool \"issue_dummy\""
   }
}
```

_Metrics_:
```
mcp_initialization_duration_token_count{otel_scope_name="envoyproxy/ai-gateway",otel_scope_schema_url="",otel_scope_version="",ownerId="nai-github-hf-2-test"} 4
# HELP mcp_method_count_total Total number of MCP methods invoked
# TYPE mcp_method_count_total counter
mcp_method_count_total{mcp_method_name="initialize",otel_scope_name="envoyproxy/ai-gateway",otel_scope_schema_url="",otel_scope_version="",ownerId="nai-github-hf-2-test",status="success"} 2
mcp_method_count_total{mcp_method_name="logging/setLevel",otel_scope_name="envoyproxy/ai-gateway",otel_scope_schema_url="",otel_scope_version="",ownerId="nai-github-hf-2-test",status="success"} 2
mcp_method_count_total{mcp_method_name="notifications/initialized",otel_scope_name="envoyproxy/ai-gateway",otel_scope_schema_url="",otel_scope_version="",ownerId="nai-github-hf-2-test",status="success"} 2
mcp_method_count_total{mcp_method_name="prompts/list",otel_scope_name="envoyproxy/ai-gateway",otel_scope_schema_url="",otel_scope_version="",ownerId="nai-github-hf-2-test",status="success"} 1
mcp_method_count_total{mcp_method_name="resources/list",otel_scope_name="envoyproxy/ai-gateway",otel_scope_schema_url="",otel_scope_version="",ownerId="nai-github-hf-2-test",status="success"} 1
mcp_method_count_total{mcp_method_name="tools/call",otel_scope_name="envoyproxy/ai-gateway",otel_scope_schema_url="",otel_scope_version="",ownerId="nai-github-hf-2-test",status="success"} 5
mcp_method_count_total{mcp_method_name="tools/list",otel_scope_name="envoyproxy/ai-gateway",otel_scope_schema_url="",otel_scope_version="",ownerId="nai-github-hf-2-test",status="success"} 2
# HELP mcp_request_duration Duration of MCP requests
# TYPE mcp_request_duration histogram
mcp_request_duration_bucket{otel_scope_name="envoyproxy/ai-gateway",otel_scope_schema_url="",otel_scope_version="",ownerId="nai-github-hf-2-test",le="0.005"} 0
mcp_request_duration_bucket{otel_scope_name="envoyproxy/ai-gateway",otel_scope_schema_url="",otel_scope_version="",ownerId="nai-github-hf-2-test",le="0.01"} 0
mcp_request_duration_bucket{otel_scope_name="envoyproxy/ai-gateway",otel_scope_schema_url="",otel_scope_version="",ownerId="nai-github-hf-2-test",le="0.025"} 4
mcp_request_duration_bucket{otel_scope_name="envoyproxy/ai-gateway",otel_scope_schema_url="",otel_scope_version="",ownerId="nai-github-hf-2-test",le="0.05"} 4
mcp_request_duration_bucket{otel_scope_name="envoyproxy/ai-gateway",otel_scope_schema_url="",otel_scope_version="",ownerId="nai-github-hf-2-test",le="0.1"} 4
mcp_request_duration_bucket{otel_scope_name="envoyproxy/ai-gateway",otel_scope_schema_url="",otel_scope_version="",ownerId="nai-github-hf-2-test",le="0.25"} 5
mcp_request_duration_bucket{otel_scope_name="envoyproxy/ai-gateway",otel_scope_schema_url="",otel_scope_version="",ownerId="nai-github-hf-2-test",le="0.5"} 15
mcp_request_duration_bucket{otel_scope_name="envoyproxy/ai-gateway",otel_scope_schema_url="",otel_scope_version="",ownerId="nai-github-hf-2-test",le="1"} 17
mcp_request_duration_bucket{otel_scope_name="envoyproxy/ai-gateway",otel_scope_schema_url="",otel_scope_version="",ownerId="nai-github-hf-2-test",le="2.5"} 17
mcp_request_duration_bucket{otel_scope_name="envoyproxy/ai-gateway",otel_scope_schema_url="",otel_scope_version="",ownerId="nai-github-hf-2-test",le="5"} 17
mcp_request_duration_bucket{otel_scope_name="envoyproxy/ai-gateway",otel_scope_schema_url="",otel_scope_version="",ownerId="nai-github-hf-2-test",le="10"} 17
mcp_request_duration_bucket{otel_scope_name="envoyproxy/ai-gateway",otel_scope_schema_url="",otel_scope_version="",ownerId="nai-github-hf-2-test",le="+Inf"} 17
mcp_request_duration_sum{otel_scope_name="envoyproxy/ai-gateway",otel_scope_schema_url="",otel_scope_version="",ownerId="nai-github-hf-2-test"} 4.688333919
mcp_request_duration_count{otel_scope_name="envoyproxy/ai-gateway",otel_scope_schema_url="",otel_scope_version="",ownerId="nai-github-hf-2-test"} 17

```

---

#### What this means?
one can only see, 
✅ Total number of errors (status="error")
✅ Error types (error.type dimension: invalid_param, internal_error, etc.)
✅ Error durations by error type

You cannot currently see:
❌ How many tools/call requests failed vs tools/list
❌ Error rate per method
❌ Which specific methods are causing the most errors

For proper error tracking, RecordMethodErrorCount should probably accept a methodName parameter and include it as a dimension, similar to how RecordMethodCount works.
cc: @nacx

Hritik003:
**Description**

1. Introduce a status `failed` which is triggered only for application errors, i.e, those requests when rpc response is successful but returns with `isError=true` in the response. Eg: `data: {"jsonrpc":"2.0","id":7,"result":{"content":[{"type":"text","text":"MCP error -32602: Tool model_search1 not found"}],"isError":true}}`

2. Introduce a errorType `errToolCall` for toolCalls:

```go
type errToolCall struct {
	toolName string
	backend  string
	err      error
}
```

3. Introduce a function , `func errorType(err error) metrics.MCPErrorType ` which checks in the following order:
    - first checks if its a json error ` jsonrpc.CodeInvalidParams, jsonrpc.CodeMethodNotFound,  jsonrpc.CodeInvalidRequest, jsonrpc.CodeParseError, jsonrpc.CodeInternalError`
    - secondly, it checks for specific error types `if errors.Is(err, errBackendNotFound) || errors.Is(err, errSessionNotFound) || errors.Is(err, errInvalidToolName) `
    - third, handles joined errors
    
4. In `servePost` function, add logic if the error is of type `errToolCall` : set applicationError to true. And whenever applicationError is true we `RecordMethodErrorCount( with status : failed) ` and if not(meaning protocol error) then `RecordMethodErrorCount( with status : error) `

#### things out of scope for this PR

1. Recording error counts per backend level
2. Recording errors for each SSE event received for a single request to a backend

**Testing**

<details><summary> When tools/call method calls a tool with wrong payload </summary>

- response from MCP client:
```
{
  "content": [
    {
      "type": "text",
      "text": "missing required parameter: method"
    }
  ],
  "isError": true
}
```

- npx logs:
```
Received POST message for sessionId a6a60807-0d0f-48da-b917-10c149602e04
```

- Metric recorded:
```
mcp_method_count_total{mcp_method_name="tools/call",otel_scope_name="envoyproxy/ai-gateway",otel_scope_schema_url="",otel_scope_version="",ownerId="nai-github-hf-2-test",status="failed"} 2
....

mcp_request_duration_count{error_type="internal_error",otel_scope_name="envoyproxy/ai-gateway",otel_scope_schema_url="",otel_scope_version="",ownerId="nai-github-hf-2-test"} 2
```
</details>

<details><summary> When tools/call method calls a tool part of mcp_route but tool doesnt belong to MCP server</summary>

- response from MCP client
```
(data: {"jsonrpc":"2.0","id":7,"result":{"content":[{"type":"text","text":"MCP error -32602: Tool model_search1 not found"}],"isError":true}})
```

- npx logs
```
Received POST message for sessionId ba9e0c62-26f7-47b5-a295-030dd4b38391
```

- metric Recorded:
```
mcp_method_count_total{mcp_method_name="tools/call",otel_scope_name="envoyproxy/ai-gateway",otel_scope_schema_url="",otel_scope_version="",ownerId="nai-github-hf-2-test",status="failed"} 3

....

mcp_request_duration_count{error_type="internal_error",otel_scope_name="envoyproxy/ai-gateway",otel_scope_schema_url="",otel_scope_version="",ownerId="nai-github-hf-2-test"} 2

mcp_request_duration_count{error_type="invalid_param",otel_scope_name="envoyproxy/ai-gateway",otel_scope_schema_url="",otel_scope_version="",ownerId="nai-github-hf-2-test"} 1
```
</details>

<details><summary> When tools/call method calls a tool not part of mcp route tool filters</summary>

- response from MCP client
```
(data: {"jsonrpc":"2.0","id":7,"error":{"code":-32001,"message":"Error POSTing to endpoint (HTTP 400): invalid tool name: model_search2","data":{}}})
```

- npx logs
```
Error from MCP server: Error: Error POSTing to endpoint (HTTP 400): invalid tool name: model_search2
```

- metric Recorded:
```
mcp_method_count_total{mcp_method_name="tools/call",otel_scope_name="envoyproxy/ai-gateway",otel_scope_schema_url="",otel_scope_version="",ownerId="nai-github-hf-2-test",status="failed"} 3

mcp_method_count_total{mcp_method_name="tools/call",otel_scope_name="envoyproxy/ai-gateway",otel_scope_schema_url="",otel_scope_version="",ownerId="nai-github-hf-2-test",status="error"} 1

....

mcp_request_duration_count{error_type="internal_error",otel_scope_name="envoyproxy/ai-gateway",otel_scope_schema_url="",otel_scope_version="",ownerId="nai-github-hf-2-test"} 3

mcp_request_duration_count{error_type="invalid_param",otel_scope_name="envoyproxy/ai-gateway",otel_scope_schema_url="",otel_scope_version="",ownerId="nai-github-hf-2-test"} 1
```
</details>

> [!NOTE]
> All the MCP success metrics are getting recorded and tested with the changes.

Fixes #1661 

---

cc: @nacx @mathetake @AyushSawant18588

### DIFF
--- a/internal/endpointspec/endpointspec.go
+++ b/internal/endpointspec/endpointspec.go
@@ -109,7 +109,7 @@ func (ChatCompletionsEndpointSpec) ParseBody(
 func (ChatCompletionsEndpointSpec) GetTranslator(schema filterapi.VersionedAPISchema, modelNameOverride string) (translator.OpenAIChatCompletionTranslator, error) {
 	switch schema.Name {
 	case filterapi.APISchemaOpenAI:
-		return translator.NewChatCompletionOpenAIToOpenAITranslator(schema.Prefix, modelNameOverride), nil
+		return translator.NewChatCompletionOpenAIToOpenAITranslator(schema.OpenAIPrefix(), modelNameOverride), nil
 	case filterapi.APISchemaAWSBedrock:
 		return translator.NewChatCompletionOpenAIToAWSBedrockTranslator(modelNameOverride), nil
 	case filterapi.APISchemaAzureOpenAI:
@@ -139,7 +139,7 @@ func (CompletionsEndpointSpec) ParseBody(
 func (CompletionsEndpointSpec) GetTranslator(schema filterapi.VersionedAPISchema, modelNameOverride string) (translator.OpenAICompletionTranslator, error) {
 	switch schema.Name {
 	case filterapi.APISchemaOpenAI:
-		return translator.NewCompletionOpenAIToOpenAITranslator(schema.Prefix, modelNameOverride), nil
+		return translator.NewCompletionOpenAIToOpenAITranslator(schema.OpenAIPrefix(), modelNameOverride), nil
 	default:
 		return nil, fmt.Errorf("unsupported API schema: backend=%s", schema)
 	}
@@ -161,7 +161,7 @@ func (EmbeddingsEndpointSpec) ParseBody(
 func (EmbeddingsEndpointSpec) GetTranslator(schema filterapi.VersionedAPISchema, modelNameOverride string) (translator.OpenAIEmbeddingTranslator, error) {
 	switch schema.Name {
 	case filterapi.APISchemaOpenAI:
-		return translator.NewEmbeddingOpenAIToOpenAITranslator(schema.Prefix, modelNameOverride), nil
+		return translator.NewEmbeddingOpenAIToOpenAITranslator(schema.OpenAIPrefix(), modelNameOverride), nil
 	case filterapi.APISchemaAzureOpenAI:
 		return translator.NewEmbeddingOpenAIToAzureOpenAITranslator(schema.Version, modelNameOverride), nil
 	default:
@@ -184,7 +184,7 @@ func (ImageGenerationEndpointSpec) ParseBody(
 func (ImageGenerationEndpointSpec) GetTranslator(schema filterapi.VersionedAPISchema, modelNameOverride string) (translator.OpenAIImageGenerationTranslator, error) {
 	switch schema.Name {
 	case filterapi.APISchemaOpenAI:
-		return translator.NewImageGenerationOpenAIToOpenAITranslator(schema.Prefix, modelNameOverride), nil
+		return translator.NewImageGenerationOpenAIToOpenAITranslator(schema.OpenAIPrefix(), modelNameOverride), nil
 	default:
 		return nil, fmt.Errorf("unsupported API schema: backend=%s", schema)
 	}
--- a/internal/filterapi/filterconfig.go
+++ b/internal/filterapi/filterconfig.go
@@ -12,6 +12,7 @@
 package filterapi
 
 import (
+	"cmp"
 	"os"
 	"time"
 
@@ -91,6 +92,13 @@ type VersionedAPISchema struct {
 	Version string `json:"version,omitempty"`
 	// Prefix is the prefix of the API schema. Optional. Currently, only used for OpenAI.
 	Prefix string `json:"prefix,omitempty"`
+}
+
+// OpenAIPrefix returns the OpenAI API prefix for the VersionedAPISchema.
+// This is for backwards compatibility with existing users. This won't be
+// necessary after v0.5 release when we can use Prefix directly.
+func (v VersionedAPISchema) OpenAIPrefix() string {
+	return cmp.Or(v.Version, v.Prefix)
 }
 
 // APISchemaName corresponds to APISchemaName in api/v1alpha1/api.go.
--- a/tests/extproc/extproc_test.go
+++ b/tests/extproc/extproc_test.go
@@ -34,7 +34,11 @@ const (
 )
 
 var (
-	openAISchema         = filterapi.VersionedAPISchema{Name: filterapi.APISchemaOpenAI, Prefix: "v1"}
+	openAISchema = filterapi.VersionedAPISchema{
+		Name: filterapi.APISchemaOpenAI,
+		// This is backwards compatibility for existing users. Use 'prefix' after v0.5 release.
+		Version: "v1",
+	}
 	awsBedrockSchema     = filterapi.VersionedAPISchema{Name: filterapi.APISchemaAWSBedrock}
 	awsAnthropicSchema   = filterapi.VersionedAPISchema{Name: filterapi.APISchemaAWSAnthropic, Version: "bedrock-2023-05-31"}
 	azureOpenAISchema    = filterapi.VersionedAPISchema{Name: filterapi.APISchemaAzureOpenAI, Version: "2025-01-01-preview"}


internal/mcpproxy/handlers.go
```diff
@@ -588,33 +588,60 @@ func copyProxyHeaders(resp *http.Response, w http.ResponseWriter) {
...
 				msg.ID = req.ID
+
+				// Check if this is a JSON-RPC error response
+				if msg.Error != nil {
+					// Return error so metrics are recorded as failure
+					body, _ = jsonrpc.EncodeMessage(msg)
+					w.Header().Set("Content-Length", fmt.Sprintf("%d", len(body)))
+					w.WriteHeader(resp.StatusCode)
+					_, _ = w.Write(body)
+					return fmt.Errorf("%s: %s", string(metrics.MCPErrorInternal), msg.Error.Error())
```
nacx: Here I would just return the `msg.Error` and let the caller decide what to do with it and what error code to propagate. In most cases, the error will be a `jsonrpc.Error` type and we'll be able to cast it and read the `Code` field to determine the right erorr type.
In the `handelrs.go`, for example, we already have the `errorType` method doing this. We should enhance that method to check if the error is an `jsonrpc.Error` and use the `Code` to get the right string type.

internal/mcpproxy/handlers.go
```diff
@@ -588,33 +588,60 @@ func copyProxyHeaders(resp *http.Response, w http.ResponseWriter) {
...
 				}
 				msg.ID = req.ID
+
+				// Check if this is a JSON-RPC error response
+				if msg.Error != nil {
+					// Return error so metrics are recorded as failure
+					body, _ = jsonrpc.EncodeMessage(msg)
+					w.Header().Set("Content-Length", fmt.Sprintf("%d", len(body)))
+					w.WriteHeader(resp.StatusCode)
+					_, _ = w.Write(body)
```
nacx: This is done already by default afterwards.... can we change the code slightly to reuse that and reduce code duplication?

internal/mcpproxy/handlers.go
```diff
@@ -588,33 +588,60 @@ func copyProxyHeaders(resp *http.Response, w http.ResponseWriter) {
...
+				if req.Method == "tools/call" && msg.Result != nil {
+					var toolResult mcp.CallToolResult
+					if err := json.Unmarshal(msg.Result, &toolResult); err == nil {
+						if toolResult.IsError {
+							// Tool execution failed, record as failure and return error
+							m.recordResponse(ctx, msg)
+							body, _ = jsonrpc.EncodeMessage(msg)
+							w.Header().Set("Content-Length", fmt.Sprintf("%d", len(body)))
+							w.WriteHeader(resp.StatusCode)
+							_, _ = w.Write(body)
```
nacx: All this is already done in the default case. can we redo the code to just set the error to return, so we don't duplicate the code?

internal/mcpproxy/handlers.go
```diff
@@ -588,33 +588,60 @@ func copyProxyHeaders(resp *http.Response, w http.ResponseWriter) {
...
+							return fmt.Errorf("%s: %s", string(metrics.MCPErrorInternal), "tool returned isError=true")
```
nacx: We should probably create a `errToolCall` type we can use to return here, to let callers have better structured information about the tool execution error, and just return it.
This will allow us to have a specific metric for tool calls that includes the details of the failed tool name as well.

internal/mcpproxy/handlers.go
```diff
@@ -664,6 +694,23 @@ func (m *MCPProxy) proxyResponseBody(ctx context.Context, s *session, w http.Res
 							continue
 						}
 						msg.ID = req.ID
+
+						// Check if this is a JSON-RPC error response. This error occurs when a tool name is part of the allowed list
+						// but upstream MCP Server doesn't contain this tool.
+						if msg.Error != nil {
+							responseError = fmt.Errorf("%s: %s", string(metrics.MCPErrorInvalidParam), msg.Error.Error())
```
nacx: We probably can't assume this is an invalid param. Instead, we should probably (it is unlikely that we'll get N events each with individual errors, but we should be generic and gracefully handle it):

* Collect a list of errors, so we don't miss any individual error.
* Wrap them with `errors.Join`
* Return the errors to the caller.

Update the `errType` helper to also check if the error is a joined one, unwrap it to get the error list, and act accordingly.

internal/mcpproxy/handlers.go
```diff
@@ -664,6 +694,23 @@ func (m *MCPProxy) proxyResponseBody(ctx context.Context, s *session, w http.Res
...
+						}
+
+						// Check if this is a tools/call response with isError=true
+						if req.Method == "tools/call" && msg.Result != nil {
+							var toolResult mcp.CallToolResult
+							if err = json.Unmarshal(msg.Result, &toolResult); err == nil {
+								if toolResult.IsError {
+									// Tool execution failed, record as failure
+									responseError = fmt.Errorf("%s: %s", string(metrics.MCPErrorInternal), "tool returned isError=true")
```
nacx: Same as before, I think we should create a proper error type and return to the caller so it can better decide the best way to handle it.

internal/mcpproxy/handlers.go
```diff
@@ -664,6 +694,23 @@ func (m *MCPProxy) proxyResponseBody(ctx context.Context, s *session, w http.Res
...
+						// but upstream MCP Server doesn't contain this tool.
+						if msg.Error != nil {
+							responseError = fmt.Errorf("%s: %s", string(metrics.MCPErrorInvalidParam), msg.Error.Error())
+						}
+
+						// Check if this is a tools/call response with isError=true
+						if req.Method == "tools/call" && msg.Result != nil {
+							var toolResult mcp.CallToolResult
+							if err = json.Unmarshal(msg.Result, &toolResult); err == nil {
+								if toolResult.IsError {
```
nacx: Here and in the similar code above, can we make it more concise?
```suggestion
							if err = json.Unmarshal(msg.Result, &toolResult); err == nil && toolResult.IsError {
```

internal/mcpproxy/handlers.go
```diff
@@ -1241,6 +1288,7 @@ func sendToAllBackendsAndAggregateResponsesImpl[responseType any](ctx context.Co
 			if respMsg, ok := event.messages[l-1].(*jsonrpc.Response); ok && respMsg.ID == request.ID {
 				switch {
 				case respMsg.Error != nil:
+					m.metrics.RecordMethodFailedCount(ctx, request.Method, nil)
```
nacx: We need to decide how we want to count failures here.

* If the request has been send to N backends, do we want to increment the counter for each backend where the request failed to be sent? Or since it is a single client call, we just record it once? Keep in mind that this method is called _not only_ for tool calls!
* How should we properly record these kinds of errors?
  * Should we add a `backend` attribute to the main metric?
  * Should we have a separate metric to track upstream errors like this?
* Probably here we should be recording the errors, wrapping them in an error type that properly conveys the upstream error, and return the list of errors (wrapped with `errors.Join` to the caller, so that the caller can decide what to do, and send the right metrics. This way we don't introduce metrics calls here and keep a better separation of concerns.

cc @mathetake

internal/metrics/mcp_metrics.go
```diff
@@ -205,6 +208,18 @@ func (m *mcp) RecordMethodCount(ctx context.Context, methodName string, params m
 		))
 }
 
+// RecordMethodFailedCount implements [MCPMetrics.RecordMethodFailedCount].
+func (m *mcp) RecordMethodFailedCount(ctx context.Context, methodName string, params mcpsdk.Params) {
+	if methodName == "" {
+		return
+	}
+	m.methodCount.Add(ctx, 1,
```
nacx, line 214: I think reusing the same metric this way is wrong, because on any failure, we may be incrementing it twice, with different attributes. This means that when querying the metric with no attributes, users may get the wrong results and have higher counter values than they should be.
We should probably add the missing `methodName` to the existing `MethodErrorCount` and probably add a parameter for the status type to record .

internal/mcpproxy/handlers.go
Hritik003: I had a query in this area, is it right to throw an error here? because I see in the `make test-extproc` it is expected to not receive any span exceptions?
But if we dont return an error, defer function will not get triggered to Record the failure metric.

What is the right thing here?

nacx: Can you link to the exact test you're referring to?

Hritik003: - https://github.com/envoyproxy/ai-gateway/blob/0a3ada44ecc73f04eb60e5a09aa8d2cd434e298c/tests/extproc/mcp/mcp_test.go#L76
- https://github.com/envoyproxy/ai-gateway/blob/0a3ada44ecc73f04eb60e5a09aa8d2cd434e298c/tests/extproc/mcp/mcp_test.go#L67

These basically. But I found a solution for this, i.e, to record failures whenever responsesErr is not empty and to return nil overall to not trigger the span exception. For ref: 

- https://github.com/Hritik003/ai-gateway/blob/3b8a8fa80b64b959effa92b5a7199430ca780662/internal/mcpproxy/handlers.go#L801

```go

func (m *MCPProxy) proxyResponseBody(ctx context.Context, s *session, w http.ResponseWriter, resp *http.Response,
	req *jsonrpc.Request, backend filterapi.MCPBackend,
) error {

.......
	// Record metrics for application-level errors but don't return them (to avoid span exceptions)
	if len(responseErrors) > 0 && req != nil {
		// we need to record failedCount here and not errorCount since these are application errors.
		m.metrics.RecordMethodFailedCount(ctx, req.Method, nil)
		m.metrics.RecordRequestErrorDuration(ctx, time.Now(), metrics.MCPErrorInternal, nil)
	}
	// Return nil to prevent span exceptions while still having recorded failure metrics
	return nil
```

nacx: >I had a query in this area, is it right to throw an error here? because I see in the make test-extproc it is expected to not receive any span exceptions?
But if we dont return an error, defer function will not get triggered to Record the failure metric.
>
>What is the right thing here?

If there is an actual error, the right thing is probably to record that in the span. IF we're recording a metric with a failure, why wouldn't the trace span contain that data? IMHO we should include the details of the error in the span as well, and update the tests accordingly if that's needed.

internal/mcpproxy/handlers.go
```diff
@@ -370,13 +387,45 @@ func (m *MCPProxy) servePOST(w http.ResponseWriter, r *http.Request) {
 }
 
 func errorType(err error) metrics.MCPErrorType {
+	if err == nil {
+		return ""
+	}
+
+	var joinedErrs interface{ Unwrap() []error }
```
nacx: [nit] check this last, as it's more expensive. Do first the more direct checks

nacx: Also let's add a unit test for this method now that it is doing some more stuff.

internal/mcpproxy/handlers.go
```diff
@@ -588,33 +636,66 @@ func copyProxyHeaders(resp *http.Response, w http.ResponseWriter) {
...
+						if req.Params != nil {
+							var params mcp.CallToolParams
+							if err = json.Unmarshal(req.Params, &params); err == nil {
+								toolName = params.Name
+							}
+						}
+						responseError = &errToolCall{
+							toolName: toolName,
+							backend:  backend.Name,
+							err:      errors.New("tool returned isError=true"),
```
nacx: When `isError` is true, the error details will be in the `toolResult.Content`. We should probably encapsulate that here instead of a generic error.

internal/mcpproxy/handlers.go
```diff
@@ -624,7 +705,16 @@ func (m *MCPProxy) proxyResponseBody(ctx context.Context, s *session, w http.Res
 		w.Header().Set("Content-Length", fmt.Sprintf("%d", len(body)))
 		w.WriteHeader(resp.StatusCode)
 		_, _ = w.Write(body)
-		return
+
+		// Record metrics for application-level errors but don't return them (to avoid span exceptions)
```
nacx: I wouldn't do this (see the other comment on the span topic).

internal/mcpproxy/handlers.go
```diff
@@ -664,6 +757,32 @@ func (m *MCPProxy) proxyResponseBody(ctx context.Context, s *session, w http.Res
...
 						msg.ID = req.ID
+
+						// Check if this is a JSON-RPC error response
+						if msg.Error != nil {
+							// Collect error to return to caller
+							responseErrors = append(responseErrors, msg.Error)
+						}
+
+						// Check if this is a tools/call response with isError=true
+						if req.Method == "tools/call" && msg.Result != nil {
```
nacx: This code is very similar to the code above. refactor it to a small method to favor code reuse.

internal/mcpproxy/handlers.go
Hritik003: So shall we go with an approach like, check if its an application error and if yes---> do not span an exception for it?

internal/mcpproxy/handlers.go
Hritik003: > [nit] check this last, as it's more expensive. Do first the more direct checks

So to answer that, it is inexpensive because it simply returns the slice from errors.Join() without allocating any new memory

internal/mcpproxy/handlers.go
Hritik003: @nacx @mathetake 

> If the request has been send to N backends, do we want to increment the counter for each backend where the request failed to be sent? Or since it is a single client call, we just record it once? Keep in mind that this method is called not only for tool calls!

it should be recorded once, because the metrics we show currently are mcp route level and not per backend level. Might as well consider this point at the time we have metrics per backend. But for now, since its a tools/list call to a "mcp route", it should be recorded ONCE!

> How should we properly record these kinds of errors?
Should we add a backend attribute to the main metric?

Yes that is correct, so something like

```
metric`{ method=tools/call, mcp_route = r1, backend = b1} 1
metric`{  method=tools/call, mcp_route = r1, backend = b2} 1
```

> Should we have a separate metric to track upstream errors like this?

Whats wrong in just adding a new attribute, that should suffice right?

nacx: Adding the attribute to the existing metric changes the current semantics, as you can't have "client focused metrics". If we add it to the existing metric, then we'll increment the metric once for each backend (setting the attirbute), but users that are querying the metric _without filtering by backend_ (which is what they'll be doing now) will get the sum across backends, which may not be the expected behaviour. This is why I was suggesting to add a new metric when it comes to record upstream errors/latency/etc.

internal/mcpproxy/handlers.go
nacx: It's not. It is checking errors in a loop and calling the errType recursively. It's cheaper to do the other checks first (that also are more likely to be the cause in normal execution paths)

internal/mcpproxy/handlers.go
nacx: Why wouldn't we want to include the errors in the span, even if it is an application error? We'll be recording an error in the metrics, right? What would the spans not contain this info?

Hritik003: Please correct me if I am wrong here

Application errors are something like "the system **correctly communicated** that something was invalid/failed". 
Examples:
- JSON-RPC validation errors (e.g., missing required parameters)
- Tool execution that successfully runs but returns isError: true (e.g., a file operation that fails because the file doesn't exist)
- Protocol-level errors that are part of the MCP specification

and Span exceptions are something like "the system itself broke" meaning system/infra level errors. 

we dont want to mix both of them right?

nacx: The semantic conventions on span exceptions talk about "application exceptions": https://opentelemetry.io/docs/specs/semconv/exceptions/exceptions-spans/

Also, think about a full trace an application developer may be observing:

```
ChatCompletion
  -> ToolCall1 (!)
    -> ChatCompletion
      -> ToolCall2
        -> ChatCompletion
          -> GenerateResponse
```
When users are inspecting sucha. trace, if `ToolCall1` failed, that could impact the entire conversation. Wouldn't we want to clearly flag that `ToolCall1` errored, so that any issue or side effect that comes from it can be easily identified?
Perhaps if the ToolCall1 succeeded, the ToolCall2 would never be made. A tool completing with error may have a significant impact on the conversation, and I think we should record the information as clearly as possible for users to reason about what's going on.

Hritik003:
@nacx Could you please have a relook at the PR? Thanks

internal/mcpproxy/handlers.go
nacx: Can you address this?

internal/mcpproxy/handlers.go
```diff
@@ -40,6 +40,75 @@ var (
...
+}
+
+func (e *errToolCall) Unwrap() error {
+	return e.err
+}
+
+// checkToolCallError examines a tools/call response and creates a structured error if isError is true.
+// It extracts the tool name from the request params and the error content from the tool result.
+// Returns nil if the response is not a tools/call error.
+func checkToolCallError(req *jsonrpc.Request, msg *jsonrpc.Response, backendName string) error {
```
nacx: nit: I think it's better for this method to return `*errToolCall`, so callers can easily use the structured info without a cast.

internal/mcpproxy/handlers.go
```diff
@@ -680,6 +796,17 @@ func (m *MCPProxy) proxyResponseBody(ctx context.Context, s *session, w http.Res
...
+						// Check if this is a JSON-RPC error response
+						if msg.Error != nil {
+							// Collect error to return to caller
+							responseErrors = append(responseErrors, msg.Error)
+						}
+
+						// Check if this is a tools/call response with isError=true
+						if toolErr := checkToolCallError(req, msg, backend.Name); toolErr != nil {
+							responseErrors = append(responseErrors, toolErr)
+						}
```
nacx, line 817: Let's not call the checkTool if it's not necessary?
```suggestion

						// Check if this is a JSON-RPC error response
						if msg.Error != nil {
							// Collect error to return to caller
							responseErrors = append(responseErrors, msg.Error)
						// Check if this is a tools/call response with isError=true
						} else if toolErr := checkToolCallError(req, msg, backend.Name); toolErr != nil {
							responseErrors = append(responseErrors, toolErr)
						}
```

internal/mcpproxy/handlers.go
```diff
@@ -1257,6 +1384,7 @@ func sendToAllBackendsAndAggregateResponsesImpl[responseType any](ctx context.Co
 			if respMsg, ok := event.messages[l-1].(*jsonrpc.Response); ok && respMsg.ID == request.ID {
 				switch {
 				case respMsg.Error != nil:
+					m.metrics.RecordMethodErrorCount(ctx, request.Method, nil)
```
nacx: This would be counted per-backend, so we go back to the discussion in previous comments.

From a client perspective, having this code here means that, a single "tools/list" to the gateway, would generate a count of N errors if N backends fail. This is different from what it currently behaves, and gives the wrong impression, especially if we're not adding a backend attribute to the metric.

I think we should not record this here, or introduce per-backend specific metrics to avoid misleading counts.

internal/mcpproxy/handlers_test.go
```diff
@@ -718,6 +722,138 @@ func TestHandleToolCallRequest_BackendError(t *testing.T) {
...
+		// Read the request to extract the ID
+		body, _ := io.ReadAll(r.Body)
+		req, _ := jsonrpc.DecodeMessage(body)
+		reqMsg := req.(*jsonrpc.Request)
+
+		// Return a JSON-RPC error response indicating tool not found
+		resp := &jsonrpc.Response{
+			ID: reqMsg.ID,
+			Error: &jsonrpc.Error{
+				Code:    -32601,
```
nacx: can you use the jsonrpc errror constants for better readability?

internal/mcpproxy/handlers_test.go
```diff
@@ -718,6 +722,138 @@ func TestHandleToolCallRequest_BackendError(t *testing.T) {
...
+	require.Equal(t, "test-tool", toolErr.toolName)
+	require.Equal(t, "backend1", toolErr.backend)
+	require.Contains(t, err.Error(), "missing required parameter: owner")
+
+	// The response should be written to the client (HTTP 200)
+	require.Equal(t, http.StatusOK, rr.Code)
+
+	// Verify the response contains the error message
+	require.Contains(t, rr.Body.String(), "missing required parameter: owner")
+}
```
nacx, line 855: Nice! ❤️

internal/mcpproxy/handlers.go
```diff
@@ -370,13 +439,45 @@ func (m *MCPProxy) servePOST(w http.ResponseWriter, r *http.Request) {
...
+	var joinedErrs interface{ Unwrap() []error }
+	if errors.As(err, &joinedErrs) {
+		errs := joinedErrs.Unwrap()
+		for _, e := range errs {
+			if errType := errorType(e); errType != "" && errType != metrics.MCPErrorInternal {
+				return errType
+			}
+		}
+		return metrics.MCPErrorInternal
+	}
```
nacx: Also, with the `joinedErrors` we need to account for a case and decide what to do:

We'll get a list of errors when we received N events as a response and there are several events with errors. It is unlikely to happen, but code needs to be robust. The question is: if there are N errors, should we increase the error count N times, one for each error? And account for each error type individually?

internal/metrics/mcp_metrics.go
```diff
@@ -78,6 +78,7 @@ type mcpStatusType string
 
 const (
 	mcpStatusSuccess mcpStatusType = "success"
+	mcpStatusFailed  mcpStatusType = "failed"
```
nacx: This seems to not be used anymore, but the idea was that when there is an application error, we use this code instead, right?

Hritik003: Right missed this.

internal/mcpproxy/handlers.go
Hritik003: Changes made: 

First check: jsonrpc.Error - direct type check
Second check: Specific error types with errors.Is
Last check: Joined errors - expensive check

internal/mcpproxy/handlers.go
Hritik003: ig metrics recording per backend basis should be part of another dev plan and effort. For now, whenever atleast one of the backend fails, we should be recording it only ONCE! 

something like, 

```go
func sendToAllBackendsAndAggregateResponsesImpl[responseType any](ctx context.Context, events <-chan *sseEvent, m *MCPProxy, w http.ResponseWriter, s *session, request *jsonrpc.Request, mergeFn broadCastResponseMergeFn[responseType]) error {

........

    var recordErrorMetric bool <---------declare a bool 
	var responses []broadCastResponse[responseType]
	for event := range events {
		.......
		if l := len(event.messages); l != 0 {
		........
			if respMsg, ok := event.messages[l-1].(*jsonrpc.Response); ok && respMsg.ID == request.ID {
				switch {
				case respMsg.Error != nil:
					recordErrorMetric = true <-----set it to true when atleast one of the backend throws an error
					
					
.....
     }

	if recordErrorMetric {
		m.metrics.RecordMethodErrorCount(ctx, request.Method, nil, metrics.MCPStatusError)
	}

	mergedResp := mergeFn(s, responses)
	encodedResp, err := json.Marshal(mergedResp)
	if err != nil {
		return fmt.Errorf("failed to marshal response: %w", err)
	}
```

internal/mcpproxy/handlers.go
Hritik003: Good point, didn't strike my mind.

In that case, its better to record for each error, i.e increment for n times

Hritik003: also if we are doing this, then the joined errors check needs to happen first to handle multiple errors properly https://github.com/envoyproxy/ai-gateway/pull/1663#discussion_r2629523296

Hritik003: Also if we have multiple errors coming in, do we span out each error into exception or the just the first internal error?

Hritik003:
Boiling down, we have to decide on the following two scenarios to complete this PR:

(at aggregated backends level) https://github.com/envoyproxy/ai-gateway/pull/1663#discussion_r2628253647
1. Do we want to record error count per backend or only per route level(current scenario) ? eg: tools/list to all the aggregated backends.

--> For this, we might have to refactor our RecordMethodErrorCount( to accept a backend field) and record per backend iterating through a loop of backend errors.

(at single backend level) https://github.com/envoyproxy/ai-gateway/pull/1663#discussion_r2628294583
2. When tools/list is called to a particular backend, might return N sse events. Do we want to record all N sse events ? or record just an error?

---> For this, I am kind of ambiguous if we need to consider this edge case? In terms of code, we might need an extra function 

```
errorTypes() → []metrics.MCPErrorType  // For counting multiple errors
       ↓
errorType() → metrics.MCPErrorType     // For single error type (logging, spans)
```

and just to solve this edgecase, we might have to return []{single object} everytime for normal single backend errors.

<details><summary> example errorTypes() function</summary>

```go
// This is important for proper metrics when multiple backends fail with different error types.
func errorTypes(err error) []metrics.MCPErrorType {
	if err == nil {
		return nil
	}

	// Check for joined errors first - this allows us to handle multiple errors properly
	var joinedErrs interface{ Unwrap() []error }
	if errors.As(err, &joinedErrs) {
		errs := joinedErrs.Unwrap()
		var types []metrics.MCPErrorType
		for _, e := range errs {
			// Recursively get error types from each error
			subTypes := errorTypes(e)
			types = append(types, subTypes...)
		}
		// If we found any error types, return them; otherwise return internal error
		if len(types) > 0 {
			return types
		}
		return []metrics.MCPErrorType{metrics.MCPErrorInternal}
	}

	// Check if error is a jsonrpc.Error and map the code to appropriate MCPErrorType
	var jsonrpcErr *jsonrpc.Error
	if errors.As(err, &jsonrpcErr) {
		switch jsonrpcErr.Code {
		case jsonrpc.CodeInvalidParams:
			return []metrics.MCPErrorType{metrics.MCPErrorInvalidParam}
		case jsonrpc.CodeMethodNotFound:
			return []metrics.MCPErrorType{metrics.MCPErrorUnsupportedMethod}
		case jsonrpc.CodeInvalidRequest, jsonrpc.CodeParseError:
			return []metrics.MCPErrorType{metrics.MCPErrorInvalidJSONRPC}
		case jsonrpc.CodeInternalError:
			return []metrics.MCPErrorType{metrics.MCPErrorInternal}
		default:
			return []metrics.MCPErrorType{metrics.MCPErrorInternal}
		}
	}

	// Check for specific error types
	if errors.Is(err, errBackendNotFound) || errors.Is(err, errSessionNotFound) || errors.Is(err, errInvalidToolName) {
		return []metrics.MCPErrorType{metrics.MCPErrorInvalidParam}
	}

	return []metrics.MCPErrorType{metrics.MCPErrorInternal}
}
```

</details>

3. a new question arises here, if we need to span a single error or multiple errors here?

---

cc: @nacx @mathetake

internal/mcpproxy/handlers.go
nacx: That sounds good, but then I'd probably make this method return an error so we can have the handler record the metric in the "normal path" like all other methods. This will also help preventing "double recording" by mistake in the future, and we have the metric recording centralized in the same place, and methods consistently leveraging it.

nacx:
>1. Do we want to record error count per backend or only per route level(current scenario)? eg: tools/list to all the aggregated backends.

IMO backend-specific metrics should go to another metric, not the current one, otherwise we'll create misleading counts. we can introduce an upstream mcp specific metric as a feature in other PRs.

>2. When tools/list is called to a particular backend, might return N sse events. Do we want to record all N sse events ? or record just an error?

Thinking about it, it is OK for now to return and count just 1 error, because that will align with the number of calls being made. We can later decide a proper way of counting errors for each SSE event received for a single request, but I think this is out of the scope for this PR.
For now I think it's OK to just return the first error (if any), as I highly doubt we'll find an SSE response with N error events.

>3. a new question arises here, if we need to span a single error or multiple errors here?

I think we should flag a single span (not multiple). We could merge all the errors in that span's exception details if we want.

Hritik003:
> >1. Do we want to record error count per backend or only per route level(current scenario)? eg: tools/list to all the aggregated backends.
> 
> 
> 
> IMO backend-specific metrics should go to another metric, not the current one, otherwise we'll create misleading counts. we can introduce an upstream mcp specific metric as a feature in other PRs.
> 
> 
> 
> >2. When tools/list is called to a particular backend, might return N sse events. Do we want to record all N sse events ? or record just an error?
> 
> 
> 
> Thinking about it, it is OK for now to return and count just 1 error, because that will align with the number of calls being made. We can later decide a proper way of counting errors for each SSE event received for a single request, but I think this is out of the scope for this PR.
> 
> For now I think it's OK to just return the first error (if any), as I highly doubt we'll find an SSE response with N error events.
> 
> 
> 
> >3. a new question arises here, if we need to span a single error or multiple errors here?
> 
> 
> 
> I think we should flag a single span (not multiple). We could merge all the errors in that span's exception details if we want.

Agree, will fix the changes accordingly.

Hritik003:
Summary since the last commit(updating the same in PR desc):

1. Introduce a status `failed` which is triggered only for application errors, i.e, those requests when rpc response is successful but returns with `isError=true` in the response. Eg: `data: {"jsonrpc":"2.0","id":7,"result":{"content":[{"type":"text","text":"MCP error -32602: Tool model_search1 not found"}],"isError":true}}`

2. Introduce a errorType `errToolCall` for toolCalls:

```go
type errToolCall struct {
	toolName string
	backend  string
	err      error
}
```

3. Introduce a function , `func errorType(err error) metrics.MCPErrorType ` which checks in the following order:
    - first checks if its a json error ` jsonrpc.CodeInvalidParams, jsonrpc.CodeMethodNotFound,  jsonrpc.CodeInvalidRequest, jsonrpc.CodeParseError, jsonrpc.CodeInternalError`
    - secondly, it checks for specific error types `if errors.Is(err, errBackendNotFound) || errors.Is(err, errSessionNotFound) || errors.Is(err, errInvalidToolName) `
    - third, handles joined errors
    
4. In `servePost` function, add logic if the error is of type `errToolCall` ---> set applicationError to true. And whenever applicationError is true we `RecordMethodErrorCount( with status ---> failed) ` and if not(meaning protocol error) then `RecordMethodErrorCount( with status ---> error) `


#### things out of scope for this PR

1. Recording error counts per backend level
2. Recording errors for each SSE event received for a single request to a backend

---

cc: @nacx

internal/mcpproxy/handlers.go
```diff
@@ -109,14 +179,15 @@ func doNotForwardResponseToBackends(msg *jsonrpc.Response) bool {
...
-		params        mcp.Params
+		ctx              = r.Context()
+		startAt          = time.Now()
+		s                *session
+		err              error
+		errType          metrics.MCPErrorType
+		requestMethod    string
+		span             tracing.MCPSpan
+		params           mcp.Params
+		applicationError bool
```
nacx, line 190: nit: no need to define it globally to the method as it's only used in the defered function

internal/mcpproxy/handlers.go
```diff
@@ -125,11 +196,21 @@ func (m *MCPProxy) servePOST(w http.ResponseWriter, r *http.Request) {
 				slog.String("error_type", string(errType)),
 				slog.String("duration", time.Since(startAt).String()))
 		}
+		applicationError = false
 		if err != nil {
+			var errToolCall *errToolCall
+			if errors.As(err, &errToolCall) {
+				applicationError = true
+			}
```
nacx, line 204:
```suggestion
		if err != nil {
			var errToolCall *errToolCall
			applicationError := errors.As(err, &errToolCall)
```

nacx:
Looks good! Thank you!

PR #1666, opened 2025-12-17
api: add 'prefix' in schema config

### LINKED ISSUE #1629: schema.version: "" (empty string) should allow no version prefix for OpenAI-compatible backends, but defaults to "v1"
**_Description_:**

When configuring an AIServiceBackend with schema.name: OpenAI and schema.version: "" (empty string), the AI Gateway should use no version prefix in the upstream path (e.g., /chat/completions). However, the controller converts empty string to "v1", resulting in /v1/chat/completions being sent to the backend.

This contradicts the API documentation which states:
On the other hand, DeepSeek (https://api-docs.deepseek.com/) doesn't use version prefix, so the version can be set to an empty string.

https://aigateway.envoyproxy.io/docs/api/#versionedapischema

But also states:
This defaults to "v1" if not set or empty string.
The expected behavior based on the DeepSeek example is that version: "" should produce /chat/completions, not /v1/chat/completions.

Note: in our case we are using perplexity which uses the same url format as deepseek without the v1

Root Cause seem to be in  In internal/controller/gateway.go line 158:
The cmp.Or function treats empty string as a zero value and falls back to "v1". This prevents users from configuring OpenAI-compatible backends that don't use a version prefix (e.g., Perplexity AI uses /chat/completions).

**_Repro steps_:**

  1. Create an AIServiceBackend with empty version:

```
  kind: AIServiceBackend
  metadata:
    name: perplexity
    namespace: envoy-ai-gateway-system
  spec:
    schema:
      name: OpenAI
      version: ""
    backendRef:
      name: perplexity-backend
      kind: Backend
      group: gateway.envoyproxy.io
```

  2. Create a Backend pointing to api.perplexity.ai:

```
  kind: Backend
  metadata:
    name: perplexity-backend
    namespace: envoy-ai-gateway-system
  spec:
    endpoints:
      - fqdn:
          hostname: api.perplexity.ai
          port: 443
```

  3. Create an AIGatewayRoute to route requests:

```
  kind: AIGatewayRoute
  metadata:
    name: perplexity-route
    namespace: envoy-ai-gateway-system
  spec:
    parentRefs:
      - name: my-gateway
        namespace: envoy-ai-gateway-system
        kind: Gateway
    filterConfig:
      type: ExternalProcessor
    rules:
      - matches:
          - headers:
              - type: RegularExpression
                name: x-ai-eg-model
                value: "sonar.*"
        backendRefs:
          - name: perplexity
```

  4. Send a request:

  curl -X POST https://your-gateway/v1/chat/completions \
    -H "Authorization: Bearer your-key" \
    -d '{"model": "sonar-pro", "messages": [{"role": "user", "content": "Hello"}]}'

  Expected: Gateway forwards to https://api.perplexity.ai/chat/completions

  Actual: Gateway forwards to https://api.perplexity.ai/v1/chat/completions → Returns 404

**Workaround:**

 You can use "." (dot) as the version value instead of empty string:
```
  kind: AIServiceBackend
  metadata:
    name: perplexity
  spec:
    schema:
      name: OpenAI
      version: "."  # Workaround: dot bypasses cmp.Or and path.Join normalizes to /chat/completions
    backendRef:
      name: perplexity-backend
      kind: Backend
      group: gateway.envoyproxy.io
```
  This works because:
  1. "." is non-empty, so cmp.Or(".", "v1") returns "."
  2. path.Join("/", ".", "chat/completions") normalizes to /chat/completions

  Tested and confirmed working with Perplexity AI (api.perplexity.ai).

### LINKED ISSUE #1639: API: deprecate `VersionedAPISchema.Version` and introduce explicit config per schema
_Background_:

[VersionedAPISchema](https://aigateway.envoyproxy.io/docs/api/#versionedapischema) has been used for specifying the "schema" for each AIServiceBackend. For example, 

```yaml
kind: AIServiceBackend
spec:
  schema:
    name: AWSBedrock
...
```

this will tell the AIGW that this backend speaks the AWS Bedrock schema, such as /invoke API etc.  There's another field in VersionedAPISchema which is `Version` field. For example, 

```yaml
spec:
  schema:
    name: AzureOpenAI
    version: 2025-01-01-preview
```
This specifies the version of AzureOpenAI which _can_ serve multiple API versions.

However, the use of this generic field "version" feels a bit off for OpenAI compatible endpoint's case:

```yaml
spec:
  schema:
    name: OpenAI
    version: /v1beta/openai
```
This config can be access the Geimini model hosted on GCP's OpenAI compatible endpoint where the path is expected to be /v1beta/openai/chat/completions, etc. So it's essentially not a version but just a "prefix" or "base" in their terminology (but "base" usually includes the full host name + http(s) so it's not quite right here).

So the point here is that the interpretation of the field "version" varies and it's confusing. It's sometimes not used and sometimes used in a different sense etc. In other words, it's not really always a version in real sense, and  it's not always the "path" either (anthropic case).

_Proposal_:

I would propose to deprecate the existing version field and have an explicit per-schema config field for each schema such as 

```diff
diff --git a/api/v1alpha1/shared_types.go b/api/v1alpha1/shared_types.go
index 628f34f1..1aae61b1 100644
--- a/api/v1alpha1/shared_types.go
+++ b/api/v1alpha1/shared_types.go
@@ -15,6 +15,15 @@ package v1alpha1
 type VersionedAPISchema struct {
        // Name is the name of the API schema of the AIGatewayRoute or AIServiceBackend.
        //
+       // When OpenAI is populated in this field, you can optionally configure the prefix
+       // in openai field.
+       //
+       // When AzureOpenAI is populated in this field, you can optionally configure the version
+       // in azureOpenAI field.
+       //
+       // When GCPAnthropic is populated in this field, you can optionally configure the version
+       // in gcpAnthropic field.
+       //
        // +kubebuilder:validation:Enum=OpenAI;Cohere;AWSBedrock;AzureOpenAI;GCPVertexAI;GCPAnthropic;Anthropic;AWSAnthropic
        Name APISchema `json:"name"`
 
@@ -32,9 +41,50 @@ type VersionedAPISchema struct {
        //
        // When the name is set to AzureOpenAI, this version maps to "API Version" in the
        // Azure OpenAI API documentation (https://learn.microsoft.com/en-us/azure/ai-services/openai/reference#rest-api-versioning).
+       //
+       // Deprecated: This field is deprecated and will be removed in future releases.
        Version *string `json:"version,omitempty"`
+
+       OpenAI       *OpenAISchemaConfig       `json:"openai,omitempty"`
+       AzureOpenAI  *AzureOpenAISchemaConfig  `json:"azureOpenAI,omitempty"`
+       GCPAnthropic *GCPAnthropicSchemaConfig `json:"gcpAnthropic,omitempty"`
+       AWSAnthropic *AWSAnthropicSchemaConfig `json:"awsAnthropic,omitempty"`
+       ....
 }
 
+type (
+       // OpenAISchemaConfig defines the configuration for OpenAI compatible API schema.
+       OpenAISchemaConfig struct {
+               // Prefix is the prefix for OpenAI compatible API endpoints.
+               //
+               // For example, if the prefix is set to "/v1beta/openai", the "chat completions" API endpoint
+               // will be "/v1beta/openai/chat/completions".
+               //
+               // +kubebuilder:validation:Pattern=`^/.*`
+               Prefix string `json:"prefix"`
+       }
+       AzureOpenAISchemaConfig struct {
+               // Version is the API version for Azure OpenAI API.
+               //
+               // This maps to "API Version" in the Azure OpenAI API documentation
+               // (https://learn.microsoft.com/en-us/azure/ai-services/openai/reference#rest-api-versioning).
+               Version string `json:"version"`
+       }
+       GCPAnthropicSchemaConfig struct {
+               // Version is the API version for GCP Anthropic API.
+               //
+               // This maps to "API Version" in the GCP Anthropic API documentation
+               // (https://docs.anthropic.com/en/api/claude-on-vertex-ai).
+               Version string `json:"version"`
+       }
+       AWSAnthropicSchemaConfig struct {
+               // Version is the API version for AWS Anthropic API.
+               //
+               // This maps to "API Version" in the AWS Anthropic API documentation
+               // (https://docs.claude.com/en/api/claude-on-amazon-bedrock).
+               Version string `json:"version"`
+       }
+       // ...
+)
+
```

mathetake:
**Description**

This adds a new field 'prefix' as discussed in #1639. tldr is that the existing "version" field has been overused in various and confusing ways. Notably for OpenAI, it was literally used for expressing the prefix of the endpoints and not really a version in any sense.

Closes #1639 
Closes #1629

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -51,6 +51,38 @@ You will find which credentials are required in the output of the test command. 
 credentials are skipped by default when the credentials are not provided. If you
 want to run these tests locally, please prepare the necessary credentials by yourself.
 
+## Use of generative AI policy
+
+### Goals
+
+- Keep Envoy AI Gateway code consistent and high-quality.
+- Save reviewers time. Reviewers are often in short supply and time, so try to avoid wasted time.
+
+### What is allowed
+
+Use of generative AI to assist in writing code or tests, as long as the submitter fully understands
+the code being submitted.
+
+All of the following are required:
+
+- You understand the change you are submitting.
+- You respond to questions and comments from the reviewer. If you use generative AI to help in your
+  responses, you are required to edit and proof read the AI-generated responses, and ensure it is
+  a reasonable response to the question or issue raised.
+- You are able to revise the AI-generated code if requested by the reviewer. You are responsible for
+  ensuring issues are addressed, even if your AI assistant is unable.
+- You are transparent about your AI usage. It is often helpful to a reviewer to know that an AI tool
+  was used; please include that information in the PR description.
+- All generated code must be released under the same [license](LICENSE) as Envoy. You are responsible
+  for ensuring that the tools you use to generate code do not add any additional licensing restrictions.
+
+### What is not allowed
+
+- PRs which the submitter does not understand or take full ownership of.
+- Code comments should be valuable to the codebase. Any comments which only help the AI interact with
+  the code must be removed before the PR is submitted. Comments which explain what straightforward code
+  does are not useful.
+
 ## Opening a Pull Request
 
 We employ almost the same guidelines as the [envoyproxy/envoy](https://github.com/envoyproxy/envoy/blob/main/CONTRIBUTING.md#submitting-a-pr).


site/docs/capabilities/llm-integrations/supported-providers.md
```diff
@@ -12,24 +12,24 @@ Below is a table of currently supported providers and their respective configura
...
+| [Together AI](https://docs.together.ai/docs/openai-api-compatibility)                                     |                                   `{"name":"OpenAI","prefix":"/v1"}`                                   |                         [API Key]                         |   ✅   |                                                                                                                                                        |
+| [Cohere](https://docs.cohere.com/v2/docs/compatibility-api)                                               |                `{"name":"Cohere","version":"v2"}` or `{"name":"OpenAI","prefix":"/v1"}`                |                         [API Key]                         |   ✅   | Supports native Cohere v2 (e.g., /cohere/v2/rerank) and OpenAI-compatible endpoints.                                                                   |
```
gavrissh: It should be  - `{"name":"OpenAI","prefix":"/compatibility/v1"}`
ref - https://docs.cohere.com/v2/docs/compatibility-api#basic-chat-completions

PR #1668, opened 2025-12-17
fix: enable ext_proc filter and set HTTPS scheme for Azure OpenAI

andreaswachs:
**Description**

This PR fixes two critical details with Azure OpenAI integration:

1. ext_proc Filter Not Running (BackendSecurityPolicy auth skipped) The ext_proc filter was disabled at the listener level, creating a chicken-and-egg problem: ext_proc must run BEFORE routing to set the x-ai-eg-model header used for route matching, but per-route config only applies AFTER a route matches. Fixed by enabling ext_proc at the listener level.

2. Azure OpenAI Rejects Requests with :scheme http Azure OpenAI rejects proxied requests because Envoy sends :scheme: http instead of :scheme: https. This is an Envoy Gateway limitation - it correctly creates TLS transport sockets but doesn't configure upstream HTTPS scheme. Fixed by adding :scheme: https header mutation in the Azure OpenAI translator.

We've implemented the reference architecture and use the Envoy AI Gateway as an internal gateway where we talk HTTP to the proxies. We've had issues with the Azure OpenAI API only accepting HTTPS requests so we propose this change to make it such that independent of the incoming requests to the Envoy Proxies for the gateway, we always send HTTPS requests to Azure OpenAI.

Since I don't have the expertise to fully understand the consequences of the 1st change (Thanks Claude Code), so I appreciate any helpful comments to correct this if needed. And accept any scolding for doing crazy stuff.

### DIFF
--- a/internal/translator/openai_azureopenai.go
+++ b/internal/translator/openai_azureopenai.go
@@ -51,7 +51,7 @@ func (o *openAIToAzureOpenAITranslatorV1ChatCompletion) RequestBody(raw []byte, 
 	pathTemplate := "/openai/deployments/%s/chat/completions?api-version=%s"
 	newHeaders = []internalapi.Header{
 		{pathHeaderName, fmt.Sprintf(pathTemplate, modelName, o.apiVersion)},
-		{schemeHeaderName, "https"},
+		{schemeHeaderName, protocolHTTPS},
 	}
 	if req.Stream {
 		o.stream = true
--- a/internal/translator/openai_azureopenai_embeddings.go
+++ b/internal/translator/openai_azureopenai_embeddings.go
@@ -54,7 +54,7 @@ func (o *openAIToAzureOpenAITranslatorV1Embedding) RequestBody(original []byte, 
 	}
 	newHeaders = []internalapi.Header{
 		{pathHeaderName, fmt.Sprintf(pathTemplate, modelName, o.apiVersion)},
-		{schemeHeaderName, "https"},
+		{schemeHeaderName, protocolHTTPS},
 	}
 
 	if len(newBody) > 0 {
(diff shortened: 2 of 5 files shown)


mathetake:
can you narrow down the PR into the second issue? Let's resolve one by one to have a smaller context for discussion.

> Azure OpenAI Rejects Requests with :scheme http Azure OpenAI rejects proxied requests because Envoy sends :scheme: http instead of :scheme: https. This is an Envoy Gateway limitation - it correctly creates TLS transport sockets but doesn't configure upstream HTTPS scheme. Fixed by adding :scheme: https header mutation in the Azure OpenAI translator.

internal/extensionserver/post_translate_modify.go
```diff
@@ -581,17 +585,35 @@ func (s *Server) patchVirtualHostWithInferencePool(vh *routev3.VirtualHost, infe
...
+				ProcessingMode: &extprocv3.ProcessingMode{
+					RequestHeaderMode:   extprocv3.ProcessingMode_SEND,
+					RequestBodyMode:     extprocv3.ProcessingMode_BUFFERED,
+					RequestTrailerMode:  extprocv3.ProcessingMode_SKIP,
+					ResponseHeaderMode:  extprocv3.ProcessingMode_SEND,
+					ResponseBodyMode:    extprocv3.ProcessingMode_BUFFERED,
+					ResponseTrailerMode: extprocv3.ProcessingMode_SKIP,
+				},
+			},
+		},
```
yuzisun, line 606: The extension server is already getting complex, can we fix these at the envoy gateway level? It does not seem like ai gateway specific.

yuzisun:
These issues are not ai gateway specific, I would prefer fixing at the envoy gateway level instead of hacking in the extension server.

yuzisun:
We haven't ran into any issue with Azure OpenAI via proxy, are you sure it is not your configuration issue ?

andreaswachs:
> We haven't ran into any issue with Azure OpenAI via proxy, are you sure it is not your configuration issue ?

That is always a real possibility. I will triple check the gateway setup and get back to you. I have focused solely on the AI gateway part

mathetake:
Let's close it until then!

PR #1681, opened 2025-12-23
feat: add anthropic bedrock cache

alexagriffith:
**Description**

  This commit adds prompt caching support for AWS Bedrock Claude models through the existing cache_control API,
  enabling provider-agnostic caching across all Claude deployments (Direct Anthropic, GCP & AWS).

  The implementation translates OpenAI-compatible cache_control fields to AWS Bedrock's cachePoint
  format, allowing users to use the same caching syntax regardless of the backend provider. This enables
  cost optimization and performance improvements through content caching for system prompts, user messages, and tool
  definitions.

  Key changes include:
  - Extended AWS Bedrock schema to support CachePoint fields on ContentBlock, SystemContentBlock, and Tool structures
  - Added translation logic in the OpenAI-to-Bedrock translator to apply cache points when cache_control is specified
  - Test coverage for all cache control scenarios
  - Example documentation showing usage across all Claude providers


examples/cache/cache_control.md
```diff
@@ -0,0 +1,320 @@
...
+
+## Example Requests
+
+### Basic System Prompt Caching
+
+This example works identically across all providers - just change the model name:
+
+```bash
+# Direct Anthropic
+curl -X POST http://localhost:8080/v1/chat/completions \
```
yuzisun: This should be using /anthropic/v1/messages endpoint

examples/cache/cache_control.md
```diff
@@ -0,0 +1,320 @@
...
+
+To switch between providers while maintaining caching:
+
+1. **Update model names** in your requests
+2. **Review provider-specific limitations** if applicable
+3. **No other code changes required** - the cache_control syntax is identical
+
+## Implementation Notes
+
+- **Anthropic Direct**: Native cache_control support
```
yuzisun, line 309: Anthropic Native API are supported for Anthropic Direct, AWS and GCP

internal/translator/openai_awsbedrock.go
```diff
@@ -71,6 +71,34 @@ func getAwsBedrockThinkingMap(tu *openai.ThinkingUnion) map[string]any {
...
 }
 
+// createCachePointBlock creates a cache point block for AWS Bedrock when cache control is enabled.
+func createCachePointBlock() *awsbedrock.CachePointBlock {
+	return &awsbedrock.CachePointBlock{
+		Type: "default",
+	}
+}
+
+// applyCacheControlToContentBlock adds a cache point to a content block if cache control is enabled.
```
aabchoo: CachePoint is `json:"cachePoint,omitempty"`. Setting the value to nil is the same as not setting the value.

What do you think about removing the 3 apply functions -> having a function that returns cachePoint or nil.

Then you can write the following:
```
&awsbedrock.ContentBlock{
	Text: &textContentPart.Text,
     CachePoint: getCacheControl() // nil or valid
 }
```

examples/cache/cache_control.md
alexagriffith: ah good catch ty

examples/cache/cache_control.md
alexagriffith: I guess what im saying is that the others are using openai translation

alexagriffith: is what i have now better?

internal/translator/openai_awsbedrock.go
alexagriffith: yeah thats better

alexagriffith: fixed

aabchoo:
Overall looks good to me. Deferring to @yuzisun for remaining comments

yuzisun (approved):

PR #1684, opened 2025-12-24
feat: cel based mcp authorization

zhaohuabing:
**Description**

This PR introudces a more flexible CEL-based MCP authorization, which algins with the k8s networking agentic API proposal: https://github.com/kubernetes-sigs/kube-agentic-networking/pull/17/changes .

Unlike the Agentic API, our API is primarily intended for end-user authorization. For this use case, a more type-safe and user-friendly interface should probably be the primary option, since these policies are more likely to be configured by an administrator after deployment and tailored to individual users. 

However, CEL is still a valuable addition as it provides a dynamic and flexible way to configure authorization for more advanced use cases.

### DIFF
--- a/.github/workflows/build_and_test.yaml
+++ b/.github/workflows/build_and_test.yaml
@@ -165,7 +165,6 @@ jobs:
           TEST_GROK_API_KEY: ${{ secrets.ENVOY_AI_GATEWAY_GROK_API_KEY }}
           TEST_SAMBANOVA_API_KEY: ${{ secrets.ENVOY_AI_GATEWAY_SAMBANOVA_API_KEY }}
           TEST_DEEPINFRA_API_KEY: ${{ secrets.ENVOY_AI_GATEWAY_DEEPINFRA_API_KEY }}
-          TEST_GITHUB_ACCESS_TOKEN: ${{ secrets.TEST_GITHUB_ACCESS_TOKEN }}
         run: make test-data-plane
 
   test_data_plane_mcp:
@@ -193,6 +192,8 @@ jobs:
             ~/go/bin
           key: data-plane-tests-${{ hashFiles('**/go.mod', '**/go.sum', '**/Makefile') }}
       - run: make test-data-plane-mcp
+        env:
+          TEST_GITHUB_ACCESS_TOKEN: ${{ secrets.TEST_GITHUB_ACCESS_TOKEN }}
 
   test_e2e:
     needs: changes
--- a/internal/apischema/anthropic/anthropic.go
+++ b/internal/apischema/anthropic/anthropic.go
@@ -576,3 +576,20 @@ func (m *MessagesStreamChunk) UnmarshalJSON(data []byte) error {
 	}
 	return nil
 }
+
+// ErrorResponse represents an error response from the Anthropic API.
+// https://platform.claude.com/docs/en/api/errors
+type ErrorResponse struct {
+	Error     ErrorResponseMessage `json:"error"`
+	RequestID string               `json:"request_id"`
+	// Type is always "error".
+	Type string `json:"type"`
+}
+
+// ErrorResponseMessage represents the error message in an Anthropic API error response
+// which corresponds to the HTTP status code.
+// https://platform.claude.com/docs/en/api/errors#http-errors
+type ErrorResponseMessage struct {
+	Message string `json:"message"`
+	Type    string `json:"type"`
+}
--- a/internal/translator/anthropic_anthropic.go
+++ b/internal/translator/anthropic_anthropic.go
@@ -12,6 +12,7 @@ import (
 	"fmt"
 	"io"
 	"strconv"
+	"strings"
 
 	"github.com/tidwall/sjson"
 
@@ -188,12 +189,52 @@ func (a *anthropicToAnthropicTranslator) updateTotalTokens() {
 }
 
 // ResponseError implements [AnthropicMessagesTranslator] for Anthropic to AWS Bedrock Anthropic translation.
-func (a *anthropicToAnthropicTranslator) ResponseError(map[string]string, io.Reader) (
+func (a *anthropicToAnthropicTranslator) ResponseError(respHeaders map[string]string, r io.Reader) (
 	newHeaders []internalapi.Header,
 	mutatedBody []byte,
 	err error,
 ) {
-	// TODO: implement the non-anthropic error conversion logic here. For now, we just return the original error
-	// 	from the upstream as-is.
+	statusCode := respHeaders[statusHeaderName]
+	if !strings.Contains(respHeaders[contentTypeHeaderName], jsonContentType) {
+		buf, err := io.ReadAll(r)
+		if err != nil {
+			return nil, nil, fmt.Errorf("failed to read error body: %w", err)
+		}
+		var typ string
+		switch statusCode {
+		case "400":
+			typ = "invalid_request_error"
+		case "401":
+			typ = "authentication_error"
+		case "403":
+			typ = "permission_error"
+		case "404":
+			typ = "not_found_error"
+		case "413":
+			typ = "request_too_large"
+		case "429":
+			typ = "rate_limit_error"
+		case "500":
+			typ = "internal_server_error"
+		case "503":
+			typ = "service_unavailable_error"
+		case "529":
+			typ = "overloaded_error"
+		default:
+			typ = "internal_server_error"
+		}
+		anthropicError := anthropic.ErrorResponse{
+			Type:  "error", // Always "error" at the top level.
+			Error: anthropic.ErrorResponseMessage{Type: typ, Message: string(buf)},
+		}
+		mutatedBody, err = json.Marshal(anthropicError)
+		if err != nil {
+			return nil, nil, fmt.Errorf("failed to marshal error body: %w", err)
+		}
+		newHeaders = append(newHeaders,
+			internalapi.Header{contentTypeHeaderName, jsonContentType},
+			internalapi.Header{contentLengthHeaderName, strconv.Itoa(len(mutatedBody))},
+		)
+	}
 	return
 }
--- a/internal/translator/anthropic_anthropic_test.go
+++ b/internal/translator/anthropic_anthropic_test.go
@@ -6,6 +6,7 @@
 package translator
 
 import (
+	"encoding/json"
 	"strconv"
 	"strings"
 	"testing"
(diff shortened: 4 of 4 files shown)


api/v1alpha1/mcp_route.go
```diff
@@ -268,6 +268,23 @@ type MCPRouteAuthorizationRule struct {
...
+	//   - `request.headers["X-Custom-Header"] == "AllowedValue"`
+	//   - `request.mcp.tool in ["toolA", "toolB"]`
+	//
+	// Note: The CEL expression support is experimental, and the attributes
+	// available to the expression may change in future releases.
+	//
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:MaxLength=4096
+	// +optional
+	CEL *string `json:"cel,omitempty"`
```
zhaohuabing, line 300: I previously discussed with @nacx that this should be mutually exclusive with source / target. However, on further thought, it might make more sense to allow CEL to coexist with source / target as an optional additional condition.

If both source / target and CEL are specified, all of them would need to match for the rule to apply. This would let users define a simple rule using source / target, while adding extra conditions that aren’t supported there but are available through a CEL expression, like "request.headers".

api/v1alpha1/mcp_route.go
```diff
@@ -320,15 +337,6 @@ type ToolCall struct {
...
 	Tool string `json:"tool"`
-
-	// When is a CEL expression that must evaluate to true for the rule to match.
-	// The expression is evaluated with a single variable "args" bound to the tool call arguments as a dynamic object.
-	// Guard against missing fields with null checks (e.g., args["foo"] != null && args["foo"]["bar"] == "val").
-	//
-	// +kubebuilder:validation:Optional
-	// +kubebuilder:validation:MaxLength=4096
-	// +optional
-	When *string `json:"when,omitempty"`
```
zhaohuabing, line 343: We can replace "when" with an explicit "args" field here if needed.

api/v1alpha1/mcp_route.go
nacx: Looks good to me. Can you add the available CEL variables in the API docs, so users know what can be used in the expressions? A reasonable initial proposal, IMHO would be:

```
# General HTTP-related properties that may be useful (especially the auth one)
request.method       # GET / POST, etc
request.headers      # Map of request headers msp[string]string (no multi-value headers support probably)
request.path         # Path like /mcp, etc
request.auth.jwt     # map[string]string with all JWT token claims, if a bearer token is present AND it is a JWT (not opaque) token

# MCP specific properties
request.mcp.method         # tools/list, tools/call, etc
request.mcp.backend        # name of the upstream backend (kiwi, github)
request.mcp.tool           # tool being called without the backend prefix (list_issues, etc)
request.mcp.params         # params of the MCP method. Include keys as "_meta", "arguments" in tool calls, etc.
```

internal/mcpproxy/authorization.go
```diff
@@ -106,30 +108,40 @@ func (m *MCPProxy) authorizeRequest(authorization *compiledAuthorization, header
...
-		scopeSet = sets.New(extractScopes(tokenClaims)...)
+		scopeSet = sets.New(extractScopes(claims)...)
 
 		// Scopes are handled separately, remove them from the claims map to avoid interference.
-		delete(tokenClaims, "scope")
+		delete(claims, "scope")
 	}
 
 	var requiredScopesForChallenge []string
+	requestForCEL := buildRequestForCEL(req, claims, scopeSet)
```
mathetake: maybe we should avoid this map allocation when there's no CEL

internal/mcpproxy/authorization.go
```diff
@@ -200,7 +249,28 @@ func extractScopes(claims jwt.MapClaims) []string {
 	}
 }
 
-func (m *MCPProxy) toolMatches(backend, tool string, tools []compiledToolCall, args any) bool {
+func (m *MCPProxy) evalRuleCEL(rule compiledAuthorizationRule, request map[string]any) (bool, bool) {
+	if rule.celProgram == nil {
+		return true, true
+	}
+	result, _, err := rule.celProgram.Eval(map[string]any{"request": request})
```
mathetake: this map can be allocated once outside, instead of allocating here?

api/v1alpha1/mcp_route.go
```diff
@@ -269,6 +269,36 @@ type MCPRouteAuthorizationRule struct {
 	// +kubebuilder:validation:Optional
 	Target *MCPAuthorizationTarget `json:"target,omitempty"`
 
+	// CEL specifies a Common Expression Language (CEL) expression evaluated for this rule.
+	// The expression must return a boolean; evaluation errors or non-boolean results
+	// are treated as "no match".
+	//
+	// Example CEL expressions:
+	//	* `request.method == "POST"`
```
mathetake, line 277: curious to know what would be the other prefixes than `request` to be added later?

zhaohuabing: Currently, I can only think of `request`. kube-agentic-networking also supports Identities, but since their `Identities` are workload-scoped, I don’t think we need to support that. We could probably also support some context information from the LLM conversation in the future.

nacx: Most are related to the request, so the prefix makes sense. Also, I think it is good to keep attributes close (use the same where possible) to the [Envoy upstream ones](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/advanced/attributes.html) so that it's easy to use the same features upstream when available.

api/v1alpha1/mcp_route.go
```diff
@@ -269,6 +269,36 @@ type MCPRouteAuthorizationRule struct {
...
+	//	* `request.mcp.tool in ["toolA", "toolB"]`
+	//
+	// Available attributes in the CEL expression:
+	//
+	//	* request.method: HTTP method such as GET or POST. Type: string.
+	//	* request.headers: map of headers with lowercased keys, first value only. Type: map[string]string.
+	//	* request.headers_all: map of headers with lowercased keys, all values. Type: map[string][]string.
+	//	* request.path: request path such as /mcp. Type: string.
+	//	* request.auth.jwt.claims: JWT claims when a bearer JWT is present. Type: map[string]string.
+	//	* request.auth.jwt.scopes: JWT scopes when a bearer JWT is present. Type: []string.
```
nacx: Should we collapse this into a single one? (claims values could be of any type)
```
request.auth.jwt: JWT claims when a bearer JWT is present. Type: map[string]any.
```

internal/mcpproxy/authorization.go
```diff
@@ -29,15 +29,23 @@ type compiledAuthorization struct {
...
+	Headers    http.Header
+	HTTPMethod string
+	Host       string
+	HTTPPath   string
+	MCPMethod  string
 	Backend    string
 	Tool       string
-	Expression string
-	program    cel.Program
+	Params     any
```
nacx: This should probably be `mcp.Params` as all param types implement it.
```suggestion
	Params     mcp.Params
```

internal/mcpproxy/authorization.go
```diff
@@ -200,7 +256,28 @@ func extractScopes(claims jwt.MapClaims) []string {
 	}
 }
 
-func (m *MCPProxy) toolMatches(backend, tool string, tools []compiledToolCall, args any) bool {
+func (m *MCPProxy) evalRuleCEL(rule compiledAuthorizationRule, activation map[string]any) (bool, bool) {
+	if rule.celProgram == nil {
+		return true, true
+	}
```
nacx: nit: This is unnecessary as we only call this when the program is present.

internal/mcpproxy/authorization.go
```diff
@@ -200,7 +256,28 @@ func extractScopes(claims jwt.MapClaims) []string {
 	}
 }
 
-func (m *MCPProxy) toolMatches(backend, tool string, tools []compiledToolCall, args any) bool {
+func (m *MCPProxy) evalRuleCEL(rule compiledAuthorizationRule, activation map[string]any) (bool, bool) {
```
nacx: Can we make the method return `(bool, error)` instead? I think the method will be easier to understand.
This way, we can better return the error when evaluating the rule.
In the case of a non-boolean result, we can also return an error indicating that.

internal/mcpproxy/authorization.go
```diff
@@ -106,30 +108,43 @@ func (m *MCPProxy) authorizeRequest(authorization *compiledAuthorization, header
...
-		if _, _, err := jwt.NewParser().ParseUnverified(token, tokenClaims); err != nil {
+		if _, _, err := jwt.NewParser().ParseUnverified(token, claims); err != nil {
 			m.l.Info("failed to parse JWT token", slog.String("error", err.Error()))
 		}
-		scopeSet = sets.New(extractScopes(tokenClaims)...)
+		scopeSet = sets.New(extractScopes(claims)...)
 
 		// Scopes are handled separately, remove them from the claims map to avoid interference.
-		delete(tokenClaims, "scope")
+		delete(claims, "scope")
```
nacx: nit: we don't need to do this if we fail to parse the JWT token ( it is maybe an opaque token).

internal/mcpproxy/handlers.go
```diff
@@ -663,7 +663,20 @@ func (m *MCPProxy) handleToolCallRequest(ctx context.Context, s *session, w http
...
+		}
+		allowed, requiredScopes := m.authorizeRequest(route.authorization, authorizationRequest{
+			Headers:    r.Header,
+			HTTPMethod: r.Method,
+			Host:       r.Host,
+			HTTPPath:   httpPath,
+			MCPMethod:  req.Method,
+			Backend:    backendName,
+			Tool:       toolName,
+			Params:     p.Arguments,
```
nacx: I think this should be just `p`, otherwise we won't give access to `_meta` to the CEL expressions.

nacx (commented):
Thanks! Some small nits!

api/v1alpha1/mcp_route.go
zhaohuabing: Change the API comment of the request.jwt.claims to map[string]any. It was a typo - thanks for catching it!
 
> Should we collapse this into a single one? (claims values could be of any type)

I'm inclined to keep them as two separate fields. Scopes are a bit special since they represent permissions and are treated differently elsewhere as well - for example, the MCP spec explicitly defines scopes related authorization, and both EG and AIEG use a dedicated scopes field for authorization.

nacx: Sounds good!

PR #1687, opened 2025-12-25
extproc: adds annotation when log level changes

aabchoo:
**Description**

Annotations are only added to deamonset and deployment when extproc's pod image does not match the controller's expected extproc image. The annotation helps restart the extproc which results in the extproc image updating.

I want to have that added for log levels as well. Right now, updating the extproc log level via helm values will not trigger a redeployment, so extproc will not have the expected log level. This PR adds logic to add annotations if there is a mismatch between extproc pod and controller's expected log level.

Needed to update separate test `deploymen1` -> `deployment2` (already passing bc the annotation was added from previous test case).

### DIFF
--- a/site/src/data/talks.json
+++ b/site/src/data/talks.json
@@ -1,5 +1,13 @@
 [
     {
+        "date": "2026-03-23",
+        "title": "Scaling LLMs at Bloomberg Using Envoy AI Gateway: A Hybrid Cloud Approach",
+        "url": "https://colocatedeventseu2026.sched.com/event/2DY8Z/scaling-llms-at-bloomberg-using-envoy-ai-gateway-a-hybrid-cloud-approach-xiaolin-lin-bloomberg",
+        "event": "KubeCon Europe 2026",
+        "speaker": "Xiaolin Lin (Bloomberg), Aaron Choo (Bloomberg)",
+        "description": "This session examines how financial institutions navigate the complex security and governance requirements of GenAI by building resilient infrastructure that spans cloud and on-premise environments. It details Bloomberg's production implementation of the Envoy AI Gateway, showcasing how features like intelligent routing and secure authentication enable the safe, scalable delivery of real-time AI traffic."
+    },
+    {
         "date": "2025-11-13",
         "title": "Inference Awakens: Tools for the Age of GenAI",
         "url": "https://www.youtube.com/watch?v=nlTk5q3MPZ4",


internal/controller/gateway.go
```diff
@@ -701,6 +703,13 @@ func (c *GatewayController) annotateGatewayPods(ctx context.Context,
 				// If there's an extproc sidecar container with the current target image, we don't need to roll out the deployment.
 				if podSpec.InitContainers[i].Name == extProcContainerName && podSpec.InitContainers[i].Image == c.extProcImage {
 					rollout = false
+					for j := range podSpec.InitContainers[i].Args {
+						// logLevel arg should be indexed 2 based on gateway_mutator.go, but we check all args to be safe.
+						if j > 0 && podSpec.InitContainers[i].Args[j-1] == "-logLevel" && podSpec.InitContainers[i].Args[j] != c.extProcLogLevel {
+							rollout = true
+							break
+						}
+					}
```
mathetake, line 775: while there could you share the implementation between init container and normal container

mathetake (approved):
I think this makes sense -- JFYI once https://github.com/envoyproxy/ai-gateway/pull/1623 lands, log level etc should be added to the GatewayConfig CRD so that we naturally reconcile the config change like this automatically

internal/controller/gateway.go
nacx: +1. Ideally, we could have a list of predicates, like `func(corev1.Container) bool` to be used to determine if a rollout is needed, and have both init and normal containers evaluate those predicates.
Then we can easily define the predicate variables to reuse them easily, test them, etc.

mathetake: gently ping @aabchoo for addressing the comment here

aabchoo: Yeah, I will work on that within the next week! Was a bit distracted with some other things and haven't had time to read over the new changes. Apologies for the delay

internal/controller/gateway.go
```diff
@@ -772,6 +781,12 @@ func (c *GatewayController) annotateGatewayPods(ctx context.Context,
 				// If there's an extproc container with the current target image, we don't need to roll out the deployment.
 				if podSpec.Containers[i].Name == extProcContainerName && podSpec.Containers[i].Image == c.extProcImage {
 					hasSideCar = true
+					for j := range podSpec.Containers[i].Args {
+						if j > 0 && podSpec.Containers[i].Args[j-1] == "-logLevel" && podSpec.Containers[i].Args[j] != c.extProcLogLevel {
+							hasSideCar = false
```
aabchoo, line 786: Hey @mathetake, could you confirm that this change is fine? 

Added this extra side car bool override w.r.t your latest changes

mathetake: looks fine to me

PR #1689, opened 2025-12-26
test: uses only k8s latest for CRD/Controller tests

mathetake:
**Description**

It is redundant to use multiple k8s versions with envtest driven e2e tests, so this switches to run only on the latest


aabchoo:
In [Build and Test / Controller Test (ubuntu-latest) (pull_request)](https://github.com/envoyproxy/ai-gateway/actions/runs/20526838996/job/58971621663?pr=1689) I see `fatal: No names found, cannot describe anything.`

mathetake:
wow good catch

PR #1692, opened 2025-12-26
feat: adds version matching check for filter config

mathetake:
**Description**

This adds a version matching check in filter config loading on the extproc side, which allows us to introduce any breaking change in filter config after the release of the next version.


aabchoo:
You need to update this filter config https://github.com/envoyproxy/ai-gateway/blob/a55f4687feacf5ad799ed13d0e6a97bdafbd48a2/tests/extproc/real_providers_test.go#L41

mathetake:
ah

PR #1699, opened 2025-12-28
test: reorganizes  extproc tests

mathetake:
**Description**

This refactors tests/extproc integration tests:
1. Renamed as tests/data-plane to be usable for Dynamic Module too #1564 
2. Extract tests/extproc/mcp into tests/data-plane-mcp to have a separate concern in each integration tests

Preparation for #90

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -46,7 +46,7 @@ For example,
 Other than that, everything will be automatically managed and installed via `make` targets,
 and you should not need to worry about the dependencies (tell us if you do).
 
-Additionally, some of the test cases in `test-e2e` and `test-extproc` might require some credentials.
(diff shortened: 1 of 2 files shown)


.github/workflows/build_and_test.yaml
```diff
@@ -102,7 +102,7 @@ jobs:
             ~/.cache/go-build
             ~/go/pkg/mod
             ~/go/bin
-          key: celvalidation-test-${{ hashFiles('**/go.mod', '**/go.sum', '**/Makefile') }}-${{ matrix.os }}
+          key: controller-test-${{ hashFiles('**/go.mod', '**/go.sum', '**/Makefile') }}-${{ matrix.os }}
```
mathetake, line 105: use the same key as controller test for reducing cache purging

PR #1703, opened 2025-12-29
test(refactor): executes testupstream in process to be consistent

mathetake:
**Description**

This refactors the testupstream code so that it can be executed in-process like fakeopenai, etc.


tests/internal/testupstreamlib/server.go
mathetake, line 1: this is extracted from the original main.go and nothing has changed

tests/internal/dataplaneenv/dataplaneenv.go
```diff
@@ -28,22 +28,11 @@ import (
 	testsinternal "github.com/envoyproxy/ai-gateway/tests/internal"
 )
 
-var (
-	// extprocBin holds the path to the compiled extproc binary.
-	extprocBin string
-
-	// TestupstreamBin holds the path to the compiled testupstream binary.
-	TestupstreamBin string
```
mathetake, line 36: this is the main purpose that was inconsistent (vs fakeopenai, etc)

PR #1707, opened 2025-12-30
fix: helm install to support global.imagePullSecrets fallback for umbrella charts

### LINKED ISSUE #1706: support global.imagePullSecrets fallback for controller/extProc
_Description_:

> In umbrella charts, users want to set a single image pull secret list once (under `global.imagePullSecrets`) and have all subcharts consume it. Today `ai-gateway-helm` only reads `controller.imagePullSecrets` and `extProc.imagePullSecrets`, so umbrella charts must repeat the same values per-subchart.

[optional *Relevant Links*:]
https://github.com/envoyproxy/ai-gateway/blob/main/manifests/charts/ai-gateway-helm/values.yaml#L34
https://github.com/envoyproxy/ai-gateway/blob/main/manifests/charts/ai-gateway-helm/values.yaml#L98

_Proposal_:
> If .Values.controller.imagePullSecrets is empty, fallback to .Values.global.imagePullSecrets.
> If .Values.extProc.imagePullSecrets is empty, fallback to .Values.global.imagePullSecrets.

Why: Standard Helm pattern; improves UX; avoids secret duplication across charts.

_Acceptance criteria_:
- helm template works when only global.imagePullSecrets is set.
- Existing per-component fields still override global values.

Example values:
```
global:
    imagePullSecrets:
      - name: regcred
```

vijaypal89:
**Description**

Allow umbrella charts to set imagePullSecrets once via global.imagePullSecrets. If controller.imagePullSecrets / extProc.imagePullSecrets are unset, the chart falls back to global.imagePullSecrets for:
- Deployment spec.imagePullSecrets (controller)
- --extProcImagePullSecrets arg generation

Fixes #1706


manifests/charts/ai-gateway-helm/templates/_helpers.tpl
```diff
@@ -113,9 +113,24 @@ Convert extraEnvVars array to semicolon-separated string for extProc
 Convert imagePullSecrets array to semicolon-separated string for extProc
 */}}
 {{- define "ai-gateway-helm.extProc.imagePullSecretsString" -}}
+{{- $src := .Values.extProc.imagePullSecrets -}}
+{{- if not $src -}}
+{{- $src = .Values.global.imagePullSecrets -}}
+{{- end -}}
```
nacx:
```suggestion
{{- $src := default .Values.global.imagePullSecrets .Values.extProc.imagePullSecrets -}}
```

manifests/charts/ai-gateway-helm/templates/_helpers.tpl
```diff
@@ -113,9 +113,24 @@ Convert extraEnvVars array to semicolon-separated string for extProc
...
+
+{{/*
+Returns controller imagePullSecrets if defined, otherwise falls back to global.imagePullSecrets
+*/}}
+{{- define "ai-gateway-helm.controller.imagePullSecrets" -}}
+{{- if .Values.controller.imagePullSecrets -}}
+{{- .Values.controller.imagePullSecrets | toYaml -}}
+{{- else if .Values.global.imagePullSecrets -}}
+{{- .Values.global.imagePullSecrets | toYaml -}}
+{{- end -}}
```
nacx:
```suggestion
{{- default .Values.global.imagePullSecrets .Values.controller.imagePullSecrets | toYaml -}}
```

manifests/charts/ai-gateway-helm/templates/deployment.yaml
```diff
@@ -24,9 +24,9 @@ spec:
       labels:
         {{- include "ai-gateway-helm.controller.selectorLabels" . | nindent 8 }}
     spec:
-      {{- with .Values.controller.imagePullSecrets }}
+      {{- if include "ai-gateway-helm.controller.imagePullSecrets" . }}
```
nacx: Here we are executing the helm template twice, when it's unnecessary... Probably we don't even need to include a helper and just use here the proposed default with the `with` statement>

manifests/charts/ai-gateway-helm/templates/deployment.yaml
```diff
@@ -46,7 +46,7 @@ spec:
             - --extProcImage={{ .Values.extProc.image.repository }}:{{ .Values.extProc.image.tag | default .Chart.AppVersion }}
             - --extProcImagePullPolicy={{ .Values.extProc.imagePullPolicy }}
             - --extProcLogLevel={{ .Values.extProc.logLevel }}
-            {{- if .Values.extProc.imagePullSecrets }}
+            {{- if include "ai-gateway-helm.extProc.imagePullSecretsString" . }}
```
nacx: Same here, we're executing the thing twice, when it's not needed.

manifests/charts/ai-gateway-helm/templates/_helpers.tpl
vijaypal89: Addressed

manifests/charts/ai-gateway-helm/templates/_helpers.tpl
vijaypal89: Addressed

manifests/charts/ai-gateway-helm/templates/deployment.yaml
vijaypal89: This makes sense. updated

manifests/charts/ai-gateway-helm/templates/deployment.yaml
vijaypal89: don.

manifests/charts/ai-gateway-helm/values.yaml
```diff
@@ -5,6 +5,14 @@
 
 # Default values for ai-gateway-helm.
 
+# Global values shared across this chart and its subcharts (umbrella charts can set these once).
+global:
+  # global.imagePullSecrets -- references to pre-created imagePullSecrets in the target namespace
```
mathetake, line 10: this needs a comment about the precedences over the component specific secrets

vijaypal89: comments updated.

Issue #1571, opened 2025-11-25
Support for token quota with failover


yanavlasov:
_Description_:

A token quota is defined on the AIGatewayRouteRuleBackendRef. Token quota will include maximum number of tokens, a conversion function from consumed tokens to consumed quota. Initially conversion function will be a simple multiplication factor and later on can be a CEL expression. quota is specified similarly to rate limit as the number of tokens per time interval.

If a quota is exceeded request is automatically failed over to the next backend ref in the AIGatewayRouteSpec. If all backend refs have exceeded their quota an 429 error response is served.

mathetake:
I like this one. Will this quota be configured per backend/backendRef instead of per downstream user (attribute) right? One thing i feel tricky to implement is that the "rejection" by nature will happen at cluster filters but at that point ext_proc cannot force a retry since the upstream filters' local reply are not retry-able... i hope there is some workaround

aabchoo:
I think we should configure per backendRef as quota calculation can differ per model with the same provider. 

Would be great if we can add attributes to distinguish usage for quota similar to [rate limit distinct header type](https://gateway.envoyproxy.io/v1.2/api/extension_types/#headermatch). Ie goal is each user consumes quota based on their usage

johnugeorge:
Thanks @yanavlasov for this

Yes. quota calculation is different for each provider and per backendref config makes sense

Can we reuse our cost concept for quota ? It is flexible with CEL, client selectors from headers etc 

eg:  quota definition to be similar to our cost definition (Can we reuse in https://github.com/envoyproxy/ai-gateway/blob/a2ba2c9ed8e7488d6fd76dbcbe08ac7008170e5d/examples/token_ratelimit/token_ratelimit.yaml#L50-L63)

And quota usage to similar to our current cost consumption https://github.com/envoyproxy/ai-gateway/blob/a2ba2c9ed8e7488d6fd76dbcbe08ac7008170e5d/examples/token_ratelimit/token_ratelimit.yaml#L106-L206

yuzisun:
One important thing to note is that we should differentiate token rate limit from token quota failover, the former is between downstream client and ai gateway to answer how much user quota is left, the other is between gateway and upstream model provider to make the route decision based on how much remaining quota is left from provider.

aabchoo:
Hi @yanavlasov, do you have any updates on this issue?

yanavlasov:
[...]

PR #1709, opened 2025-12-30
api: add backend quota policy

yanavlasov:
**Description**

Add policy for a backed token quota. Configuring multiple backends in an AIGatewayRouteRule allows request that exceeds allotted token quota on one service to fall back to the next service in the list.

Part of: #1571

This PR contains just API change to make sure this is the right direction. I will add tests once we agree on the API.
Split configuration between QutaPolicy with the metadata keys and additional BackendTrafficPolicy is cumbersome. Perhaps BackendTrafficPolicy content should just be made part of the QutaPolicySpec for simplicity.

### DIFF
--- a/docs/proposals/007-batch-processing/proposal.md
+++ b/docs/proposals/007-batch-processing/proposal.md
@@ -1,0 +1,624 @@
+# Batch Processing API Support
+
+## Table of Contents
+
+<!-- toc -->
+
+- [Summary](#summary)
+- [Goals](#goals)
+- [Background](#background)
+- [Provider Batch API Comparison](#provider-batch-api-comparison)
+  - [OpenAI and Azure OpenAI](#openai-and-azure-openai)
+  - [AWS Bedrock](#aws-bedrock)
+  - [GCP Vertex AI](#gcp-vertex-ai)
+  - [Anthropic](#anthropic)
+- [Proposed API Endpoints](#proposed-api-endpoints)
+- [API Endpoint Specification](#api-endpoint-specification)
+  - [File Management](#file-management)
+  - [Batch Management](#batch-management)
+- [Batch Request Routing Stickiness](#batch-request-routing-stickiness)
+  - [Introduce FileID and BatchID Encoding](#introduce-fileid-and-batchid-encoding)
+- [Request and Response Transformation](#request-and-response-transformation)
+  - [OpenAI to AWS Bedrock](#openai-to-aws-bedrock)
+  - [OpenAI to GCP Vertex AI](#openai-to-gcp-vertex-ai)
+- [Out of Scope](#out-of-scope)
+- [Future Work](#future-work)
+- [Batch Token Usage and Rate Limiting](#batch-token-usage-and-rate-limiting)
+- [Open Questions for Community Discussion](#open-questions-for-community-discussion)
+
+<!-- /toc -->
+
+## Summary
+
+This proposal introduces support for batch processing APIs in the Envoy AI Gateway, enabling users to submit large-scale inference jobs for asynchronous processing. Cloud providers including OpenAI, Azure OpenAI, AWS Bedrock, and GCP Vertex AI all support batch processing with similar file-based approaches.
+
+This proposal focuses on implementing a unified Batch API following the OpenAI specification for these providers, with Anthropic's native Batch API deferred to future work due to its fundamentally different design.
+
+## Goals
+
+- Implement a unified Batch API following the OpenAI specification
+- Support OpenAI, Azure OpenAI, AWS Bedrock, and GCP Vertex AI providers
+- Provide seamless request/response transformation between OpenAI format and provider-specific formats
+- Enable file management operations (upload, retrieve, delete) with proper model routing
+- Defer Anthropic Batch API support to future work due to architectural differences
+
+## Background
+
+Batch processing is a cost-effective approach for running large-scale AI inference workloads that don't require immediate responses. Most major cloud providers offer batch processing capabilities with significant cost savings (typically 50% compared to real-time inference) and higher rate limits.
+
+The Envoy AI Gateway currently provides a unified OpenAI API interface for real-time inference. Extending this to support batch processing will enable users to:
+
+- Submit batch inference jobs through a unified API
+- Leverage cost savings from batch processing across different providers
+- Manage batch jobs (create, retrieve, cancel) consistently
+- Handle large-scale workloads with higher throughput limits
+
+## Provider Batch API Comparison
+
+### OpenAI and Azure OpenAI
+
+OpenAI and Azure OpenAI use a file-based approach with two main API groups:
+
+#### File Management Endpoints (`/v1/files`)
+
+- **Upload File** (`POST /v1/files`): Upload a JSONL file where each line contains a batch request with `custom_id`, `method`, `url`, and `body`. Returns a `file_id`.
+- **Get File Metadata** (`GET /v1/files/{file_id}`): Retrieve file information and metadata.
+- **Download File Content** (`GET /v1/files/{file_id}/content`): Download file contents (used for retrieving batch results).
+- **Delete File** (`DELETE /v1/files/{file_id}`): Remove an uploaded file.
+
+**Input File Format:**
+
+```jsonl
+{"custom_id": "request-1", "method": "POST", "url": "/v1/chat/completions", "body": {"model": "gpt-3.5-turbo-0125", "messages": [{"role": "system", "content": "You are a helpful assistant."},{"role": "user", "content": "Hello world!"}],"max_tokens": 1000}}
+{"custom_id": "request-2", "method": "POST", "url": "/v1/chat/completions", "body": {"model": "gpt-3.5-turbo-0125", "messages": [{"role": "system", "content": "You are an unhelpful assistant."},{"role": "user", "content": "Hello world!"}],"max_tokens": 1000}}
+```
+
+**Output File Format:**
+
+```jsonl
+{"id": "batch_req_abc123", "custom_id": "request-1", "response": {"status_code": 200, "request_id": "req_abc123", "body": {...}}, "error": null}
+{"id": "batch_req_def456", "custom_id": "request-2", "response": {"status_code": 200, "request_id": "req_def456", "body": {...}}, "error": null}
+```
+
+#### Batch Management Endpoints (`/v1/batches`)
+
+- **Create Batch** (`POST /v1/batches`): Create a batch job by referencing an `input_file_id`. Specify the `endpoint` (e.g., `/v1/chat/completions`) and `completion_window` (e.g., `24h`). Returns a `batch_id`.
+- **List Batches** (`GET /v1/batches`): Retrieve all batch jobs with pagination support (`limit`, `after` parameters).
+- **Get Batch Status** (`GET /v1/batches/{batch_id}`): Retrieve batch job details including status, progress counters, and output file IDs.
+- **Cancel Batch** (`POST /v1/batches/{batch_id}/cancel`): Cancel an in-progress batch job.
+
+**Batch Status Lifecycle:**
+
+- `validating` → `in_progress` → `finalizing` → `completed`
+- Alternative states: `failed`, `expired`, `cancelling`, `cancelled`
+
+**Workflow Summary:**
+
+1. Upload input JSONL file → receive `file_id`
+2. Create batch with `input_file_id` → receive `batch_id`
+3. Poll batch status until `completed`
+4. Download results using `output_file_id` from batch response
+
+**Reference**: See [OpenAI Batch API Documentation](https://platform.openai.com/docs/guides/batch) for detailed curl examples and complete request/response schemas.
+
+### AWS Bedrock
+
+AWS Bedrock uses a similar file-based approach but with S3:
+
+1. **Upload to S3**: Upload JSONL input file to S3
+2. **Create Batch Job**: Reference S3 URIs
+
+```json
+{
+  "clientRequestToken": "string",
+  "inputDataConfig": {
+    "s3InputDataConfig": {
+      "s3Uri": "s3://input-bucket/abc.jsonl"
+    }
+  },
+  "jobName": "my-batch-job",
+  "modelId": "anthropic.claude-3-haiku-20240307-v1:0",
+  "outputDataConfig": {
+    "s3OutputDataConfig": {
+      "s3Uri": "s3://output-bucket/"
+    }
+  },
+  "roleArn": "arn:aws:iam::123456789012:role/MyBatchInferenceRole"
+}
+```
+
+**Reference**: See [AWS Bedrock Batch Inference](https://docs.aws.amazon.com/bedrock/latest/userguide/batch-inference-create.html) and [AWS Bedrock CreateModelInvocationJob](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_CreateModelInvocationJob.html)
+
+### GCP Vertex AI
+
+GCP Vertex AI uses Google Cloud Storage (GCS):
+
+1. **Upload to GCS**: Upload JSONL input file to GCS
+2. **Create Batch Prediction Job**: Reference GCS URIs
+
+```json
+{
+  "displayName": "my-cloud-storage-batch-inference-job",
+  "model": "publishers/google/models/gemini-2.5-flash",
+  "inputConfig": {
+    "instancesFormat": "jsonl",
+    "gcsSource": {
+      "uris": ["gs://bucket/input.jsonl"]
+    }
+  },
+  "outputConfig": {
+    "predictionsFormat": "jsonl",
+    "gcsDestination": {
+      "outputUriPrefix": "gs://bucket/output/"
+    }
+  }
+}
+```
+
+**Reference**: See [GCP Vertex Batch Inference](https://docs.cloud.google.com/vertex-ai/generative-ai/docs/multimodal/batch-prediction-from-cloud-storage#create-batch-job-drest) for detailed request/response schema.
+
+### Anthropic
+
+Anthropic takes a different approach with inline requests for batch inference processing:
+
+1. **Create Batch**: Include complete message API payloads directly
+   ```json
+   {
+     "requests": [
+       {
+         "custom_id": "req-1",
+         "params": {
+           "model": "claude-3-5-sonnet-20241022",
+           "max_tokens": 1024,
+           "messages": [...]
+         }
+       }
+     ]
+   }
+   ```
+
+**Key Difference**: Anthropic doesn't use file references - all request data is included inline in the batch creation request.
+
+## Proposed API Endpoints
+
+The following table outlines the unified Batch API endpoints following OpenAI's specification:
+
+| Endpoint                        | Method | Description                        |
+| ------------------------------- | ------ | ---------------------------------- |
+| `/v1/files`                     | POST   | Upload a file for batch processing |
+| `/v1/files/{file_id}`           | GET    | Retrieve file metadata             |
+| `/v1/files/{file_id}/content`   | GET    | Download file content              |
+| `/v1/files/{file_id}`           | DELETE | Delete a file                      |
+| `/v1/batches`                   | POST   | Create a batch processing job      |
+| `/v1/batches`                   | GET    | List all batch jobs                |
+| `/v1/batches/{batch_id}`        | GET    | Retrieve batch job status          |
+| `/v1/batches/{batch_id}/cancel` | POST   | Cancel a batch job                 |
+
+## API Endpoint Specification
+
+### File Management
+
+#### Upload File
+
+```http
+POST /v1/files
+Content-Type: multipart/form-data
+
+file: <JSONL file>
+purpose: "batch"
+```
+
+**Response:**
+
+```json
+{
+  "id": "file-abc123",
+  "object": "file",
+  "bytes": 120000,
+  "created_at": 1677610602,
+  "filename": "batch_input.jsonl",
+  "purpose": "batch"
+}
+```
+
+#### Get File Metadata
+
+```http
+GET /v1/files/{file_id}
+```
+
+#### Download File Content
+
+```http
+GET /v1/files/{file_id}/content
+```
+
+#### Delete File
+
+```http
+DELETE /v1/files/{file_id}
+```
+
+### Batch Management
+
+#### Create Batch
+
+```http
+POST /v1/batches
+Content-Type: application/json
+
+{
+  "input_file_id": "file-abc123",
+  "endpoint": "/v1/chat/completions",
+  "completion_window": "24h",
+  "metadata": {
+    "description": "Monthly analysis job"
+  }
+}
+```
+
+**Response:**
+
+```json
+{
+  "id": "batch-abc123",
+  "object": "batch",
+  "endpoint": "/v1/chat/completions",
+  "errors": null,
+  "input_file_id": "file-abc123",
+  "completion_window": "24h",
+  "status": "validating",
+  "output_file_id": null,
+  "error_file_id": null,
+  "created_at": 1677610602,
+  "in_progress_at": null,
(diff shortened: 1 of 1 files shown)


api/v1alpha1/quota_policy.go
```diff
@@ -0,0 +1,121 @@
...
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+	Spec              QuotaPolicySpec `json:"spec,omitempty"`
+	// Status defines the status details of the QuotaPolicy.
+	Status QuotaPolicyStatus `json:"status,omitempty"`
+}
+
+// QuotaPolicySpec specifies rules for computing token based costs of requests.
+type QuotaPolicySpec struct {
+	// TargetRefs are the names of the AIServiceBackend resources this QuotaPolicy is being attached to.
+	// Attaching multiple QuotaPolicies to the same AIServiceBackend is invalid and will result in an error
```
yuzisun: I think we want to attach the QuotaPolicies to the AIGatewayRoute backendRefs level mainly because the cost function can be different for models and capacity type (provision throughput vs on demand), e.g for VertexAI `AIServiceBackend` the supported models have different burndown rates.
https://docs.cloud.google.com/vertex-ai/generative-ai/docs/provisioned-throughput/supported-models

api/v1alpha1/quota_policy.go
```diff
@@ -0,0 +1,121 @@
...
+	//	  type: InputToken
+	//	- metadataKey: llm_output_token
+	//	  type: OutputToken
+	//	- metadataKey: llm_total_token
+	//	  type: TotalToken
+	//	- metadataKey: llm_cached_input_token
+	//	  type: CachedInputToken
+	// ```
+	// The token costs computed using this policy are debited from the quota specificed by the BackendTrafficPolicy
+	// attached to the same AIServiceBackend. In the BackendTrafficPolicy you can have multiple buckets
```
yuzisun: The BackendTrafficPolicy limit is for client to gateway, I think here we should define own model quota limit of the AIServiceBackend.

yuzisun:
My main question is the relationship between BackendTrafficPolicy and QuotaPolicy,  I think the former is the rate limit check between client to AI gateway and the later is for remaining quota based routing between gateway to upstream AI service. Let's think about the following scenario:

We have 10M provisioned throughput bought from AWS and GCP for claude sonnet 4 model, 6M from GCP and 4M from AWS, anything beyond 10M we process as PayGO (20M quota). So the route can be setup as following:

```yaml
    - modelNameOverride: us.anthropic.claude-sonnet-4-20250514-v1:0
      name: bedrock-us-east-1
      priority: 0
      quota: 4M
    - headerMutation:
        set:
        - name: X-Vertex-AI-LLM-Request-Type
          value: dedicated
      modelNameOverride: claude-sonnet-4@20250514
      name: gcp-us-east1-gcpanthropic
      priority: 0
      quota: 6M
    - headerMutation:
        set:
        - name: X-Vertex-AI-LLM-Request-Type
          value: shared
      modelNameOverride: claude-sonnet-4@20250514
      name: gcp-us-east5-gcpanthropic
      priority: 1
      quota: 20M
```
As AI gateway manages a multi-tenant environment, the 10M PT and 20M OD quota are shared among many users and each user we setup a token rate limit rule (BackendTrafficPolicy) to prevent abusing the system. Sometimes we can have different tier of services, e.g default tier gives 1M token quota and premium tier gives 5M quota. I think the steps should look like following:
 - gateway checks if the user exceeds the token rate limit quota against BackendTrafficPolicy.
 - once it passes the rate limit check gateway tries to route to available PT capacity as much as possible by checking the remaining quota from each provider where you bought the capacity, backends are defined on AIGatewayRoute.
 - when PT capacity is exhausted then gateway spillover to PayGo.

johnugeorge:
Agree with Dan. We need modelNameOverride field to send requests to multiple supported models of same AIServicebackend.   Hence, quota policy makes more sense with the backendrefs of the AIGatewayRoute. 

@yuzisun  In your example, where is quota measured?  eg: For `quota: 20M`, what is the unit and how is it calculated?  For each backendRef, won't we need both quota definition and limit like in https://github.com/envoyproxy/ai-gateway/blob/c434c4eb2e542b736fb1dcf200f11c1abcc6bc58/examples/token_ratelimit/token_ratelimit.yaml#L111-L127 ?  


In AIGatewayRoute, Something like 
```
  llmRequestCosts:
      type: InputToken
    ....
 
  
  rules:
     -   modelNameOverride: us.anthropic.claude-sonnet-4-20250514-v1:0
         name: bedrock-us-east-1
         priority: 0
         quota: 
             key: llm_input_token #can be any llmRequestCost metadata key: cel, input token, output token, cached
             # Configures the number of "tokens" allowed per hour, per user. 
             requests: 10 
             unit: Hour 
```

yuzisun:
On the other thought we can define the quota on AIServiceBackend as the models quota policy logically belongs to the backend
  ```
  kind: AIServiceBackend                                                                                                                                                                                                                                                                  
  metadata:                                                                                                                                                                                                                                                                               
    name: openai                                                                                                                                                                                                                                                               
    namespace: default                                                                                                                                                                                                                                                                    
  spec:                                                                                                                                                                                                                                                                                   
    schema:                                                                                                                                                                                                                                                                               
      name: OpenAI                                                                                                                                                                                                                                                                        
    backendRef:                                                                                                                                                                                                                                                                           
      group: gateway.envoyproxy.io                                                                                                                                                                                                                                                        
      kind: Backend                                                                                                                                                                                                                                                                       
      name: openai-backend                                                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                          
    # NEW: Token quota configuration                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                              
       # Model name this quota applies to (matches request model)                                                                                                                                                                                                                          
    - model: "gpt-4"                                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                          
      # Provisioned throughput capacity (reserved/guaranteed)                                                                                                                                                                                                                             
      provisioned:                                                                                                                                                                                                                                                                        
        inputTokensLimit: 100000      # 100K TPM provisioned   
        outputTokensLimit: 10000                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                          
      # On-demand capacity (pay-per-use, best effort)                                                                                                                                                                                                                                     
      onDemand:                                                                                                                                                                                                                                                                           
        inputTokensLimit: 50000       # 50K TPM on-demand limit (0 = unlimited)
        outputTokensLimit: 5000                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                          
      # Sliding window configuration                                                                                                                                                                                                                                                      
      window:                                                                                                                                                                                                                                                                             
        duration: 60s                # Window size (default: 30s) 
```

File: api/v1alpha1/quota_policy.go [hunk 3]
```diff
@@ -0,0 +1,121 @@
...
+	// tracking quot ausage for each distinct x-user-id header value. Quota can be specified in terms of total (input + output)
```
siddharth1036: nit, typo `quot ausage` -> `quota usage`

api/v1alpha1/quota_policy.go
```diff
@@ -0,0 +1,121 @@
...
+	//	     name: usage-rate-limit
+	//	  rateLimit:
+	//	    type: Global
+	//	    global:
+	//	      rules:
+	//	        - clientSelectors:
+	//	            # Do the rate limiting based on the x-user-id header.
+	//	            - headers:
+	//	                - name: x-user-id
+	//	                  type: Distinct
```
siddharth1036: I think if header based distinction is to be done in `backendtrafficpolicy`, wont we require a `clientSelectors` interface on `quotapolicyspec` as well to translate the same to `backendtrafficpolicy`

yanavlasov: Yes, I have added them to API.

yanavlasov:
I have updated the API to include per model breakdown and client selection in the quota definition.

api/v1alpha1/quota_policy.go
```diff
@@ -0,0 +1,123 @@
...
+	// Attaching multiple QuotaPolicies to the same AIServiceBackend is invalid and will result in an error
+	// during the reconciliation of AIServiceBackend.
+	//
+	// +optional
+	// +kubebuilder:validation:MaxItems=16
+	// +kubebuilder:validation:XValidation:rule="self.all(ref, ref.group == 'aigateway.envoyproxy.io' && ref.kind == 'AIServiceBackend')", message="targetRefs must reference AIServiceBackend resources"
+	TargetRefs []gwapiv1a2.LocalPolicyTargetReference `json:"targetRefs,omitempty"`
+	// PerModelQuotas specifies quota for different models served by the AIServiceBackend(s) where this
+	// policy is attached.
+	PerModelQuotas []PerModelQuota `json:"perModelQuotas,omitempty"`
```
johnugeorge, line 50: How many quotas can be created? Is there a limit?

api/v1alpha1/quota_policy.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+type PerModelQuota struct {
+	// Model name for which the quota is specified.
+	//
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinLength=1
+	ModelName string `json:"modelName"`
```
johnugeorge: Is model name always necesaary? How can global quota be set (irrespective of models, set at provider level) ?  If global quota is set, same quota bucket is used for all models of that provider.

johnugeorge: cc @yuzisun

siddharth1036:
@yanavlasov should we record a rough idea or proposal on how are we planning to use quotapolicy to generate envoy configurations (either backendtrafficpolicy or cluster configurations) so that it would be easier for reviewers to review with the context of implementation.
 https://github.com/envoyproxy/ai-gateway/issues/1571#issuecomment-3711026606 (is it on the similar lines to this?)

api/v1alpha1/quota_policy.go
yanavlasov: Added limit of 128 similar to rate limit rules.

api/v1alpha1/quota_policy.go
yanavlasov: I have made it to not be required with empty model name applied to all models.

yanavlasov:
> @yanavlasov should we record a rough idea or proposal on how are we planning to use quotapolicy to generate envoy configurations (either backendtrafficpolicy or cluster configurations) so that it would be easier for reviewers to review with the context of implementation. [#1571 (comment)](https://github.com/envoyproxy/ai-gateway/issues/1571#issuecomment-3711026606) (is it on the similar lines to this?)

Yes, this is API for the proposal you have linked.

johnugeorge:
@yanavlasov @yuzisun 

Few questions.

Lets consider a multi-tenant usecase. 

There is one AIServiceBackend per AI provider. There is 1 AIGatewayRoute configured per tenant for the same AIServiceBackend.

Eg: 

AIServiceBackend: 

```
kind: AIServiceBackend
metadata:
     name: openai-backend
spec:
  backendRef:
    group: gateway.envoyproxy.io
    kind: Backend
    name: openai-backend
  schema:
    name: OpenAI
```

There are 2 tenants each having 1 AIGatewayRoute pointing to the same AIServiceBackend `openai-backend`

AIGatewayRoute of tenant 1

```
kind: AIGatewayRoute
metadata:
  name: envoy-ai-gateway-openai-1
  namespace: default
spec:
  parentRefs:
    - name: envoy-ai-gateway-basic
      kind: Gateway
  rules:
    - matches:
        - headers:
            - type: Exact
              name: x-ai-eg-model
              value: gpt-4o-mini
      backendRefs:
        - modelNameOverride: gpt-5
           name: openai-backend   <-- quota to be applied for tenant 1
 
```
 This means that if you get a request with model `gpt-4o-mini`, send request to  backend `openai-provider` with the model name overwritten as `gpt-5`
           
AIGatewayRoute of tenant 2

```
kind: AIGatewayRoute
metadata:
  name: envoy-ai-gateway-openai-2
  namespace: default
spec:
  parentRefs:
    - name: envoy-ai-gateway-basic
      kind: Gateway
  rules:
    - matches:
        - headers:
            - type: Exact
              name: x-ai-eg-model
              value: gpt-4o
      backendRefs:
        - modelNameOverride: gpt-5
           name: openai-backend   <-- quota to be applied for tenant 2  
```


Questions
1. `modelName` in `perModelQuota` described in QuotaPolicy must refer to final model name(ie `modelNameOverride` if present, else it is `x-ai-eg-model`). Is that correct?
2. I assume, there is 1 envoy quota bucket per `perModelQuota` entry.  
Why is there a restriction of `Attaching multiple QuotaPolicies to the same AIServiceBackend is invalid and will result in an error` ? 
Tenants/AIGatewayRoutes are added on the fly. Instead of a single global QuotaPolicy, we should allow multiple quotapolicies against the same AIServiceBackend.  Allowing Multiple QuotaPolicy against one AIServiceBackend means merging all perModelQuota entries(across all QuotaPolicy) against same AIServiceBackend.

api/v1alpha1/quota_policy.go
johnugeorge: Is this a hard limit from envoy side? Since quota policies are added against the same AIServiceBackend, limit should be much higher.

yanavlasov:
> @yanavlasov @yuzisun
> 
> Few questions.
> 
> Lets consider a multi-tenant usecase.
> 
> There is one AIServiceBackend per AI provider. There is 1 AIGatewayRoute configured per tenant for the same AIServiceBackend.
> 
> Eg:
> 
> AIServiceBackend:
> 
> ```
> apiVersion: aigateway.envoyproxy.io/v1alpha1
> kind: AIServiceBackend
> metadata:
>      name: openai-backend
> spec:
>   backendRef:
>     group: gateway.envoyproxy.io
>     kind: Backend
>     name: openai-backend
>   schema:
>     name: OpenAI
> ```
> 
> There are 2 tenants each having 1 AIGatewayRoute pointing to the same AIServiceBackend `openai-backend`
> 
> AIGatewayRoute of tenant 1
> 
> ```
> apiVersion: aigateway.envoyproxy.io/v1alpha1
> kind: AIGatewayRoute
> metadata:
>   name: envoy-ai-gateway-openai-1
>   namespace: default
> spec:
>   parentRefs:
>     - name: envoy-ai-gateway-basic
>       kind: Gateway
>       group: gateway.networking.k8s.io
>   rules:
>     - matches:
>         - headers:
>             - type: Exact
>               name: x-ai-eg-model
>               value: gpt-4o-mini
>       backendRefs:
>         - modelNameOverride: gpt-5
>            name: openai-backend   <-- quota to be applied for tenant 1
>  
> ```
> 
> This means that if you get a request with model `gpt-4o-mini`, send request to backend `openai-provider` with the model name overwritten as `gpt-5`
> 
> AIGatewayRoute of tenant 2
> 
> ```
> apiVersion: aigateway.envoyproxy.io/v1alpha1
> kind: AIGatewayRoute
> metadata:
>   name: envoy-ai-gateway-openai-2
>   namespace: default
> spec:
>   parentRefs:
>     - name: envoy-ai-gateway-basic
>       kind: Gateway
>       group: gateway.networking.k8s.io
>   rules:
>     - matches:
>         - headers:
>             - type: Exact
>               name: x-ai-eg-model
>               value: gpt-4o
>       backendRefs:
>         - modelNameOverride: gpt-5
>            name: openai-backend   <-- quota to be applied for tenant 2  
> ```
> 
> Questions
> 
> 1. `modelName` in `perModelQuota` described in QuotaPolicy must refer to final model name(ie `modelNameOverride` if present, else it is `x-ai-eg-model`). Is that correct?

I think we will need to make it work with the model name override. Otherwise if multiple original model names are rewritten to the same name the quota will not be applied correctly.

> 2. I assume, there is 1 envoy quota bucket per `perModelQuota` entry.

If there are client selectors configured, then one perModelQuota entry can result in multiple buckets for each distinct client selector match.

>    Why is there a restriction of `Attaching multiple QuotaPolicies to the same AIServiceBackend is invalid and will result in an error` ?
>    Tenants/AIGatewayRoutes are added on the fly. Instead of a single global QuotaPolicy, we should allow multiple quotapolicies against the same AIServiceBackend.  Allowing Multiple QuotaPolicy against one AIServiceBackend means merging all perModelQuota entries(across all QuotaPolicy) against same AIServiceBackend.

Sure, I think I have carried this restriction from rate limit configuration. But I do not think there is a technical reason to have it. I have removed it.

api/v1alpha1/quota_policy.go
```diff
@@ -0,0 +1,123 @@
...
+	TargetRefs []gwapiv1a2.LocalPolicyTargetReference `json:"targetRefs,omitempty"`
+	// PerModelQuotas specifies quota for different models served by the AIServiceBackend(s) where this
+	// policy is attached.
+	//
+	// +kubebuilder:validation:MaxItems=128
+	PerModelQuotas []PerModelQuota `json:"perModelQuotas,omitempty"`
+}
+
+type PerModelQuota struct {
+	// Model name for which the quota is specified. If the name is empty the quota rule is applied to all models
```
aabchoo: Could we make this mandatory and create a generic costExpression in QuotaPolicySpec? Cost can be overridden if defined in PerModelQuota?

api/v1alpha1/quota_policy.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+// QuotaPolicySpec specifies rules for computing token based costs of requests.
+type QuotaPolicySpec struct {
+	// TargetRefs are the names of the AIServiceBackend resources this QuotaPolicy is being attached to.
+	//
+	// +optional
+	// +kubebuilder:validation:MaxItems=16
+	// +kubebuilder:validation:XValidation:rule="self.all(ref, ref.group == 'aigateway.envoyproxy.io' && ref.kind == 'AIServiceBackend')", message="targetRefs must reference AIServiceBackend resources"
+	TargetRefs []gwapiv1a2.LocalPolicyTargetReference `json:"targetRefs,omitempty"`
```
yuzisun, line 39: what happens if multiple QuotaPolicy rules are attached to the same AIServiceBackend, do we merge them or each AIServiveBackend can only have one QuotaPolicy?

api/v1alpha1/quota_policy.go
```diff
@@ -0,0 +1,123 @@
...
+	// +optional
+	CostExpression *string `json:"costExpression,omitempty"`
+
+	// Rules are a list of client selectors and quotas. If a request
+	// matches multiple rules, each of their associated quotas get applied, so a
+	// single request might burn down the quota for multiple rules
+	// if selected. The quota service will return a logical OR of the individual
+	// quota checks of all matching rules. For example, if a request
+	// matches two rules, one has available quota and one not, the final decision will be
+	// to allow the request.
```
yuzisun: If we have a global quota rule and a tenant quota rule, the global quota is still available but the limit is reached for tenant quota, we should reject the request for this backend instead of allowing?

api/v1alpha1/quota_policy.go
```diff
@@ -0,0 +1,123 @@
...
+	// The request always succeeds, even if the configured quota is exceeded.
+	//
+	// +optional
+	ShadowMode *bool `json:"shadowMode,omitempty"`
+}
+
+// QuotaValue defines the quota limits using sliding window.
+type QuotaValue struct {
+	// Number of tokens alloted for a specified time window.
+	Tokens uint `json:"tokens"`
```
yuzisun: need to think about a good name here, after CEL expression calculation this is no longer in unit of `tokens`. Maybe we can just call it `limit`.

api/v1alpha1/quota_policy.go
johnugeorge: It has to be a merge of all perModelQuotas. If there are conflicting rules for the same AIServiceBackend(eg: same client selector in same or different QuotaPolicy), behaviour will be undefined.  Ref: https://github.com/envoyproxy/ai-gateway/pull/1709#issuecomment-3724441102

api/v1alpha1/quota_policy.go
johnugeorge: If both quotas are configured, Default case IMO should be that request should be allowed only if both quotas are available  (tenant quota + global quota ).  As discussed, this logic can be switched over a boolean value.

api/v1alpha1/quota_policy.go
yanavlasov: Done

api/v1alpha1/quota_policy.go
yanavlasov: Added config as discussed

api/v1alpha1/quota_policy.go
```diff
@@ -0,0 +1,152 @@
...
+	// using the "BucketRules" configuration.
+	//
+	// +optional
+	GlobalBucket QuotaValue `json:"globalBucket"`
+	// BucketRules are a list of client selectors and quotas. If a request
+	// matches multiple rules, each of their associated quotas get applied, so a
+	// single request might burn down the quota for multiple rules.
+	//
+	// +optional
+	BucketRules []QuotaRule `json:"bucketRules"`
```
johnugeorge, line 107: Do we set only one of GlobalBucket or  BucketRules?  ie GlobalBucket is set only with ServiceQuota configuration(Global Quota)  and BucketRules are set with PerModelQuota?

api/v1alpha1/quota_policy.go
yanavlasov: It is becoming harder to reason about various case. The API defines one special "global" bucket that matches all request and then request specific buckets. Ideally a request either matches one request bucket bucket or the global bucket.

Perhaps for simplicity at this point the matching process should stop once it a process specific bucket matches. Otherwise it is hard to reason how multiple request specific matches should behave.

api/v1alpha1/quota_policy.go
yanavlasov: Per model quota has its own global bucket. Per model quota is independent from the model agnostic quota.

api/v1alpha1/quota_policy.go
```diff
@@ -0,0 +1,152 @@
...
+	// In the "exclusive" mode the quota is charged to matching BucketRules or the GlobalBucket
+	// if no BucketRules match the request. The request is denied if all matching buckets are out of quota.
+	// In the "shared" mode the quota is charged to all matching "BucketRules" AND the "GlobalBucket"
+	// and request is allowed only if the quota is available in all matching buckets.
+	Mode QuotaBucketMode `json:"mode"`
+	// Quota applicable to all traffic. This value can be overridden for specific classes of requests
+	// using the "BucketRules" configuration.
+	//
+	// +optional
+	GlobalBucket QuotaValue `json:"globalBucket"`
```
nacx: This seems to be more about a default quota for this model when no BucketRules match (or none is defined). Would it make sense to rename this field to `DefaultBucket`, to make it a bit less confusing with the concept of the "service quota" ?

johnugeorge:
From our last discussion 
1. For now, we can simplify GlobalQuota to remove ClientSelectors. 

Instead of,

~~ServiceQuota QuotaDefinition `json:"serviceQuota,omitempty"`~~

Something like

    ServiceQuota ServiceQuotaDefinition json:"serviceQuotaDefinition,omitempty"`

    type ServiceQuotaDefinition struct {

      CostExpression *string `json:"costExpression,omitempty"`
      Quota QuotaValue `json:"quota"`
    }

2. As in the usecase described in https://github.com/envoyproxy/ai-gateway/pull/1709#issuecomment-3724441102, Final QuotaPolicy should be a merge of all bucket rules described across quotapolicies. eg: QuotaPolicy1(for tenantA)  can have GPT5 with bucket rule1. QuotaPolicy2(for tenant B)  can have same GPT5 with a different bucketrule2.  Effective quotaPolicy is equivalent to GPT5 having 2 bucket rules.

api/v1alpha1/quota_policy.go
yanavlasov: Done

yuzisun (approved):

PR #1710, opened 2025-12-30
test: adds more cases for small/medium/large/xlarge req/res pairs; part2

mathetake:
**Description**

This adds additional variety in the bench_test.go test cases where instead of having tons of messages but a few large messages, which are more realistic given that Azure OpenAI limits # of message to 2048.

### DIFF
--- a/internal/apischema/openai/openai.go
+++ b/internal/apischema/openai/openai.go
@@ -1076,11 +1076,27 @@ const (
 	ToolTypeFunction            ToolType = "function"
 	ToolTypeImageGeneration     ToolType = "image_generation"
 	ToolTypeEnterpriseWebSearch ToolType = "enterprise_search"
+	ToolTypeGoogleSearch        ToolType = "google_search"
 )
 
+// GCPGoogleSearchConfig contains GCP-specific configuration for Google Search grounding.
+// https://pkg.go.dev/google.golang.org/genai#GoogleSearch
+type GCPGoogleSearchConfig struct {
+	ExcludeDomains     []string            `json:"exclude_domains,omitempty"`     //nolint:tagliatelle
+	BlockingConfidence string              `json:"blocking_confidence,omitempty"` //nolint:tagliatelle
+	TimeRangeFilter    *GCPTimeRangeFilter `json:"time_range_filter,omitempty"`   //nolint:tagliatelle
+}
+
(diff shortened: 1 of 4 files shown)


tests/data-plane/bench_test.go
```diff
@@ -106,19 +110,37 @@ func BenchmarkChatCompletions(b *testing.B) {
...
 	}
-	largeRequestBody := fmt.Sprintf(`{
-		"model": "gpt-4",
-		"messages": [%s],
-		"max_tokens": 100
-	}`, strings.Join(messages, ","))
+	const template = `{
+	"model": "gpt-4",
+	"messages": [%s],
+	"max_tokens": 100
```
aabchoo: Use max_completion_tokens?

tests/data-plane/bench_test.go
```diff
@@ -65,10 +61,18 @@ func BenchmarkChatCompletions(b *testing.B) {
 				requestBody          string
 				fakeResponseBodyType string
 			}{
-				{name: "small", requestBody: smallRequest, fakeResponseBodyType: "small"},
-				{name: "medium", requestBody: mediumRequest, fakeResponseBodyType: "medium"},
-				{name: "large", requestBody: largeRequest, fakeResponseBodyType: "large"},
-				{name: "xlarge", requestBody: xlargeRequest, fakeResponseBodyType: "large"},
+				{name: "small/100-messages-large-messages", requestBody: createChatCompletionRequest(100, 1), fakeResponseBodyType: "small"},
```
aabchoo: Why is the test for small `large-message`? Similar for medium

aabchoo: What does small/medium/large prefix indicate?

tests/data-plane/bench_test.go
mathetake: good catch

tests/data-plane/bench_test.go
mathetake: made a change to make the test case name more explicit about the type

mathetake:
omg i messedup the commits ... let me do the force push

mathetake:
@aabchoo PTAL

mathetake:
ah drift

mathetake:
😎

PR #1715, opened 2025-12-31
chore: use native CustomResponse for www-authenticated header

zhaohuabing:
**Description**

Implemented direct header-based WWW-Authenticate handling for MCP OAuth 401 responses and removed the extension-server logic that rewrote bodies into headers. This uses Envoy Gateway’s header support directly in BackendTrafficPolicy CustomResponse, and simplifies the extension-server implementation.


nacx (approved):

nacx:
Nice!

PR #1717, opened 2026-01-01
fix: cachePoint should be in separate content block 

yuzisun:
**Description**
For AWS Bedrock `cachePoint` should be in its own content block, as currently we are getting following error:
```
{"message":"ContentBlock object at messages.0.content.0 can only set one of the following keys: text, image, toolUse, toolResult, document, video, cachePoint, reasoningContent, citationsContent, searchResult."}
```


yuzisun:
> Almost happy new year

Last PR for 2025!

PR #1719, opened 2026-01-02
extproc: add cache writes

aabchoo:
**Description**

Anthropic cache writes cost different from cache reads. Cost calculation should be updated to account for writes vs reads. Adding a new cost type. Updated similarly for AWS.

https://platform.claude.com/docs/en/build-with-claude/prompt-caching
https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_TokenUsage.html

Vertex AI and OpenAI themselves do not support cache write response so cache writes will be set to 0. 

**Changes**
Dynamic metadata will now include cache writes.
Separates cache reads and writes.
User is returned new usage cached writes.

Updated tests to match -- hopefully I caught them all. Updated wherever I saw CachedInputTokens.

### DIFF
--- a/tests/data-plane/vcr/prometheus_metrics_test.go
+++ b/tests/data-plane/vcr/prometheus_metrics_test.go
@@ -127,7 +127,7 @@ func verifyPrometheusTokenUsage(t *testing.T, metric *dto.MetricFamily, expected
 	}
 	require.NotNil(t, inputMetric, "Input metric not found")
 	require.NotNil(t, cachedInputMetric, "Cached Input metric not found")
-	require.NotNil(t, cacheCreationInputMetric, "Cached Creation Input metric not found")
+	require.NotNil(t, cacheCreationInputMetric, "Cache Creation Input metric not found")
 	require.NotNil(t, outputMetric, "Output metric not found")
 
 	type testCase struct {


internal/translator/anthropic_anthropic_test.go
```diff
@@ -95,7 +95,7 @@ func TestAnthropicToAnthropic_ResponseBody_non_streaming(t *testing.T) {
 	require.NoError(t, err)
 	require.Nil(t, headerMutation)
 	require.Nil(t, bodyMutation)
-	expected := tokenUsageFrom(9, 0, 16, 25)
+	expected := tokenUsageFrom(9, 0, -1, 16, 25)
```
yuzisun: we should not have negative cache writes ?

internal/apischema/openai/openai.go
```diff
@@ -1382,6 +1382,8 @@ type PromptTokensDetails struct {
 	AudioTokens int `json:"audio_tokens,omitzero"`
 	// Cached tokens present in the prompt.
 	CachedTokens int `json:"cached_tokens,omitzero"`
+	// Tokens written to the cache.
+	CachedWriteTokens int `json:"cached_write_tokens,omitzero"`
```
yuzisun: litellm named it `cache_creation_input_tokens`
https://docs.litellm.ai/docs/completion/prompt_caching

internal/translator/anthropic_anthropic_test.go
aabchoo: Some cases we have tokenUsageFrom(-1, -1, -1, -1, -1). I copied the values wrong for this. Should be 0

internal/apischema/openai/openai.go
aabchoo: AWS calls it `cacheWriteInputTokens` and Anthropic calls it `cache_creation_input_tokens`. Can opt to name it creation instead of write

yuzisun: I think either `cache_creation_input_tokens` or `cache_write_input_tokens` is fine.

aabchoo: updated to use cache creation everywhere

internal/translator/anthropic_anthropic_test.go
```diff
@@ -89,13 +89,13 @@ func TestAnthropicToAnthropic_ResponseHeaders(t *testing.T) {
 func TestAnthropicToAnthropic_ResponseBody_non_streaming(t *testing.T) {
 	translator := NewAnthropicToAnthropicTranslator("", "")
 	require.NotNil(t, translator)
-	const responseBody = `{"model":"claude-sonnet-4-5-20250929","id":"msg_01J5gW6Sffiem6avXSAooZZw","type":"message","role":"assistant","content":[{"type":"text","text":"Hi! 👋 How can I help you today?"}],"stop_reason":"end_turn","stop_sequence":null,"usage":{"input_tokens":9,"cache_creation_input_tokens":0,"cache_read_input_tokens":0,"cache_creation":{"ephemeral_5m_input_tokens":0,"ephemeral_1h_input_tokens":0},"output_tokens":16,"service_tier":"standard"}}`
+	const responseBody = `{"model":"claude-sonnet-4-5-20250929","id":"msg_01J5gW6Sffiem6avXSAooZZw","type":"message","role":"assistant","content":[{"type":"text","text":"Hi! 👋 How can I help you today?"}],"stop_reason":"end_turn","stop_sequence":null,"usage":{"input_tokens":9,"cache_creation_input_tokens":0,"cache_read_input_tokens":0,"cached_creation":{"ephemeral_5m_input_tokens":0,"ephemeral_1h_input_tokens":0},"output_tokens":16,"service_tier":"standard"}}`
```
yuzisun: it should be `cache_creation` based on the anthropic doc
https://platform.claude.com/docs/en/build-with-claude/prompt-caching#1-hour-cache-duration

internal/apischema/openai/openai.go
```diff
@@ -1382,6 +1382,8 @@ type PromptTokensDetails struct {
 	AudioTokens int `json:"audio_tokens,omitzero"`
 	// Cached tokens present in the prompt.
 	CachedTokens int `json:"cached_tokens,omitzero"`
+	// Tokens written to the cache.
+	CachedCreationTokens int `json:"cached_creation_input_tokens,omitzero"`
```
yuzisun: I think you are trying to align with the name `cached_tokens` but `cache_creation_input_tokens` reads better.

yuzisun (approved):

PR #1726, opened 2026-01-05
fix: update the signatures for gemini models

hustxiayang:
**Description**

According to the description in https://ai.google.dev/gemini-api/docs/thought-signatures, gemini enforces users to use signature in multi-turn conversations. 

Discussion:
In the google's specification doc, it appends the signatures into different parts, which is quite different than anthropic claude models. I would suggest to make it compatible with anthropic claude models, and use a single signature for each message. Specifically,

- For text parts in the message, we combine the signatures of text parts into one
- For tool calls, we only need to extract the first tool call's signature, and we do not need to combine it with the signatures from the text parts.

### DIFF
--- a/internal/translator/gemini_helper.go
+++ b/internal/translator/gemini_helper.go
@@ -287,7 +287,7 @@ func assistantMsgToGeminiParts(msg *openai.ChatCompletionAssistantMessageParam) 
 
 		funcCallPart := genai.NewPartFromFunctionCall(toolCall.Function.Name, parsedArgs)
 
-		// Attach signature to FIRST tool call if signature exists
+		// According to https://ai.google.dev/gemini-api/docs/thought-signatures, if the model generates parallel function calls in a response, the thought_signature is attached only to the first functionCall part. Subsequent  functionCall parts in the same response will not contain a signature.
 		if i == 0 && thoughtSignature != nil {
 			funcCallPart.ThoughtSignature = thoughtSignature
 		}


internal/translator/gemini_helper.go
```diff
@@ -256,16 +257,42 @@ func toolMsgToGeminiParts(msg openai.ChatCompletionToolMessageParam, knownToolCa
...
 		knownToolCalls[*toolCall.ID] = toolCall.Function.Name
 		var parsedArgs map[string]any
 		if err := json.Unmarshal([]byte(toolCall.Function.Arguments), &parsedArgs); err != nil {
 			return nil, nil, fmt.Errorf("function arguments should be valid json string. failed to parse function arguments: %w", err)
 		}
-		parts = append(parts, genai.NewPartFromFunctionCall(toolCall.Function.Name, parsedArgs))
+
+		funcCallPart := genai.NewPartFromFunctionCall(toolCall.Function.Name, parsedArgs)
+
+		// Attach signature to FIRST tool call if signature exists
```
yuzisun: according to the doc:
```
Parallel Function Calls: If the model generates parallel function calls in a response, the
 thought_signature is attached only to the first functionCall part. Subsequent 
functionCall parts in the same response will not contain a signature.
```

internal/translator/gemini_helper_test.go
```diff
@@ -440,6 +440,101 @@ func TestAssistantMsgToGeminiParts(t *testing.T) {
...
+					},
+					{
+						ID: ptr.To("call_time"),
+						Function: openai.ChatCompletionMessageToolCallFunctionParam{
+							Name:      "get_time",
+							Arguments: `{"timezone":"PST"}`,
+						},
+						Type: openai.ChatCompletionMessageToolCallTypeFunction,
+					},
+				},
```
yuzisun, line 496: Function calls should have their own signatures if they are sequential steps

internal/translator/openai_gcpvertexai_test.go
```diff
@@ -1708,7 +1708,7 @@ func TestExtractToolCallsStreamIndexing(t *testing.T) {
 	}
 	o := NewChatCompletionOpenAIToGCPVertexAITranslator("gemini-2.0-flash-001").(*openAIToGCPVertexAITranslatorV1ChatCompletion)
 
-	calls, err := o.extractToolCallsFromGeminiPartsStream(toolCalls, parts, json.MarshalForDeterministicTesting)
+	calls, _, err := o.extractToolCallsFromGeminiPartsStream(toolCalls, parts, json.MarshalForDeterministicTesting)
```
yuzisun, line 1733: add signature tests for streaming case

internal/translator/gemini_helper.go
```diff
@@ -256,16 +257,42 @@ func toolMsgToGeminiParts(msg openai.ChatCompletionToolMessageParam, knownToolCa
...
+	if v, ok := msg.Content.Value.([]openai.ChatCompletionAssistantMessageParamContent); ok {
+		for _, contPart := range v {
+			if contPart.Type == openai.ChatCompletionAssistantMessageParamContentTypeThinking {
+				if contPart.Signature != nil && *contPart.Signature != "" {
+					// Decode base64 signature
+					sigBytes, err := base64.StdEncoding.DecodeString(*contPart.Signature)
+					if err != nil {
+						return nil, nil, fmt.Errorf("failed to decode thought signature: %w", err)
+					}
+					thoughtSignature = sigBytes
```
yuzisun, line 272: is it true that for a given message param there could be only one thoughtSignature ?

internal/translator/gemini_helper_test.go
yuzisun: I guess here the case is for parallel tool calls which can only have one signature

internal/translator/gemini_helper.go
hustxiayang: I think currently it's implemented according to this specification. Do you mean you want to add this into the comment?

internal/translator/gemini_helper.go
hustxiayang: yes, and this is on purpose: anthropic's reasoning models have just one `signature`, and I want to make it compatible

internal/translator/gemini_helper_test.go
hustxiayang: `I guess here the case is for parallel tool calls which can only have one signature`, this is correct.  `Function calls should have their own signatures if they are sequential steps`, in this test, it's a parallel tool call. The models returns two function calls in just one message.

internal/translator/openai_gcpvertexai_test.go
hustxiayang: added two stream tests: signature with text and tool calls. Thanks a lot for the comment!

internal/translator/gemini_helper.go
yuzisun: yes, can you update the comment ?

hustxiayang: added, thanks!

PR #1742, opened 2026-01-08
chore: enable gocritic perf tag

mathetake:
**Description**

This enables "perf" gocritic tag to enforce some rules that give us performance gains


.golangci.yml
```diff
@@ -14,6 +14,9 @@ linters:
     - testifylint
     - unconvert
   settings:
+    gocritic:
+      enabled-tags:
+        - performance
```
mathetake, line 19: this is the main change

PR #1744, opened 2026-01-08
chore: clean up translator package

mathetake:
**Description**

This cleans up the translator package: moved test only constructs into test files properly as well as share the SSE related variable across multiple places instead of having multiple ones including inline byte slices that might cause allocations


internal/translator/openai_awsbedrock_test.go
```diff
@@ -32,6 +32,25 @@ import (
 	"github.com/envoyproxy/ai-gateway/internal/metrics"
 )
 
+const releaseDateUnix = 1731679200 // Represents 2024-11-15 09:00:00 UTC
+
+func getChatCompletionResponseChunk(body []byte) []openai.ChatCompletionResponseChunk {
```
mathetake, line 37: these are moved from the non test file despite being only used in tests

missBerg (approved):

PR #1749, opened 2026-01-09
extproc: support service tier for bedrock

aabchoo:
**Description**
AWS supports Service tiers similar to Openai[1]. Following the service tier definition based off the converse API[2].

Updated translation for openai -> aws bedrock in the request body and response body. Updated the tests to tests translations.

1. https://docs.aws.amazon.com/bedrock/latest/userguide/service-tiers-inference.html
2. https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_Converse.html

Successful setting of service_body will return:
```
{..., "service_tier":"default", ...}
```

A **valid** service tier that is unsupported for the model will return a 400:
```
{"type":"error","error":{"type":"ValidationException:http://internal.amazon.com/coral/com.amazon.bedrock/","code":"400","message":"Reservation not found."}}
```
Only specific models have support for reserved.

Dropped openai.ChatCompletionNewParamsServiceTier in favour of string to keep consistent with the other definitions of service_tier in that file

### DIFF
--- a/tests/internal/testopenai/server_test.go
+++ b/tests/internal/testopenai/server_test.go
@@ -17,6 +17,13 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+func TestMain(m *testing.M) {
+	// Unset API keys to ensure tests behave consistently regardless of environment.
+	_ = os.Unsetenv("OPENAI_API_KEY")
+	_ = os.Unsetenv("AZURE_OPENAI_API_KEY")
+	os.Exit(m.Run())
+}
+
 func TestServer_ExistingCassette(t *testing.T) {
 	// Test that an existing cassette (chat-basic) works.
 	server := newTestServer(t)


internal/apischema/openai/openai.go
```diff
@@ -975,11 +975,12 @@ type ChatCompletionRequest struct {
 	// If set to 'auto', then the request will be processed with the service tier configured in the Project settings. Unless otherwise configured, the Project will use 'default'.
 	// If set to 'default', then the request will be processed with the standard pricing and performance for the selected model.
 	// If set to 'flex' or 'priority', then the request will be processed with the corresponding service tier.
+	// Includes support for bedrock's 'reserved' service_tier.
 	// When the service_tier parameter is set, the response body will include the service_tier value based on the processing mode actually used to serve the request.
 	// This response value may be different from the value set in the parameter.
 	// Docs: https://platform.openai.com/docs/api-reference/chat/create#chat-create-service_tier
-	// Any of "auto", "default", "flex", "scale", "priority".
-	ServiceTier openai.ChatCompletionNewParamsServiceTier `json:"service_tier,omitzero"`
+	// Any of "auto", "default", "flex", "scale", "priority" (and "reserved" for bedrock).
```
yuzisun: let's document the list the options by provider

internal/apischema/awsbedrock/awsbedrock.go
```diff
@@ -401,6 +412,12 @@ type ConverseOutput struct {
 	Message Message `json:"message,omitempty"`
 }
 
+// ServiceTier is defined in the AWS Bedrock API:
+// https://docs.aws.amazon.com/bedrock/latest/userguide/service-tiers-inference.html
+type ServiceTier struct {
+	Type string `json:"type"`
```
yuzisun, line 418: make it optional for backwards compatibility as currently we do not set it

yuzisun: nvm, this is a type, it was marked optional above

internal/apischema/openai/openai.go
aabchoo: Updated the comments! Please let me know if I can make it clearer

PR #1753, opened 2026-01-10
deps: bumps tool versions

mathetake:
**Description**

This upgrades tool versions in tools/go.mod. Most notable change is that due to the golangci-lint upgrade, this renames the internal/tracing/api package to internal/tracing/tracingapi.


internal/tracing/api/api_test.go
```diff
@@ -3,7 +3,7 @@
 // The full text of the Apache license is available in the LICENSE file at
 // the root of the repo.
 
-package api
+package api // nolint: revive
```
mathetake: this is hilarious cc @codefromthecrypt 
```
internal/tracing/api/api_test.go:6:9: var-naming: avoid meaningless package names (revive)
package api
        ^
```

nacx (approved):

PR #1755, opened 2026-01-12
fix: handles weight zero backendRef gracefully

### LINKED ISSUE #1664: Weights in aigatewayroutes set to 0 affects translation

mathetake:
**Description**

Envoy Gateway skips the weight zero backendRef when constructing localityLbEndoints slices. This starts handling such cases gracefully rather than taking them as errors.

Close #1664
Supersedes #1694

### DIFF
--- a/tests/e2e/testdata/testupstream.yaml
+++ b/tests/e2e/testdata/testupstream.yaml
@@ -52,6 +52,8 @@ spec:
       kind: Gateway
       group: gateway.networking.k8s.io
   rules:
+    # The backend refs with weight zero should be ignored by the extension server.
+    # See https://github.com/envoyproxy/ai-gateway/issues/1664
     - matches:
         - headers:
             - type: Exact
@@ -65,9 +67,12 @@ spec:
               name: x-ai-eg-model
               value: another-cool-model
       backendRefs:
+        - name: translation-testupstream-cool-model-backend
+          weight: 0
         - name: translation-testupstream-another-cool-model-backend
-        # This should be ignored since weight is 0 by the extension server.
-        # See https://github.com/envoyproxy/ai-gateway/issues/1664
+          weight: 100
+        - name: translation-testupstream-cool-model-backend
+          weight: 0
         - name: translation-testupstream-cool-model-backend
           weight: 0
     - matches:
@@ -77,8 +82,6 @@ spec:
               value: whatever-model
       backendRefs:
         - name: translation-testupstream-cool-model-backend
-        # This should be ignored since weight is 0 by the extension server.
-        # See https://github.com/envoyproxy/ai-gateway/issues/1664
         - name: translation-testupstream-another-cool-model-backend
           weight: 0
 ---


internal/extensionserver/post_translate_modify.go
```diff
@@ -207,14 +207,16 @@ func (s *Server) maybeModifyCluster(cluster *clusterv3.Cluster) error {
...
-		if len(cluster.LoadAssignment.Endpoints) != len(httpRouteRule.BackendRefs) {
-			s.log.Info("LoadAssignment endpoints length does not match backend refs length",
-				"cluster_name", cluster.Name, "endpoints_length", len(cluster.LoadAssignment.Endpoints), "backend_refs_length", len(httpRouteRule.BackendRefs))
-			return nil
-		}
 		// Populate the metadata for each endpoint in the LoadAssignment.
-		for i, endpoints := range cluster.LoadAssignment.Endpoints {
-			backendRef := httpRouteRule.BackendRefs[i]
+		var lbEndpointIndex int
+		for i, backendRef := range httpRouteRule.BackendRefs {
```
aabchoo, line 212: as a sanity check, is the `i` expected to be the lbEndpointIndex or backend ref index: https://github.com/envoyproxy/ai-gateway/blob/15080b001d3f4d92944828e08cd055900eed38dc/internal/extensionserver/post_translate_modify.go#L243C78-L243C96

mathetake: yes and good catch anyways https://github.com/envoyproxy/ai-gateway/blob/15080b001d3f4d92944828e08cd055900eed38dc/internal/controller/gateway.go#L339-L342

aabchoo (approved):

PR #1757, opened 2026-01-12
mcp: do not instantiate proxy per request

mathetake:
**Description**

Previously the MCPProxy struct was create per request, which caused a recreation of http client for example, and resulted in not reusing the connection pool.

This did not result in a noticeable benchmark result diff but it will improve the perf in majority of use cases.

### DIFF
--- a/internal/controller/gateway.go
+++ b/internal/controller/gateway.go
@@ -111,12 +111,6 @@ func (c *GatewayController) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		return mcpRoutes.Items[i].CreationTimestamp.Before(&mcpRoutes.Items[j].CreationTimestamp)
 	})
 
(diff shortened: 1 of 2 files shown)


internal/mcpproxy/mcpproxy.go
```diff
@@ -52,16 +51,7 @@ func NewMCPProxy(l *slog.Logger, mcpMetrics metrics.MCPMetrics, tracer tracingap
 		// For example, if we mistakenly set /mcp here, only the route with prefix /mcp will be matched, and other routes
 		// with different prefixes will not be matched, which is not what we want.
 		"/", func(w http.ResponseWriter, r *http.Request) {
-			proxy := &MCPProxy{
```
mathetake, line 55: previously this was created per request, which was unnecessary and caused recreation of http clients

Issue #1765, opened 2026-01-14
Bug: Stale filter config after all routes are deleted from Gateway


mtparet:
## Summary

When all `AIGatewayRoute` or `MCPRoute` resources are deleted from a Gateway, the filter config secret retains stale backends and models instead of being cleared. This can cause ext-proc to route requests to non-existent backends.

## Labels

`kind/bug`, `area/controller`

## Description

The `GatewayController` has an early-return optimization that skips processing when no routes are attached to a Gateway. However, this optimization prevents the filter config from being updated to an empty state when routes are deleted.

**Bug location:** `internal/controller/gateway.go:114-118`

```go
if len(aiRoutes.Items) == 0 && len(mcpRoutes.Items) == 0 {
    c.logger.Info("No AIGatewayRoute or MCPRoute attached to the Gateway", ...)
    return ctrl.Result{}, nil  // ← Returns without updating filter config
}
```

## Steps to Reproduce

1. Create a Gateway with an AIGatewayRoute and AIServiceBackend
2. Verify the filter config secret contains the backend and model
3. Delete the AIGatewayRoute
4. Observe the filter config secret still contains the stale backend/model

## Expected Behavior

When all routes are deleted from a Gateway, the filter config secret should be updated to contain empty backends and models arrays.

## Actual Behavior

The filter config secret retains the previous backends and models, leading to:
- Stale routing configuration in ext-proc
- Potential routing to deleted backends
- Inconsistent state between Kubernetes resources and ext-proc config

## Test Case

A failing test demonstrating this bug:

```go
func TestGatewayController_StaleFilterConfigAfterAllRoutesDeleted(t *testing.T) {
    // Setup: Create gateway, route, backend, pod, deployment
    // ...

    // Reconcile to create filter config
    _, err = c.Reconcile(ctx, req)
    // Filter config has 1 backend, 1 model

    // Delete the route
    fakeClient.Delete(ctx, route)

    // Reconcile again
    _, err = c.Reconcile(ctx, req)

    // BUG: Filter config still has 1 backend, 1 model (should be 0)
}
```

## Root Cause

The early-return optimization at line 114-118 was added to skip unnecessary processing when a Gateway has no routes. However, this optimization fails to account for the case where routes previously existed and have now been deleted.

The downstream code (`reconcileFilterConfigSecret`) correctly handles empty route lists by producing an empty config, but it's never called due to the early return.

## Proposed Fix

[...]

PR #1766, opened 2026-01-14
fix: clear filter config when all routes are deleted from Gateway

mtparet:
**Description**

Fix stale filter config remaining after all AIGatewayRoutes/MCPRoutes are deleted from a Gateway.

## Related Issues

Fixes  : https://github.com/envoyproxy/ai-gateway/issues/1765

## Problem

When all `AIGatewayRoute` or `MCPRoute` resources are deleted from a Gateway, the `GatewayController` returns early without updating the filter config secret. This leaves stale backends and models in the ext-proc configuration.

**Bug location:** `internal/controller/gateway.go:114-118`

```go
if len(aiRoutes.Items) == 0 && len(mcpRoutes.Items) == 0 {
    c.logger.Info("No AIGatewayRoute or MCPRoute attached to the Gateway", ...)
    return ctrl.Result{}, nil  // ← Returns without updating filter config
}
```

**Impact:**
- Stale routing configuration in ext-proc after route deletion
- Potential routing to non-existent backends
- Inconsistent state between Kubernetes resources and ext-proc config

## Solution

Remove the early-return optimization and let the controller proceed to update the filter config with empty routes. The downstream code (`reconcileFilterConfigSecret`) correctly handles empty route lists by producing an empty config.

```diff
-	if len(aiRoutes.Items) == 0 && len(mcpRoutes.Items) == 0 {
-		c.logger.Info("No AIGatewayRoute or MCPRoute attached to the Gateway", ...)
-		return ctrl.Result{}, nil
-	}
+	// Note: We intentionally do NOT return early when there are no routes.
+	// We must update the filter config to empty state to clear any stale backends/models.
```

## Changes

| File | Change |
|------|--------|
| `internal/controller/gateway.go` | Remove early return, update comment |
| `internal/controller/gateway_stale_config_test.go` | New test file with 3 tests |
| `internal/controller/gateway_test.go` | Update existing test expectation |

## Test Plan

- [x] `TestGatewayController_StaleFilterConfigAfterAllRoutesDeleted` - Verifies filter config is cleared when all routes deleted
- [x] `TestGatewayController_PartialRouteDeletion` - Verifies partial deletion still works correctly
- [x] `TestGatewayController_RouteRecreationAfterDeletion` - Verifies route recreation works after deletion
- [x] All existing `TestGatewayController_*` tests pass
- [x] Full controller test suite passes

```bash
$ go test ./internal/controller/
ok  	github.com/envoyproxy/ai-gateway/internal/controller	2.213s
```

### DIFF
--- a/internal/controller/gateway.go
+++ b/internal/controller/gateway.go
@@ -111,11 +111,8 @@ func (c *GatewayController) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		return mcpRoutes.Items[i].CreationTimestamp.Before(&mcpRoutes.Items[j].CreationTimestamp)
 	})
 
-	if len(aiRoutes.Items) == 0 && len(mcpRoutes.Items) == 0 {
-		// This means that the gateway is not attached to any AIGatewayRoute or MCPRoute.
-		c.logger.Info("No AIGatewayRoute or MCPRoute attached to the Gateway", "namespace", gw.Namespace, "name", gw.Name)
-		return ctrl.Result{}, nil
-	}
+	// Note: We intentionally do NOT return early when there are no routes.
+	// We must update the filter config to empty state to clear any stale backends/models.
 
 	namespace, pods, deployments, daemonSets, err := c.getObjectsForGateway(ctx, gw)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


mtparet:
https://github.com/envoyproxy/ai-gateway/commit/32b737ef37e74d6ed3d5bc66fec9d545392e696d contains the fix already

nacx:
thx for taking the time to file a fix anyway!

PR #1779, opened 2026-01-16
feat: support batch inference APIs

xiaolin593:
**Description**

This proposal introduces Batch Processing API support for Envoy AI Gateway, enabling users to submit large-scale asynchronous inference jobs with significant cost savings (typically 50% vs real-time inference). See details in `007-batch-processing/proposal.md` file.

### DIFF
--- a/docs/proposals/007-batch-processing/proposal.md
+++ b/docs/proposals/007-batch-processing/proposal.md
@@ -23,6 +23,7 @@
   - [OpenAI to GCP Vertex AI](#openai-to-gcp-vertex-ai)
 - [Out of Scope](#out-of-scope)
 - [Future Work](#future-work)
+- [Batch Token Usage and Rate Limiting](#batch-token-usage-and-rate-limiting)
 - [Open Questions for Community Discussion](#open-questions-for-community-discussion)
 
 <!-- /toc -->


nacx:
Thanks for the detailed proposal!

I think there should also be a section reasoning about token usage. When using the batch APIs, token usage is not available until the client explicitly queries the results. This means that the gateway will not "see" token information unless the job is done and the client explicitly asks for the result, which could lead to metrics not properly reflecting the reality of token usage. This has implications for observability, usage accountability, and also rate limiting and quotas.

I don't think this should be a blocker for the initial API and translation, but I think it is important to discuss this topic as well in this proposal.

docs/proposals/007-batch-processing/proposal.md
```diff
@@ -0,0 +1,493 @@
...
+
+```http
+POST /v1/batches/{batch_id}/cancel
+```
+
+## Batch Request Routing Stickiness
+
+Batch APIs routing presents unique challenges compared to real-time inference because batch call(which
+happens after file request) must reference to targeted backend where file was uploaded, otherwise batch request will result
+in a "file not found" error. For instance, file request hits Azure backend and subsequent batch request hits OpenAI
```
yuzisun, line 309: I think we can create a generic batch AIGatewayRoute which routes the batch request based on the backend header.

docs/proposals/007-batch-processing/proposal.md
```diff
@@ -0,0 +1,493 @@
...
+### Batch Management
+
+#### Create Batch
+
+```http
+POST /v1/batches
+Content-Type: application/json
+
+{
+  "input_file_id": "file-abc123",
```
yuzisun, line 251: For AWS and GCP where to specify the output file for S3 and GCS url?  the input file can also be s3/gcs url

nacx (approved):

PR #1780, opened 2026-01-16
fix: support per-route LLMRequestCosts for different cost formulas

### LINKED ISSUE #1688: Different CEL calculation formulas are used across multiple models.
_Description_:

I want to implement rate limiting based on usage costs in the Envoy AI Gateway.

I have two models with slightly different billing methods. I added a `metadataKey` to provide the billing metrics for the models.

This configuration does not achieve the desired effect. In the free model, the CEL seems to be using the CEL calculation formula of the other model.

## Sample：

I accessed the free GLM45 model, but I noticed in the response headers that the "billing_charges" quota was decreasing.

```
kind: AIGatewayRoute
metadata:
  name: envoy-ai-gateway-tke-glm45
  namespace: default
spec:
  llmRequestCosts:
      type: InputToken
      type: OutputToken
      type: TotalToken
    - metadataKey: billing_charges
      type: CEL
      cel: "0"
---
kind: AIGatewayRoute
metadata:
  name: envoy-ai-gateway-tke-glm47
  namespace: default
spec:
  llmRequestCosts:
      type: InputToken
      type: OutputToken
      type: TotalToken
    - metadataKey: billing_charges
      type: CEL
      cel: "int(double(input_tokens) / 1000.0 * 30.0 * 7.2 + double(output_tokens) / 1000.0 * 150.0 * 7.2)"
---
kind: BackendTrafficPolicy
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"gateway.envoyproxy.io/v1alpha1","kind":"BackendTrafficPolicy","metadata":{"annotations":{},"name":"envoy-ai-gateway-default","namespace":"default"},"spec":{"rateLimit":{"global":{"rules":[{"clientSelectors":[{"headers":[{"name":"x-user-id","type":"Distinct"}]}],"cost":{"request":{"from":"Number","number":0},"response":{"from":"Metadata","metadata":{"key":"billing_charges","namespace":"io.envoy.ai_gateway"}}},"limit":{"requests":100000,"unit":"Day"}},{"cost":{"request":{"from":"Number","number":0},"response":{"from":"Metadata","metadata":{"key":"billing_charges","namespace":"io.envoy.ai_gateway"}}},"limit":{"requests":1000000,"unit":"Day"}}]},"type":"Global"},"targetRefs":[{"group":"gateway.networking.k8s.io","kind":"Gateway","name":"envoy-ai-gateway-default"}]}}
  creationTimestamp: "2025-12-11T15:32:59Z"
  generation: 1
  name: envoy-ai-gateway-default
  namespace: default
  resourceVersion: "19951812"
  uid: b4986c34-cfe2-4758-afb0-10668815df71
spec:
  rateLimit:
    global:
      rules:
      - clientSelectors:
        - headers:
          - invert: false
            name: x-user-id
            type: Distinct
        cost:
          request:
            from: Number
            number: 0
          response:
            from: Metadata
            metadata:
              key: billing_charges
              namespace: io.envoy.ai_gateway
        limit:
          requests: 100000
          unit: Day
      - cost:
          request:
            from: Number
            number: 0
          response:
            from: Metadata
            metadata:
              key: billing_charges
              namespace: io.envoy.ai_gateway
        limit:
          requests: 1000000
          unit: Day
    type: Global
```

jonasHanhan:
**Description**

This commit enables per-route LLMRequestCosts configuration, allowing different routes to use different CEL calculation formulas for the same metadataKey.

Previously, LLMRequestCosts were gateway-scoped and deduplicated by metadataKey, causing incorrect cost calculations when multiple routes needed different formulas (e.g., free model with `cost=0` vs paid model with `cost=input_tokens*30+output_tokens*150`).

Changes:
- Add `LLMRequestCosts` field to `Backend` struct
- Copy route's `LLMRequestCosts` to each backend during reconciliation
- Compile backend-level CEL programs in `NewRuntimeConfig`
- Use backend-level costs in `ProcessResponseBody` with fallback to global config

Fixes #1688

- AI tools were used to assist with test generation and code review
- All unit tests pass (`make precommit` passes)
- Backward compatible: if backend has no LLMRequestCosts, falls back to global config


jonasHanhan:
The `Data Plane MCP Test (ubuntu-latest)` failure appears to be a flaky test - the same test passes on macOS. All other tests pass including Controller Test, Unit Test, Data Plane Test, and all E2E Tests.

Could a maintainer please re-run the failed job? Thank you!

internal/controller/gateway.go
```diff
@@ -398,6 +433,14 @@ func (c *GatewayController) reconcileFilterConfigSecret(
 					}
 				}
 
+				// Assign the route's LLMRequestCosts to this backend.
+				// This ensures each backend has its own cost calculation configuration,
+				// allowing different routes to use different CEL expressions for the same metadataKey.
+				b.LLMRequestCosts, err = llmRequestCostsToFilterAPI(aiGatewayRoute.Spec.LLMRequestCosts)
```
mathetake: let's construct this once instead of repeating it inside the for loop here

internal/extproc/processor_impl.go
```diff
@@ -430,8 +433,13 @@ func (u *upstreamProcessor[ReqT, RespT, RespChunkT, EndpointSpecT]) ProcessRespo
 		u.metrics.RecordTokenUsage(ctx, u.costs, u.requestHeaders)
 	}
 
-	if body.EndOfStream && len(u.parent.config.RequestCosts) > 0 {
-		metadata, err := buildDynamicMetadata(u.parent.config, &u.costs, u.requestHeaders, u.backendName)
+	// Use backend-level request costs if available, otherwise fall back to global config.
+	requestCosts := u.backendRequestCosts
+	if len(requestCosts) == 0 {
+		requestCosts = u.parent.config.RequestCosts
+	}
```
mathetake: when this happens? We can delete the global level ones if we merge this after v0.5 which comes with the proper config version checking which allows us to avoid breaking changes

nacx (commented):
Can you add a section on the website page that explains cost-based rate limiting, with a small example on how to achieve different costs per backend? This could be a useful feature but it's not easy to spot that the way of doing that with the current API is creating N AIGatewayRoutes.

internal/controller/gateway.go
jonasHanhan: Done! Moved the llmRequestCostsToFilterAPI call outside the backendRef loop. Thanks for the review!

internal/extproc/processor_impl.go
jonasHanhan: The global fallback happens when the backend doesn't have any LLMRequestCosts configured (for backward compatibility with existing configurations that only use the global LLMRequestCosts).

If you'd like me to remove the global level fallback since this will be merged after v0.5 with the config version checking, I'm happy to do that. Just let me know!

internal/filterapi/filterconfig.go
```diff
@@ -148,6 +148,9 @@ type Backend struct {
 	HeaderMutation *HTTPHeaderMutation `json:"httpHeaderMutation,omitempty"`
 	// Body mutations to be applied to the request before sending to the backend. Optional.
 	BodyMutation *HTTPBodyMutation `json:"httpBodyMutation,omitempty"`
+	// LLMRequestCosts configures the cost calculation for this backend. Optional.
+	// This allows different routes/backends to have different cost calculation formulas.
+	LLMRequestCosts []LLMRequestCost `json:"llmRequestCosts,omitempty"`
```
yuzisun: If you want to track per backend then it should use the new Quota API. The  BackendTrafficPolicy is applied at the route level.

yuzisun:
@jonasHanhan @nacx i do not think this is the right way to implement per backend quota which should use the new Quota API designed for this purpose. The current token rate limit is at the route level so we can’t copy it from route to backend.

site/docs/capabilities/traffic/usage-based-ratelimiting.md
```diff
@@ -84,6 +84,66 @@ spec:
...
+      type: CEL
+      cel: "input_tokens * uint(30) + output_tokens * uint(150)" # Pay per token
+```
+
+:::tip
+When using this pattern:
+
+- Each `AIGatewayRoute` has its own `llmRequestCosts` configuration
+- Different routes can use the same `metadataKey` (e.g., `billing_charges`) with different CEL expressions
+- The cost calculation is automatically applied per-backend, allowing fine-grained cost control
```
yuzisun: If the quota needs to be applied at backend level then it should attach the policy to AIServiceBackend instead of AIGatewayRoute.

internal/filterapi/filterconfig.go
jonasHanhan: Thanks for the feedback! I see there's a TODO in `ai_service_backend.go:78-79` suggesting backend-level LLMRequestCost configuration:

```go
// TODO: maybe add backend-level LLMRequestCost configuration that overrides the AIGatewayRoute-level LLMRequestCost.
```

My current implementation copies the route's LLMRequestCosts to each backend to match @aabchoo's per-backendRef proposal mentioned in #1688.

Would you prefer I implement this at the AIServiceBackend level instead (as the TODO suggests)? I'm happy to refactor if that's the preferred direction. Could you also point me to the Quota API you mentioned?

internal/controller/gateway.go
jonasHanhan: Done! Moved the llmRequestCostsToFilterAPI call outside the backendRef loop. Thanks for the review!

internal/extproc/processor_impl.go
jonasHanhan: The global fallback happens when the backend doesn't have any LLMRequestCosts configured (for backward compatibility with existing configurations that only use the global LLMRequestCosts).

If you'd like me to remove the global level fallback since this will be merged after v0.5 with the config version checking, I'm happy to do that. Just let me know!

internal/filterapi/filterconfig.go
yuzisun: Take a look at this PR https://github.com/envoyproxy/ai-gateway/pull/1709

jonasHanhan: Thanks for pointing to PR #1709! The QuotaPolicy approach looks comprehensive. I noticed it's been in progress for a while now — looking forward to seeing it land.

yuzisun:
@jonasHanhan i think we still need this PR but the fix is to aggregate the cost expressions at the route level instead of backend in filter config.

jonasHanhan:
Thanks for the clarification - makes sense to keep this at route level.

I will rework this PR to remove backend-level LLMRequestCosts in filter config and instead aggregate cost expressions at route level.

Proposed behavior:
- Keep llmRequestCosts defined on AIGatewayRoute.
- During filter-config generation, aggregate per-route costs by metadataKey into route-level expressions.
- Use the selected backend context only inside the aggregated expression (instead of storing costs on each backend entry).

Before I implement, can you confirm this is the expected aggregation model?
Also, should we keep or remove the global fallback path after v0.5 config version checks?

If this matches your expectation, I will push the refactor and update docs accordingly.

yuzisun:
> Thanks for the clarification - makes sense to keep this at route level.
> 
> I will rework this PR to remove backend-level LLMRequestCosts in filter config and instead aggregate cost expressions at route level.
> 
> Proposed behavior:
> 
> * Keep llmRequestCosts defined on AIGatewayRoute.
> * During filter-config generation, aggregate per-route costs by metadataKey into route-level expressions.
> * Use the selected backend context only inside the aggregated expression (instead of storing costs on each backend entry).
> 
> Before I implement, can you confirm this is the expected aggregation model? Also, should we keep or remove the global fallback path after v0.5 config version checks?
> 
> If this matches your expectation, I will push the refactor and update docs accordingly.

Hi @jonasHanhan, per backend cost tracking is handled in my PR https://github.com/envoyproxy/ai-gateway/pull/1869 with QuotaPolicy API(applies to AIServiceBackend). This one here is for tracking quota at the route level defined with BackendTrafficPolicy (only applies to HTTPRoute). We can aggregate per-route costs across all the routes and maintain a map, we still need a request attribute that tells which route is matched and currently we only have the information of the backend. 

A few items we need to do here:
  1. internal/internalapi/internalapi.go — Add route name constants
 - Add InternalMetadataRouteNameKey = "route_name" constant
 - Add XDSRouteMetadataRouteNamePath = "xds.route_metadata.filter_metadata['aigateway.envoy.io']['route_name']" constant
  2. internal/llmcostcel/cel.go — Add route_name CEL variable
  3. internal/controller/gateway.go — Route-level cost aggregation
  4. internal/extensionserver/post_translate_modify.go — Add route metadata to xDS routes
   - In enableRouterLevelAIGatewayExtProcOnRoute: for each AI-Gateway-generated route, add route name to route.Metadata.FilterMetadata['aigateway.envoy.io']['route_name']
   - Parse the AIGatewayRoute namespace/name from route.Name (format: httproute/<namespace>/<name>/rule/...)
   - Store as namespace/name in the metadata
   - In maybeModifyCluster (upstream ext_proc config, ~line 295): add internalapi.XDSRouteMetadataRouteNamePath to extProcConfig.RequestAttributes
   5. internal/extproc/server.go — Resolve route name from attributes
   6. internal/extproc/processor_impl.go — Store and use route name

internal/controller/gateway.go
```diff
@@ -203,6 +203,101 @@ func bodyMutationToFilterAPI(m *aigv1a1.HTTPBodyMutation) *filterapi.HTTPBodyMut
...
+	ret := make([]filterapi.LLMRequestCost, 0, len(metadataOrder))
+	for _, metadataKey := range metadataOrder {
+		costs := costsByMetadata[metadataKey]
+		if len(costs) == 0 {
+			continue
+		}
+		expr := "uint(0)"
+		// Keep "last definition wins" semantics for duplicate metadata keys by
+		// layering conditions in declaration order.
+		for i := 0; i < len(costs); i++ {
```
yuzisun: could use the simpler for _, cost := range costs since it doesn't need the index

internal/extensionserver/post_translate_modify.go
```diff
@@ -737,6 +748,57 @@ func (s *Server) isRouteGeneratedByAIGateway(route *routev3.Route) bool {
 	return false
 }
 
+func routeNameFromRouteConfigName(routeConfigName string) string {
```
yuzisun, line 750: Consider documenting the route name format assumption, the assumption that route names follow `httproute/<namespace>/<name>/...` should be documented.

internal/controller/gateway.go
```diff
@@ -203,6 +203,101 @@ func bodyMutationToFilterAPI(m *aigv1a1.HTTPBodyMutation) *filterapi.HTTPBodyMut
...
+func aggregateRouteLLMRequestCosts(
+	costsByMetadata map[string][]routeLLMRequestCost,
+	metadataOrder *[]string,
+	routeName string,
+	hasRouteBackend bool,
+	routeCosts []aigv1a1.LLMRequestCost,
+) error {
+	if !hasRouteBackend || len(routeCosts) == 0 {
+		return nil
+	}
```
yuzisun: hasRouteBackend can be checked outside of the function instead of passing as function parameters.

internal/extensionserver/post_translate_modify.go
```diff
@@ -579,6 +580,7 @@ func (s *Server) patchVirtualHostWithInferencePool(vh *routev3.VirtualHost, infe
 // for those routes. It returns true if any route was modified.
 func (s *Server) enableRouterLevelAIGatewayExtProcOnRoute(routeConfig *routev3.RouteConfiguration) (bool, error) {
 	enabled := false
+	defaultRouteName := routeNameFromRouteConfigName(routeConfig.Name)
```
aabchoo: nit: you can directly call `routeNameFromRouteConfigName` in line 604. no need to define the variable here only to reference it once

internal/extensionserver/post_translate_modify.go
aabchoo: @jonasHanhan could you please address this?

aabchoo:
@jonasHanhan could you please not force push each time you make a change? It's very difficult to track what has been modified.

aabchoo:
Hey @jonasHanhan, I'm going to take over this PR as this change is urgent.

internal/controller/gateway.go
```diff
@@ -390,43 +487,21 @@ func (c *GatewayController) reconcileFilterConfigSecret(
...
+					routeBackendNamesSet[b.Name] = struct{}{}
+					routeBackendNames = append(routeBackendNames, b.Name)
 				}
-				ec.LLMRequestCosts = append(ec.LLMRequestCosts, fc)
-				llmCosts[cost.MetadataKey] = struct{}{}
 			}
 		}
+		if len(routeBackendNames) > 0 {
+			if err = aggregateRouteLLMRequestCosts(llmCostsByMetadata, &llmCostMetadataOrder, routeName, aiGatewayRoute.Spec.LLMRequestCosts); err != nil {
+				return false, fmt.Errorf("failed to aggregate LLMRequestCosts for route %s: %w", aiGatewayRoute.Name, err)
```
aabchoo, line 498: once bad llm request cost can prevent the entire filter config from updating

thoughts cc @yuzisun

yuzisun: extproc also can not come up if there is bad cel expression 
https://github.com/envoyproxy/ai-gateway/blob/main/internal/extproc/server.go#L82. Either we need to validate the cel expression upfront or user need to fix the cel expression to unblock the reconcilation.

sukumargaonkar:
hey, @jonasHanhan thanks for working on this! I've been looking at the per-backend LLM cost implementation and noticed a potential issue with the current approach of aggregating costs into a single CEL expression per metadataKey. workshopped my suggestion in https://github.com/jonasHanhan/ai-gateway/pull/1

## The issue

When multiple routes define costs for the same metadataKey, the controller builds one nested ternary CEL expression like:

```
route_name == 'ns/route1' ? (uint(input_tokens)) : (route_name == 'ns/route2' ? (uint(input_tokens)) : (uint(0)))
```

This grows linearly with the number of routes. In clusters with many routes sharing the same metadata key, this can hit CEL-Go compile limits or make expressions unwieldy to debug.

## Alternative approach

I've pushed an alternative implementation to my fork and a PR to your branch (See https://github.com/jonasHanhan/ai-gateway/pull/1) that stores per-route costs instead of aggregating them into a single expression.

### Changes

**Filter API** (`internal/filterapi/filterconfig.go`)
- Added `RouteName` field to `LLMRequestCost` (empty = wildcard). The controller sets this to `namespace/name` for each route's cost rules.

**Controller** (`internal/controller/gateway.go`)
- Removed the aggregation helpers (`aggregateRouteLLMRequestCosts`, `aggregatedLLMRequestCosts`, `routeLLMRequestCost`, `routeNameMatchCondition`, `llmRequestCostToCELExpression`).
- Replaced with `aigwLLMRequestCostToFilterAPI` that converts each CRD cost to a filter-config row with its `RouteName`, preserving the original type (`InputToken`, `OutputToken`, etc.) instead of converting everything to synthetic CEL.
- Same-route duplicate metadataKeys are deduped at config-build time (last definition wins), so the extproc never needs to resolve duplicates at request time.

**Extproc** (`internal/extproc/processor_impl.go`)
- `buildDynamicMetadata` is now a single forward scan: for each cost row, if the route matches, evaluate and write to metadata. Rows scoped to a different route are skipped entirely (no zero default emitted for non-matching routes).
- Extracted `evalRuntimeRequestCost` and `routeMatchesCost` helpers for clarity.

**API docs** (`api/v1alpha1/ai_gateway_route.go`, `api/v1beta1/ai_gateway_route.go`)
- Updated the `LLMRequestCosts` field comment to describe per-route scoping instead of the old "pick one and ignore the rest" language.

### Benefits

- No expression size limits regardless of route count
- Simpler per-route CEL expressions (user-authored only, no synthetic routing logic)
- Easier validation errors (points at a specific route's rule)
- Route dispatch in Go is straightforward and testable
- Non-matching routes don't emit spurious zero-valued metadata keys

### Trade-offs

- More rows in filter config (one per route-cost pair vs. one per unique metadataKey)

### Backward compatibility

The new `routeName` field doesn't require coordinated deployment. The existing `Config.Version` guard in the extproc config watcher rejects any config where the version doesn't match its own binary version, so during a rolling upgrade the extproc simply keeps its previous config until both components are at the same version.

### Tests

Updated all controller and extproc tests to reflect the new structure:
- Controller tests assert flat per-route rows with `routeName` instead of aggregated CEL strings
- Duplicate metadataKey test verifies controller-side dedup (last wins)
- New extproc tests for route-scoped matching, wildcard routes, and cross-route key isolation

Happy to discuss if this direction makes sense or if there are aspects of the aggregated approach I'm missing. Let me know if you'd like me to open a separate PR or if you'd prefer to incorporate these changes here.

sukumargaonkar:
Notes from Community meeting on 6th April
- changes from https://github.com/envoyproxy/ai-gateway/pull/1780#issuecomment-4187300400 to be incorporated in this branch
- add field in GatewayConfig CR to allow configuring globalRequestCosts that apply to all routes

sukumargaonkar:
we can close this PR as above PR was merged

PR #1781, opened 2026-01-16
test: add token usage in testupstream fake response

fbalicchia:
**Description**
Hi, 
in this PT, I'll fix OpenAI response format in test upstream lib to correctly expose token usage for AI Gateway.
With this change, the example configuration works out of the box without additional modifications.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,7 +1,7 @@
 module github.com/envoyproxy/ai-gateway
 
 // Explicitly specify the Go patch version to be able to purge the CI cache correctly.
-go 1.25.5
+go 1.25.6
 
 require (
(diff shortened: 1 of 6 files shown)


examples/basic/aws.yaml
```diff
@@ -99,7 +99,7 @@ spec:
         hostname: bedrock-runtime.us-east-1.amazonaws.com
         port: 443
 ---
-apiVersion: gateway.networking.k8s.io/v1alpha3
```
mathetake, line 102: Can you not conflate irrelevant stuff in a single PR? This messes up the commit history

mathetake:
@fbalicchia can you fix DCO (you would need to do the force push)

PR #1789, opened 2026-01-20
fix: avoid admission cache race for extproc sidecar injection

### LINKED ISSUE #1495: Race Condition During Deployment: extProc sidecar not injected
_Description_:

We experience an interesting race condition during initial deployment of the `envoy-ai-gateway` Helm chart and a `Gateway` resource.

We have an umbrella chart with which we install the `envoy-ai-gateway` Helm chart  as dependency, and the `Gateway` resource as a template of our Helm chart.

Most of the time (approximately 90%), the `extProc` sidecar won't be injected into the AI Gateway pod after the initial installation of the Helm chart. Waiting for a while (approximately 30 minutes), then ultimately resolved the issue and the `extProc` sidecar is injected.

However, waiting for 30 minutes is, of course, not acceptable from a user's perspective.

Manually deleting the AI Gateway Pod also resolves the issue. The `extProc` side car is then injected after the deletion of the pod and we see the following logs, indicating a successful injection.
```
found routes for gateway
mutating gateway pod
```

Initially, we thought that the issue was related to the `envoy-ai-gateway` Helm chart being installed before the `AIGatewayRoute`s. However, we don't see any of the logs which would prove that during the initial installation (from `gateway_mutator.go`s method `mutatePod`).
```
failed to list routes
AIGatewayRoutes or MCPRoutes found for gateway
failed to get filter config secret
```
None of those are shown.

We tried adding an init container to the AI Gateway waiting for the AI Gateway Controller to be up and running, but that doesn't help either.

Initially after the deployment, we see several error messages like
```
ERROR	controller.gateway	failed to get backend auth from backend security policy. Skipping this backend
```
from the method `reconcileFilterConfigSecret´ (`gateway.go`).

But immediately afterwards, this "error" is resolved
```
INFO	controller.backend-security-policy.azure-token-rotator	creating a new azure access token into secret
INFO	controller.backend-security-policy	successfully rotated credential for ...
INFO	controller.backend-security-policy	Syncing AIServiceBackend ...
INFO	controller.secret	Reconciling Secret ...
INFO	envoy-gateway-extension-server	inserting AI Gateway extproc filter into listener
```
But although it says `inserting AI Gateway extproc filter`, the `extProc` sidecar isn't injected.

Do you have any idea what could be causing this interesting and inconsistent behavior and how we could resolve it? To me, this seems to be a bug.

Let us know if we should run any further tests or send you more logs.

_Repro steps_:

1. Create a new Helm chart
2. Add envoy-ai-gateway and gateway-helm as dependencies
3. Configure both to only watch their own namespace and configure Envoy Gateway to be deployed in namespace mode
4. Add a few AI Backend Routes and a Gateway
5. Install and uninstall a few times to observe that the `extProc` sidecar is injected sometimes, and sometimes it isn't

_Environment_:

`gateway-helm`: `1.5.1`
`ai-gateway-helm`: `0.0.0-2fa6b27fe9bfdb0be4f108f9dec595735050cd1d` (latest dev as of yesterday)

_Logs_:
See above.

jonasHanhan:
**Description**
Use the API reader in the gateway pod mutating webhook to avoid cache race during admission that can skip extproc injection. Add a regression test to ensure route lookups use the API reader. Testing: make precommit; go test ./internal/controller.

Fixes #1495

AI assistance used (OpenAI Codex) for analysis and implementation; reviewed and verified.


internal/controller/gateway_mutator.go
```diff
@@ -31,10 +32,12 @@ import (
...
 // gatewayMutator implements [admission.CustomDefaulter].
 type gatewayMutator struct {
-	codec  serializer.CodecFactory
-	c      client.Client
-	kube   kubernetes.Interface
-	logger logr.Logger
+	codec serializer.CodecFactory
+	c     client.Client
+	// apiReader avoids cache sync races during admission.
+	apiReader client.Reader
```
nacx: Rename to `noCacheReader` to make it explicit in the code?

internal/controller/gateway_mutator.go
```diff
@@ -111,6 +114,100 @@ func newGatewayMutator(c client.Client, kube kubernetes.Interface, logger logr.L
 	}
 }
 
+func (g *gatewayMutator) reader() client.Reader {
+	if g.apiReader != nil {
```
nacx: This would probably never be `nil` right? In any case, I'd remove this helper and let the code explicitly use the desired client every time.

internal/controller/gateway_mutator.go
```diff
@@ -111,6 +114,100 @@ func newGatewayMutator(c client.Client, kube kubernetes.Interface, logger logr.L
...
+	})
+	if cacheErr == nil && len(routes.Items) > 0 {
+		return routes, nil
+	}
+	if g.apiReader == nil {
+		return routes, cacheErr
+	}
+
+	var all aigv1a1.AIGatewayRouteList
+	if err := g.apiReader.List(ctx, &all); err != nil {
```
nacx: Can we still use the indices in the no-cached reader, instead of filtering afterwards?

internal/controller/gateway_mutator.go
```diff
@@ -111,6 +114,100 @@ func newGatewayMutator(c client.Client, kube kubernetes.Interface, logger logr.L
...
+func (g *gatewayMutator) listAIGatewayRoutesForGateway(ctx context.Context, gatewayName, gatewayNamespace string) (aigv1a1.AIGatewayRouteList, error) {
+	var routes aigv1a1.AIGatewayRouteList
+	key := fmt.Sprintf("%s.%s", gatewayName, gatewayNamespace)
+	cacheErr := g.c.List(ctx, &routes, client.MatchingFields{
+		k8sClientIndexAIGatewayRouteToAttachedGateway: key,
+	})
+	if cacheErr == nil && len(routes.Items) > 0 {
+		return routes, nil
+	}
+	if g.apiReader == nil {
```
nacx: This can probably never happen?

internal/controller/gateway_mutator.go
jonasHanhan: Renamed `apiReader` to `noCacheReader` across the mutator and tests to make the intent explicit.

internal/controller/gateway_mutator.go
jonasHanhan: Removed the `reader()` helper and now explicitly choose the reader. Also, `noCacheReader` is normalized in the constructor (falls back to `c`), so the nil branch is gone.

internal/controller/gateway_mutator.go
jonasHanhan: `noCacheReader` doesn’t share cache indexes, so it can’t use `MatchingFields`. We keep the indexed cache lookup as the fast path; the no-cache path lists then filters (with a comment explaining why).

internal/controller/gateway_mutator.go
jonasHanhan: Agreed; the nil case is now handled in the constructor, so the runtime branch is removed.

internal/controller/gateway_mutator.go
```diff
@@ -88,8 +91,12 @@ func newGatewayMutator(c client.Client, kube kubernetes.Interface, logger logr.L
 		}
 	}
 
+	if noCacheReader == nil {
+		noCacheReader = c
+	}
```
nacx: Let's not do this. If someone passes `nil` to the constructor, then we would end up executing the list calls twice with the cached reader, which is not what we want. we also don't want a variable named `noCacheReader` being a cached reader.

Instead, let this be `nil`, and let the methods that do the best effort List (first with the cached client) only attempt the non-cached if this is not nil.

jonasHanhan: Thanks for the suggestion. I kept `noCacheReader` optional (no normalization to the cached client) and only attempt the non-cached fallback when it’s non-nil, so we avoid double cached list calls and keep the intent explicit.

PR #1790, opened 2026-01-20
refactor: remove duplicate config watcher initialization

fengxsong:
**Description**

I think there might be a duplicate here... so removed redundant config watcher start call.


mathetake (approved):
Oh boy … good catch…

PR #1649, opened 2025-12-12
feat: initial implementation of openai responses api

sivanantha321:
**Description**

This PR adds initial support for OpenAI's Responses API.

**Done:**
- [x] Implement /v1/responses Endpoint & translator
- [x] Implement Metrics
- [x] Implement Traces

**Todo:**
- [ ] Replace OpenAI SDK Union types
- [ ] Improve and add more openinference traces req/res atrributes

**Related Issues/PRs (if applicable)**

Fixes #980

PR #1791, opened 2026-01-20
feat: responses api phase 2

sivanantha321:
**Description**
This PR is a follow up of initial support for OpenAI's Responses API PR and implements the following pending work.

- [x] Replace OpenAI SDK Union types with our own
- [x] Improve and add more openinference traces req/res atrributes and configuration.

**Related PRs**
Related PR: #1649


yuzisun:
We should consider build the unified API based on the open responses https://www.openresponses.org/specification

sivanantha321:
> We should consider build the unified API based on the open responses https://www.openresponses.org/specification

It seems very similar. Can we address it in a separate PR ?

PR #1792, opened 2026-01-20
feat: gemini context caching using Anthropic style API (prefix caching)

VarSuren:
**Description**

Gemini models has 2 endpoint allowing you to create cache  and re-use it in subsequent calls, this proposal describes on how would we encapsulate cache creation and subsequent usage leveraging existing Anthropic style cache keys.


docs/proposals/007-gemini-context-caching/proposal.md
```diff
@@ -0,0 +1,602 @@
...
+{
+  "error": {
+    "message": "Cannot specify both cache_control on messages and explicit cachedContent field",
+    "type": "invalid_request_error",
+    "code": "invalid_cache_config"
+  }
+}
+```
+
+## Cache Service API
```
yuzisun, line 245: We can have a contract for Cache Service API, but I think we should also implement an out of the box solution (e.g call vertex AI cache creation API directly and store the mapping in redis) in envoy ai gateway which does not depend on this service.

PR #1793, opened 2026-01-20
feat(metrics): extend TTFT histogram buckets to 60s

skongum02:
**Description**

  Extend time-to-first-token histogram bucket boundaries from 10s max to 60s max
  to capture slow responses from upstream providers. Adds buckets: 15.0, 20.0, 30.0, 45.0, 60.0.

### DIFF
--- a/.github/workflows/build_and_test.yaml
+++ b/.github/workflows/build_and_test.yaml
@@ -30,12 +30,13 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v4
-      - uses: dorny/paths-filter@de90cc6fb38fc0963ad72b210f1f284cd68cea36 # v3
+      - uses: dorny/paths-filter@668c092af3649c4b664c54e4b704aa46782f6f7c
         id: changes
         with:
           filters: |
             code:
               - '!**/*.md'
+              - '!docs/**'
               - '!site/**'
               - '!netlify.toml'
           predicate-quantifier: every # Make the filters be AND-ed


internal/metrics/genai.go
```diff
@@ -99,7 +99,7 @@ func newGenAI(meter metric.Meter) *genAI {
 			genaiMetricServerTimeToFirstToken,
 			metric.WithDescription("Time to receive first token in streaming responses."),
 			metric.WithUnit("s"),
-			metric.WithExplicitBucketBoundaries(0.001, 0.005, 0.01, 0.02, 0.04, 0.06, 0.08, 0.1, 0.25, 0.5, 0.75, 1.0, 2.5, 5.0, 7.5, 10.0),
+			metric.WithExplicitBucketBoundaries(0.001, 0.005, 0.01, 0.02, 0.04, 0.06, 0.08, 0.1, 0.25, 0.5, 0.75, 1.0, 2.5, 5.0, 7.5, 10.0, 15.0, 20.0, 30.0, 45.0, 60.0),
```
mathetake, line 102: note that this boundaries are originally from https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-metrics/#metric-gen_aiservertime_to_first_token but given that it is SHOULD, we can extend on our end

skongum02: Ah thanks!  I was not aware the original boundaries were based on otel.  Thanks for approval and merge.  I think given that larger models like DeepSeekV3 median times are well over 10s it makes sense to extend it.  Thanks!

eg
https://docs.vllm.ai/projects/recipes/en/latest/DeepSeek/DeepSeek-V3_2.html#tp8-benchmark-output

PR #1811, opened 2026-01-24
extproc: stream response to include service tier

aabchoo:
**Description**

ServiceTier is part of the stream metadata. It was not parse properly which is addressed in this PR.

https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_ConverseStream.html

**Related Issues/PRs (if applicable)**
https://github.com/envoyproxy/ai-gateway/pull/1749

**Special notes for reviewers (if applicable)**

Tested locally and confirmed service tier is set.

PR #1812, opened 2026-01-24
fix: remove request body restoration in body mutation after request translation

xiaolin593:
**Description**

Discovered a bug when sending `streaming` request to AWS Bedrock with body mutation set with `serviceTier:default` at route level. The request translation for AWS Bedrock request is incorrect due to the bug and resulting in an error from AWS Bedrock

Malformed message structure for AWS Bedrock
```
{
  ...
  "messages" : [ 
    {
      "role" : "user",
      "content" : "Explain what AWS Bedrock is in one sentence."
    } 
  ],
  "serviceTier" : {
    "type" : "default"
  }
  ....
}
```

AWS Bedrock expects message content structure [1]
```
{
 ...
  "messages" : [ {
    "role" : "user",
    "content" : [ {
      "text" : "Explain what AWS Bedrock is in one sentence."
    } ],
  } ],
  "serviceTier" : {
    "type" : "default"
  }
}
```

Error response due to malformed request body for AWS Bedrock
```
'Response code: 400\n'
{'Date': 'Sat, 24 Jan 2026 20:59:22 GMT', 'Content-Type': 'application/json', 'Content-Length': '35', 'x-amzn-RequestId': '1140c7eb-bedd-42f7-aeed-657bd6fdfc46', 'x-amzn-ErrorType': 'SerializationException:http://internal.amazon.com/coral/com.amazon.coral.service/', 'Cache-Control': 'proxy-revalidate', 'Connection': 'keep-alive'}

{"Message":"Unexpected field type"}
```

** Root Cause **
After OpenAI request format -> Bedrock request format translation is done, current logic informs BodyMutator to use original request instead of translated request to mutate with forceBodyMutation flag set to true for initial request, this causes the undesired behavior. 

**My Debugging and Analysis**:

in `endpointspec.go`'s ParseBody method return non-nil mutatedBody when it's streaming request and cost is configured which result in non-nil `mutatedBody` at [2]

in processor_impl.go's ProcessRequestBody method, `forceBodyMutation` is set to true for initial request at [3] later cause problem in 
```
if onRetry && b.originalBody != nil {
	// On retry, restore the original body first
	requestBody = b.originalBody
}
```
that resets translated body(supposed to Bedrock format) back to original OpenAI format.

**My proposed solution**:
Remove request body restoration in body mutation.

**Disclaimer**:
Debug analysis and code change in processor_impl.go were done myself and test cases were added with Claude Code and tweaked manually.

Related PR: #1600

1: https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_ConverseStream.html
2: https://github.com/envoyproxy/ai-gateway/blob/main/internal/endpointspec/endpointspec.go#L90
3: https://github.com/envoyproxy/ai-gateway/blob/main/internal/extproc/processor_impl.go#L174-L176


internal/extproc/processor_impl.go
```diff
@@ -279,8 +279,7 @@ func (u *upstreamProcessor[ReqT, RespT, RespChunkT, EndpointSpecT]) ProcessReque
 	}
 
 	// Apply body mutations from the route and also restore original body on retry.
-	bodyMutation = applyBodyMutation(u.bodyMutator, bodyMutation,
-		u.parent.originalRequestBodyRaw, forceBodyMutation, u.logger)
+	bodyMutation = applyBodyMutation(u.bodyMutator, bodyMutation, u.parent.originalRequestBodyRaw, u.onRetry(), u.logger)
```
xiaolin593: this is the gist of change.

xiaolin593:
@mathetake can you help take a look?

internal/extproc/processor_impl.go
yuzisun: the question here is that if we need to restore to original body when it is retry to apply the body mutation since the translation already happened before this call.

internal/extproc/processor_impl_test.go
```diff
@@ -1005,4 +1005,196 @@ func TestChatCompletionProcessorUpstreamFilter_ProcessRequestHeaders_WithBodyMut
...
+		inferenceConfig, ok := result["inferenceConfig"].(map[string]interface{})
+		require.True(t, ok, "inferenceConfig should be an object")
+		require.Equal(t, float64(1000), inferenceConfig["maxTokens"], "Original maxTokens from Bedrock translation should be preserved")
+
+		// Verify OpenAI-specific fields are NOT present (confirming translation wasn't erased)
+		require.NotContains(t, result, "stream", "OpenAI format should not be present - translation was erased if this fails")
+		require.NotContains(t, result, "stream_options", "OpenAI format should not be present - translation was erased if this fails")
+	})
+
+	t.Run("retry request should restore original body before applying mutations", func(t *testing.T) {
```
yuzisun: I think we need to add test case for cross format retry. This works for same-format retries (OpenAI→OpenAI) but breaks for cross-format retries (OpenAI→Bedrock→Bedrock).

internal/bodymutator/body_mutator.go
```diff
@@ -101,18 +96,17 @@ func (b *BodyMutator) Mutate(requestBody []byte, onRetry bool) ([]byte, error) {
...
-	replaceInPlace := onRetry
 	if len(b.bodyMutations.Set) > 0 {
 		for _, field := range b.bodyMutations.Set {
 			if field.Path != "" {
 				// Check value type to determine appropriate sjson method
 				// TODO handle JSON value check in configuration load time too.
 				if isJSONValue(field.Value) {
 					// Use SetRawBytes for JSON values (quoted strings, numbers, booleans, objects, arrays)
-					mutatedBody, err = sjson.SetRawBytesOptions(mutatedBody, field.Path, []byte(field.Value), &sjson.Options{ReplaceInPlace: replaceInPlace})
+					mutatedBody, err = sjson.SetRawBytesOptions(mutatedBody, field.Path, []byte(field.Value), &sjson.Options{ReplaceInPlace: true})
```
yuzisun, line 106: is it safe to override to true here? need to make sure there are no other places referencing the original data

aabchoo:
could you update the tests https://github.com/envoyproxy/ai-gateway/pull/1811/changes#diff-dc499811af8b33a60048de834f419fcd7132e36ba8ab93aba1b9412d23263b7f

internal/bodymutator/body_mutator.go
aabchoo: should be fine since the requestBody is new each retry

PR #1815, opened 2026-01-26
feat(mcp): support JWT claim forwarding to backend MCP servers

### LINKED ISSUE #1725: Support JWT claim forwarding to backend MCP servers
Title: MCP: Support JWT claim forwarding to backend MCP servers

**Description:**

The MCP proxy currently creates new HTTP requests to backends without forwarding JWT claims or the Authorization header. This prevents backends from accessing user context for authorization, auditing, or personalization.

**Current Behavior:**
- Client sends JWT to AI Gateway
- Gateway validates JWT and extracts claims
- MCP proxy creates NEW requests to backends
- Backends receive NO user context (email, ID, groups)

**Desired Behavior:**
Support one of:
1. Forward JWT Authorization header to backends
2. Extract JWT claims as HTTP headers (X-User-Email, X-User-ID, etc.) that are forwarded
3. Add `claimToHeaders` support in MCPRoute OAuth configuration

**Use Case:**
- Backend-level authorization based on user identity
- Audit logging in backends
- User-specific data filtering
- Multi-tenant isolation

**Environment:**
- AI Gateway: v0.4.0
- Envoy Gateway: v1.6.1

**Related:**
- Source code: `internal/mcpproxy/session.go:sendRequestPerBackend()`
- The MCP proxy uses `http.NewRequestWithContext()` which starts with empty headers

mtparet:
**Description**

This commit adds JWT claim forwarding support to MCP (Model Context Protocol) routes, allowing gateway operators to extract claims from validated JWTs and forward them as HTTP headers to backend MCP servers.

Key features:
- New claimToHeaders field in MCPRouteOAuth API to define claim-to-header mappings
- Leverages Envoy Gateway's native ClaimToHeaders in SecurityPolicy.JWTProvider
- Support for nested claims using dot notation (e.g., realm_access.roles)
- Security: client-provided headers matching configured names are stripped to prevent forgery

This enables backends to access user identity for authorization, audit logging, or personalization without parsing JWTs themselves.

Example configuration:
```yaml
spec:
  securityPolicy:
    oauth:
      claimToHeaders:
        - claim: sub
          header: X-User-Id
        - claim: email
          header: X-User-Email
        - claim: realm_access.roles
          header: X-User-Roles
```

Fixes #1725

  - Implementation leverages Envoy's JWT filter for claim extraction and validation - claims are read from headers set by Envoy, not re-extracted from the JWT
- Maximum of 16 claim-to-header mappings are allowed per route to prevent abuse
  - Claim headers stored in session and forwarded on all backend requests (initialize + subsequent calls)

### DIFF
--- a/internal/extensionserver/mcproute.go
+++ b/internal/extensionserver/mcproute.go
@@ -71,6 +71,10 @@ func (s *Server) createBackendListener(mcpHTTPFilters []*httpconnectionmanagerv3
 	httpConManager := &httpconnectionmanagerv3.HttpConnectionManager{
 		StatPrefix: fmt.Sprintf("%s-http", mcpBackendListenerName),
 		AccessLog:  accessLogConfig,
+		// Match the :scheme pseudo-header to the upstream transport protocol.
+		SchemeHeaderTransformation: &corev3.SchemeHeaderTransformation{
+			MatchUpstream: true,
+		},
 		RouteSpecifier: &httpconnectionmanagerv3.HttpConnectionManager_Rds{
 			Rds: &httpconnectionmanagerv3.Rds{
 				RouteConfigName: fmt.Sprintf("%s-route-config", mcpBackendListenerName),
--- a/internal/extensionserver/mcproute_test.go
+++ b/internal/extensionserver/mcproute_test.go
@@ -89,6 +89,7 @@ func TestServer_createBackendListener(t *testing.T) {
 
 			hcm, _, err := findHCM(listener.FilterChains[0])
 			require.NoError(t, err)
+			require.True(t, hcm.GetSchemeHeaderTransformation().GetMatchUpstream(), "SchemeHeaderTransformation.MatchUpstream should be true")
 			require.Len(t, hcm.AccessLog, len(tt.accessLogConfig))
 			for i := range tt.accessLogConfig {
 				require.Equal(t, tt.accessLogConfig[i].Name, hcm.AccessLog[i].Name)


fbalicchia:
Hi @mtparet, thanks for your work. Does this PR also forward the JWT Authorization header? If not, do you plan to add support for it?

mtparet:
@fbalicchia this PR does not enable forwarding the original JWT token, it could be added as another feature.

internal/mcpproxy/handlers.go
```diff
@@ -1188,6 +1193,91 @@ func extractSubject(r *http.Request) string {
 	return claims.Subject
 }
 
+// extractClaimHeaders extracts JWT claims from the Authorization header and maps them to HTTP headers.
+// This is used for forwarding user identity to backend MCP servers.
+// The function assumes the JWT has already been validated by the gateway's security policy.
+func extractClaimHeaders(r *http.Request, claimToHeaders []filterapi.ClaimToHeader) map[string]string {
```
zhaohuabing: Instead of extracting claims inside the mcp proxy, a simpler and more envoy-native way could be leveraging the envoy's jwt filter.

We can set `JWTProvider.ClaimToHeaders` in EG's SecurityPolicy here: https://github.com/envoyproxy/ai-gateway/blob/b3608a8169661d325dc02edd60ae12b4762c5b47/internal/controller/mcp_route_security_policy.go#L121-L151.

These headers still need to be copied form the incoming requests and forwarded to the backends.

internal/mcpproxy/handlers.go
```diff
@@ -326,7 +326,12 @@ func (m *mcpRequestContext) servePOST(w http.ResponseWriter, r *http.Request) {
 				onErrorResponse(w, http.StatusInternalServerError, "missing route header")
 				return
 			}
-			err = m.handleInitializeRequest(ctx, w, msg, params.(*mcp.InitializeParams), route, extractSubject(r), span)
+			// Extract JWT claim headers configured for this route to forward to backends.
+			var claimHeaders map[string]string
+			if routeConfig := m.routes[route]; routeConfig != nil {
+				claimHeaders = extractClaimHeaders(r, routeConfig.claimToHeaders)
+			}
+			err = m.handleInitializeRequest(ctx, w, msg, params.(*mcp.InitializeParams), route, extractSubject(r), claimHeaders, span)
```
zhaohuabing: Question: would subsequent session requests also need the claim headers?

internal/mcpproxy/handlers.go
mtparet: Thank you, I will take a look.

mtparet: I updated the PR to ease this better way.

internal/mcpproxy/handlers.go
mtparet: I suppose yes because we cannot suppose MCP servers will handle session storage with these headers informations on their side.

api/v1alpha1/mcp_route.go
```diff
@@ -232,6 +232,40 @@ type MCPRouteOAuth struct {
...
+	//
+	// Security considerations:
+	// - Any client-provided headers matching the configured header names will be stripped to prevent forgery
+	// - Only the specified claims are extracted; the full JWT is not forwarded to backends
+	// - Consider using a header prefix (e.g., "X-Jwt-Claim-") to avoid conflicts with other headers
+	//
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:MaxItems=16
+	// +optional
+	ClaimToHeaders []ClaimToHeader `json:"claimToHeaders,omitempty"`
```
zhaohuabing: Nit: can we use egv1a1.ClaimToHeader here? AIGW and EG are two very closed projects, and the two `ClaimToHeader` structs are the same.

mtparet: I adjusted this too, thanks!

manifests/charts/ai-gateway-crds-helm/templates/aigateway.envoyproxy.io_aigatewayroutes.yaml
```diff
@@ -1,8 +1,3 @@
-# Copyright Envoy AI Gateway Authors
-# SPDX-License-Identifier: Apache-2.0
-# The full text of the Apache license is available in the LICENSE file at
-# the root of the repo.
```
zhaohuabing, line 4: Why the header is removed?

mtparet: That's a mistake, going to take it back.

mtparet:
> Could you take a look at the e2e tests and the internal testMCP server and see if you can add an end to end test that validates, on the server side, that the headers extracted from the claims are received?

> You may be able to update the MCP OAuth e2e test to configure header extraction and update the test server to verify extracted headers are present?

@nacx I tried to setup some of this, let me know :)

internal/mcpproxy/config.go
```diff
@@ -39,9 +39,10 @@ type (
 	}
 
 	mcpProxyConfigRoute struct {
-		backends      map[filterapi.MCPBackendName]filterapi.MCPBackend
-		toolSelectors map[filterapi.MCPBackendName]*toolSelector
-		authorization *compiledAuthorization
+		backends       map[filterapi.MCPBackendName]filterapi.MCPBackend
+		toolSelectors  map[filterapi.MCPBackendName]*toolSelector
+		authorization  *compiledAuthorization
+		claimToHeaders []filterapi.ClaimToHeader
```
nacx: With the recent change and having the JWT filter already push the claims to headers, we just need to know the "list of headers to forward" (we don't need to know the details that they're related to JWT claims).

Can we make this just a `[]string` to be the `forwardHeaders` or something like that, so that it is a bit more generic? This would allow us to easily reuse this to forward other headers in the future, even if they don't come from a JWT token.

internal/mcpproxy/handlers.go
```diff
@@ -493,9 +498,9 @@ func errorType(err error) metrics.MCPErrorType {
 }
 
 // handleInitializeRequest handles the "initialize" JSON-RPC method.
-func (m *mcpRequestContext) handleInitializeRequest(ctx context.Context, w http.ResponseWriter, req *jsonrpc.Request, p *mcp.InitializeParams, route, subject string, span tracingapi.MCPSpan) error {
+func (m *mcpRequestContext) handleInitializeRequest(ctx context.Context, w http.ResponseWriter, req *jsonrpc.Request, p *mcp.InitializeParams, route, subject string, claimHeaders map[string]string, span tracingapi.MCPSpan) error {
```
nacx: The `claimHeaders` is already available as it is accessible from the `mcpRequestContext`. Instead of passing it as a parameter, can we just read it from `m.ProxyConfig` where needed? This would avoid polluting method signatures with unnecessary stuff and keeping the code a bit cleaner.
(Apply this to the other modified signatures where the information is already available).

internal/mcpproxy/session.go
```diff
@@ -45,6 +45,9 @@ type session struct {
 	reqCtx             *mcpRequestContext
 	mu                 sync.RWMutex
 	perBackendSessions map[filterapi.MCPBackendName]*compositeSessionEntry
+	// claimHeaders contains JWT claims extracted from the initial request and forwarded to backends.
+	// The key is the HTTP header name, the value is the claim value.
+	claimHeaders map[string]string
```
nacx: Similar to the previous comment, I'd name this `extraHeaders` or something like this.

Also, some important things about the comment describing the property:

1. Can you comment that this property will NOT be encoded in the session ID? We encode the backend and other routing information, and it is important to mention thiat this will not be included in the session IDs.
2. When you say "initial request", I think it's not entirely correct, because the headers belong to the request that triggered the call to "newSession". When receiving a request, some of the information is derived from the ID (not the current request, but from the ID that was generated on Initialize), but these headers to forward are always taken from the "current" request, so the "initial" wording could be misleading.

nacx:
There is great progress in this PR. Any chance you could take a look at the final round of comments?

mtparet:
Yes sorry I was pretty busy these latest days, will adress latest comments soon.

mtparet:
@nacx I addressed your last comments, let me know :)

Makefile
nacx, line 1: Can you undo the change to this file? It is the one breaking the build

mtparet:
(I pushed a fix for e2e test)

Makefile
mtparet: I did it initially because `make api-gen` was removing all header without this but I will revert it, yes.

mtparet: Done!

PR #1816, opened 2026-01-26
chore: removes deprecated metadata use

mathetake:
**Description**

This removes the deprecated metadata use since #1660.


internal/extproc/server.go
```diff
@@ -316,45 +315,18 @@ func (s *Server) setBackend(ctx context.Context, p Processor, internalReqID stri
 	}
 	// metadataFieldKey is the key for the entire metadata field in the attributes for backward compatibility.
```
aabchoo, line 316: remove comment

aabchoo: @mathetake I think you remove the wrong comment 😅

Issue #902, opened 2025-07-17
Bring back Kubernetes Service support in AIServiceBackend's ref


mathetake:
Kubernetes Service was temporarily disabled as a BackendRef in AIserviceBackend. We can bring it back by support of the EG native endpoint metadata (https://github.com/envoyproxy/gateway/pull/5875) via " xds.upstream_host_locality_metadata" https://github.com/envoyproxy/envoy/pull/39604 in setBackend function: https://github.com/envoyproxy/ai-gateway/blob/18379e87ed56e3fd9ad9124df61154a20e46d7dc/internal/extproc/server.go#L281

mathetake:
After spending a few hours, this seems like much more nuanced than I thought. Currently, upstream_host_locality_metadata only carries the BackendRef's name (+namespace, kind, port) https://github.com/envoyproxy/gateway/pull/5875 but that doesn't include the exact index of the backend reference in a specific rule. That makes sense because EG doesn't have any configuration knob on the backend reference vs us having priorities and modelNameOverride. Even if that index info is available in the metadata, we still cannot support the per-backendRef level priority as the endpoints are not passed to extension server (note that with Service, EG distributes endpoints via EDS, not together with CDS.). 

So, i think I would defer this to v0.4.0 with the endpoints support in the extension server after EG v1.5. With that extension server support, we won't need to rely on upstream_host_locality_metadata but just we can construct our own metadata just like the current implementation details for EG's Backend support.

cc @arkodg

PR #1822, opened 2026-01-29
fix: resolve backend name from cluster metadata when LoadAssignment is nil

codefromthecrypt:
**Description**

In standalone mode (`aigw run`), EDS-managed clusters arrive with `LoadAssignment=nil`. The extension server logged and returned without attaching backend metadata, so the upstream ext_proc filter could not resolve a backend and request processing failed.

This change keeps the metadata flow intact:

1. **Extension server** (`post_translate_modify.go`): when `LoadAssignment` is nil,  set cluster-level metadata with `per_route_rule_backend_name` so upstream ext_proc can fall back to cluster metadata. When endpoints are present, continue to stamp the same metadata on every `LbEndpoint`.

2. **ext_proc server** (`server.go`): simplify `resolveBackendName` to the current two-path logic—use upstream-host metadata for router filters, cluster metadata for endpoint-picker filters, and (for router filters) fall back to cluster metadataa when host metadata is missing.

Tests touch both sides: ext_proc `TestResolveBackendName` now mirrors the new path/fallback rules, and extension server tests cover the cluster-level metadata fallback when `LoadAssignment` is nil.

N/A

### DIFF
--- a/tests/e2e-aigw/otel_test.go
+++ b/tests/e2e-aigw/otel_test.go
@@ -3,18 +3,15 @@
 // The full text of the Apache license is available in the LICENSE file at
 // the root of the repo.
 
-package vcr
+package e2emcp
 
 import (
-	"bytes"
-	_ "embed"
 	"fmt"
 	"io"
 	"net/http"
 	"testing"
 
 	"github.com/stretchr/testify/require"
-	"github.com/tidwall/sjson"
 	commonv1 "go.opentelemetry.io/proto/otlp/common/v1"
 	metricsv1 "go.opentelemetry.io/proto/otlp/metrics/v1"
 	tracev1 "go.opentelemetry.io/proto/otlp/trace/v1"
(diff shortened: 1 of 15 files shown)


codefromthecrypt:
```
                                |
                      +---------+---------+
                      |  Extension Server |
                      |  (xDS/gRPC)       |
                      +---------+---------+
                                |
                maybeModifyCluster(cluster)
                                |
                LoadAssignment == nil ?
                /                       \
             YES                         NO
   (EDS-managed, no endpoints)   (static/inline endpoints)
              |                          |
   set cluster-level metadata     set endpoint-level metadata
   (per_route_rule_backend_name   (per_route_rule_backend_name
    under cluster filter meta)     on each LbEndpoint)
              |
              v
      +-------+--------+
      |     Envoy       |
      +-------+--------+
              |
      upstream ext_proc
      setBackend -> resolveBackendName()
              |
     isEndpointPicker?
         /             \
       YES             NO
  use xds.cluster_metadata   use xds.upstream_host_metadata
       path                      path
                                (fallback to cluster path
                                 if host metadata missing)
              |
     backend name found?
         /             \
       YES             NO -> error: missing backend name
              |
              v
      Backend resolved
      OTel spans + metrics exported

```

codefromthecrypt:
@nacx @mathetake this was really hard for me to hunt down! innocently stumbled across this trying to get a test in.

tests/e2e-aigw/otel_test.go
```diff
@@ -0,0 +1,247 @@
...
+	require.NoError(t, err)
+	defer resp.Body.Close()
+
+	body, err := io.ReadAll(resp.Body)
+	require.NoError(t, err)
+
+	require.Equal(t, http.StatusOK, resp.StatusCode, "Response body: %s", string(body))
+
+	// Get the span to extract actual token counts and duration.
+	span := env.collector.TakeSpan()
```
codefromthecrypt, line 74: this test triggered the bug. it was always nil. couldn't figure out why until setting the otel span exporter to console which showed missing stuff, eventually leading to the metadata miss

codefromthecrypt:
due to #1816 the logic is simpler. I updated this PR and and desc and diagrams

mathetake (approved):
So ideally, EG should add support for manipulating EDS via extension server so that we could resolve #902 #950 (including this) but i guess at least this works for the standalone mode

codefromthecrypt:
@mathetake thanks for the help. I tried to elaborate what you might have meant. I don't have enough experience so any project lead could probably edit the issue desc and fix up my mistakes cc @arkodg @zirain @zhaohuabing @nacx https://github.com/envoyproxy/gateway/issues/8150

PR #1831, opened 2026-02-01
feat: add text-to-speech OpenAI endpoint

xiaolin593:
**Description**

Add OpenAI /v1/audio/speech endpoint support - https://platform.openai.com/docs/api-reference/audio/createSpeech

  Implements the OpenAI text-to-speech endpoint with support for both binary audio responses (MP3, WAV, FLAC, etc.) and SSE streaming mode. The
  implementation follows the established AI Gateway patterns with full observability support including OpenTelemetry tracing and GenAI metrics.

  Features:
  - Binary and SSE streaming response modes
  - Model override and routing support
  - All standard voices (alloy, ash, ballad, coral, echo, fable, onyx, nova, sage, shimmer, verse, marin, cedar)
  - Audio format options (mp3, opus, aac, flac, wav, pcm)

  Note:
  - Custom voice IDs not yet supported


xiaolin593:
@yuzisun @aabchoo can you help take a look?

yuzisun (approved):

PR #1843, opened 2026-02-09
api: anthropic messages schema tracing

### LINKED ISSUE #1799: Complete Anthropic /messages schema and tracing
_Description_:
 
We have quite a few TODOs in our Anthropic schema, e.g. here  https://github.com/envoyproxy/ai-gateway/blob/fefb039dc230c2da481254d88b27ed034bcf45a4/internal/apischema/anthropic/anthropic.go#L149

which eventually should be used in enriching the tracing attributes e.g. https://github.com/envoyproxy/ai-gateway/blob/fefb039dc230c2da481254d88b27ed034bcf45a4/internal/tracing/openinference/anthropic/messages.go#L154

changminbark:
**Description**
This PR adds more struct definitions and their corresponding JSONMarshal and JSONUnmarshal methods according to the Anthropic Messages API. These are metrics that are provided in the latest Anthropic messages API. Currently, these metrics/metadata are not being tracked for tracing purposes. Eventually, these metrics should be used in openinference tracing as seen in 
https://github.com/envoyproxy/ai-gateway/blob/fefb039dc230c2da481254d88b27ed034bcf45a4/internal/tracing/openinference/anthropic/messages.go#L154

It does not introduce any tracing as that should be opened in a new issue. This means that another issue should be opened to include tracking these metrics.

Fixes #1799 

Claude code was used to write most of the code here, but I manually checked all of code that it generated. One thing that I noticed is that some of the previous API schema for Anthropic implemented in this repository were moved to beta status by Anthropic. This is something that the maintainers should discuss.

1. https://platform.claude.com/docs/en/api/messages#body-messages-content
2. https://platform.claude.com/docs/en/api/messages/create
3. https://platform.claude.com/docs/en/api/beta/messages#body-messages-content
4. https://platform.claude.com/docs/en/api/beta/messages/create

### DIFF
--- a/internal/controller/gateway.go
+++ b/internal/controller/gateway.go
@@ -142,11 +142,12 @@ func (c *GatewayController) Reconcile(ctx context.Context, req ctrl.Request) (ct
 	// Finally, we need to annotate the pods of the gateway deployment with the new uuid to propagate the filter config Secret update faster.
 	// If the pod doesn't have the extproc container, it will roll out the deployment altogether which eventually ends up
 	// the mutation hook invoked.
-	if err := c.annotateGatewayPods(ctx, pods, deployments, daemonSets, uid, hasEffectiveRoutes, len(mcpRoutes.Items) > 0); err != nil {
+	result, err := c.annotateGatewayPods(ctx, pods, deployments, daemonSets, uid, hasEffectiveRoutes, len(mcpRoutes.Items) > 0)
+	if err != nil {
 		c.logger.Error(err, "Failed to annotate gateway pods", "namespace", gw.Namespace, "name", gw.Name)
 		return ctrl.Result{}, err
 	}
-	return ctrl.Result{}, nil
+	return result, nil
 }
 
 // schemaToFilterAPI converts an aigv1a1.VersionedAPISchema to filterapi.VersionedAPISchema.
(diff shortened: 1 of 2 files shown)


changminbark:
@nacx I wanted to confirm with the new Anthropic beta messages API and how we should deal with that. Some of the old fields from the schema were migrated to beta status by Anthropic. Should I go ahead and implement the rest of the structs?

changminbark:
@nacx I went ahead and implemented every schema parameter (mainly using Claude Code since it is simple struct defining and UnmarshalJSON/MarshalJSON). I checked through everything and it looks good. Some of the MarshalJSON functions use value receivers instead of pointer receivers because they are required fields in the parent structs (does not make sense to make them pointers as that would imply that they are optional fields). 

I am unsure about how to run the code coverage test but I can add more unit tests if necessary.

ehfd:
@nacx Can the CI be approved?

changminbark:
@gavrissh I don't think the failing e2e tests are due to my commits. Or am I wrong?

image

adirags:
I tried hitting an EAIGW with this change from a Claude Code client and was getting
```
time=2026-02-28T02:01:43.036Z level=ERROR msg="returning user-facing error for malformed request" request_id=9bfc5c6c-1c51-4e8e-aec1-cc5d64d07baa is_upstream_filter=false isUpstreamFilter=false error="malformed request: failed to parse JSON for /v1/messages"
```

I'm running the claude code CLI from a linux machine
```
$ claude --version
2.1.62 (Claude Code)
-----
❯ hi
  ⎿  API Error: 400 {"type":"error","error":{"type":"BadRequest","code":"400","message":"malformed request: failed to
     parse JSON for /v1/messages"}}
```

I added debugs and was able to narrow the issue down to the `(*ToolUnion).UnmarshalJSON()` function, where we were hitting `return errors.New("missing type field in tool")`.

I haven't read the anthropic spec, but do we need to allow for clients coming in without the `"type"` field as well? I'm not sure if the behavior differs on Claude Code across versions, or clients
Can we treat the tool as "custom" if the `type` is not provided?  
```
        if !typ.Exists() {
-               return errors.New("missing type field in tool")
+               // No type field: treat as a custom (user-defined) tool
+               var tool Tool
+               if err := json.Unmarshal(data, &tool); err != nil {
+                       return fmt.Errorf("failed to unmarshal tool: %w", err)
+               }
+               t.Tool = &tool
+               return nil
        }
```
Or even just fall through the the `default` case and do nothing?

ehfd:
@adirags 
I think we can follow up in another issue. I need to troubleshoot something else as well.

changminbark:
@adirags How is your Claude Code set up? What model server/upstream is the EAIGW pointing to? There is a new PR for helping translate Anthropic requests to OpenAI compatible schema and back #1878. It might be best to try it out again after that PR is merged. If it still does not work then, I think it's better to open a new issue.

adirags:
@changminbark, I had the EAIGW pointing to a Sonnet 4.6 model on Vertex AI.
I can retry once #1878 is merged. Thanks

PR #1844, opened 2026-02-09
feat: update reasoning effort mapping for gemini-3-flash 

hustxiayang:
**Description**

For gemini-3-flash-preview, there are some additional reasoning efforts: https://docs.cloud.google.com/vertex-ai/generative-ai/docs/start/get-started-with-gemini-3#openai-example

### DIFF
--- a/internal/apischema/openai/openai.go
+++ b/internal/apischema/openai/openai.go
@@ -2163,6 +2163,9 @@ type ResponseRequest struct {
 	// [Learn more](https://platform.openai.com/docs/guides/background).
 	Background *bool `json:"background,omitempty"`
 
+	// Context management configuration for this request.
+	ContextManagement []ResponseContextManagement `json:"context_management,omitzero"`
+
 	// A system (or developer) message inserted into the model's context.
 	//
 	// When using along with `previous_response_id`, the instructions from a previous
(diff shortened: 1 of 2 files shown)


internal/translator/gemini_helper.go
```diff
@@ -568,19 +568,41 @@ func reasoningEffortAvailable(requestModel internalapi.RequestModel) bool {
...
+// The mapping depends on the model type:
+// - "none" → ThinkingLevelMinimal (Gemini Flash only)
 // - "low" → ThinkingLevelLow
-// - "medium" → ThinkingLevelHigh
-// https://ai.google.dev/gemini-api/docs/gemini-3?thinking=low#openai_compatibility
-func mapReasoningEffortToThinkingLevel(reasonEffort openaisdk.ReasoningEffort) (genai.ThinkingLevel, error) {
+// - "medium" → ThinkingLevelMedium for Flash, ThinkingLevelHigh for Pro
+// - "high" → ThinkingLevelHigh
+// https://docs.cloud.google.com/vertex-ai/generative-ai/docs/start/get-started-with-gemini-3#openai-example
+func mapReasoningEffortToThinkingLevel(reasonEffort openaisdk.ReasoningEffort, model internalapi.RequestModel) (genai.ThinkingLevel, error) {
```
yuzisun, line 583: this function is only called for gemini models right as in the gemini_helper.go

yuzisun: so do we need the check for isGeminiFlashModel ?

hustxiayang: `this function is only called for gemini models right as in the gemini_helper.go` yes; `so do we need the check for isGeminiFlashModel ?` I think so, because the mapping is different, it's kind of weird that the medium is mapped to different levels for pro and flash

hustxiayang:
@yuzisun Any further concerns?

PR #1846, opened 2026-02-09
feat: add structured output for claude models

hustxiayang:
**Description**
Currently only claude-4.5 and claude-4.6 models support direct `output_config.format`, for other models, we would add work-arounds.

### DIFF
--- a/cmd/extproc/mainlib/main.go
+++ b/cmd/extproc/mainlib/main.go
@@ -282,6 +282,7 @@ func Main(ctx context.Context, args []string, stderr io.Writer) (err error) {
 	embeddingsMetricsFactory := metrics.NewMetricsFactory(meter, metricsRequestHeaderAttributes, metrics.GenAIOperationEmbedding)
 	imageGenerationMetricsFactory := metrics.NewMetricsFactory(meter, metricsRequestHeaderAttributes, metrics.GenAIOperationImageGeneration)
 	responsesMetricsFactory := metrics.NewMetricsFactory(meter, metricsRequestHeaderAttributes, metrics.GenAIOperationResponses)
+	speechMetricsFactory := metrics.NewMetricsFactory(meter, metricsRequestHeaderAttributes, metrics.GenAIOperationSpeech)
 	rerankMetricsFactory := metrics.NewMetricsFactory(meter, metricsRequestHeaderAttributes, metrics.GenAIOperationRerank)
 	mcpMetrics := metrics.NewMCP(meter, metricsRequestHeaderAttributes)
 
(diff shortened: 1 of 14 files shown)


internal/translator/anthropic_helper.go
```diff
@@ -608,6 +619,23 @@ func buildAnthropicParams(openAIReq *openai.ChatCompletionRequest) (params *anth
 		ToolChoice: toolChoice,
 	}
 
+	// 5. Handle structured outputs (ResponseFormat -> OutputConfig).
+	// See: https://platform.claude.com/docs/en/build-with-claude/structured-outputs
+	if openAIReq.ResponseFormat != nil && openAIReq.ResponseFormat.OfJSONSchema != nil && outputConfigAvailable(openAIReq.Model) {
+		// Convert OpenAI JSON schema to Anthropic OutputConfig format
+		var schemaMap map[string]any
```
yuzisun, line 627: I think this map introduces the ordering issue we had before

xiaolin593: +1. please see https://github.com/envoyproxy/ai-gateway/pull/1299

internal/translator/anthropic_helper.go
```diff
@@ -576,6 +576,17 @@ func getThinkingConfigParamUnion(tu *openai.ThinkingUnion) *anthropic.ThinkingCo
 	return result
 }
 
+// outputConfigAvailable checks if the model supports structured outputs (OutputConfig).
+// Structured outputs are available on Claude Opus 4.6, Claude Sonnet 4.5, Claude Opus 4.5, and Claude Haiku 4.5.
+// See: https://platform.claude.com/docs/en/build-with-claude/structured-outputs
+func outputConfigAvailable(model internalapi.RequestModel) bool {
+	modelLower := strings.ToLower(model)
+	return strings.Contains(modelLower, "4.5") ||
+		strings.Contains(modelLower, "4.6") ||
```
xiaolin593: model name in request will be `claude-opus-4-6` only?

hustxiayang: currently they only released `claude-opus-4-6`, you have concerns on `claude-sonnet-4-6` etc do not support it?

hustxiayang: could limit it to be just `claude-opus-4-6` if you think it's safer

xiaolin593: https://platform.claude.com/docs/en/about-claude/models/migration-guide 
model name will be `claude-opus-4-6` and `claude-opus-4-5` with hyphen without dot in the name.

yuzisun: is it true for both AWS and GCP as this logic is now shared

xiaolin593: aws: anthropic.claude-opus-4-6-v1
gcp: claude-opus-4-6

hustxiayang:
@yuzisun @xiaolin593 verified that it works on bedrock and not work for gcp, so I skipped gcp schema. For others I think I have resolved them.

Issue #1857, opened 2026-02-12
extProc sidecar not injected in some race condition scenarios


gavrissh:
_Description_:
In scenarios where all routes are removed, the extProc sidecar is removed with a rollout. Any new routes added during the rollout will not restore the extProc sidecar. Resulting in a state where extProc is not present, with routes being present
 
_Repro steps_:
AIGW Deployment with 1 replica
- Create AIGW/MCP Routes -> injects extProc sidecar
- After a while, remove all Routes -> triggers a rollout without the extProc sidecar
- Create Routes again when the rollout is in progress -> As there is a pod with a sidecar(Terminating state) the sidecar check for rollout gets bypassed
- After rollout is complete, sidecar won't be present

AIGW Deployment with >1 replicas
- Create AIGW/MCP Routes -> injects extProc sidecar
- After a while, remove all Routes -> triggers a rollout without the extProc sidecar
- Create Routes again when the rollout is in progress
-  After rollout is complete, some replicas will have a sidecar, and some will not.

_Environment_:

aigw version: latest main as of 12-02-2026
eg version: 1.6.3
gw api: 1.4.1

PR #1860, opened 2026-02-12
fix: condition when extProc sidecar not injected when deployment rollout inprogress 

gavrissh:
**Description**
There have been sequences when the routes have been deleted and created. The extProc sidecar would not be injected. We noticed this due to a condition when the rollout is in progress, and a few pods are in a terminating state but have an extProc sidecar in them. It does not inject the sidecar in some cases (e.g., with multiple replicas).

To fix this issue, we add a check to look for deployment/daemonset status if they are in progress. Requeue with a 5-second timeout. Additionally, check whether the sidecar is present in each pod. This is to ensure that, in multi-replica cases, there are no scenarios with inconsistent state (some with sidecar and some without). If there is an inconsistent state, we force rollout.

Changes:
- Skip terminating pods from checks and Consolidated checking of pods with sidecar into `checkPodHasSideCar()`
- Included `isRolloutInProgress()` for checking inprogress

Fixes: #1857 

Tested this change by running a sequence of creation -> deletion of routes. After the sequence was complete. The replicas were inconsistent state which triggered a force rollout

```
INFO	controller.gateway	pods are inconsistent while rollout is stable, forcing rollout	{"podsWithSidecarSeen": true, "podsWithoutSidecarSeen": true}
```

Requesting thoughts/feedback on this approach

### DIFF
--- a/internal/translator/gemini_helper.go
+++ b/internal/translator/gemini_helper.go
@@ -568,19 +568,41 @@ func reasoningEffortAvailable(requestModel internalapi.RequestModel) bool {
 	return strings.Contains(requestModel, "gemini") && strings.Contains(requestModel, "3")
 }
 
+// isGeminiFlashModel checks if the model is a Gemini Flash model.
+func isGeminiFlashModel(model internalapi.RequestModel) bool {
+	return strings.Contains(strings.ToLower(model), "flash")
+}
+
 // mapReasoningEffortToThinkingLevel converts OpenAI reasoning effort levels to Gemini thinking levels.
-// Currently supports:
+// The mapping depends on the model type:
+// - "none" → ThinkingLevelMinimal (Gemini Flash only)
 // - "low" → ThinkingLevelLow
-// - "medium" → ThinkingLevelHigh
-// https://ai.google.dev/gemini-api/docs/gemini-3?thinking=low#openai_compatibility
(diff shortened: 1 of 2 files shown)


internal/controller/gateway.go
```diff
@@ -743,10 +744,105 @@ func (c *GatewayController) getBSPForInferencePool(ctx context.Context, namespac
...
+	for i := range deployments {
+		dep := &deployments[i]
+		desired := ptr.Deref(dep.Spec.Replicas, int32(1))
+		if desired == 0 {
+			continue
+		}
+		if dep.Status.ObservedGeneration < dep.Generation {
+			return true
+		}
+		if dep.Status.UpdatedReplicas < desired || dep.Status.AvailableReplicas < desired {
```
nacx: How safe is this?
I mean, there could be several reasons why the available replicas are fewer than the desired ones. For example, one node could have gone offline, and anti-affinity rules prevent the deployment of a replica to satisfy the desired ones...
In such cases, is it OK to consider the rollout in progress? This may cause a constant requeuing without the pods being annotated...

gavrissh: I have updated the changes to make it stricter, checking only the possible status replicas specific to the rollout.
Not checking for Available Replicas, but still, it can requeue if there are resource constraints during rollout, which is when new template pods are stuck in a pending state. I think that is okay

PR #1865, opened 2026-02-13
fix: split the ext proc config to avoid k8s size limits

### LINKED ISSUE #1613: Avoid exceeding Kubernetes object size limits for Authorization config storage
Not for this PR, but for the future:
This config is stored on a k8s secret, and in scenarios where there are many MCPRoutes and many Authorization rules, the contents may exceed the max sicze of a k8s object.
We may need to come up with a different way to store this to avoid reaching these limits. We'll create a follow-up issue to track this.

_Originally posted by @nacx in https://github.com/envoyproxy/ai-gateway/pull/1482#discussion_r2584506786_

### LINKED ISSUE #2341: support compressed filter config to avoid 1MB Kubernetes secret limit
_Description_:

filter config is stored as secret and in extproc container, it contains information includes model route, model name override, auth, head mutation and cost cel information. The size of filter config will increase by factor of number of model routes and the size of it's plain text format can exceeds 1MB Kubernetes secret limit. 

Propose ai gateway controller to write filter config in a compressed format and reader will decompress to unpack it at runtime. 

This new feature should be extension via config flag rather than overriding existing gateway behavior. 

Simply gzip compress will reduce size from 1.3MB to 79KB bytes well under 1MB limit. 

Example of error message:
```
Secret \"[masked]\" is invalid: data: Too long: may not be more than 1048576 bytes"}
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).reconcileHandler
    /home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.24.1/pkg/internal/controller/controller.go:494 
```

zhaohuabing:
**Description**

This PR splits the ext proc/mcpproxy config into multiple secrets, so it can bypass the k8s size limits.

Fixes #1613

cc @nacx


cmd/aigw/testdata/translate_basic.out.yaml
```diff
@@ -264,4 +324,52 @@ spec:
     - name: http
       port: 8888
       protocol: HTTP
-
+---
+apiVersion: v1
```
nacx: Why are all these empty parts here? Is it due to the `maxFilterConfigBundleSlots` ?

nacx (commented):
Overall, it looks good.

So the summary of the PR is that:
* It will split the secret if it exceeds the max object k8s size, and split to a max of 8 chunks (for now).
* If there is no size overflow, empty chunks up to the 8 max ones will be created to avoid having to remount and cycle the pod.
* The extproc receives a new parameter, `-configBundlePath`, that if present will start the new watcher to watch for the new chunked config.

If this understanding is OK :) I have a few questions:

* Can we avoid creating the empty chunks? We could still add the volumes and secret projections to the pod, but mark those `optional`? Would that work to avoid having to create the empty secrets?
* The controller will pass the new arg `-configBundlePath` to the extproc if it finds the "index" secret. That secret is always created now, correct?
  * Could we provide a way to opt-in to this new feature, so that the controller does not create this secret (and extproc does not instantiate the new watchers, etc, etc)? This is quite a sensitive feature, and it would be great to have an option to explicitly opt-in.

Thanks for this PR!

zhaohuabing:
> Can we avoid creating the empty chunks? We could still add the volumes and secret projections to the pod, but mark those optional? Would that work to avoid having to create the empty secrets?

Fixed.

> The controller will pass the new arg -configBundlePath to the extproc if it finds the "index" secret. That secret is always created now, correct?
Could we provide a way to opt-in to this new feature, so that the controller does not create this secret (and extproc does not instantiate the new watchers, etc, etc)? This is quite a sensitive feature, and it would be great to have an option to explicitly opt-in.

Both the chunked secrets are always created, and the legacy secret is kept for upgrade where both the old and new version of the AIGW could coexist. We can remove the legacy one in the next release. 

In theory, we could opt-in the chunked secrets, but I have a few concerns to this approach:

1. The chunked Secret layout is intended to be an internal controller/extproc implementation detail. Adding an opt-in would expose that implementation detail as user-facing configuration.

2. It would also introduce another supported mode that we need to keep testing across the controller, mutator, and extproc startup paths. It may not double the test effort, but it would add meaningful test coverage(translate output, unit tests, e2e tests to cover both modes) and ongoing maintenance cost.

nacx (approved):

PR #1869, opened 2026-02-16
feat: inject backend quota rate limit filter for QuotaPolicy

yuzisun:
**Description**
This is the first step for quota aware routing by injecting the backend quota rate limit filter when there is QuotaPolicy attached to the AIServiceBackend specified in the upstream cluster.

End-to-End Data Flow

  User creates QuotaPolicy CR
      ↓
  QuotaPolicy Controller reconciles
      ↓ translator.BuildRateLimitConfigs()
  RateLimitConfig protobuf built (descriptor tree)
      ↓ runner.UpdateConfigs()
  xDS snapshot pushed to rate limit service via gRPC
      ↓
  Extension Server injects filter + actions into Envoy xDS
   
  Request arrives at Envoy
      ↓ (request-time rate limit check)
  If no available quota → 429; else continue
      ↓
  ext_proc processes request/response, extracts token usage
      ↓ (stream-done rate limit action)
  Rate limit filter sends HitsAddend to service
      ↓
  Rate limit service increments Redis counter by token count

Related PR: #1709


internal/extensionserver/quota_ratelimit.go
```diff
@@ -0,0 +1,653 @@
...
+			GrpcService: &corev3.GrpcService{
+				TargetSpecifier: &corev3.GrpcService_EnvoyGrpc_{
+					EnvoyGrpc: &corev3.GrpcService_EnvoyGrpc{
+						ClusterName: quotaRateLimitClusterName,
+					},
+				},
+			},
+			TransportApiVersion: corev3.ApiVersion_V3,
+		},
+		Timeout:                        &durationpb.Duration{Seconds: 5},
```
xiaolin593: any rationale behind the fixed timeout?

internal/extensionserver/quota_ratelimit.go
```diff
@@ -0,0 +1,653 @@
...
+		FailureModeDeny:                false,
+		DisableXEnvoyRatelimitedHeader: true,
+		EnableXRatelimitHeaders:        ratelimitfilterv3.RateLimit_DRAFT_VERSION_03,
+		RateLimitedAsResourceExhausted: false,
```
xiaolin593: The boolean values are hardcoded. Any chance they can be configurable when needed?

yuzisun:
@yanavlasov @johnugeorge @nacx

cmd/controller/main.go
```diff
@@ -412,6 +413,14 @@ func main() {
 		}
 	}()
 
+	// Start the rate limit xDS config server.
+	rlRunner := runner.New(ctrl.Log, runner.DefaultPort)
+	go func() {
+		if err := rlRunner.Start(ctx); err != nil {
+			setupLog.Error(err, "failed to start rate limit xDS server")
+		}
+	}()
```
nacx, line 443: How will this work when used in combination with Envoy Gateway?
IIRC the XDS config in the ratelimit service is 1 URL. How will the ratelimit service be configured, and how would it affect the EnvotyGateway installation and normal (non-ai related) EG rate limit configurations?
Or are we deploying another, different rate limit service to handle the quota?

internal/extensionserver/quota_ratelimit.go
```diff
@@ -0,0 +1,729 @@
...
+
+const (
+	// quotaRateLimitClusterName is the Envoy cluster name for the AI Gateway rate limit service.
+	quotaRateLimitClusterName = "ai_gateway_ratelimit_cluster"
+	// quotaRateLimitFilterName is the name of the rate limit HTTP filter inserted into the
+	// HCM filter chain for QuotaPolicy enforcement. The filter is disabled by default and
+	// enabled per-route via TypedPerFilterConfig on routes with quota backends.
+	quotaRateLimitFilterName = "envoy.filters.http.ratelimit"
+	// defaultQuotaRateLimitServiceHost is the default hostname for the AI Gateway rate limit service.
+	defaultQuotaRateLimitServiceHost = "envoy-ai-gateway-ratelimit.envoy-gateway-system"
```
nacx: If I'm not wrong, the name of the rate limit service is configurable via Helm values?

yuzisun: will update.

cmd/controller/main.go
yuzisun: We setup separate rate limit service and XDS server for Envoy AI gateway with QuotaPolicy. No impact to EG installation.

johnugeorge:
Correct me if I am wrong
1. Currently, it doesn't handle multiple quotaPolicies on the same backend. We will have to merge descriptors under the same domain. Can you create an issue to track it?
2. We should keep new rate limit service optional under a quota helm flag.

Question: How much latency is added if the request passes through the second ratelimit filter(if enabled)?

internal/controller/quota_policy.go
```diff
@@ -0,0 +1,170 @@
...
+// set to the xDS runner.
+func (c *QuotaPolicyController) syncQuotaPolicy(ctx context.Context, _ *aigv1a1.QuotaPolicy) error {
+	return c.rebuildAllConfigs(ctx)
+}
+
+// rebuildAllConfigs lists all QuotaPolicies, resolves their target backends,
+// builds rate limit configs, and pushes the full snapshot to the xDS runner.
+func (c *QuotaPolicyController) rebuildAllConfigs(ctx context.Context) error {
+	var allPolicies aigv1a1.QuotaPolicyList
+	if err := c.client.List(ctx, &allPolicies); err != nil {
```
johnugeorge: Isn't rebuildAllConfigs expensive if done this way? 

Currently, 
func Reconcile(policy) {
    allPolicies = listAllPolicies()           
    for each policy in allPolicies {          
        backends = getBackends(policy)        
        buildConfig(policy, backends)        
    }
    pushAllConfigs()                   
}
It will need P*B calls


Instead. Can we do this? 


func Reconcile(changedPolicy) {
    backends = getBackends(changedPolicy)    
    config = buildConfig(changedPolicy, backends)  
    pushSingleConfig(config)                
}
It will be B calls

yuzisun:
> Correct me if I am wrong
> 
> 1. Currently, it doesn't handle multiple quotaPolicies on the same backend. We will have to merge descriptors under the same domain. Can you create an issue to track it?

created https://github.com/envoyproxy/ai-gateway/issues/1971
> 2. We should keep new rate limit service optional under a quota helm flag.

will address

> Question: How much latency is added if the request passes through the second ratelimit filter(if enabled)?

The second ratelimit filter adds 5ms 99% and 3ms 50% to the total latency.

internal/controller/quota_policy.go
```diff
@@ -0,0 +1,189 @@
...
+	}
+
+	// Update cache and push merged configs to xDS.
+	cacheKey := fmt.Sprintf("%s/%s", policy.Namespace, policy.Name)
+	c.mu.Lock()
+	c.configCache[cacheKey] = configs
+	allConfigs := c.getMergedConfigsLocked()
+	c.mu.Unlock()
+
+	return c.rateLimitRunner.UpdateConfigs(ctx, allConfigs)
```
johnugeorge, line 144: won't out of order execution of UpdateConfigs create issue in correctness? (race condition) eg: UpdateConfigs of second event called before the UpdateConfigs call of the first event ?

internal/extensionserver/quota_ratelimit.go
```diff
@@ -0,0 +1,729 @@
...
+	httpconnectionmanagerv3 "github.com/envoyproxy/go-control-plane/envoy/extensions/filters/network/http_connection_manager/v3"
+	matcherv3 "github.com/envoyproxy/go-control-plane/envoy/type/matcher/v3"
+	metadatav3 "github.com/envoyproxy/go-control-plane/envoy/type/metadata/v3"
+	"github.com/envoyproxy/go-control-plane/pkg/wellknown"
+	"google.golang.org/protobuf/types/known/anypb"
+	"google.golang.org/protobuf/types/known/durationpb"
+	"google.golang.org/protobuf/types/known/wrapperspb"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	aigv1a1 "github.com/envoyproxy/ai-gateway/api/v1alpha1"
```
nacx, line 35: Should we use the beta APIs here by default?

internal/controller/controller.go
```diff
@@ -238,6 +241,16 @@ func StartControllers(ctx context.Context, mgr manager.Manager, config *rest.Con
 		return fmt.Errorf("failed to create controller for GatewayConfig: %w", err)
 	}
 
+	// QuotaPolicy controller for backend quota rate limiting.
+	if options.RateLimitRunner != nil {
+		quotaPolicyC := NewQuotaPolicyController(c, kube, logger.WithName("quota-policy"), options.RateLimitRunner)
+		if err = TypedControllerBuilderForCRD(mgr, &aigv1a1.QuotaPolicy{}).
+			Watches(&aigv1a1.AIServiceBackend{}, handler.EnqueueRequestsFromMapFunc(quotaPolicyC.BackendToQuotaPolicy)).
```
nacx: should we watch for beta1 api here now?

internal/ratelimit/translator/translator.go
```diff
@@ -0,0 +1,348 @@
...
+
+// parseDurationSimple returns the best-fit unit and adjusts the limit accordingly.
+// For example, a duration of "5m" with limit 100 means 100 requests per 5 minutes.
+// The rate limit service only supports per-unit rates (per second, minute, hour, day),
+// so we express this as: requests_per_unit = limit, unit = closest match.
+//
+// Note: The rate limit service's sliding window is per-unit, so "100 per 5m" is
+// approximated as "100 per MINUTE" with the understanding that the window resets
+// each minute. For exact multi-unit windows, the limit should be divided.
+func ParseDurationAndAdjustLimit(limit uint, duration string) (uint32, rlsconfv3.RateLimitUnit, error) {
```
nacx: This funciton seems that is not being called, and it is meant to adjust the limits to the right values for the rate limit service. Is this intentional, a leftover?

internal/extensionserver/quota_ratelimit.go
```diff
@@ -0,0 +1,729 @@
...
+	routeName := parts[2]
+	ruleIndexStr := parts[4]
+	ruleIndex, err := strconv.Atoi(ruleIndexStr)
+	if err != nil {
+		return false
+	}
+
+	// Fetch the AIGatewayRoute to get the actual backend refs.
+	var aigwRoute aigv1a1.AIGatewayRoute
+	if err := s.k8sClient.Get(context.Background(), client.ObjectKey{
```
nacx: Here and in other cases, should we propagate the incoming context from the extension server call and use it instead of using the unbounded `context.Background()`?

internal/ratelimit/translator/translator_test.go
```diff
@@ -0,0 +1,872 @@
...
+		require.Equal(t, rlsconfv3.RateLimitUnit_MINUTE, policy.Unit)
+	})
+
+	t.Run("5 minutes multiplies limit", func(t *testing.T) {
+		qv := &aigv1a1.QuotaValue{Limit: 20, Duration: "5m"}
+		policy, err := quotaValueToPolicy(qv)
+		require.NoError(t, err)
+		// 20 * 5 = 100 requests per MINUTE
+		require.Equal(t, uint32(100), policy.RequestsPerUnit)
+		require.Equal(t, rlsconfv3.RateLimitUnit_MINUTE, policy.Unit)
```
nacx: If I understand correctly how the rate limit service works and the comment in one of the functions, the rate limit service sliding window is "per unit", which, from my understanding, means that when the unit is "minutes", the sliding window will reset "every minute".

If that is true, then I think the `quotaValueToPolicy` does not correctly approximate the desired outcome, as producing `20 * 5 = 100 requests per MINUTE` as a result will mean, effectively 100 requests perminute, which would be `500 requests each 5 minutes`. Is that what `&aigv1a1.QuotaValue{Limit: 20, Duration: "5m"}` should translate to?

internal/ratelimit/translator/translator.go
```diff
@@ -0,0 +1,348 @@
...
+	if quota.DefaultBucket.Limit > 0 {
+		defaultPolicy, err := quotaValueToPolicy(&quota.DefaultBucket)
+		if err != nil {
+			return nil, err
+		}
+		defaultKey := DefaultBucketDescriptorKey(modelName, len(quota.BucketRules))
+		nested = append(nested, &rlsconfv3.RateLimitDescriptor{
+			Key:       defaultKey,
+			Value:     defaultKey,
+			RateLimit: defaultPolicy,
```
nacx, line 271: Should this one also have the `QuotaMode: true`?

internal/ratelimit/translator/translator.go
```diff
@@ -0,0 +1,348 @@
...
+// all models (when no PerModelQuota matches). Uses only the key without a
+// specific value so that any model name will match.
+func buildServiceQuotaDescriptor(sq *aigv1a1.ServiceQuotaDefinition) (*rlsconfv3.RateLimitDescriptor, error) {
+	policy, err := quotaValueToPolicy(&sq.Quota)
+	if err != nil {
+		return nil, err
+	}
+	return &rlsconfv3.RateLimitDescriptor{
+		Key:       ModelNameDescriptorKey,
+		RateLimit: policy,
```
nacx, line 303: Just to confirm, it is expected that this doesn't set the `QuotaMode: true`, right? In that case, can we be explicit in the code and set it to false, even if it is redundant (here and in my previous comment) to make it explicit that it's not a miss? It will make the code more easy to maintain.

internal/extensionserver/quota_ratelimit.go
```diff
@@ -0,0 +1,729 @@
...
+	for _, currChain := range filterChains {
+		httpConManager, hcmIndex, err := findHCM(currChain)
+		if err != nil {
+			continue
+		}
+
+		// Check if the filter already exists.
+		alreadyExists := false
+		for _, f := range httpConManager.HttpFilters {
+			if f.Name == quotaRateLimitFilterName {
```
nacx, line 201: If I'm not wrong, this filter name is the same as the upstream filter names. Could this be a problem if other existing rate limits exist? Are we trying to catch that, or _only_ look for the quota-specific one? It it's the latter we probably want to use a different filter name for the quota rate limits?

cmd/controller/main.go
```diff
@@ -420,6 +433,14 @@ func main() {
 		}
 	}()
 
+	// Start the rate limit xDS config server.
+	rlRunner := runner.New(ctrl.Log, runner.DefaultPort)
+	go func() {
+		if err := rlRunner.Start(ctx); err != nil {
+			setupLog.Error(err, "failed to start rate limit xDS server")
```
nacx, line 440: If this fails to start, should we return an error to trigger shutdown, or is it OK to leave it this way? The rate limit server will be configured to connect to this xDS server, but that will fail, and the rate limit config cache will fail to push....

How will the system behave if this fails to start?

nacx (commented):
Did another review. Here are some additional notes to the comments:

* The `CostExpression` seems not yet to be implemented. Should we add a comment in the API saying the field will be implemented in the future, so users know the field is not taken into account yet?
* The `mode` seems not to be used, but it is `required` in the API. Is this a leftover or something that needs to be taken into account?

internal/ratelimit/translator/translator.go
yuzisun: I think we only need `QuotaMode` for the rate limit configurations to return the backends in dynamic metadata that exceeds quota. On the response path we only reduce the quota.

internal/extensionserver/quota_ratelimit.go
yuzisun: QuotaPolicy is still v1alpha1

internal/extensionserver/quota_ratelimit.go
yuzisun: Good catch, lost the name suffix, added back now.

internal/extensionserver/quota_ratelimit.go
nacx: I mean, for the uses of AIGatewayRoute, etc, all of them should be using v1beta1. QuotaPolicy v1alpha1, but we ned to check that all references to other objects are using v1beta1.

johnugeorge: I agree. We should reference v1beta1 folder fir all resources for the upgraded resources. Else, we will miss future updates to the v1beta1 folder

internal/ratelimit/translator/translator_test.go
yuzisun: Fixed, you are right envoy rate limit only support unit window not per 5 minute.

internal/controller/quota_policy.go
yuzisun: right, we need to hold the lock across UpdateConfigs so older snapshot does not overwrite newer snapshot, and ensure the cache read + xDS push is atomic from the controller's perspective.

tests/e2e/testdata/backend_quota_ratelimit.yaml
```diff
@@ -0,0 +1,213 @@
...
+  provider:
+    type: Kubernetes
+    kubernetes:
+      envoyDeployment:
+        container:
+          # Note: this is to clear the large default memory/cpu requirements for local tests.
+          # In production, you should set these to values that make sense for your environment.
+          resources: {}
+---
+apiVersion: aigateway.envoyproxy.io/v1alpha1
```
nacx: Can we start using `v1beta1` for all resources? I see the code has been upated accordingly to the previous comments, but I'd also update e2e tests, and start using `v1beta1` by wherever we add new code/tests.

internal/extensionserver/quota_ratelimit.go
nacx: Any comments on this?

internal/ratelimit/translator/translator.go
```diff
@@ -0,0 +1,328 @@
...
+// Since Envoy rate limit only supports standard units (SECOND, MINUTE, HOUR, DAY),
+// non-standard durations like "5m" are approximated by dividing the limit by the
+// number of base units. For example, limit=20 with duration="5m" becomes 4 per MINUTE.
+func quotaValueToPolicy(qv *aigv1a1.QuotaValue) (*rlsconfv3.RateLimitPolicy, error) {
+	divisor, unit, err := parseDuration(qv.Duration)
+	if err != nil {
+		return nil, fmt.Errorf("invalid duration %q: %w", qv.Duration, err)
+	}
+	return &rlsconfv3.RateLimitPolicy{
+		RequestsPerUnit: uint32(qv.Limit) / divisor, //nolint:gosec
```
nacx: I think this is still not correct. Say a user configures a quota such as the following to configure 20 requests every 90 seconds.

```go
Limit:    20,
Duration: "90s",
```

This would produce the result, which is wrong:

```go
rlsconfv3.RateLimitPolicy{
		RequestsPerUnit: 0,
		Unit:            rlsconfv3.RateLimitUnit_SECOND,
}
```
When the `divisor > limit`, the result of this division will be `0`, leading to an incorrect value for the limit.

nacx (commented):
Re-chedked the updated code and added some comments.

I see the general comment I added about the `mode` and `CostExpression` are not yet addressed. Is that still a AIP?

Let me know when I can do another review pass!

internal/ratelimit/runner/runner.go
```diff
@@ -0,0 +1,129 @@
...
+	if port == 0 {
+		port = DefaultPort
+	}
+	return &Runner{
+		logger: logger.WithName("ratelimit-xds-runner"),
+		port:   port,
+	}
+}
+
+// Start starts the xDS gRPC server. It blocks until ctx is cancelled.
```
aabchoo, line 58: What does this block?

internal/ratelimit/runner/runner.go
```diff
@@ -0,0 +1,129 @@
...
+func (r *Runner) Start(ctx context.Context) error {
+	r.mu.Lock()
+	r.cache = cachev3.NewSnapshotCache(false, cachev3.IDHash{}, nil)
```
aabchoo, line 61: where is the config specifying rate limit?

internal/ratelimit/runner/runner.go
```diff
@@ -0,0 +1,129 @@
...
+	"google.golang.org/grpc"
+)
+
+const (
+	// NodeID is the xDS node identifier that the rate limit service uses
+	// when connecting to this xDS server.
+	NodeID = "envoy-ai-gateway-ratelimit"
+
+	// DefaultPort is the default listening port for the rate limit xDS config server.
+	DefaultPort = 18002
```
aabchoo, line 32: Might need to change to avoid conflict with the wasm port

internal/ratelimit/translator/translator.go
```diff
@@ -0,0 +1,328 @@
...
+// specific value so that any model name will match.
+func buildServiceQuotaDescriptor(sq *aigv1a1.ServiceQuotaDefinition) (*rlsconfv3.RateLimitDescriptor, error) {
+	policy, err := quotaValueToPolicy(&sq.Quota)
+	if err != nil {
+		return nil, err
+	}
+	return &rlsconfv3.RateLimitDescriptor{
+		Key:       ModelNameDescriptorKey,
+		RateLimit: policy,
+		QuotaMode: true,
```
aabchoo, line 304: Does this value depend on Mode https://github.com/envoyproxy/ai-gateway/blob/main/api/v1alpha1/quota_policy.go#L91-L96

api/v1alpha1/registry.go
```diff
@@ -18,6 +18,7 @@ func init() {
 	SchemeBuilder.Register(&BackendSecurityPolicy{}, &BackendSecurityPolicyList{})
 	SchemeBuilder.Register(&MCPRoute{}, &MCPRouteList{})
 	SchemeBuilder.Register(&GatewayConfig{}, &GatewayConfigList{})
+	SchemeBuilder.Register(&QuotaPolicy{}, &QuotaPolicyList{})
```
aabchoo, line 21: quota policy needs to be updated properly to be optional https://github.com/envoyproxy/ai-gateway/blob/main/api/v1alpha1/quota_policy.go#L106-L107C33

internal/extensionserver/quota_ratelimit.go
```diff
@@ -0,0 +1,731 @@
...
+	rateLimitActions = append(rateLimitActions, &routev3.RateLimit{
+		Actions:           baseDescriptorActions(),
+		HitsAddend:        quotaHitsAddend(),
+		ApplyOnStreamDone: true,
+	})
+
+	// Append RateLimit entries for bucket rules from each policy's per-model quotas.
+	for i := range policies {
+		for _, pmq := range policies[i].Spec.PerModelQuotas {
+			if pmq.ModelName == nil || len(pmq.Quota.BucketRules) == 0 {
```
aabchoo: should be &&?

aabchoo:
List of fixes needed:

```
  1. Descriptor ordering fix (quota_ratelimit.go)                                                                                                                                                                                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                                                                                                                                                                                         
  Route-side actions now produce descriptors in the order backend_name → model_name_override → bucket_rule_key, matching the translator's service config tree. Request-time entries use GenericKey with static values; stream-done entries use Metadata from ext_proc.                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                                                                                                                                                                         
  2. Model name resolved from AIGatewayRoute (quota_ratelimit.go)                                                                                                                                                                                                                                                                                                                                                                                        
                                                            
  New resolveRouteModelInfo() method fetches the AIGatewayRoute and returns both route names (for filtering) and backend→ModelNameOverride mapping (for descriptors). Request-time GenericKeys use the actual ModelNameOverride from the route instead of the QuotaPolicy's modelName.                                                                                                                                                                   
                                                            
  3. Model-name filtering by route name (quota_ratelimit.go)                                                                                                                                                                                                                                                                                                                                                                                             
                                                            
  enableQuotaRateLimitOnRoute filters PerModelQuotas by AIGatewayRoute name (not ModelNameOverride). This matches the QuotaPolicy's modelName (e.g., claude-sonnet-4-6) to the route name, preventing sonnet rules from leaking to haiku routes.                                                                                                                                                                                                         
                                                            
  4. Config merging — most strict wins (quota_policy.go, translator/merge.go)                                                                                                                                                                                                                                                                                                                                                                            
                                                            
  getMergedConfigsLocked() produces a single RateLimitConfig (fixing xDS Name deduplication). MergeDescriptors() recursively merges descriptor trees, keeping the stricter rate limit when paths collide.                                                                                                                                                                                                                                                
                                                            
  5. Translator uses resolved model names (translator.go, quota_policy.go)                                                                                                                                                                                                                                                                                                                                                                               
                                                            
  BuildRateLimitConfigs accepts a backendModelOverrides map. The controller resolves AIGatewayRoutes matching the policy's model names and passes the ModelNameOverride values to the translator. The service config tree now uses provider-specific names (e.g., us.anthropic.claude-sonnet-4-6) instead of the policy's modelName.                                                                                                                     
                                                            
  6. Descriptor depth alignment (quota_ratelimit.go)                                                                                                                                                                                                                                                                                                                                                                                                     
                                                            
  Simple case (no bucket rules): 2-level descriptors. Bucket rules case: 3-level descriptors. Both match the translator's tree structure.                                                                                                                                                                                                                                                                                                                
                                                            
  7. Stream-done deduplication (quota_ratelimit.go)                                                                                                                                                                                                                                                                                                                                                                                                      
                                                            
  One stream-done entry per unique model name, appended at the end — not per policy/target. Prevents multi-counting when multiple policies target different backends for the same model.                                                                                                                                                                                                                                                                 
   
  8. Per-backend cost expressions (gateway.go, filterconfig.go, processor_impl.go)                                                                                                                                                                                                                                                                                                                                                                       
                                                            
  - LLMRequestCost has a new internal-only Backend field                                                                                                                                                                                                                                                                                                                                                                                                 
  - Gateway controller injects QuotaPolicy CostExpression as LLMRequestCost entries with the Backend filter, one per target backend
  - ext_proc skips entries where Backend doesn't match the serving backend                                                                                                                                                                                                                                                                                                                                                                               
  - Defaults to total_tokens when CostExpression is nil                                                                                                                                                                                                                                                                                                                                                                                                  
  - HitsAddend reads from per-model metadata key (quota_cost_<modelName>)                                                                                                                                                                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                                                                                                                                                                                         
  9. Helm chart (values.yaml, deployment.yaml)                                                                                                                                                                                                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                                                                                                                                                                                                         
  Added controller.quotaRateLimitServiceHost config (defaults to envoy-ai-gateway-ratelimit.envoy-gateway-system) and wired it into the deployment template.                                                                                                                                                                                                                                                                                             
                                                            
  10. New files                                                                                                                                                                                                                                                                                                                                                                                                                                          
                                                            
  - internal/ratelimit/translator/merge.go — MergeDescriptors() + isStricter()                                                                                                                                                                                                                                                                                                                                                                           
  - internal/ratelimit/translator/merge_test.go — merge logic tests
```

Will address comments and cleaning up some confusion

aabchoo:
Currently mode and client selectors are not fully implemented. Will add that soon

internal/ratelimit/translator/translator.go
aabchoo: Currently wondering if we can instead update the rate limit service to support "custom" duration

aabchoo: Personally leaning towards having strict durations ie has to be 1s, 1m, 1h, etc for now. We are doing the math to convert it behind the scenes which to me is not what "90s" implies since it will be x per second or x per minute.

internal/extensionserver/quota_ratelimit.go
aabchoo: agreed, will add this change

internal/ratelimit/translator/translator.go
aabchoo: I've opt'd to default to 1s, 1m, 1h as that is ultimately what the math does.

aabchoo:
Quota Policy will be built on a per route rate limit

On the Envoy descriptor side:

1. Quota policy per model default bucket will create a set of descriptor -> request path is specific but response path is generic based on metadata (may be a possible issue)
2. Any bucketRules will generate a set of descriptors

We deduplicate descriptors for same models to avoid extra rate limit increments.

On the config side:

Rate limit config is a nested format in quota mode


-----

Currently the rate limit config needs a quick fix + I need to confirm Yan's Envoy change but our implementation should be good here.

----

Mode will by default only be Shared mode. Duration will be strict 1s, 1m, and 1h.

tests/e2e/testdata/backend_quota_ratelimit.yaml
aabchoo: Updated to use v1beta1

aabchoo:
Will figure out why Test_Examples_BackendQuotaRateLimit is failing but for now @johnugeorge @nacx could you please review this again?

aabchoo:
Quota Policy's model name is now based on the model name override. Updated documentation to outline that.

RateLimitPerRoute is based on the httproute rule and cross-references the backend name + model name override. Multiple backends for a rule means the route will send multiple descriptors for the potential "match". Response path will only update the override+backend routed.

aabchoo:
I will follow up with a PR to add rate limit server deployment + service in the ai-gateway-helm chart.

internal/ratelimit/translator/translator.go
```diff
@@ -0,0 +1,433 @@
...
+		return *header.Value
+	}
+	return ""
+}
+
+// BucketRuleDescriptorKey returns the descriptor key for a bucket rule's header match.
+// Both the backend name and model name are included so that each backend's bucket rule
+// has a unique key, preventing cross-backend matching in the rate limit service config.
+func BucketRuleDescriptorKey(backendName, modelName string, ruleIndex, matchIndex int) string {
+	return fmt.Sprintf("rule-%s-%s-%d-match-%d", backendName, modelName, ruleIndex, matchIndex)
```
yuzisun: Do we need to include backendName and modelName here as they are already in the key/value level

internal/ratelimit/translator/translator.go
```diff
@@ -0,0 +1,451 @@
...
+// parseDuration accepts exactly "1s", "1m", or "1h".
+func parseDuration(s string) (rlsconfv3.RateLimitUnit, error) {
+	switch s {
+	case "1s":
+		return rlsconfv3.RateLimitUnit_SECOND, nil
+	case "1m":
+		return rlsconfv3.RateLimitUnit_MINUTE, nil
+	case "1h":
+		return rlsconfv3.RateLimitUnit_HOUR, nil
+	default:
```
nacx, line 414: The API field allows `1d` as a value. It should be either added here or removed from the API.

internal/extensionserver/quota_ratelimit.go
```diff
@@ -0,0 +1,1016 @@
...
+							streamDoneActions = append(streamDoneActions, &routev3.RateLimit{
+								Actions:           append(baseDescriptorActions(), clientActions...),
+								HitsAddend:        quotaHitsAddend(modelName),
+								ApplyOnStreamDone: true,
+							})
+						}
+					}
+				}
+				// Default bucket: 3-level stream-done with GenericKey (always fires).
+				// In Exclusive mode this also decrements the default bucket for requests
```
nacx: Exclusive mode is not yet implemented. Should we still mention it here?

api/v1alpha1/quota_policy.go
```diff
@@ -103,18 +107,21 @@ type QuotaDefinition struct {
...
 //
-// +kubebuilder:validation:Enum=Exclusive;Shared
+// TODO: Add Exclusive mode in the future.
+// +kubebuilder:validation:Enum=Shared
 type QuotaBucketMode string
 
 const (
-	QuoteBucketModeShared    QuotaBucketMode = "Shared"
-	QuoteBucketModeExclusive QuotaBucketMode = "Exclusive"
+	QuoteBucketModeShared QuotaBucketMode = "Shared"
```
nacx: Typo:
```suggestion
	QuotaBucketModeShared QuotaBucketMode = "Shared"
```

internal/extensionserver/quota_ratelimit.go
```diff
@@ -0,0 +1,1016 @@
...
+				bucketActions := buildBucketRuleLimitEntries(modelName, policy.Namespace, &pmq.Quota, policy.Spec.TargetRefs, backendModels)
+				rateLimitActions = append(rateLimitActions, bucketActions...)
+				// Bucket rules: one stream-done per (target, rule). The parent
+				// backend_name descriptor ensures the rate limit service only matches
+				// the entry under the actual backend's tree.
+				for rIdx, rule := range pmq.Quota.BucketRules {
+					firstHdr, firstVal := firstSelectorHeader(rule.ClientSelectors)
+					for _, target := range policy.Spec.TargetRefs {
+						targetName := string(target.Name)
+						dupKey := targetName + "|" + modelName + "|" + translator.BucketRuleDescriptorKey(rIdx, 0, firstHdr, firstVal)
```
nacx: Two bucket rules with the same `rIdx` and the same first sorted header but different remaining headers would collide, and one of the stream-done entries would be silently dropped.
Is this a case that can actually happen?

manifests/charts/ai-gateway-helm/templates/deployment.yaml
```diff
@@ -100,6 +100,7 @@ spec:
             {{- end }}
             - --cacheSyncTimeout={{ .Values.controller.watch.cacheSyncTimeout }}
             - --watchNamespaces={{ join "," .Values.controller.watch.namespaces }}
+            - --quotaRateLimitServiceHost={{ .Values.controller.quotaRateLimitServiceHost }}
```
nacx: Should the other two new quota flags (timeout and failure mode) be added here as well?

internal/ratelimit/runner/runner_test.go
```diff
@@ -0,0 +1,177 @@
...
+		snap, err = r.cache.GetSnapshot(NodeID)
+		require.NoError(t, err)
+		resources := snap.GetResources(resourcev3.RateLimitConfigType)
+		require.Len(t, resources, 1)
+	})
+}
+
+func TestStart(t *testing.T) {
+	t.Run("starts and stops with context cancellation", func(t *testing.T) {
+		r := New(logr.Discard(), 0) // port 0 lets the OS pick a free port.
```
nacx: OS never picks the port as it is fixed to the default one if 0 is passed as a value. Is this the desired behavior? If it is, then the comment should be updated, otherwise the constructor should be changed.

internal/extensionserver/quota_ratelimit.go
```diff
@@ -0,0 +1,1016 @@
...
+	if len(parts) != 5 || parts[0] != "httproute" {
+		return false
+	}
+	namespace := parts[1]
+	routeName := parts[2]
+	ruleIndexStr := parts[4]
+	ruleIndex, err := strconv.Atoi(ruleIndexStr)
+	if err != nil || ruleIndex < 0 {
+		return false
+	}
```
nacx, line 364: This identical logic is repeated in 3 places. Let's create a small helper method so that it is easier to maintain and less error-prone.

internal/extensionserver/quota_ratelimit.go
```diff
@@ -0,0 +1,1016 @@
...
+const (
+	// quotaRateLimitClusterName is the Envoy cluster name for the AI Gateway rate limit service.
+	quotaRateLimitClusterName = "ai_gateway_ratelimit_cluster"
+	// quotaRateLimitFilterName is the name of the rate limit HTTP filter inserted into the
+	// HCM filter chain for QuotaPolicy enforcement. The name uses a suffix to avoid
+	// conflicting with Envoy Gateway's own rate limit filter (e.g., from BackendTrafficPolicy).
+	// The per-route TypedPerFilterConfig keys on this name to target the quota filter specifically.
+	quotaRateLimitFilterName = "envoy.filters.http.ratelimit/ai-gateway-quota"
+	// defaultQuotaRateLimitServiceHost is the default hostname for the AI Gateway rate limit service.
+	defaultQuotaRateLimitServiceHost = "envoy-ai-gateway-ratelimit.envoy-gateway-system"
```
nacx: This is not used anymore, apparently. Let's remove it?

internal/extensionserver/quota_ratelimit.go
```diff
@@ -0,0 +1,1016 @@
...
+	// defaultQuotaRateLimitServicePort is the default gRPC port for the rate limit service.
+	defaultQuotaRateLimitServicePort = 8081
```
nacx, line 49: We allow configuring the host now, but not the port. Should we also allow configuring the port instead of hardcoding it? We could use the same flag and split the value.

internal/controller/quota_policy.go
```diff
@@ -0,0 +1,271 @@
...
+	defer c.mu.Unlock()
+	delete(c.configCache, cacheKey)
+	allConfigs := c.getMergedConfigsLocked()
+
+	return c.rateLimitRunner.UpdateConfigs(ctx, allConfigs)
+}
+
+// getMergedConfigsLocked merges all cached configs into a single RateLimitConfig.
+// When multiple QuotaPolicies define the same descriptor path, the policy whose
+// namespace/name is alphabetically first takes precedence. Keys are sorted to
```
nacx, line 161: Should this comment also be explained in the API comments, so it is visible in the user-facing docs?

nacx (commented):
There is one additional comment that Claude has caught, but I don't fully understand it, so I'm sharing it here verbatim for your consideration:

> There are three different views of how PerModelQuota.modelName relates to AIGatewayRouteRuleBackendRef.ModelNameOverride:
> 
> * API doc (quota_policy.go:70-72) and gateway.go:832 -> modelName must equal the override
> * Extension server (quota_ratelimit.go:569) -> Skips the policy unless override == modelName
> * Controller resolveBackendModelOverrides (quota_policy.go:193-231) -> Picks up overrides from AIGatewayRoutes whose route name matches the policy modelName; the translator then substitutes the override into the descriptor value
> 
> So, for the only non-trivial case the controller supports (route name = policy modelName, but ModelNameOverride differs), the translator produces descriptors keyed on the override, while the extension server filters the policy out entirely because override ≠ modelName → no rate-limit actions ever get attached to the route, even though the rate-limit-service config has the descriptors. The e2e test (backend_quota_ratelimit.yaml) sidesteps this by making route name = override = modelName, so the bug is invisible there.
> * If the doc is authoritative (modelName == ModelNameOverride), then resolveBackendModelOverrides and the descriptorModelNames := resolveModelNames(...) branch in translator.go:161-169 are dead code that should be deleted, and TestQuotaPolicyController_resolveBackendModelOverrides should be removed.
> * If the controller's "route-name convention" is real, the extension server filter needs to match on route name (not override), and the API doc must change.

internal/extensionserver/quota_ratelimit.go
```diff
@@ -0,0 +1,1016 @@
...
+	}
+	return keys
+}
+
+// routeModelInfo holds the resolved model information for an Envoy route.
+type routeModelInfo struct {
+	// backendModels maps backend name → ModelNameOverride from the AIGatewayRoute.
+	// Used for filtering (a QuotaPolicy's target+modelName must match) and for
+	// request-time descriptors (the actual value ext_proc will set).
+	backendModels map[string]string
```
johnugeorge: This will not support multiple modelOverrides of same backends? eg: Two backendRefs - GPT-5 and GPT-4 of same OpenAI AIServiceBackend

internal/ratelimit/translator/translator.go
johnugeorge: Day is needed in many cases

aabchoo: Agreed

internal/extensionserver/quota_ratelimit.go
aabchoo: Good catch; thank you

api/v1alpha1/quota_policy.go
aabchoo: Amazing catch

internal/extensionserver/quota_ratelimit.go
aabchoo: Removed

internal/controller/quota_policy.go
aabchoo: Surfaced

internal/ratelimit/runner/runner_test.go
aabchoo: I've removed the comment

internal/extensionserver/quota_ratelimit.go
aabchoo: Accounting for that

internal/extensionserver/quota_ratelimit.go
aabchoo: I've updated to use the entire selectors in case

internal/extensionserver/quota_ratelimit.go
aabchoo: added

manifests/charts/ai-gateway-helm/templates/deployment.yaml
aabchoo: another great catch thank you

aabchoo:
Added logic to trigger `postTranslateModify` as the xDS was not being updated previously. QuotaPolicy will trigger AIGateway to reconcile for associated aiservicebackends. Rate limit config were updating properly.

nacx (approved):

PR #1280, opened 2025-10-06
End-to-end OpenAI Image Generation support: processor, translator, tracing, metrics

nutanix-Hrushikesh:
### Title
End-to-end OpenAI Image Generation support: processor, translator, tracing, metrics, Brotli, docs, and repo hygiene

### Overview
This PR adds complete support for OpenAI’s image generation endpoint (/v1/images/generations) across the Envoy AI Gateway. It introduces a processor, translation layer, tracing and metrics instrumentation, Brotli decoding, example client/service updates, and repo hygiene improvements. 

### Changes
- **ExtProc (image generation)**
  - Added `imageGenerationProcessorFactory` and registered it in ExtProc main.
  - Implemented `imageGenerationProcessorRouterFilter` and `imageGenerationProcessorUpstreamFilter`.
  - Added request/response header and body processing, including retry handling and auth passthrough.
  - Introduced `imageGenerationMetrics` and integrated with the processor to record image-specific telemetry.
  - Improved diagnostics with debug logs for processor selection/instantiation when matching routes.

- **Translator (OpenAI to OpenAI)**
  - Added `ImageGenerationTranslator` interface and `ImageGenerationMetadata` to standardize request/response translation for images.
  - Implemented OpenAI to OpenAI translator:
    - Request body transformation (model overrides, forced mutation).
    - Response headers/body parsing with OpenAI SDK v2 schema to avoid drift.

- **Tracing**
  - Extended Tracing API to support image generation with `ImageGenerationTracer` and `ImageGenerationRecorder`.
  - Implemented `imageGenerationTracer` and `imageGenerationSpan`:
    - Span start, header injection, response recording, and well-defined error termination paths.
  - Added OpenInference-based `ImageGenerationRecorder` for router filter instrumentation and a Noop tracer.

- **Metrics**
  - Implemented `ImageGenerationMetrics` with methods to record lifecycle events, model/backend selection, token usage, and image generation stats.
  - Extended GenAI metrics with image-specific attributes:
    - `genaiAttributeImageCount`, `genaiAttributeImageModel`, `genaiAttributeImageSize`
  - Added operation type: `genaiOperationImageGeneration`.

- **Utilities**
  - `decodeContentIfNeeded` to support `brotli` encoding alongside gzip for modern upstreams.

- **CLI, docs, and examples**
  - `cmd/aigw/docker-compose.yaml`: added `image-generation` service (curl-based client) modeled after chat/embeddings.
[...]

PR #1870, opened 2026-02-16
feat: add OpenAI image edit endpoint support

### LINKED ISSUE #1746: Openai image edit endpoint
_Description_:
Alongside OpenAI API for image generation would be great to have an image edit endpoint. Similarly to generations, part of v1/images

> Describe the desired behavior, what scenario it enables and how it
> would be used.
Support the OpenAI schema for image editing.
Endpoint path: v1/images/edits

Minimal set of payload fields:
model
image
prompt
response_format

Allow extra fields by: extra_body

[optional *Relevant Links*:]

> Any extra documentation required to understand the issue.

Aman123lug:
**Description**

This PR adds support for the OpenAI /v1/images/edits endpoint across the Envoy AI Gateway. The implementation closely follows the pattern established by the image generation endpoint (merged in #1280).

Resolves #1746.

**Changes**

**API Schema** (internal/apischema/openai/image_edit.go)
- Added ImageEditRequest with all fields from the OpenAI API reference: image, prompt, model, mask, n, size, response_format, user.
- Added ImageEditResponse and ImageEditResponseData matching the images object schema.

**Translator** (internal/translator/imageedit_openai_openai.go)
- Implemented OpenAI-to-OpenAI translator for image edits:
  - Request body transformation with model override support.
  - Path header routing to /v1/images/edits.
  - Response body parsing and response model attribution (using request model since the images API omits it in responses).
  - Error handling via the existing convertErrorOpenAIToOpenAIError helper.
- Added OpenAIImageEditTranslator type alias in translator.go.

**Tracing** (internal/tracing/openinference/openai/image_edit.go)
- Added ImageEditRecorder following OpenInference semantic conventions.
- Records request/response attributes with proper input/output redaction support.
- Wired up ImageEditTracer, ImageEditSpan, and NoopImageEditTracer in the tracing API and implementation.

**Metrics** (internal/metrics/genai.go)
- Added GenAIOperationImageEdit operation type for observability.

**Endpoint Registration** (internal/endpointspec/endpointspec.go)
- Added ImageEditEndpointSpec with ParseBody and GetTranslator methods.

**ExtProc** (cmd/extproc/mainlib/main.go)
- Registered /v1/images/edits processor with its metrics factory, tracer, and endpoint spec.

**Tests**

- Unit tests for the translator covering:
  - Path header routing
  - Model override behavior
  - Response body parsing with model attribution
  - Error response handling
- All existing tests continue to pass.

**AI Disclosure**

AI tooling (GitHub Copilot) was used to assist with scaffolding this implementation. I fully understand the code, reviewed every change, and take ownership of this PR.

Resolves #1746
Related PR: #1280

OpenAI API reference for the image edit endpoint:
1: https://platform.openai.com/docs/api-reference/images/createEdit

### DIFF
--- a/cmd/extproc/mainlib/main.go
+++ b/cmd/extproc/mainlib/main.go
@@ -261,6 +261,7 @@ func Main(ctx context.Context, args []string, stderr io.Writer) (err error) {
 	completionMetricsFactory := metrics.NewMetricsFactory(meter, metricsRequestHeaderAttributes, metrics.GenAIOperationCompletion)
 	embeddingsMetricsFactory := metrics.NewMetricsFactory(meter, metricsRequestHeaderAttributes, metrics.GenAIOperationEmbedding)
 	imageGenerationMetricsFactory := metrics.NewMetricsFactory(meter, metricsRequestHeaderAttributes, metrics.GenAIOperationImageGeneration)
+	imageEditMetricsFactory := metrics.NewMetricsFactory(meter, metricsRequestHeaderAttributes, metrics.GenAIOperationImageEdit)
 	responsesMetricsFactory := metrics.NewMetricsFactory(meter, metricsRequestHeaderAttributes, metrics.GenAIOperationResponses)
 	rerankMetricsFactory := metrics.NewMetricsFactory(meter, metricsRequestHeaderAttributes, metrics.GenAIOperationRerank)
 	mcpMetrics := metrics.NewMCP(meter, metricsRequestHeaderAttributes)
@@ -284,6 +285,8 @@ func Main(ctx context.Context, args []string, stderr io.Writer) (err error) {
 		responsesMetricsFactory, tracing.ResponsesTracer(), endpointspec.ResponsesEndpointSpec{}))
 	server.Register(path.Join(flags.rootPrefix, endpointPrefixes.OpenAI, "/v1/images/generations"), extproc.NewFactory(
 		imageGenerationMetricsFactory, tracing.ImageGenerationTracer(), endpointspec.ImageGenerationEndpointSpec{}))
+	server.Register(path.Join(flags.rootPrefix, endpointPrefixes.OpenAI, "/v1/images/edits"), extproc.NewFactory(
+		imageEditMetricsFactory, tracing.ImageEditTracer(), endpointspec.ImageEditEndpointSpec{}))
 	server.Register(path.Join(flags.rootPrefix, endpointPrefixes.Cohere, "/v2/rerank"), extproc.NewFactory(
 		rerankMetricsFactory, tracing.RerankTracer(), endpointspec.RerankEndpointSpec{}))
 	server.Register(path.Join(flags.rootPrefix, endpointPrefixes.OpenAI, "/v1/models"), extproc.NewModelsProcessor)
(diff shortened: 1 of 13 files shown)


missBerg (commented):
Nice work following the established image generation pattern, this is clean and consistent. A few things I noticed while reading through:

**1. Multipart/form-data handling**

The OpenAI `/v1/images/edits` endpoint requires `multipart/form-data` (since `image` and `mask` are file uploads), unlike `/v1/images/generations` which accepts JSON. The gateway's extproc pipeline is JSON-only, `ParseBody` does `json.Unmarshal` directly on the raw bytes from Envoy with no content-type awareness or multipart parsing. This means a real client sending a multipart request would fail at the JSON parse step, and even if it somehow got through, OpenAI would reject a non-multipart body upstream. **How is this intended to work? This may need more design thinking.**

**2. Response type reuse with image generation**

`ImageEditResponse` / `ImageEditResponseData` are structurally very similar to `ImageGenerationResponse` / `ImageGenerationResponseData` (generation has extra `Usage`, `OutputFormat`, `Size`, `Background` fields). Would it make sense to reuse the generation types here, or is the intent to keep them separate so they can evolve independently?

**3. Token usage handling**

The image generation translator populates `tokenUsage` from `resp.Usage` (for gpt-image-1). The edit response type doesn't include a `Usage` field, which matches the current OpenAI API behavior. Might be worth a brief comment or TODO noting that if/when OpenAI adds usage info to the edits response, this would need updating.

**4. Minor: `imageEditStartOpts` var comment**

```go
// startOpts sets trace.SpanKindInternal as that's the span kind used in
// OpenInference.
var imageEditStartOpts = ...
```
The comment says `startOpts` but the variable is `imageEditStartOpts` which looks like a copy-paste artifact from image generation.

Overall this looks solid and well-structured. The main question is #1 around multipart handling — curious to hear how that's intended to work.

Aman123lug:
Thanks for the thorough review @missBerg! Great catches across the board. I have addressed all four points in the latest commit:

**1. Multipart/form-data handling**

Good question — with gpt-image-1, OpenAI now accepts JSON bodies for /v1/images/edits (images are passed as URLs or file_ids instead of file uploads). This is the same approach the image generation endpoint already uses in the gateway. The extproc pipeline handles JSON naturally for gpt-image-1 clients. I have updated the ImageEditRequest struct comment to clarify this, and added the newer gpt-image-1 fields (quality, background, output_format, moderation, output_compression, partial_images). The legacy DALL-E 2 multipart/form-data flow would bypass the extproc JSON parsing, consistent with how the gateway handles non-JSON content types today.

**2. Response type reuse with image generation**

Agreed — since OpenAI returns the same ImagesResponse schema for both endpoints, I have made ImageEditResponse a type alias for ImageGenerationResponse. Both endpoints now share the same response structure including Usage, OutputFormat, Size, and Background fields. No duplication, and if the shared schema evolves upstream, both benefit automatically.

**3. Token usage handling**

Since ImageEditResponse now reuses ImageGenerationResponse (which includes the Usage field), I have added token usage population in the image edit translator ResponseBody, matching the exact same pattern used in image generation. gpt-image-1 edit responses with usage info will now be properly tracked for metrics.

**4. Comment fix**

Fixed the copy-paste comment from startOpts to imageEditStartOpts. Also noticed the image edit tracing was missing the LLMInvocationParameters attribute that image generation has, so added that for consistency.

I also expanded the test coverage to match the image generation test structure — usage, force mutation, decode error, span recording, model propagation, error passthrough. All tests passing.

PR #1847, opened 2026-02-10
feat: add compatibility with open responses api

sivanantha321:
**Description**
This PR adds support for OpenResponse API[^1] on top of existing OpenAI's responses API.

[^1]: https://www.openresponses.org/

Other than the mentioned differences in the comments, Open Responses also does not support OpenAI specific fields like `user`, `shell_call`, `file_search_call`,`computer_call`,`websearch_call`,`compaction`, `image_generation_call`, `code_interpreter_call`, `local_shell_call`, `apply_patch_call`, `mcp`, `custom tool call`, `hosted_tools`, `prompt_cache_retention`, `prompt` and `conversation`

**Related PRs**
-  #1649 
-  #1791 

**Open Responses Compliance Test**
```shell
$ bun run bin/compliance-test.ts --base-url "http://localhost:1975/v1" -v --api-key"xxxx" --json
{
  "summary": {
    "passed": 6,
    "failed": 0,
    "total": 6
  },
  "results": [
    {
      "id": "multi-turn",
      "name": "Multi-turn Conversation",
      "description": "Send assistant + user messages as conversation history",
      "status": "passed",
      "duration": 834,
      "request": {
        "model": "gpt-4o-mini",
        "input": [
          {
            "type": "message",
            "role": "user",
            "content": "My name is Alice."
          },
          {
            "type": "message",
            "role": "assistant",
            "content": "Hello Alice! Nice to meet you. How can I help you today?"
          },
          {
            "type": "message",
            "role": "user",
            "content": "What is my name?"
          }
        ]
      },
      "response": {
        "id": "resp_07327ffeea44a26500699e9aa8cb2c81a3bbda037ae5fc082f",
        "object": "response",
        "created_at": 1772001960,
        "completed_at": 1772001961,
        "status": "completed",
        "incomplete_details": null,
        "model": "gpt-4o-mini-2024-07-18",
        "previous_response_id": null,
        "instructions": null,
        "output": [
          {
            "type": "message",
            "id": "msg_07327ffeea44a26500699e9aa9150881a3921860a23d13d85c",
            "status": "completed",
            "role": "assistant",
            "content": [
              {
                "type": "output_text",
                "text": "Your name is Alice.",
                "annotations": [],
                "logprobs": []
              }
            ]
          }
        ],
        "error": null,
        "tools": [],
        "tool_choice": "auto",
        "truncation": "disabled",
        "parallel_tool_calls": true,
        "text": {
          "format": {
            "type": "text"
          },
[...]

PR #1879, opened 2026-02-20
fix: support assistant messages without id in Responses API input parsing

xiaolin593:
**Description**

The `ResponseInputItemUnionParam.UnmarshalJSON` discriminator for `type: "message"` only checked for an `"id"` field to identify output messages.
  Assistant messages sent back in multi-turn conversation history (e.g., from Codex CLI) carry `output_text` content but no `"id"`. Without the `"id"`, parsing fell through to input message types that don't support output_text, causing `"failed to parse JSON for /v1/responses"`.

  The fix adds a secondary check: if `role` is `"assistant"`, try parsing as `ResponseOutputMessage` before falling through to input message types.

Specific request payload block cause the issue:

```
{
      "type": "message",
      "role": "assistant",
      "content": [
        {
          "type": "output_text",
          "text": "Hi! I’m here and working. How can I help you with the codex repo or anything else?"
        }
      ]
    },
```

Related PR: #1791

### DIFF
--- a/examples/mcp/mcp_example.yaml
+++ b/examples/mcp/mcp_example.yaml
@@ -28,10 +28,18 @@ spec:
         apiKey:
           secretRef:
             name: github-access-token
-    - name: context7
-      kind: Backend
-      group: gateway.envoyproxy.io
-      path: "/mcp"
+    # TODO(nacx): Context7 started giving errors due to its certificate:
+    # time=2026-02-20T12:14:12.555+01:00 level=ERROR msg="failed to create MCP session" component=mcp-proxy backend=context7
+    # error="MCP initialize request failed with status code 503 and body=upstream connect error or disconnect/reset before headers.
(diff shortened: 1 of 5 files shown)


mathetake (approved):
+1

internal/apischema/openai/openai.go
```diff
@@ -3674,6 +3676,13 @@ func (r *ResponseInputItemUnionParam) UnmarshalJSON(data []byte) error {
 			r.OfOutputMessage = &om
 			return nil
 		}
+		if gjson.GetBytes(data, "role").String() == "assistant" {
```
yuzisun, line 3679: Looks like @sivanantha321 made the same change here and also marked id field optional. https://github.com/envoyproxy/ai-gateway/pull/1847/changes#diff-d792316034d6659a28a1e83a59cd15cfd3d3ff0d1c27dff750251d22950be784R3673

PR #1884, opened 2026-02-23
fix: set ext_proc circuit breakers to MaxInt32

PatilHrushikesh:
**Description**

This commit configures only the ext_proc UDS cluster and InferencePool EPP ext_proc cluster circuit breaker thresholds `max_connections` and `max_requests` to `math.MaxInt32`. Envoy's built-in default of 1024 concurrent requests is too low for this hop: each request opens a gRPC stream to ext_proc, which produces overflow errors (`ext_proc_error_gRPC_error_14{...overflow}`) under concurrency. Limits should be applied on the downstream/upstream side instead of at ext_proc.

`max_pending_requests` and `max_retries` are intentionally left unconfigured so Envoy's defaults continue to apply to those thresholds.

This follows the review feedback to hardcode MaxInt32 rather than adding a Helm/flag option.


internal/extensionserver/extensionserver.go
```diff
@@ -69,6 +80,13 @@ func (s *Server) List(context.Context, *grpc_health_v1.HealthListRequest) (*grpc
 	}}, nil
 }
 
+func (s *Server) getExtProcMaxRequests() uint32 {
```
nacx: Is this getter needed? The logic in this getter is already handled in hte constructor.

nacx:
Can you take a look at the failing tests?

mathetake (changes requested):
why don't you just set MAXINT32 instead of adding an option? The limit could be effectively applied on the downstream/upstream side config instead of at extrproc

PatilHrushikesh:
> why don't you just set MAXINT32 instead of adding an option? The limit could be effectively applied on the downstream/upstream side config instead of at extrproc


This can also be done, but this will bring back upgrade failure issue, where after upgrade is cluster configuration is different which causes CDS update and streaming requests in progress are dropped by extproc

pratheekrebala:
Hi @nacx - would you mind taking a look at this PR when you have a chance?

We are using a vLLM through KServe as part of our document extraction pipeline. Our workflow involves submitting a lot of small image patches concurrently. Right now, EPP isn't really usable for us because we run into the gRPC overflow errors almost immediately.

We're also unable to adjust this value using a `BackendTrafficPolicy` so we're having to bypass EPP/Envoy and just using a round robin service.

william-rom:
Hitting connection overflow in production with InferencePool/EPP setup as well. Unable to use EnvoyPatchPolcies to fix this either as the EPP cluster does not exist yet when the patches are applied.

twandja:
Hi, any plan on merging this PR? We are encountering the same connection exhaustion when benching our kserve llminferenceservice x llm-d setup.

ngyale-mirakl:
Hi, we are having the same issue on our side using KServe LLMInferenceService with the Endpoint Picker scheduler as ext_proc in a basic scenario testing with 2 replicas of Qwen2.5-7B-Instruct on 2 L40S with the following load produced by inference-perf:


```yaml
load:
  type: constant
  stages:
    - {rate: 10, duration: 120}
    - {rate: 20, duration: 120}
    - {rate: 30, duration: 120}
```

```
                         Throughput and Goodput Summary
┏━━━━━━━┳━━━━━━━━━━┳━━━━━━━━━┳━━━━━━━━━━┳━━━━━━━┳━━━━━━━━━┳━━━━━━━━━━┳━━━━━━━━━┓
┃       ┃          ┃ Achiev… ┃    Error ┃       ┃      In ┃      Out ┃     Tot ┃
┃ Stage ┃ Req Rate ┃    Rate ┃     Rate ┃ Req/s ┃ Tokens… ┃ Tokens/s ┃ Tokens… ┃
┡━━━━━━━╇━━━━━━━━━━╇━━━━━━━━━╇━━━━━━━━━━╇━━━━━━━╇━━━━━━━━━╇━━━━━━━━━━╇━━━━━━━━━┩
│     0 │     10.0 │    10.0 │    58.9% │   2.4 │  5249.3 │      5.8 │  5255.1 │
│     1 │     20.0 │    20.0 │    82.0% │   2.3 │  5092.2 │      8.5 │  5100.8 │
│     2 │     30.0 │    30.0 │    88.9% │   2.1 │  4544.6 │     17.0 │  4561.6 │
└───────┴──────────┴─────────┴──────────┴───────┴─────────┴──────────┴─────────┘


                         Request Error Summary
┏━━━━━━━┳━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
┃ Stage ┃ Successes ┃ 500 - Internal Server Error ┃ 503 - Service Unavailable ┃
┡━━━━━━━╇━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━┩
│     0 │       493 │                         707 │                         0 │
│     1 │       433 │                        1966 │                         1 │
│     2 │       398 │                        3202 │                         0 │
└───────┴───────────┴─────────────────────────────┴───────────────────────────┘
```

The same scenario without the scheduler produced this (No error at all):
```
                         Throughput and Goodput Summary
┏━━━━━━━┳━━━━━━━━━━┳━━━━━━━━━┳━━━━━━━━━━┳━━━━━━━┳━━━━━━━━━┳━━━━━━━━━━┳━━━━━━━━━┓
┃       ┃          ┃ Achiev… ┃    Error ┃       ┃      In ┃      Out ┃     Tot ┃
┃ Stage ┃ Req Rate ┃    Rate ┃     Rate ┃ Req/s ┃ Tokens… ┃ Tokens/s ┃ Tokens… ┃
┡━━━━━━━╇━━━━━━━━━━╇━━━━━━━━━╇━━━━━━━━━━╇━━━━━━━╇━━━━━━━━━╇━━━━━━━━━━╇━━━━━━━━━┩
│     0 │     10.0 │    10.0 │     0.0% │   8.0 │ 17507.1 │    781.9 │ 18289.0 │
│     1 │     20.0 │     9.5 │     0.0% │   8.2 │ 17951.3 │    764.6 │ 18715.8 │
│     2 │     30.0 │     9.1 │     0.0% │   8.3 │ 17967.0 │    786.8 │ 18753.8 │
└───────┴──────────┴─────────┴──────────┴───────┴─────────┴──────────┴─────────┘

                         Request Error Summary
┏━━━━━━━┳━━━━━━━━━━━┓
┃ Stage ┃ Successes ┃
┡━━━━━━━╇━━━━━━━━━━━┩
│     0 │      1200 │
│     1 │      2400 │
│     2 │      3600 │
└───────┴───────────┘
```

We are really happy with the Envoy AI Gateway and KServe but it's hard to go to production with the LLMInferenceService (and its benefits from llm-d intelligent routing) with the current status.

Hope you will be able to merge this PR, thank you !

PR #1892, opened 2026-02-24
feat: add reasoning_effort mapping for claude models with anthropic api.

hustxiayang:
**Description**
This is to add the mapping of `reasoning_effort` from openai into claude's `Effort`:
Current mapping:
```
"low" → OutputConfigEffortLow
"medium" → OutputConfigEffortMedium
"high" → OutputConfigEffortHigh
"xhigh" → OutputConfigEffortMax
```

Upgrade openai version for "xhigh".

### DIFF
--- a/internal/extproc/processor_impl.go
+++ b/internal/extproc/processor_impl.go
@@ -6,9 +6,12 @@
 package extproc
 
 import (
+	"bytes"
 	"cmp"
 	"context"
+	"errors"
 	"fmt"
+	"io"
 	"log/slog"
 	"strconv"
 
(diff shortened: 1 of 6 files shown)


extproc
nacx, line 1: Remove this

hustxiayang: thanks!!

internal/translator/anthropic_helper.go
```diff
@@ -585,6 +586,27 @@ func outputConfigAvailable(model internalapi.RequestModel) bool {
 		strings.Contains(modelLower, "4-6")
 }
 
+// mapReasoningEffortToOutputConfigEffort converts OpenAI reasoning effort levels to Anthropic output config effort levels.
+// Currently supports:
+// - "low" → OutputConfigEffortLow
+// - "medium" → OutputConfigEffortMedium
+// - "high" → OutputConfigEffortHigh
+// - "xhigh" → OutputConfigEffortMax
```
aabchoo, line 607: Thoughts on updating this to max? We don’t need to map the exact OpenAI value since xhigh doesn’t mean max

hustxiayang: I would prefer to have an unified interface here as we do not have any information exposed by anthropic on what is the propriate mapping to openai's reasoning_effort, for example "low" in reasoning_effort should be mapped to "low" in OutputConfigEffortLow? Thus, I think it's better just to use the same unified interface, and not to introduce new.

aabchoo: I don't think we should be arbitrarily mapping `xhigh` to `max`. There is no guarantee that openai treats xhigh the same as max.

We already have examples of custom values such as `service_tier` having the option to be `reserved` which is a bedrock specific field.

hustxiayang: `There is no guarantee that openai treats xhigh the same as max.` As mentioned above, openai's `low` is not necessary to be same as claude's low, and `we do not have any information exposed by anthropic on what is the appropriate mapping to openai's reasoning_effort`.  If we want users to know what exactly they're using, then we have to use something like `claude-low`, `claude-medium`, `claude-high`.  You may be able to refer https://github.com/envoyproxy/ai-gateway/blob/860e8bccb907fdf82f8eb030cf0c471f7f636c97/internal/translator/gemini_helper.go#L576, here openai's medium is mapped to "high" for gemini3-pro, because they exposed the details to users. When we do not have such information, we just need to distinguish them and not necessary to add just one to change the interface, as we do before, for example,  medium resolution 's low is not necesssary the same as detail low, we just use them to distinguish between details. @yuzisun

yuzisun: litellm uses the vendor field directly instead of translation.
https://docs.litellm.ai/blog/claude_opus_4_6#effort-levels

yuzisun: all reasoning_effort map to adaptive

yuzisun: Discussed offline to document the mapping on the website.

PR #1901, opened 2026-02-26
examples: fix invalid CR schema in one of the examples

herewasmike:
**Description**

In current examples manifests for EnvoyProxy resource are not matching CR schema https://gateway.envoyproxy.io/docs/api/extension_types/#envoyproxy

Thus when I tried to deploy it I had to figure this out

I'll update it everywhere once someone confirms that this is the right track


mathetake (approved):

herewasmike:
@mathetake I take it as I was right.

I'll update the rest of the examples, then

nacx:
Thanks!

PR #1906, opened 2026-03-01
feat: add eager_input_stream support for Anthropic

xiaolin593:
**Description**

- Adds a new `eager_input_streaming` field to ChatCompletionRequest that, when true, opts the request into Anthropic's fine-grained tool streaming so tool input parameters stream incrementally instead of being buffered until complete.
- API behavior: sets EagerInputStreaming=true on every translated Anthropic ToolParam when enabled

PS: Re-purpose the PR due to accidentally force-push

### DIFF
--- a/internal/translator/anthropic_helper.go
+++ b/internal/translator/anthropic_helper.go
@@ -971,13 +971,6 @@ func (p *anthropicStreamParser) handleAnthropicStreamEvent(eventType []byte, dat
 		if input, ok := usage.InputTokens(); ok {
 			p.tokenUsage.SetInputTokens(input)
 		}
-		if cached, ok := usage.CachedInputTokens(); ok {
-			p.tokenUsage.SetCachedInputTokens(cached)
-		}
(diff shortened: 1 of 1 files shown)


hustxiayang:
1) would it be better that users can ignore it if we choose to continue with the stream 2) is it safe that we always continue the stream for anthropic errors during stream?

internal/translator/anthropic_helper.go
```diff
@@ -723,6 +724,13 @@ func newAnthropicStreamParser(requestModel string) *anthropicStreamParser {
 func (p *anthropicStreamParser) writeChunk(eventBlock []byte, buf *[]byte) error {
 	chunk, err := p.parseAndHandleEvent(eventBlock)
 	if err != nil {
+		var streamErr anthropicStreamErrorEvent
+		if errors.As(err, &streamErr) {
+			*buf = append(*buf, sseDataPrefix...)
+			*buf = append(*buf, streamErr.payload...)
+			*buf = append(*buf, '\n', '\n')
```
yuzisun: Do we still append the DONE event after the error type event?

PR #1907, opened 2026-03-02
fix: default tool type to customer when type is absent

aripurnama0705-engineer:
**Description**

This commit sets a default value for the type field in /v1/messages requests when using tools.

Based on the current Anthropic API behavior, the type field is required. If it is omitted from the API request, the server returns a 400 BadRequest error:

```bash
{
  "type": "error",
  "error": {
    "type": "BadRequest",
    "code": "400",
    "message": "malformed request: failed to parse JSON for /v1/messages"
  }
}
```
By assigning a default value, we ensure compatibility with the API requirement and prevent malformed request errors when the type field is not explicitly provided.

### DIFF
--- a/internal/apischema/anthropic/anthropic.go
+++ b/internal/apischema/anthropic/anthropic.go
@@ -1113,10 +1113,13 @@ const (
 
 func (t *ToolUnion) UnmarshalJSON(data []byte) error {
 	typ := gjson.GetBytes(data, "type")
-	if !typ.Exists() {
-		return errors.New("missing type field in tool")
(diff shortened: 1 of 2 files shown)


mathetake:
why do you think it's AIGW's responsibility to modify the API behavior/contract between client<>anthropic? like if you think about it, are you planning to set the default like this to every single one of the field Anthropic requires? I don't think so - you can always tell your clients/app set the field correctly following Anthropic rather than having a very fragile implicit behavior here.

mathetake:
also this is not impacting the runtime behavior - anthropic messages' unmarshal is not used for constructing requests. I would suggest you solve the problem that you actually have and verify the patch actually resolves that beforehand

PR #1878, opened 2026-02-19
feat: anthropic endpoint support and translation for openai backend

changminbark:
**Description**

This commit adds a translator that will convert a request sent to `/anthropic/v1/messages` and `/v1/messages` endpoints for OpenAI schema backends. It does not matter whether the OpenAI schema backend natively supports the endpoint (e.g. vLLM) as translating should be a light/fast enough process. This approach is also more versatile and future-proof than just passing through the Anthropic Message Request to a backend that natively supports it. It also follows the already-existing structure for adding translators, path processor factories, and schema translation.

A major example use case would be using AI Gateway to route requests from Claude Code to several AI backends like locally hosted vLLM models with LoRA adapters.

NOTE: vLLM is only used for local testing as I do not have access to compute. The intended goal for this PR is to support any OpenAI compatible backend/services using an Anthropic interface.

**Related Issues/PRs (if applicable)**

Fixes #1372 
Fixes #1867 

**Special notes for reviewers (if applicable)**
Claude Code was used to write most of the tests but were verified. It would also be nice if the maintainers could review the other PR #1843 as some of the Anthropic apischema here can be updated once #1843 is merged.

<details>

<summary>Functional Test Results</summary>

Test for anthropic endpoints for OpenAI schema backends that natively support it:
```
$ curl -v http://localhost:8080/v1/messages   -H "Content-Type: application/json"   -d '{
    "model": "Qwen/Qwen2.5-0.5B-Instruct",
    "messages": [
      {"role": "user", "content": "Say hello!"}
    ],
    "max_tokens": 100
  }'
* Host localhost:8080 was resolved.
* IPv6: ::1
* IPv4: 127.0.0.1
*   Trying [::1]:8080...
* Connected to localhost (::1) port 8080
> POST /v1/messages HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/8.5.0
> Accept: */*
> Content-Type: application/json
> Content-Length: 143
> 
< HTTP/1.1 200 OK
< date: Fri, 20 Feb 2026 18:46:04 GMT
< server: uvicorn
< content-type: application/json
< content-length: 331
< 
* Connection #0 to host localhost left intact
[...]

PR #1909, opened 2026-03-02
Add header mutation for anthropic version in sample

jimangel:
Add header mutation to set anthropic version (https://platform.claude.com/docs/en/api/versioning).

Otherwise:

```
{"type":"error","error":{"type":"invalid_request_error","message":"anthropic-version: header is required"},"request_id":"..."}
```

**Description**


mathetake:
i mean why? you should send from your client in the first place, no?

mathetake:
if this is about the translation case (#1878), then the translator should set it, not via the API settings like this

PR #1920, opened 2026-03-05
ci: upgrade trivy action to fix scan issues

nacx:
**Description**

Upgrades the Trivy action to fix the recent scan job failures.

N/A

I tested running the action on my fork and it completed fine.

### DIFF
--- a/.github/workflows/trivy.yaml
+++ b/.github/workflows/trivy.yaml
@@ -31,7 +31,7 @@ jobs:
           make docker-build.${{ matrix.target.command_name }} TAG=${{ github.sha }}
 
       - name: Run Trivy vulnerability scanner for ${{ matrix.target.command_name }}
-        uses: aquasecurity/trivy-action@e368e328979b113139d6f9068e03accaed98a518 # master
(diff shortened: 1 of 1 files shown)


.github/workflows/trivy.yaml
```diff
@@ -31,7 +31,7 @@ jobs:
           make docker-build.${{ matrix.target.command_name }} TAG=${{ github.sha }}
 
       - name: Run Trivy vulnerability scanner for ${{ matrix.target.command_name }}
-        uses: aquasecurity/trivy-action@e368e328979b113139d6f9068e03accaed98a518 # master
+        uses: aquasecurity/trivy-action@97e0b3872f55f89b95b2f65b3dbab56962816478 # v0.34.2
```
mathetake, line 34: just to be clear this is not compromised right? https://github.com/aquasecurity/trivy/discussions/10265

mathetake: though this workflow has only read permission, so no issue anyway

mathetake (approved):

.github/workflows/trivy.yaml
nacx: I think this version is fine.

PR #1927, opened 2026-03-06
fix: handle anthropic gzip streaming response

yuzisun:
**Description**
Add upstream test case to reproduce the anthropic streaming error in vendor e2e tests.

  1. The test upstream sends gzip-compressed SSE chunks (one per event with gz.Flush())
  2. Envoy delivers each chunk to the ext_proc in STREAMED mode
  3. decodeContentIfNeeded creates a new gzip.NewReader for the 2nd+ chunks, which lack valid gzip headers
  4. The gzip reader fails with unexpected EOF
  5. This propagates through ProcessResponseBody → translator → gRPC error
  6. Envoy sees the gRPC error, terminates the downstream connection
  7. The client gets status 200 but an empty body (0 bytes) — the test times out waiting for actual content

### DIFF
--- a/tests/internal/testupstreamlib/server_test.go
+++ b/tests/internal/testupstreamlib/server_test.go
@@ -8,6 +8,7 @@ package testupstreamlib
 import (
 	"bufio"
 	"bytes"
+	"compress/gzip"
 	"encoding/base64"
 	"fmt"
 	"io"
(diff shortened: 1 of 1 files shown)


yuzisun:
@nacx @mathetake can you review and unblock the main branch releases.

internal/extproc/processor_impl.go
```diff
@@ -530,6 +545,29 @@ func (u *upstreamProcessor[ReqT, RespT, RespChunkT, EndpointSpecT]) ProcessRespo
 	return resp, nil
 }
 
+// decodeStreamingContent handles decompression for streaming responses with content-encoding.
+// It accumulates raw compressed bytes across chunks and re-decompresses from the beginning each time,
+// returning only the newly decompressed data. This is necessary because gzip streams are stateful
+// and a new decompressor cannot be created mid-stream without the full preceding data.
+func (u *upstreamProcessor[ReqT, RespT, RespChunkT, EndpointSpecT]) decodeStreamingContent(chunk []byte, endOfStream bool) (contentDecodingResult, error) {
+	u.compressedBuf = append(u.compressedBuf, chunk...)
+	decodingResult, err := decodeContentIfNeeded(u.compressedBuf, u.responseEncoding)
```
mathetake, line 554: i am not following where compressedBuf is shrunk. Isn't this buffering the entire stream, which could be huge MBs?

yuzisun: Unlikely for output tokens, frontier models generally support tens of thousands of output tokens, with a few pushing toward ~128k output token.

internal/extproc/processor_impl.go
```diff
@@ -530,6 +545,29 @@ func (u *upstreamProcessor[ReqT, RespT, RespChunkT, EndpointSpecT]) ProcessRespo
...
+	if err != nil {
+		return contentDecodingResult{}, err
+	}
+	allDecompressed, readErr := io.ReadAll(decodingResult.reader)
```
mathetake, line 558: also the reader is created per responseBody callback on the []byte buffering all the historical stream. I am not sure how that works - isn't this duplicating the content when each event is compressed? Can you double check the testupstream behavior? (I would suggest to write a unit test in testupstream/server_test.go to verify it actually compresses each event

yuzisun: @mathetake since we do flush on each event it does compression for each of them, see the test case which verifies this https://github.com/envoyproxy/ai-gateway/pull/1927/changes/7a8c56d7ad10fef155e3435ca9e0938891938154

PR #1930, opened 2026-03-09
fix(controller): use server-side apply when patching finalizers

### LINKED ISSUE #1116: controller: do serverside apply on finalizer
*Description*:

There's a wip PR here https://github.com/envoyproxy/ai-gateway/pull/1088 but it seems some tests are failing, so it's missing something.

KaveeshKhattar:
**Description**

Switch to server-side apply via a new `patchFinalizersWithServerSideApply` helper. Server-side apply (SSA) does not require checking `resourceVersion`, so Kubernetes can merge the change even if there are concurrent modifications.

The helper does three things:
1. Looks up the GVK via `apiutil.GVKForObject`: required because controller-runtime strips TypeMeta when reading objects from the API, but SSA requires it to be set
2. Stamps the GVK back onto the object before patching
3. Patches using `client.RawPatch` with `types.ApplyPatchType`, `ForceOwnership`, and field owner `aigateway-controller`

Closes #1116
Related PR: #1088 

**Validation**

`make precommit test` passed locally.

This is my first PR to `envoyproxy/ai-gateway`, so I’m happy to iterate based on any feedback. Thanks!

### DIFF
--- a/.gemini/config.yaml
+++ b/.gemini/config.yaml
@@ -1,0 +1,12 @@
+# Copyright Envoy AI Gateway Authors
+# SPDX-License-Identifier: Apache-2.0
+# The full text of the Apache license is available in the LICENSE file at
+# the root of the repo.
+
+code_review:
+  disable: false
+  pull_request_opened:
+    help: false
+    summary: false
+    code_review: false
+    include_drafts: false
--- a/internal/extensionserver/extensionserver_test.go
+++ b/internal/extensionserver/extensionserver_test.go
@@ -1418,6 +1418,36 @@ func TestPostRouteModify(t *testing.T) {
 		require.Equal(t, wrapperspb.Bool(false), route.GetRoute().GetAutoHostRewrite())
 		require.NotNil(t, route.TypedPerFilterConfig)
 	})
+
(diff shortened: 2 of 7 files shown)


internal/controller/controller.go
```diff
@@ -523,9 +526,25 @@ func newConditions(conditionType, message string) []metav1.Condition {
 
 // aiGatewayControllerFinalizer is the name of the finalizer added to various AI Gateway resources.
 const aiGatewayControllerFinalizer = "aigateway.envoyproxy.io/finalizer"
+const aiGatewayControllerFieldOwner = "aigateway-controller"
+
+func patchFinalizersWithServerSideApply(ctx context.Context, c client.Client, o client.Object) error {
```
siddharth1036, line 533: Should we use merge patch over SSA?

KaveeshKhattar: I considered merge patch. The main reason I went with SSA is field ownership; SSA gives us managedFields semantics for the finalizer.

A merge patch is still a read‑modify‑write of the finalizers list and can overwrite concurrent changes without optimistic locking.

That said, if merge patch is preferred for simplicity, I can switch to it.

mathetake:
Tests & format is failing. could you check locally first before pushing changes?

PR #1933, opened 2026-03-09
feat: configure default max_tokens for anthropic translator

herewasmike:
**Description**

max_tokens and max_completion_tokens are optional in openAI spec
https://developers.openai.com/api/reference/resources/chat/subresources/completions/methods/create

However when I try to send the request from the source I get 422 if my AIServiceBackend has AWSAnthropic schema.
It's impossible to add this field with bodyMutation due to ordering (I've considered mutating body before the translation but it seems to be a whole redesign rather than small fix with providing default)

Per gen AI policy I disclose that claude did help me with setting up this PR


herewasmike:
I'm not sure if arbitrary default value + bodyMutator is the best approach.
Though, no idea how anthropic api reacts if supplied with max_tokens value greater than the model supports.
Happy to take suggestions here

nutanix-Hrushikesh:
> Though, no idea how anthropic api reacts if supplied with max_tokens value greater than the model supports.

it returns 400 Bad Request

johnugeorge:
This might be a bit confusing if low default is observed by default. Isn't it better for the client to be aware of and use the right values? Are you using a client where it is not configurable?

herewasmike:
Yes, I'm connecting 3rd party code in this particular case so it's not easy to change the client's behavior.
I'll rework this PR following up the conversation on slack, though, to avoid the confusion
> Simply allow requests to pass through and fail on the provider side (one would be able to detect it and mutate request body)

nacx (approved):

PR #1934, opened 2026-03-10
docs: mark version and prefix optional in VersionedAPISchema

herewasmike:
**Description**

Since they have "omitempty" tag they should be optional
This is a bit confusing as it is now


johnugeorge:
Can you rebase the changes?

herewasmike:
These tests seem to be flaky, how can I re-trigger them?

PR #1945, opened 2026-03-13
fix: handle email addresses in JWT sub claim for MCP session ID parsing

### LINKED ISSUE #1940: MCP session ID parsing breaks when JWT sub claim contains "@" (e.g. email addresses)
_Description_:

The MCP proxy session ID parser splits on "@" and expects exactly 3 parts, but fails when the JWT sub claim is an email address (e.g. user@example.com), which itself contains "@".

All MCP requests fail for any user whose identity provider uses email addresses as the sub claim. 

**Expected behavior**: MCP session initializes and requests are processed normally when the sub claim contains `@`.

**Component**: `internal/mcpproxy/session.go` 

The session ID is built as "routeName@subject@backends". When subject is user@example.com, splitting on "@" yields 4 parts instead of 3, causing the parser to reject the session ID.

**Workaround**: Configure the IdP to emit a `sub` claim without `@`. 

_Repro steps_:

1. Configure an MCPRoute with an OAuth security policy backed by Okta (or any IdP that uses email as sub)
2. Authenticate as a user whose sub claim is an email address (e.g. user@example.com)
3. Send any MCP request with the obtained Bearer token

**Actual result**: All MCP requests return an error

```
error="invalid session ID: missing '@' separator"
```

_Environment_:

- Envoy AI Gateway version: v0.5.0
- Envoy Gateway version: v1.7.0-rc.1
- IdP: Okta emits email as sub

_Logs_:

```
level=ERROR msg="invalid session ID in POST request" component=mcp-proxy error="invalid session ID: missing '@' separator"
```

deobieta:
**Description**

The MCP session ID parser used strings.Split on @ and expected exactly 3 parts ({routeName}@{subject}@{backends}), but failed when the JWT sub claim was an email address (e.g. user@example.com), producing 4 parts instead of 3 and returning invalid session ID: missing '@' separator on every request.

Fix by replacing strings.Split with a strings.LastIndex-based approach: the backends segment is always the suffix after the last @ (safe because backend session IDs are base64-encoded and cannot contain @), and the route name is the prefix before the first @ in the remainder. No change to the session ID format. 

Add TestBackendSessionIDs_EmailSubject covering email subject and empty string.

Fixes #1940

The subject field (parts[1] in the original code, discarded with _ = parts[1]) is retained in the session ID format for anti-hijacking purposes as documented in the spec reference on line 462. Its value is never needed during parsing — security is provided by the symmetric encryption wrapping the session ID, not by the plaintext content.

### DIFF
--- a/manifests/charts/ai-gateway-helm/templates/deployment.yaml
+++ b/manifests/charts/ai-gateway-helm/templates/deployment.yaml
@@ -78,7 +78,7 @@ spec:
               {{- $endpointPrefixes = append $endpointPrefixes (printf "anthropic:%s" $anthropic) -}}
             {{- end -}}
             {{- if $endpointPrefixes }}
-            - --endpointPrefixes={{ join "," $endpointPrefixes }}
+            - "--endpointPrefixes={{ join "," $endpointPrefixes }}"
             {{- end }}
             {{- if .Values.extProc.extraEnvVars }}
             - --extProcExtraEnvVars={{ include "ai-gateway-helm.extProc.envVarsString" . }}


internal/mcpproxy/session_test.go
```diff
@@ -59,6 +59,34 @@ func TestBackendSessionIDs_Success(t *testing.T) {
...
+	t.Parallel()
+	backendA := "backendA"
+	backendB := "backendB"
+	idA := "session-a"
+	idB := "session-b"
+	routeName := "some-route"
+	for _, subject := range []string{
+		"user@example.com",
+		"user@mail.example.com",
+		"firstname.lastname@company.org",
```
nacx: These two cases are redundant. Can you repalce these two for the empty string? So er test an empty subject as well, which is a valid input value tot he method.

deobieta: Of course. Done.

PR #1950, opened 2026-03-14
fix: anthropicOpenAi translator support for custom prefix

johnugeorge:
**Description**

Use OpenAIPrefix() for Anthropic→OpenAI translator so that /v1/messages uses the correct chat completions path instead of hardcoded path.

### DIFF
--- a/internal/translator/anthropic_openai.go
+++ b/internal/translator/anthropic_openai.go
@@ -26,7 +26,7 @@ import (
 
 // NewAnthropicToChatCompletionOpenAITranslator implements [Factory] for Anthropic to OpenAI ChatCompletion translation.
 // This translator converts Anthropic API format to OpenAI ChatCompletion API requests.
-// The prefix parameter is the OpenAI API prefix used to construct the translated path
+// The prefix parameter is the prefix field set in the OpenAI VersionAPISchema used to construct the translated path
 // (e.g., "v1" produces "/v1/chat/completions", "enterpriseai/v1" produces "/enterpriseai/v1/chat/completions").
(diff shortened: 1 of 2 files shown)


internal/translator/anthropic_openai.go
```diff
@@ -25,13 +26,15 @@ import (
...
 // NewAnthropicToChatCompletionOpenAITranslator implements [Factory] for Anthropic to OpenAI ChatCompletion translation.
 // This translator converts Anthropic API format to OpenAI ChatCompletion API requests.
-func NewAnthropicToChatCompletionOpenAITranslator(version string, modelNameOverride internalapi.ModelNameOverride) AnthropicMessagesTranslator {
-	// TODO: use "version" in APISchema struct to set the specific prefix if needed like OpenAI does. However, two questions:
-	// 	* Is there any "Anthropic compatible" API that uses a different prefix like OpenAI does?
-	// 	* Even if there is, we should refactor the APISchema struct to have "prefix" field instead of abusing "version" field.
-	_ = version
-	passthroughTranslator := NewAnthropicToAnthropicTranslator(version, modelNameOverride)
-	return &anthropicToOpenAIV1ChatCompletionTranslator{passthroughTranslator: &passthroughTranslator, modelNameOverride: modelNameOverride}
+// The prefix parameter is the OpenAI API prefix used to construct the translated path
```
gavrissh: Maybe reword to this?
```suggestion
// The prefix parameter is the prefix field set in the OpenAI VersionAPISchema used to construct the translated path
```

PR #1961, opened 2026-03-17
feat: support Gemini native client endpoints with prefix-based routing and token usage extraction

### LINKED ISSUE #1960: feat: support Gemini native client endpoints (/v1beta/models/{model}:generateContent)
## Summary

Gemini CLI and the Google AI SDK hard-code the native Gemini API path format:

```
POST /v1beta/models/{model}:generateContent
POST /v1beta/models/{model}:streamGenerateContent
```

The model name is embedded in the URL, not the request body. The current extproc server uses an exact-path hash map (`processorFactories`), so these paths fall through with a 404 and cannot be routed.

This issue tracks adding first-class support for Gemini native client endpoints so that tools like Gemini CLI can be pointed at the gateway without modification.

### 1. Prefix-based path dispatch in `Server` (`internal/extproc/server.go`)

Add `RegisterPrefix(prefix string, factory ProcessorFactory)` alongside the existing `Register`. When an exact match is not found, fall back to a longest-prefix match over all registered prefix factories. This is a general mechanism — not Gemini-specific.

### 2. `NewGeminiProcessorFactory` (`internal/extproc/processor_impl.go`)

A `ProcessorFactory` that:
- Parses the model name and streaming flag from the URL path (`extractGeminiModelFromPath`)
- Passes them as a pre-populated `GenerateContentEndpointSpec` to `newRouterProcessor` (router branch) or `newUpstreamProcessor` (upstream branch)
- Uses `NoopTracer` since Gemini passthrough does not need tracing

### 3. `GenerateContentEndpointSpec` (`internal/endpointspec/gemini_generatecontent.go`)

Endpoint spec carrying `ModelFromPath string` and `Streaming bool`. The existing `GeminiToGCPVertexAI` translator reads these to build the correct Vertex AI path suffix.

### 4. `GeminiToGCPVertexAITranslator` (`internal/translator/gemini_gcpvertexai.go`)

A near-passthrough translator: rewrites `:path` to the full Vertex AI `generateContent` / `streamGenerateContent` URL. Also strips `FunctionResponse.ID` from all parts before forwarding — the Google AI SDK (used by Gemini CLI in `gemini-api` auth mode) populates this field, but Vertex AI `/v1` rejects it as an unknown field. Vertex AI matches `function_response` to `function_call` by name, not id, so stripping is safe.

### 5. `EndpointPrefixes.Gemini` (`internal/internalapi/internalapi.go`)

Adds a `gemini` key to `ParseEndpointPrefixes`, defaulting to `/v1beta`, so operators can remap the prefix if needed.

### 6. Wire-up in `cmd/extproc/mainlib/main.go`

```go
geminiModelPrefix := strings.TrimRight(path.Join(flags.rootPrefix, endpointPrefixes.Gemini), "/") + "/models/"
server.RegisterPrefix(geminiModelPrefix, extproc.NewGeminiProcessorFactory(generateContentMetricsFactory))
```

## Tests

- `TestServer_ProcessorForPath_PrefixMatch` — exact match wins over prefix, longest prefix wins, unknown paths return `errNoProcessor`
- `TestExtractGeminiModelFromPath` — table-driven: standard, streaming, with root prefix, missing colon, empty string
- `TestNewGeminiProcessorFactory` — router (non-streaming), router (streaming), upstream branch

## Notes

- No changes to `filterapi`, controller, or CRD schema — the `GCPVertexAI` backend schema is reused as-is
- The translator is a passthrough for response bodies; token usage extraction is not implemented (Gemini native responses use a different token field structure — can be a follow-up)
- Streaming detection uses the `:streamGenerateContent` method suffix in the URL path, not a body field

cc @mathetake @yuzisun — would love your thoughts on the prefix dispatch approach and whether `RegisterPrefix` belongs on `Server` or should be a separate routing layer.

adirags:
**Description**

Closes #1960

## Summary

Gemini CLI and the Google AI SDK hard-code the native Gemini API path format:

```
POST /v1beta/models/{model}:generateContent
POST /v1beta/models/{model}:streamGenerateContent
```

The model name is embedded in the URL, not the request body. This PR adds full support for these endpoints by:

- **Prefix-based path dispatch** in the extproc server — `RegisterPrefix` complements the existing exact-match `Register`, using longest-prefix matching as a fallback. This is a general mechanism, not Gemini-specific.
- **`NewGeminiProcessorFactory`** — parses model name and streaming flag from the URL path (`extractGeminiModelFromPath`) and pre-populates `GenerateContentEndpointSpec` before handing off to `newRouterProcessor` or `newUpstreamProcessor`.
- **`GenerateContentEndpointSpec`** — carries `ModelFromPath` and `Streaming` from the factory through to the translator. `GetTranslator` currently supports `GCPVertexAI`; other backends can be added as follow-ups.
- **`GeminiToGCPVertexAITranslator`** — near-passthrough translator that rewrites `:path` to the full Vertex AI URL. Also strips `FunctionResponse.ID` from all parts: the Google AI SDK (used by Gemini CLI in `gemini-api` auth mode with `GOOGLE_GEMINI_BASE_URL`) populates this field, but Vertex AI `/v1` rejects it as unknown. Safe to strip — Vertex AI matches `function_response` to `function_call` by name, not id.
- **Token usage extraction** from `usageMetadata` — both non-streaming (JSON) and streaming (SSE, scans all events, takes the last one with `usageMetadata`). Extracts `promptTokenCount`, `candidatesTokenCount`, `totalTokenCount`, and `cachedContentTokenCount`.
- **`EndpointPrefixes.Gemini`** — adds `gemini` key to `--endpointPrefixes` flag, defaulting to `/v1beta`, so operators can remap the prefix alongside OpenAI/Anthropic/Cohere.

## Test plan

- `TestServer_ProcessorForPath_PrefixMatch` — exact match wins over prefix, longest prefix wins, shorter prefix fallback, unknown path returns `errNoProcessor`
- `TestExtractGeminiModelFromPath` — table-driven: standard, streaming, with root prefix, missing colon, empty string
- `TestNewGeminiProcessorFactory` — router (non-streaming), router (streaming), upstream branch
- `TestGeminiToGCPVertexAITranslator_RequestBody` — path rewrite, streaming path rewrite, `FunctionResponse.ID` stripping (single and multiple parts)
- `TestGeminiToGCPVertexAITranslator_ResponseBody` — non-streaming token extraction, cached token extraction, streaming SSE token extraction from final event, empty body, reader error
- `TestParseEndpointPrefixes` — existing tests cover the new `gemini` key via the updated allowed-keys error message

cc @mathetake @yuzisun


adirags:
## Validation

Tested end-to-end with Gemini CLI pointed at an Envoy AI Gateway instance backed by GCP Vertex AI (`locations/global`). The following are sanitized Envoy access log entries.

**Non-streaming** (`generateContent`) — token counts and cost units populated correctly:

```json
{
  "method": "POST",
  "path": "/v1/projects/[redacted]/locations/global/publishers/google/models/gemini-2.5-flash-lite:generateContent",
  "response_code": 200,
  "llm_input_token": 3265,
  "llm_output_token": 61,
  "llm_total_token": 3518,
  "llm_cached_input_token": 0,
  "llm_cost_units": 35700,
  "duration": 6778,
  "user_agent": "GeminiCLI/0.33.1/gemini-3.1-pro-preview (linux; x64)"
}
```

**Streaming** (`streamGenerateContent`) — routed and proxied correctly; token counts extracted from `usageMetadata` in the final SSE event:

```json
{
  "method": "POST",
  "path": "/v1/projects/[redacted]/locations/global/publishers/google/models/gemini-2.5-flash:streamGenerateContent?alt=sse",
  "response_code": 200,
  "llm_input_token": 4821,
  "llm_output_token": 312,
  "llm_total_token": 5133,
  "llm_cached_input_token": 0,
  "llm_cost_units": 11052,
  "duration": 2474,
  "user_agent": "GeminiCLI/0.33.1/gemini-3.1-pro-preview (linux; x64)"
}
```

Both requests matched `httproute/envoy-gateway-system/gemini-vertex-route/rule/0` via the new prefix-based dispatch (`/v1beta/models/` prefix → `NewGeminiProcessorFactory`).

cmd/extproc/mainlib/main.go
```diff
@@ -310,6 +310,12 @@ func Main(ctx context.Context, args []string, stderr io.Writer) (err error) {
 	server.Register(path.Join(flags.rootPrefix, endpointPrefixes.Anthropic, "/v1/messages"), extproc.NewFactory(
 		messagesMetricsFactory, tracing.MessageTracer(), endpointspec.MessagesEndpointSpec{}))
 
+	generateContentMetricsFactory := metrics.NewMetricsFactory(meter, metricsRequestHeaderAttributes, metrics.GenAIOperationChat)
+	// IMPORTANT: do NOT use path.Join for the trailing "/models/" segment —
+	// path.Join would strip the trailing slash needed to prevent prefix collisions.
+	geminiModelPrefix := strings.TrimRight(path.Join(flags.rootPrefix, endpointPrefixes.Gemini), "/") + "/models/"
```
PatilHrushikesh, line 316: This way would be better:
```suggestion
	geminiModelPrefix := path.Join(flags.rootPrefix, endpointPrefixes.Gemini, "models") + "/"
```

internal/translator/gemini_gcpvertexai.go
```diff
@@ -0,0 +1,159 @@
...
+	model     string
+	streaming bool
+}
+
+// RequestBody implements [GeminiGenerateContentTranslator.RequestBody].
+// Rewrites :path to the full GCP Vertex AI path. Also strips FunctionResponse.ID
+// from all parts: the Google AI SDK (used by Gemini CLI in gemini-api auth mode)
+// populates this field, but Vertex AI /v1 rejects it as an unknown field.
+// Safe to strip: Vertex AI matches function_response to function_call by name, not id.
+func (g *geminiToGCPVertexAITranslator) RequestBody(_ []byte, req *gcp.GenerateContentRequest, _ bool) (
```
JAORMX, line 46: **Blocker. Retries with tool calls will 400 on every attempt.**

So, the third parameter here is `forceBodyMutation` and it's being ignored (named `_`). That alone wouldn't be the problem... but combined with the in-place mutation below, it is.

The chain:
- `Content.Parts` is `[]*Part` (pointer slice; verified in `google.golang.org/genai@v1.46.0/types.go:1232-1235`).
- `Part.FunctionResponse` is `*FunctionResponse` (line 1129).
- So `for _, part := range content.Parts` gives a `*Part`, and `part.FunctionResponse.ID = ""` mutates the heap object that `routerProcessor.originalRequestBody` still points to.

First call: IDs cleared in place, `mutatedBody` returned. Good.

Retry: `forceBodyMutation` comes in as `true` from `processor_impl.go:328-329` (`u.onRetry() || u.parent.forceBodyMutation`). But the in-memory IDs were already cleared on the first call, so `needsMutation = false`, `mutatedBody = nil`. With no `BodyMutation` configured, `applyBodyMutation` falls back to `originalRequestBodyRaw` (`internal/extproc/util.go:107-110`). Which still has the IDs. Vertex AI v1 rejects. Same 400 every retry.

The canonical pattern is in `internal/translator/openai_openai.go:76-78`:

```go
if forceBodyMutation && len(newBody) == 0 {
    newBody = original
}
```

Three options, in order of preference:

1. Move the strip into `ParseBody` so the spec layer's `mutatedOriginalBody` handling takes over (already correct). Cleanest.
2. Use `sjson.DeleteBytes` on the raw body, mirror the pattern in `internal/translator/anthropic_gcpanthropic.go:54-56`. This also fixes the `cachedContent`/`labels`/`modelArmorConfig` drop on the same code path (separate comment on `gcp.go`).
3. Honor `forceBodyMutation` here and re-marshal unconditionally on retry.

Note that there's no test exercising the retry path (`forceBodyMutation=true` after a successful first call). Adding one would have caught this and is worth doing regardless of which fix lands.

cmd/extproc/mainlib/main.go
```diff
@@ -310,6 +310,12 @@ func Main(ctx context.Context, args []string, stderr io.Writer) (err error) {
 	server.Register(path.Join(flags.rootPrefix, endpointPrefixes.Anthropic, "/v1/messages"), extproc.NewFactory(
 		messagesMetricsFactory, tracing.MessageTracer(), endpointspec.MessagesEndpointSpec{}))
 
+	generateContentMetricsFactory := metrics.NewMetricsFactory(meter, metricsRequestHeaderAttributes, metrics.GenAIOperationChat)
```
JAORMX, line 313: **Blocker. Wrong OTel GenAI operation name.**

The OTel GenAI semantic-conventions registry defines `generate_content` for exactly this surface. From the live page (https://opentelemetry.io/docs/specs/semconv/registry/attributes/gen-ai/, fetched fresh during review):

> `generate_content` ... "Multimodal content generation operation such as Gemini Generate Content"
> `chat` ... "Chat completion operation such as OpenAI Chat API"

Verbatim. So `chat` is wrong here.

Anthropic uses the dedicated `GenAIOperationMessages` constant on line 280 of this same file, for the same reason. The repo convention is "if the spec defines a dedicated value, use it." Right now there's no `GenAIOperationGenerateContent` in `internal/metrics/genai.go:27-34` because nothing has needed it before.

Two-line fix:

```go
// internal/metrics/genai.go
GenAIOperationGenerateContent GenAIOperation = "generate_content"
```

```go
// here:
metrics.NewMetricsFactory(meter, metricsRequestHeaderAttributes, metrics.GenAIOperationGenerateContent)
```

Without this, dashboards filtering on `gen_ai.operation.name == "chat"` will conflate Gemini-native traffic with OpenAI chat completions, and CEL request-cost rules keyed off `chat` will fire for both providers when the operator only intended one. People wiring up GenAI-semconv-aware receivers (OpenLLMetry, otel-collector connectors) will mis-bucket.

internal/translator/gemini_gcpvertexai.go
```diff
@@ -0,0 +1,159 @@
...
+		var resp gcp.GenerateContentResponse
+		if jsonErr := internaljson.Unmarshal(mutatedBody, &resp); jsonErr == nil && resp.UsageMetadata != nil {
+			applyUsageMetadata(resp.UsageMetadata, &tokenUsage)
+		}
+	}
+	return
+}
+
+// applyUsageMetadata populates a TokenUsage from a Gemini UsageMetadata response field.
+func applyUsageMetadata(u *gcp.GenerateContentUsageMetadata, usage *metrics.TokenUsage) {
```
JAORMX, line 112: **Blocker. Reasoning tokens silently dropped. Total-tokens invariant breaks.**

So, the struct in `internal/apischema/gcp/gcp.go:56-63` carries `ThoughtsTokenCount` and `ToolUsePromptTokenCount`, but `applyUsageMetadata` only reads four of the six fields. The two thinking-related counts never reach `metrics.TokenUsage`, never reach CEL request-cost evaluation, never reach `gen_ai.client.token.usage`.

This isn't just a missing call though. `metrics.TokenUsage` has no `SetReasoningTokens` method at all. `grep -rn "Reasoning\|Thinking\|Thoughts" internal/metrics/` returns zero hits. So this is a missing capability gateway-wide, not a one-line fix.

The shape of the problem, per the Vertex AI doc (https://docs.cloud.google.com/vertex-ai/generative-ai/docs/reference/rest/v1/GenerateContentResponse#UsageMetadata):

> `totalTokenCount` ... "The total number of tokens for the entire request. This is the sum of `promptTokenCount`, `candidatesTokenCount`, `toolUsePromptTokenCount`, and `thoughtsTokenCount`."

So `usage.SetTotalTokens(u.TotalTokenCount)` on line 115 records a total that exceeds `input + output + cachedInput` for any thinking-enabled request. CEL rules of `output_token` type under-bill by exactly the reasoning count. Dashboards summing `input + output` to cross-check `total` will get a delta they can't explain.

Minimum fix for this PR: fold thoughts into output for parity with the OpenAI shim that already does this in the response body at `internal/translator/gemini_helper.go:967`:

```go
CompletionTokens: int(metadata.CandidatesTokenCount) + int(metadata.ThoughtsTokenCount),
```

So `applyUsageMetadata` becomes:

```go
usage.SetOutputTokens(u.CandidatesTokenCount + u.ThoughtsTokenCount)
```

Better long-term fix: add a real `SetReasoningTokens` setter on `metrics.TokenUsage` and surface a `gen_ai.token.type = "reasoning"` recording so CEL rules can target reasoning explicitly. That's a follow-up though... please at least do the fold for now and leave a `// TODO:` so it isn't lost. The Gemini-native path is the most likely place thinking-heavy traffic is going to land (CLI use, native SDK use), so it's the right surface to flag this on.

internal/apischema/gcp/gcp.go
```diff
@@ -41,6 +41,27 @@ type GenerateContentRequest struct {
 	SafetySettings []*genai.SafetySetting `json:"safetySettings,omitempty"`
 }
 
+// GenerateContentResponse is a minimal representation of the Gemini generateContent response,
+// used only to extract usageMetadata for token accounting. The full response body is passed
+// through to the client unchanged.
+//
+// https://docs.cloud.google.com/vertex-ai/generative-ai/docs/reference/rest/v1/GenerateContentResponse
+type GenerateContentResponse struct {
```
JAORMX, line 49: **Blocker. The request struct quietly drops three valid Vertex AI fields when re-marshaled.**

(Anchoring close to the new `GenerateContentResponse` here. The concern is the `GenerateContentRequest` struct above, lines 14-42.)

It carries `Contents`, `Tools`, `ToolConfig`, `GenerationConfig`, `SystemInstruction`, `SafetySettings`. Per the Vertex AI streamGenerateContent reference (https://docs.cloud.google.com/vertex-ai/generative-ai/docs/reference/rest/v1/projects.locations.publishers.models/streamGenerateContent), the request body also accepts `cachedContent`, `labels`, and `modelArmorConfig`. Those don't exist on the struct.

Where it bites: when `FunctionResponse.id` triggers `json.Marshal(req)` at `internal/translator/gemini_gcpvertexai.go:59`, those three fields silently disappear from the upstream request body. The data loss is conditional on tool-call traffic... which is exactly the workload most likely to use context caching. Pretty nasty inconsistency to debug.

Two fixes:

1. **Cleanest**: drop the round-trip. Use `sjson.DeleteBytes` on the raw body to strip `FunctionResponse.id`, the same way `internal/translator/anthropic_gcpanthropic.go:54-56` does it. This also fixes the retry bug on the same code path (separate comment on `gemini_gcpvertexai.go:46`).
2. Extend the struct with `CachedContent`, `Labels`, `ModelArmorConfig`. But you'll keep paying this tax every time Vertex AI ships a new request field, and there's no `Extra map[string]json.RawMessage` catch-all to absorb future additions.

Side note on the same struct: `Tools []genai.Tool \`json:"tools"\`` on line 22 is missing `omitempty`, so the round-trip flips an absent `tools` into `"tools":null` in the upstream request. Vertex AI tolerates it, but it's a behavioral diff from what the client sent.

site/docs/capabilities/llm-integrations/supported-endpoints.md
```diff
@@ -277,6 +277,80 @@ curl -H "Content-Type: application/json" \
...
+
+**Supported Providers:**
+
+- Google Vertex AI (with automatic path rewriting)
+
+**Example:**
+
+```bash
+curl -s -H "Content-Type: application/json" \
+  -H "x-goog-api-key: $X_GOOG_API_KEY" \
```
JAORMX, line 305: **Blocker. The Gemini API key is being logged in clear text.**

So, the docs here tell users to send `-H "x-goog-api-key: $X_GOOG_API_KEY"`. That's the right Gemini convention. But that header isn't in the gateway's redaction list at `internal/extproc/server.go:36`:

```go
sensitiveHeaderKeys = []string{"authorization", "x-api-key"}
```

The list is consulted by `filterSensitiveHeadersForLogging` and `redactHeaderMutation` (used at `server.go` lines 278, 318, 340, 376, 497, 578, 605). The match is exact lowercase. So `x-goog-api-key` won't match, and under `--logLevel=debug` it lands in the log line verbatim.

One-line fix:

```go
sensitiveHeaderKeys = []string{"authorization", "x-api-key", "x-goog-api-key"}
```

Worth considering also adding `x-goog-user-project` and `x-goog-iam-authorization-token` if you expect those Google auth flows to land on this endpoint. The new endpoint is the very thing causing this header to start arriving in the gateway's request path, so the fix lives with this PR.

internal/endpointspec/gemini_generatecontent.go
```diff
@@ -0,0 +1,76 @@
...
+	switch schema.Name {
+	case filterapi.APISchemaGCPVertexAI:
+		return translator.NewGeminiToGCPVertexAITranslator(effectiveModel, s.Streaming), nil
+	default:
+		return nil, fmt.Errorf("Gemini native endpoint only supports GCPVertexAI backend, got: %s", schema.Name)
+	}
+}
+
+// RedactSensitiveInfoFromRequest implements [Spec.RedactSensitiveInfoFromRequest].
+func (GenerateContentEndpointSpec) RedactSensitiveInfoFromRequest(
```
JAORMX, line 66: **Blocker. Redaction is a placeholder, not redaction.**

```go
func (GenerateContentEndpointSpec) RedactSensitiveInfoFromRequest(
    req *gcp.GenerateContentRequest,
) (*gcp.GenerateContentRequest, error) {
    // Placeholder — contents and systemInstruction could be redacted in a future pass.
    return req, nil
}
```

The flow this lands in is at `internal/extproc/processor_impl.go:233-241`:

```go
if r.debugLogEnabled && r.enableRedaction {
    if redactedBody, err := r.eh.RedactSensitiveInfoFromRequest(body); err != nil {
        ...
    } else {
        if jsonBody, err := json.Marshal(redactedBody); ...
        logger.Debug("request body processing", slog.Any("request", string(jsonBody)))
```

So when an operator runs with `--enableRedaction --logLevel=debug` (the contract being "PII won't end up in logs"), Gemini bodies get JSON-marshaled and logged unredacted. `Contents[].Parts[].Text`, `SystemInstruction.Parts[].Text`, `FunctionResponse.Response`, inline blob data, all of it.

`internal/endpointspec/endpointspec_test.go:1186-1205` even pins the placeholder behavior in a test ("Placeholder: content is NOT redacted in current implementation"). So the no-op is intentional, but the user-facing semantics are surprising: I turned redaction on and got no redaction.

`ChatCompletionsEndpointSpec` (`internal/endpointspec/endpointspec.go:148-200`) and `MessagesEndpointSpec` actually redact. Mirror that pattern. Walk `Contents[].Parts[]` and `SystemInstruction.Parts[]`, replace each `Text` (and `InlineData`) with `[REDACTED LENGTH=n HASH=...]` like the Anthropic implementation does.

If a real redactor is too big for this PR, at minimum return an error from this method so the caller skips the log line. Don't ship a "Fully Supported" badge on the docs page with a placeholder redactor... that violates the redaction contract silently for one endpoint only. Worst kind of partial protection.

internal/extproc/processor_impl.go
```diff
@@ -612,6 +617,50 @@ func (u *upstreamProcessor[ReqT, RespT, RespChunkT, EndpointSpecT]) SetBackend(c
 	return
 }
 
+// NewGeminiProcessorFactory returns a ProcessorFactory for the Gemini native
+// generateContent / streamGenerateContent endpoints. The model name and
+// streaming flag are extracted from the URL path rather than the request body.
+func NewGeminiProcessorFactory(f metrics.Factory) ProcessorFactory {
+	noopTracer := tracingapi.NoopTracer[gcp.GenerateContentRequest, struct{}, struct{}]{}
```
JAORMX, line 624: **Blocker. NoopTracer disables OTel tracing for the entire Gemini surface.**

The justifying comment in `internal/endpointspec/gemini_generatecontent.go:73-76` says "passthrough translator does not parse response structure deeply enough to emit meaningful spans." But that's not quite accurate. The translator already calls `io.ReadAll` on response bodies and parses `usageMetadata` (`internal/translator/gemini_gcpvertexai.go:103-106`). Model name, token counts, finish reason, all reachable.

Anthropic Messages is also a passthrough surface and wires a real tracer at `cmd/extproc/mainlib/main.go:310-311`:

```go
server.Register(path.Join(flags.rootPrefix, endpointPrefixes.Anthropic, "/v1/messages"), extproc.NewFactory(
    messagesMetricsFactory, tracing.MessageTracer(), endpointspec.MessagesEndpointSpec{}))
```

Its recorder at `internal/tracing/openinference/anthropic/messages.go:58-93` emits OpenInference spans with model + usage + finish reason. A minimal `GenerateContentRecorder` matching that shape is the right move.

The dead `GenerateContentTracer` type alias at `gemini_generatecontent.go:73-76` is the natural plumbing point... `grep -rn GenerateContentTracer` confirms it's only declared, never referenced. Wire it through `tracing.GenerateContentTracer()` analogous to `tracing.MessageTracer()`.

Concrete user impact: anyone running Phoenix, Langfuse, or a generic OTel collector against the gateway will see spans for OpenAI, Anthropic, Cohere, and Bedrock... and a hole where Gemini should be. No `gen_ai.request.model` span attribute, no `gen_ai.usage.input_tokens`, no token-latency annotations.

If you'd rather ship the recorder as a follow-up, that's reasonable, but please drop the "Fully Supported" badge from `site/docs/capabilities/llm-integrations/supported-endpoints.md` until it lands. Right now the docs promise more than the code delivers.

internal/extproc/processor_impl.go
```diff
@@ -612,6 +617,50 @@ func (u *upstreamProcessor[ReqT, RespT, RespChunkT, EndpointSpecT]) SetBackend(c
...
+
+// extractGeminiModelFromPath parses /…/models/{model}:(generateContent|streamGenerateContent)
+// and returns the model name and whether the request is a streaming request.
+func extractGeminiModelFromPath(path string) (model string, streaming bool) {
+	const modelsSegment = "/models/"
+	if i := strings.LastIndex(path, modelsSegment); i != -1 {
+		rest := path[i+len(modelsSegment):]
+		if j := strings.LastIndex(rest, ":"); j != -1 {
+			model = rest[:j]
+			streaming = rest[j+1:] == "streamGenerateContent"
```
JAORMX, line 658: **Major. `:countTokens`, `:embedContent`, `:batchEmbedContents`, all silently rewritten to `:generateContent` upstream.**

```go
streaming = rest[j+1:] == "streamGenerateContent"
```

Anything that isn't `streamGenerateContent` falls through with `streaming=false`, and the translator at `internal/translator/gemini_gcpvertexai.go:64-69` then unconditionally rewrites `:path` to `:generateContent`. So a client request to `:countTokens`, `:embedContent`, `:batchEmbedContents`, `:cachedContents`, all real Gemini API verbs, matches the registered `/v1beta/models/` prefix, gets the model parsed out, and silently lands at `:generateContent` upstream. Wrong-shape response, no 4xx.

The "wrong action suffix" test at `processor_impl_test.go:1442-1446` even pins this:

```go
{
    name:      "wrong action suffix",
    path:      "/v1beta/models/gemini-flash:wrongAction",
    model:     "gemini-flash",
    streaming: false,
},
```

Reject unknown suffixes. Either return `errNoProcessor` from the factory so Envoy returns 404, or signal back to the spec so it returns an `internalapi.ErrMalformedRequest` with a 400. Both are honest answers. Silent re-routing isn't.

Concrete suggestion: change the function to return an explicit action enum (or a third bool/error) so the factory can decline up front:

```go
func extractGeminiModelFromPath(path string) (model string, action string, ok bool) {
    // ... return ok=false for unknown actions
}
```

Then in the factory, `if !ok { return nil, errNoProcessor }`.

internal/internalapi/internalapi.go
```diff
@@ -220,8 +225,10 @@ func ParseEndpointPrefixes(s string) (EndpointPrefixes, error) {
 			out.Cohere = value
 		case "anthropic":
 			out.Anthropic = value
+		case "gemini":
```
JAORMX, line 228: **Major. `endpointPrefixes=gemini:` collapses the registered prefix to `/models/`.**

The docstring on `ParseEndpointPrefixes` (line 194) promises *"Unknown keys cause an error; values must be non-empty."* But the implementation only rejects empty *pairs*, not empty *values*. `internal/internalapi/internalapi_test.go:38-42` (`TestParseEndpointPrefixes_EmptyValue`) actually pins the lenient behavior:

```go
res, err := ParseEndpointPrefixes("openai:")
require.NoError(t, err)
require.Equal(t, "", res.OpenAI)
```

Same code path runs for `gemini:`. So a user setting `--endpointPrefixes=gemini:` ends up with `EndpointPrefixes.Gemini == ""`. Then in `cmd/extproc/mainlib/main.go:316`:

```go
geminiModelPrefix := strings.TrimRight(path.Join("", ""), "/") + "/models/"
                  // => "/models/"
```

And the new prefix-match dispatch in `server.go:127-141` then matches every request whose path starts with `/models/`. Same outcome for `gemini:/`.

Why this matters now and not before: Gemini is the first prefix-match registration in the gateway. Up until this PR every `Register` call has been exact-match, so empty values were a no-op. Not anymore.

Two-part fix:

1. Reject empty values in `ParseEndpointPrefixes`. Just enforce what the docstring already promises:
   ```go
   if value == "" {
       return EndpointPrefixes{}, fmt.Errorf("empty value for key %q", key)
   }
   ```
2. In `mainlib/main.go:316`, defensively guard against `Gemini == ""` or `Gemini == "/"` and either skip the registration with a warning or fail-fast at startup. Belt and suspenders.

While you're in `internal/internalapi/internalapi_test.go`, please also extend `TestParseEndpointPrefixes_Success` with a `gemini:/x` case and assert the new `/v1beta` default in `TestParseEndpointPrefixes_EmptyInput`. Right now neither covers the new key.

internal/translator/gemini_gcpvertexai.go
```diff
@@ -0,0 +1,159 @@
...
+	newHeaders []internalapi.Header, mutatedBody []byte, tokenUsage metrics.TokenUsage, responseModel string, err error,
+) {
+	mutatedBody, err = io.ReadAll(body)
+	if err != nil {
+		return nil, nil, metrics.TokenUsage{}, "", err
+	}
+	newHeaders = []internalapi.Header{
+		{contentLengthHeaderName, strconv.Itoa(len(mutatedBody))},
+	}
+	responseModel = g.model
```
JAORMX, line 96: **Major. `responseModel` always echoes the request model. `modelVersion` thrown away.**

```go
responseModel = g.model
```

`gen_ai.response.model` will never differ from `gen_ai.request.model` even when GCP transparently routes to a versioned/regional variant. The OpenAI-shim Vertex translator already does the right thing for the same backend at `internal/translator/openai_gcpvertexai.go:155-161`:

```go
responseModel = o.requestModel
if gcpResp.ModelVersion != "" {
    responseModel = gcpResp.ModelVersion
}
```

Five-line fix. Add `ModelVersion string \`json:"modelVersion,omitempty"\`` to `GenerateContentResponse` in `internal/apischema/gcp/gcp.go`, read it in `applyUsageMetadata` (or alongside it), and conditionally override `responseModel`. For streaming, the SSE final chunk carries it the same way `usageMetadata` does, so the existing parser can pick it up with one extra read.

internal/extproc/server.go
```diff
@@ -93,6 +100,14 @@ func (s *Server) Register(path string, newProcessor ProcessorFactory) {
 	s.processorFactories[path] = newProcessor
 }
 
+// RegisterPrefix registers a ProcessorFactory for all paths that have the given prefix.
+// When processorForPath cannot find an exact match, it performs a longest-prefix match
+// among all registered prefix factories.
+func (s *Server) RegisterPrefix(prefix string, factory ProcessorFactory) {
```
JAORMX, line 106: **Minor cluster on `Server`/dispatch.**

A few things to fix while you're in this file:

1. `RegisterPrefix` accepts duplicate prefixes silently. A second call to `RegisterPrefix("/v1beta/models/", ...)` would just append, and the longest-prefix loop in `processorForPath` uses strict `>` (`server.go:132`), so two equal-length prefixes resolve first-registration-wins with no warning. Compare with `Register` (line 100) which at least overwrites a map slot... a second call there would visibly stomp. The slice-append pattern hides the duplicate entirely. Suggest either logging a warning or returning an error on duplicate prefix.

2. Nothing enforces the trailing-slash invariant that `mainlib/main.go:314-315` already calls out as load-bearing. A future caller registering `/v1beta/models` (no trailing slash) would match `/v1beta/models-evil/...`. Either panic in this function if the prefix lacks `/`, or switch the matcher to a path-segment match. Preventive only, not currently exploitable.

3. Doc comment on `processorForPath` (line 113-114) still says *"Only exact path matching is supported currently."* Stale now... one-line update.

4. Inside `NewGeminiProcessorFactory` (`processor_impl.go:625-647`), the path is parsed and a `GenerateContentEndpointSpec` is constructed even when `isUpstreamFilter == true`, but the upstream branch never uses the spec... the upstream processor reads the spec via `u.parent.eh` set in `SetBackend`. Move the parse into the `if !isUpstreamFilter` branch, with a one-line comment in the `else` so the next maintainer doesn't think the upstream factory has its own spec.

internal/translator/gemini_gcpvertexai.go
```diff
@@ -0,0 +1,159 @@
+// Copyright Envoy AI Gateway Authors
+// SPDX-License-Identifier: Apache-2.0
+// The full text of the Apache license is available in the LICENSE file at
+// the root of the repo.
+
+package translator
+
+import (
+	"bytes"
+	"encoding/json"
```
JAORMX, line 10: **Minor cluster on the translator.**

A few small things in this file:

1. `encoding/json` (line 10) for Marshal vs `internaljson` (line 15) for Unmarshal in the same file. The rest of `internal/translator/*.go` uses the sonic-backed `internal/json` for both. `grep -ln '"encoding/json"' internal/translator/*.go | grep -v _test` returns only this file. Standardize to `internaljson.Marshal`.

2. `const dataPrefix = "data: "` (line 124) is space-strict. Per the WHATWG SSE spec the space is optional ("If value starts with U+0020 SPACE, remove it"), so a `data:{...}` line gets silently skipped. In practice Gemini emits the standard `data: ` form, but this is fragile. Either use `bytes.HasPrefix(line, []byte("data:"))` and strip at most one space, or use `bufio.Scanner` with `ScanLines` (which already handles `\r\n` correctly) and parse properly.

3. Comment on line 122 says *"Vertex AI includes usageMetadata in the final data event."* The code at line 137 actually takes-last-wins and works whether usageMetadata is on the final chunk only or on every chunk... which is the right behavior either way. Worth a short comment update so the next reader doesn't get confused.

4. Empty model produces `publishers/google/models/:generateContent` (when `effectiveModel==""`). The test at `gemini_gcpvertexai_test.go:215` even pins this. Better to fail fast in `GetTranslator` when `effectiveModel == ""` and return an error wrapped with `internalapi.ErrMalformedRequest` so the client gets a 400 instead of a guaranteed-bad upstream request.

5. Related: the error at `gemini_generatecontent.go:61` (`"Gemini native endpoint only supports GCPVertexAI backend, got: %s"`) isn't wrapped with `internalapi.ErrMalformedRequest`, so a backend-schema misconfig surfaces as a 5xx instead of a 4xx. Debuggability hit, since the misconfig actually lives in user config (`AIGatewayRoute`), not server state.

6. Test file is the only translator test in the repo with zero table-driven loops (`openai_gcpvertexai_test.go` has 8, `anthropic_gcpanthropic_test.go` has 6). Coverage gaps that table-driven would naturally fill: CRLF SSE, malformed-JSON event, multiple `usageMetadata` events in one stream, retry with `forceBodyMutation=true` (this would have caught the retry bug flagged on `gemini_gcpvertexai.go:46`), and redaction with non-empty content (the current test pins the no-op).

JAORMX (changes requested):
So, what this PR does is genuinely useful and we need it!

There are merge conflicts which need to be fixed. Will you have time to get back to this?

A few things to pull on before merge though, mostly production-correctness and privacy gotchas. Inline comments on each. Quick map:

**Blockers** (real bugs in production, not nits):

1. Retries with tool calls will 400 on every attempt. (`internal/translator/gemini_gcpvertexai.go:46`)
2. `gen_ai.operation.name` is `chat`, but the OTel GenAI semconv defines `generate_content` for exactly this surface. (`cmd/extproc/mainlib/main.go:313`)
3. Reasoning tokens (`thoughtsTokenCount`) parsed but never recorded. Total invariant breaks for thinking models. (`internal/translator/gemini_gcpvertexai.go:112`)
4. The struct silently drops `cachedContent`, `labels`, `modelArmorConfig` whenever the FunctionResponse strip triggers a re-marshal. (`internal/apischema/gcp/gcp.go`)
5. Privacy. `x-goog-api-key` lands in debug logs in clear text, and `RedactSensitiveInfoFromRequest` is a placeholder that lets prompts and system instructions through. (`site/docs/capabilities/llm-integrations/supported-endpoints.md:305` and `internal/endpointspec/gemini_generatecontent.go:65`)
6. `NoopTracer` disables OTel tracing for the entire Gemini surface. (`internal/extproc/processor_impl.go:624`)

**Majors:**

7. `:countTokens`, `:embedContent`, etc. silently rewritten to `:generateContent` upstream. (`internal/extproc/processor_impl.go:658`)
8. `endpointPrefixes=gemini:` collapses to `/models/`. (`internal/internalapi/internalapi.go:228`)
9. Helm chart not wired for the new prefix. The chart at `manifests/charts/ai-gateway-helm/templates/deployment.yaml:67-82` only iterates `openai`/`cohere`/`anthropic`, and `values.yaml:32-37` has no `gemini` key. So `endpointConfig.gemini` set in `values.yaml` is silently dropped. The default `/v1beta` works in the binary so this only bites operators trying to override the prefix, but please add the branch to the template, the key+comment to `values.yaml`, and a `gemini:/x` case to `internal/internalapi/internalapi_test.go`.
10. `responseModel` always echoes the request model, throwing away `modelVersion`. (`internal/translator/gemini_gcpvertexai.go:96`)

Plus a couple of cleanup clusters as inline notes on `server.go` and the translator file.

Happy to chat about any of these. Most are small, mechanical fixes. A couple (NoopTracer, the redaction story) are bigger lifts and could reasonably ship as follow-ups if they're tracked clearly and the docs status drops from "Fully Supported" until they land.

ChrisJBurns:
👋 @adirags @johnugeorge @mathetake @yuzisun — following on from the discussion in #2204.

This PR looks like the right place to centralize the native-Gemini-ingress work, and it sidesteps the concern raised on #2204 (*"adding translations for all permutations defeats the purpose of a common router"*). 

As written, #1961 is ingress + passthrough only (Gemini-native → GCP Vertex AI) - no cross-provider translation, no N×M matrix. It's the same shape as the Anthropic-messages ingress (#847): just the routing + endpoint-spec + Vertex passthrough foundation.

For context, we run a downstream build of ai-gateway with native Gemini ingress, so this is directly relevant to us and we'd be glad to help get it landed.

A framing that keeps it non-controversial and avoids the permutation explosion:

1. This PR - `RegisterPrefix` routing + `GenerateContentEndpointSpec` + Gemini→Vertex passthrough + token-usage extraction. (foundation; no translation)
2. Follow-up - Gemini→AI Studio passthrough. (still passthrough, no permutations)
3. Separate discussion, later -Gemini→OpenAI-IR translation for cross-provider routing. This is the only piece that introduces translation (and a lossy common-denominator IR boundary), so it deserves its own design review rather than being bundled here.

One thing worth flagging: `GenerateContentEndpointSpec.ParseBody` decodes the body into `gcp.GenerateContentRequest`, whose `genai.Part.ThoughtSignature` is `[]byte`. Gemini 2.5/3.x "thinking" models attach an opaque `thoughtSignature` to each reply that clients replay on every follow-up turn, and it isn't always standard base64. By our reading, a multi-turn request carrying such a signature fails body-parsing before reaching Vertex, i.e. multi-turn thinking conversations break. We ran into this on our downstream build and have a small fix that round-trips the signature byte-for-byte - happy to share it if useful.

@mathetake @yuzisun @nacx - is native Gemini ingress (this scoped, passthrough-first version) something you'd want for an upcoming release?

CodePrometheus:
I wanted to add some context from #2179.

What I’m trying to capture there is a clearer boundary for Google native traffic:

- `/gemini/*` for the Gemini Developer API / AI Studio API shape
- `/vertex-ai/*` for the Vertex AI REST shape
- `/v1/*` stays as the OpenAI-compatible normalized API

I agree with the goal of avoiding an N*M translation matrix. The direction in #2179 is provider-native passthrough lanes, not cross-provider translation routes.

In production, I’ve run into cases where schema translation loses information or subtly changes behavior, especially around provider-specific fields, streaming, tool/function-call details, and usage metadata. For those cases, preserving the native path, request shape, response shape, and error shape is important.

That is the boundary I’m trying to describe in #2179: keep `/v1/*` for OpenAI-compatible clients, and also allow provider-native passthrough lanes when clients need the original Google API semantics.

FYI @johnugeorge @yuzisun

PR #1969, opened 2026-03-17
feat: add OpenAI to AWS Bedrock embeddings translation

Flgado:
**Description**
This PR adds translation support for the OpenAI `/v1/embeddings`  endpoint targeting AWS Bedrock Titan Embed Text models (`amazon.titan-embed-text-v1` and `amazon.titan-embed-text-v2:0).

It follows the same patterns established by the existing openai_awsbedrock.go translator (JSON decoding strategy, error handling, path header construction via url.PathEscape, span recording).

Changes:
- **awsbedrock.go**: Added `TitanEmbeddingRequest` and `TitanEmbeddingResponse schema structs covering both Titan v1 and v2 fields (InputText, dimensions, normalize, embeddingTypes).
- **openai_awsbedrock_embeddings.go**:  New OpenAIEmbeddingTranslator implementation: translates EmbeddingRequest → TitanEmbeddingRequest, maps TitanEmbeddingResponse back to EmbeddingResponse, handles error responses (JSON and non-JSON), records token usage and span.
- **endpointspec.go** — Wired `NewEmbeddingOpenAIToAWSBedrockTranslator` into the APISchemaAWSBedrock embedding path.
- **openai_awsbedrock_embeddings_test.go**: 26 test cases across RequestBody, ResponseHeaders, ResponseBody, and ResponseError, covering Titan v1/v2 behavior, model override, batch rejection,  dimensions forwarding, and error paths.
- **supported-endpoints.md** — Updated embeddings provider list and compatibility table to reflect Titan support.

**What is NOT supported in this PR:**

- **Batch embeddings**: Titan InvokeModel accepts only a single inputText per request. The translator returns a 400 for len(input) != 1.
- **embeddingsByType (quantized outputs)**: Titan v2 can return quantized vectors under `embeddingsByType, but the OpenAI response schema has no equivalent. The top-level embedding float array is always present and used.
- Cohere Embed models (e.g. cohere.embed-english-v3): Cohere's embedding API uses texts (array) instead of inputText and requires a semantic `input_type field with no OpenAI equivalent. This needs its own translator, planned as a follow-up PR.

# Special notes/questions for reviewers
Two design decisions I'd like maintainer input on before considering follow-up work:

**1. Should normalize be user-configurable via vendor fields?**

Currently normalize (Titan v2 only) is always left unset, using Titan's default (true). The GCP embeddings translator exposes a similar per-provider option via `GCPVertexAIEmbeddingVendorFields.TaskType. Should we follow the same pattern and add an `AWSBedrockTitanEmbeddingVendorFields struct with normalize? Or is leaving it as the default acceptable. Same question for the embeddingsByType. 

**2. Should embeddingsByType be surfaced in the response?**

Titan v2 can return quantized binary vectors under embeddingsByType alongside the standard float embedding. This PR always discards it and returns only the float array. If the community wants to expose this, it would require either a vendor extension on EmbeddingResponse or a new field there is no OpenAI equivalent. Should this be in scope, and if so, what is the preferred approach?

Amazon Titan Embeddings docs for reference: https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-titan-embed-text.html


internal/apischema/awsbedrock/awsbedrock.go
```diff
@@ -602,3 +602,38 @@ type CachePointBlock struct {
...
+
+// TitanEmbeddingResponse is the response body returned by the Amazon Titan Embed Text models.
+//
+// See https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-titan-embed-text.html
+type TitanEmbeddingResponse struct {
+	// Embedding is the embedding vector for the input text.
+	Embedding []float64 `json:"embedding"`
+
+	// InputTextTokenCount is the number of tokens in the input text.
+	InputTextTokenCount int `json:"inputTextTokenCount"`
```
nacx, line 638: Should we add the `embeddingsByType` as well?

nacx (approved):
Just one small nit. LGTM!

internal/apischema/awsbedrock/awsbedrock.go
nacx: Nevermind. Let's keep it as-is because as you say in the description, there is no direct mapping. LGTM!

internal/translator/openai_awsbedrock_embeddings.go
```diff
@@ -0,0 +1,208 @@
...
+	contentType := respHeaders[contentTypeHeaderName]
+	awsErrorType := respHeaders[awsErrorTypeHeaderName]
+
+	rawBody, err := io.ReadAll(body)
+	if err != nil {
+		return nil, nil, fmt.Errorf("read error body: %w", err)
+	}
+
+	// Already OpenAI-style error — return nil headers to preserve original response headers.
+	if isJSON(contentType) && awsErrorType == "" {
```
gavrissh, line 145: Have we verified that if this header is not passed, it will always be OpenAI style?

Flgado: Hey, yes I verified this locally against Bedrock :)

`x-amzn-errortype` is how AWS marks Bedrock error responses, so when it’s present we translate; when it’s absent with JSON, we’re on the gateway / already-OpenAI error path and we pass the body through as-is (and return nil headers so we don’t drop other headers). That’s covered by the passthrough test.

But yeah I get youu If AWS ever didn’t follow their own contract (Bedrock JSON error but no `x-amzn-errortype`), we’d pass that body through unchanged, which could be wrong, we’re relying on the documented header behavior.

gavrissh: I think we should assume the documentation will hold true and go ahead. Thanks for verifying

sc7565:
Do we know, when this can be merged in main.. we wanted to try embedding on bedrock using AI Gateway

internal/apischema/awsbedrock/awsbedrock.go
Flgado: We can do a follow up PR too add it if it make sense :)

sc7565:
Seems, test case are failing, also once the build is success we can download latest from dockerhub and test them right?

Flgado:
@sc7565 let us know if you are happy with this 😃

sc7565:
> @sc7565 let us know if you are happy with this 😃

@Flgado 

was not able to test it as new tags including these changes are not present ... waiting for that https://hub.docker.com/r/envoyproxy/ai-gateway-controller/tags

we need this https://github.com/envoyproxy/ai-gateway/actions/runs/24831653585/job/72702417002 probably to pass

tests/data-plane/real_providers_test.go
```diff
@@ -121,6 +121,7 @@ func TestWithRealProviders(t *testing.T) {
 		t.Run("embeddings", func(t *testing.T) {
 			for _, tc := range []realProvidersTestCase{
 				{name: "openai", modelName: "text-embedding-3-small", required: internaltesting.RequiredCredentialOpenAI},
+				{name: "aws-bedrock", modelName: "amazon.titan-embed-text-v2:0", required: internaltesting.RequiredCredentialAWS},
```
Flgado, line 124: @nacx  the e2e are failing https://github.com/envoyproxy/ai-gateway/actions/runs/24831653585/job/72702417002

Do you know why ? Sorry I am not familiar with the pipeline, as I saw that we have alreay this e2e test for bedrock I was thinking this will work out of the box.
CC: @sc7565

Flgado: - https://github.com/envoyproxy/ai-gateway/pull/2081 I thinks this is the missing piece

PR #1973, opened 2026-03-19
docs: extract and validate k8s resources

herewasmike:
**Description**

Move embedded yaml from docs to separate files.
Validate them (and examples folder) with kubeconform

Not sure if we're fine with network calls to get schemas (from https://github.com/yannh/kubernetes-json-schema) for resources defined outside of this repo.
This is still wip and I want to get some feedback on the approach


herewasmike:
@missBerg what are your thoughts on this approach?

(I'll address migration to v1beta later)

PR #1977, opened 2026-03-20
fix: allow x-ai-eg-model to be overridden via headerMutation

### LINKED ISSUE #1872: Header value of `x-ai-eg-model` is not changed when headerMutation is set
Seems like `x-ai-eg-model` header value cannot be changed before sending to backend using header mutation.

This route should send `x-ai-eg-model` with `different-model`:

```yaml
kind: AIGatewayRoute
metadata:
  name: test-route
  namespace: some-ns
spec:
  parentRefs:
    kind: Gateway
    name: envoy-ai-gateway-internal
  rules:
  - backendRefs:
    - modelNameOverride: different-model
      name: test-backend
      priority: 0
      weight: 1
      headerMutation:
        set:
          - name: x-ai-eg-model
            value: different-model
    matches:
    - headers:
      - name: x-ai-eg-model
        type: Exact
        value: some-model
```

Instead server gets:

```yaml
x-ai-eg-model: some-model
x-ai-eg-original-path: /v1/chat/completions
 ```

kimjihoon3106:
Fixes #1872

**Description**

`shouldIgnoreHeader` was silently dropping all `x-ai-eg-*` headers from
`headerMutation.set`, including the user-facing `x-ai-eg-model` header.
This made it impossible to override the model name forwarded to the backend
via `headerMutation`, even though `x-ai-eg-model` is a documented routing
header that users are expected to control.

This commit adds an explicit exception for `ModelNameHeaderKeyDefault`
(`x-ai-eg-model`) in `shouldIgnoreHeader`, allowing it to be set via
`headerMutation` while keeping all other internal `x-ai-eg-*` headers protected.

Fixes #1872

Only `ModelNameHeaderKeyDefault` (`x-ai-eg-model`) is exempted from the
ignore list. All other `x-ai-eg-*` headers remain blocked. A regression
test and a guard test are included to cover both behaviors.

> Note: I identified this bug by reading through the codebase and traced it
> to `shouldIgnoreHeader` in `header_mutator.go`. I used Claude (AI) as a
> coding assistant during implementation. I fully understand and own all
> changes submitted in this PR.

### DIFF
--- a/internal/headermutator/header_mutator.go
+++ b/internal/headermutator/header_mutator.go
@@ -112,11 +112,20 @@ func (h *HeaderMutator) Mutate(headers map[string]string, onRetry bool) (sets []
 // This should be safe since these headers are managed by Envoy AI Gateway itself, not expected to be
 // modified by users via header mutation API.
 //
+// Exception: the model name header (x-ai-eg-model by default) is a user-facing header that
+// represents the effective model for a request. It can be legitimately overridden via headerMutation
+// to control what model name is forwarded to the backend.
+//
 // Also, skip Envoy pseudo-headers beginning with ':', such as ":method", ":path", etc.
 // This is important because these headers are not only sensitive to our implementation detail as well as
 // it can cause unexpected behavior if they are modified unexpectedly. User shouldn't need to
 // modify these headers via header mutation API.
 func shouldIgnoreHeader(key string) bool {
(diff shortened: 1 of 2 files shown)


mathetake:
let's understand the use case in the original issue #1872 before moving with the change

PR #1982, opened 2026-03-23
feat: openAI Files API Implementation

sivanantha321:
**Description**

This PR implements the OpenAI Files API [6] in the Envoy AI Gateway, adding support for five file operations. The Files API is a prerequisite for the Batch Processing API (proposal #007 [5]), as batch jobs reference files uploaded via this API.

### Implemented Endpoints

| Endpoint                  | Method   | Path                          | Description                                                                                                                                                                                                                                                                                                                                   |
|---------------------------|----------|-------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| **Upload file**           | `POST`   | `/v1/files`                   | Upload a file via `multipart/form-data`. `model` is required for routing/ID encoding. `backend` is optional and may be supplied to pin the upload to a specific backend (for example, in multi-backend setups where you want a non-primary backend). Upload routing inputs are read from multipart form fields (`model`, optional `backend`). |
| **List files**            | `GET`    | `/v1/files`                   | List files. Requires `backend` query parameter (e.g., `/v1/files?backend=default.openai-backend&limit=10`). `model` query parameter is rejected for this endpoint.                                                                                                                                                                            |
| **Retrieve file**         | `GET`    | `/v1/files/{file_id}`         | Fetch file metadata by ID                                                                                                                                                                                                                                                                                                                     |
| **Retrieve file content** | `GET`    | `/v1/files/{file_id}/content` | Download raw file content                                                                                                                                                                                                                                                                                                                     |
| **Delete file**           | `DELETE` | `/v1/files/{file_id}`         | Delete a file by ID                                                                                                                                                                                                                                                                                                                           |

---

## Sticky Backend Routing & File ID Encoding

A core challenge of the Files API in a gateway context is **routing stickiness**: once a file is uploaded to a specific backend (e.g., OpenAI, Azure), all subsequent operations on that file (retrieve, retrieve content, delete) **must** be routed to the same backend. Otherwise, the backend will return a "file not found" error.

This is solved with two cooperating mechanisms:

1. **Encoded file IDs** carry routing metadata (model + backend) back to the client transparently, so subsequent requests don't need extra headers. Inspired by LiteLLM's approach [1].
2. **Per-backend sticky HTTPRoutes** generated by the controller match on internal `x-ai-eg-backend` + `x-ai-eg-model` headers and pin a request to a specific backend, bypassing the regular weighted load-balancing rule.

For operations that do not naturally carry a model (list requests and raw file-id fallback), extproc injects an internal placeholder model header value solely for sticky-route matching. This placeholder is never surfaced as the original model and is not used for response encoding decisions.

Together, they make multi-backend file routing fully deterministic without requiring clients to opt in or carry routing state.

### Sticky HTTPRoute Rules (Controller)

For every unique `BackendRef` declared in an `AIGatewayRoute`, the controller (`internal/controller/ai_gateway_route.go`) emits an additional **per-backend sticky HTTPRoute** (instead of appending sticky rules to the main HTTPRoute):

- Targets a single backend with `weight = 1` (no load balancing across backends)
- Matches on two headers, both required:
  - `x-ai-eg-backend` = `<namespace>.<aiservicebackend-name>` (exact match)
  - `x-ai-eg-model` matches regex `.+` (any non-empty value)
- The two header constraints make it strictly more specific than the regular rule, so Gateway API selects it first when both headers are present
- The main HTTPRoute now keeps only regular spec rules (+ route-not-found), preserving its direct index alignment with `AIGatewayRoute.Spec.Rules`
- Sticky routes are split out to avoid hitting the 16-rule limit on a single HTTPRoute in multi-backend configurations
- Generated sticky routes are labeled with owner route identity and backend name for reverse mapping in the extension server

The upstream extproc filter writes `x-ai-eg-backend` on every request via `SetBackend`. The value is derived from `backend.Backend.Name` (which is in `PerRouteRuleRefBackendName` format `namespace/name/route/.../ref/...`) and normalized to `namespace.name` by a new helper `extractAIServiceBackendName`, so it matches the sticky rule's header value exactly.

### Extension Server: Resolving Sticky Routes Back to Spec Rules

With sticky routing moved into separate per-backend HTTPRoutes, the post-translate hook in `internal/extensionserver/post_translate_modify.go` now resolves sticky clusters by labels on the sticky route itself:

1. Detect sticky route names via `-sticky` suffix
2. Read sticky route labels to recover owning `AIGatewayRoute` (`owner`, `owner-namespace`) and sticky backend name
3. Load the owning `AIGatewayRoute` and find the original `(specRuleIdx, refIdx)` for that backend

The resolved `(specRuleIdx, refIdx)` is used to emit metadata via `PerRouteRuleRefBackendName(..., specRuleIdx, refIdx)` so the upstream extproc filter resolves the correct backend at request time.

### File ID Encoding Format

```
Original ID:  file-abc123
Encoded ID:   file-<base64url(aigw:file-abc123;model:gpt-4.1;backend:default.azure-openai)>
```

Key properties:

- **Preserves OpenAI-compatible prefixes** (`file-`, `batch_`) so the ID looks valid to clients and SDKs
- **`aigw:` payload prefix** identifies an AI Gateway encoded ID, consistent with the proposal
- **Transparent to clients** — clients use the encoded ID as-is; they don't need to know it contains routing info
- **Enables sticky routing** — on subsequent requests the gateway decodes the file ID, sets `x-ai-eg-model` and `x-ai-eg-backend` headers, and the sticky HTTPRoute rule pins the request to the original backend
- **Backend value uses the namespace-qualified `namespace.name` form** so the encoded value matches the header value written by the upstream extproc filter and the sticky rule's header match

### Routing Flow — Example API Usage

**Step 1: Upload a file** — `model` is required; `backend` is optional. Upload routing inputs are provided as multipart form fields:

```bash
# backend is optional; include it to pin upload to a specific backend
curl -X POST https://gateway.example.com/v1/files \
  -H "Authorization: Bearer $API_KEY" \
  -F "purpose=fine-tune" \
  -F "model=gpt-4.1" \
  -F "backend=default.openai-backend" \  # Optional, useful if user want to control which backend should be selected
  -F "file=@training_data.jsonl"
```

The gateway routes to the backend mapped to `gpt-4.1`, and on the **response**, encodes the file ID:

If provider fallback is configured and the primary backend fails, upload can automatically succeed on a fallback backend. In that case, the returned encoded file ID embeds the fallback backend that handled the successful upload.

```json
{
  "id": "file-YWlndzpmaWxlLWFiYzEyMzttb2RlbDpncHQtNC4xO2JhY2tlbmQ6YXp1cmUtb3BlbmFp",
  "object": "file",
  "purpose": "fine-tune",
  "filename": "training_data.jsonl"
}
```

**Step 2: Retrieve the file** — Client uses the encoded ID as-is:

```bash
curl https://gateway.example.com/v1/files/file-YWlndzpmaWxlLWFiYzEyMzttb2RlbDpncHQtNC4xO2JhY2tlbmQ6YXp1cmUtb3BlbmFp \
  -H "Authorization: Bearer $API_KEY"
```

The gateway:

1. Decodes the file ID → extracts routing fields (`model`, `backend`) and the `original ID`
2. Sets internal headers `x-ai-eg-model` and `x-ai-eg-backend`, which match the **sticky per-backend HTTPRoute** for that backend, pinning the request to the same backend that handled the upload
3. Sends `GET /v1/files/file-abc123` to the upstream (using the decoded original ID via `OriginalFileIDHeaderKey` / `DecodedFileIDHeaderKey`)
4. Re-encodes the file ID in the response before returning to the client

**Step 3: Retrieve file content:**

```bash
curl https://gateway.example.com/v1/files/file-YWlndzpmaWxlLWFiYzEyMzttb2RlbDpncHQtNC4xO2JhY2tlbmQ6YXp1cmUtb3BlbmFp/content \
  -H "Authorization: Bearer $API_KEY"
```

**Step 4: Delete the file:**

```bash
curl -X DELETE https://gateway.example.com/v1/files/file-YWlndzpmaWxlLWFiYzEyMzttb2RlbDpncHQtNC4xO2JhY2tlbmQ6YXp1cmUtb3BlbmFp \
  -H "Authorization: Bearer $API_KEY"
```

**Step 5: List files** — Backend is required via query parameter:

```bash
# backend via query parameter
curl "https://gateway.example.com/v1/files?backend=default.openai-backend&limit=10" \
  -H "Authorization: Bearer $API_KEY"
```

For list requests, the gateway validates routing context (`backend` required, `model` query param rejected), injects internal sticky-routing headers, and forwards the upstream list response as-is.

Decode/re-encode behavior applies to ID-based operations (upload/retrieve/retrieve-content/delete), not list responses.
In particular, list `data[].id` values are raw upstream file IDs and are not gateway-encoded.

### Implementation Details

The encoding/decoding is implemented in `internal/translator/util.go`:

- **`EncodeIDWithRouting(id, modelName, backendName, idType)`** — encodes file IDs with both model and backend routing metadata. Replaces the previous `EncodeIDWithModel`; the old name and the model-only payload format have been removed (no backward-compatible reader).
- **`DecodeIDWithRouting(encodedID)`** — returns `(modelName, backendName, originalID, err)`. Requires the new `aigw:` payload prefix.

The processor flow (`internal/extproc/processor_impl.go`):

- `processFileIDInPath` decodes the file ID and writes four internal headers used downstream: `OriginalFileIDHeaderKey`, `DecodedFileIDHeaderKey`, `ModelNameHeaderKeyDefault`, plus the new **`BackendNameHeaderKey` (`x-ai-eg-backend`)** when present in the encoded ID. The latter is what triggers sticky per-backend route matching.
- `SetBackend` (upstream processor) writes `x-ai-eg-backend` for every request using `extractAIServiceBackendName(backend.Backend.Name)`, normalizing the `PerRouteRuleRefBackendName` format to `namespace.name`.
- `ProcessResponseHeaders` copies `x-ai-eg-backend` from the request headers onto the response headers so the response-body translator can read the backend name when re-encoding file IDs.
- The response translator (`internal/translator/openai_file.go`) reads `x-ai-eg-backend` from response headers and embeds it into the encoded file ID returned to the client, completing the round-trip.

For `GET /v1/files` (list), there is no file ID to decode. The request requires a `backend` query parameter and rejects `model` query param usage. Extproc injects a placeholder `x-ai-eg-model` value only to satisfy sticky-route matching (`backend + model`); the list response body is not mutated.

---

## Tracing: Non-Standard OpenInference Approach

The OpenInference specification [2] does **not** define semantic conventions for file operations. OpenInference has well-defined span types for LLM/chat completions [3], embeddings [4], etc., but file upload/retrieve/delete are not part of the specification.

This PR implements tracing for file operations using a **non-standard adaptation** of the OpenInference conventions:

| Recorder                      | Span Name             | Request Type           | Response Type          | Notes                                 |
|-------------------------------|-----------------------|------------------------|------------------------|---------------------------------------|
| `CreateFileRecorder`          | `CreateFile`          | `openai.FileNewParams` | `openai.FileObject`    | Records `output.file_id` on success   |
| `RetrieveFileRecorder`        | `RetrieveFile`        | `struct{}` (no body)   | `openai.FileObject`    | Records `output.file_id` on success   |
| `RetrieveFileContentRecorder` | `RetrieveFileContent` | `struct{}` (no body)   | `struct{}` (raw bytes) | No output attributes (binary content) |
| `DeleteFileRecorder`          | `DeleteFile`          | `struct{}` (no body)   | `openai.FileDeleted`   | Records `output.file_id` on success   |

**What's non-standard:**

- All four recorders set `openinference.span.kind = "LLM"` and `llm.system = "openai"` — borrowing from existing OpenInference conventions even though file operations are not LLM inference calls. This is because OpenInference has no dedicated span kind for file/storage operations.
- Span kind is set to `trace.SpanKindInternal` (consistent with other OpenInference spans in the codebase).
- Request attributes are minimal — only `openinference.span.kind` and `llm.system` are set, since file operations don't have model/prompt/token semantics.
- Response attributes use custom keys like `output.file_id` and `output.mime_type` which are not part of the OpenInference spec.
- `RetrieveFileContent` records no output attributes at all, since the response is raw binary file data.
- All recorders embed `NoopChunkRecorder` since file operations are never streamed.

## Related Issues/PRs (if applicable)

Related: Proposal #007 — Batch Processing API Support [5]

## Special notes for reviewers (if applicable)

- The `ParseBody` interface change (added `requestHeaders` parameter) touches all existing endpoint spec implementations — each gains an unused `_ map[string]string` parameter to satisfy the interface.
- The server routing refactor from exact-match `map[string]ProcessorFactory` to a `[]Route` with regex + http method matching is a significant architectural change that affects all existing route registrations.
- Sticky routing is now implemented as separate per-backend HTTPRoutes (`<route>-<backend>-sticky`) with labels identifying owner route and backend. Logic that needs the originating `AIGatewayRoute`/backend should use these labels.
- The encoded file ID payload format changed from `id:<id>;model:<model>` to `aigw:<id>;model:<model>;backend:<backend>`. Old encoded IDs minted by previous builds will not decode — acceptable because the Files API has not been released.
- The backend name embedded in encoded file IDs and matched by sticky rules is the **namespace-qualified `AIServiceBackend` name (`namespace.name`)**, not the underlying `Backend` resource name. This is intentional so the value is stable and resolvable from `AIGatewayRoute` declarations.
- Metrics for file API endpoints currently use `NoopMetricsFactory` — a TODO is noted for adding dedicated metrics support.
- The OpenInference tracing approach for file operations is non-standard and may need to be revisited if/when the OpenInference spec adds file operation conventions.
- Routing for `POST /v1/files` currently reads multipart form fields only: `model` (required) and `backend` (optional).
- `GET /v1/files` (list) requires `backend` as query parameter and rejects `model` query parameter for deterministic backend selection and to avoid file-id/model mismatch on multi-model backends.

1: https://github.com/envoyproxy/ai-gateway/blob/main/docs/proposals/007-batch-processing/proposal.md#introduce-fileid-and-batchid-encoding
2: https://github.com/Arize-ai/openinference/tree/main/spec
3: https://github.com/Arize-ai/openinference/blob/main/spec/semantic_conventions.md
4: https://github.com/Arize-ai/openinference/blob/main/spec/embedding_spans.md
5: https://github.com/envoyproxy/ai-gateway/blob/main/docs/proposals/007-batch-processing/proposal.md
6: https://platform.openai.com/docs/api-reference/files


internal/apischema/openai/openai.go
```diff
@@ -8381,3 +8385,242 @@ const (
...
+	// - `vision`: Images used for vision fine-tuning
+	// - `user_data`: Flexible file type for any purpose
+	// - `evals`: Used for eval data sets
+	//
+	// Any of "assistants", "batch", "fine-tune", "vision", "user_data", "evals".
+	Purpose FilePurpose `json:"purpose,omitzero"`
+	// The expiration policy for a file.
+	ExpiresAfter FileNewParamsExpiresAfter `json:"expires_after,omitzero"`
+	// Used for providing extra parameters like model that is not part of the standard OpenAI Files API.
+	ExtraBody map[string]any `json:"extra_body,omitzero"`
```
xiaolin593, line 8602: Is this specifically to gateway? worth calling out in comment

internal/apischema/openai/openai.go
```diff
@@ -8381,3 +8385,242 @@ const (
...
+}
+
+func (f FileNewParams) MarshalMultipart() ([]byte, string, error) {
+	buf := &bytes.Buffer{}
+	writer := multipart.NewWriter(buf)
+
+	if f.File == nil {
+		return nil, "", fmt.Errorf("file is required")
+	}
+	filename := "anonymous_file"
```
xiaolin593, line 8686: do we want annoymous_file name? should we reject when filename is not provided by user?

internal/apischema/openai/openai.go
```diff
@@ -8381,3 +8385,242 @@ const (
...
+		return nil, "", err
+	}
+
+	if f.ExpiresAfter.Anchor != "" {
+		if err := writer.WriteField("expires_after.anchor", string(f.ExpiresAfter.Anchor)); err != nil {
+			return nil, "", err
+		}
+	}
+
+	if f.ExpiresAfter.Seconds != 0 {
```
xiaolin593, line 8725: may need a upper bound(30 days) check for expiresAfter.Seconds per 
```
The number of seconds after the anchor time that the file will expire. Must be between 3600 (1 hour) and 2592000 (30 days).

minimum3600
maximum2592000
```

internal/apischema/openai/openai.go
```diff
@@ -8381,3 +8385,242 @@ const (
...
+//
+// - `assistants`: Used in the Assistants API
+// - `batch`: Used in the Batch API
+// - `fine-tune`: Used for fine-tuning
+// - `vision`: Images used for vision fine-tuning
+// - `user_data`: Flexible file type for any purpose
+// - `evals`: Used for eval data sets
+type FilePurpose string
+
+const (
```
xiaolin593, line 8749: can you add a link as reference because `purpose` may evolve

internal/apischema/openai/openai.go
```diff
@@ -8381,3 +8385,242 @@ const (
...
+	// Any of "assistants", "assistants_output", "batch", "batch_output", "fine-tune",
+	// "fine-tune-results", "vision", "user_data".
+	Purpose FileObjectPurpose `json:"purpose"`
+	// Deprecated. The current status of the file, which can be either `uploaded`,
+	// `processed`, or `error`.
+	//
+	// Any of "uploaded", "processed", "error".
+	//
+	// Deprecated: deprecated
+	Status FileObjectStatus `json:"status"`
```
xiaolin593, line 8795: `status` is deprecated, maybe we don't add it here at all?

internal/apischema/openai/openai.go
```diff
@@ -8381,3 +8385,242 @@ const (
...
+	// Deprecated.
+	//
+	// Deprecated: deprecated
+	StatusDetails string `json:"status_details"`
+}
+
+// The intended purpose of the file. Supported values are `assistants`,
+// `assistants_output`, `batch`, `batch_output`, `fine-tune`, `fine-tune-results`,
+// `vision`, and `user_data`.
+type FileObjectPurpose string
```
xiaolin593, line 8807: add a reference link https://github.com/openai/openai-go/blob/da6db3bace003b42c2465545f47401c1fb40c646/file.go#L217-L224?

internal/extproc/processor_impl.go
```diff
@@ -38,6 +39,12 @@ import (
 // This is configured at the startup of the extproc server.
 var LogRequestHeaderAttributes map[string]string
 
+// Precomputed path prefixes for file and batch operations to avoid string concatenation on every request.
+const (
+	filePathPrefix  = "/v1/files/" + translator.FileIDPrefix
+	batchPathPrefix = "/v1/batches/" + translator.BatchIDPrefix
```
xiaolin593: leave batch path out of this PR until when we get to `/v1/batches`?

internal/extproc/processor_impl.go
```diff
@@ -179,6 +186,99 @@ func (r *routerProcessor[ReqT, RespT, RespChunkT, EndpointSpecT]) ProcessRespons
...
+
+func extractFileIDFromPath(path string) (string, bool) {
+	if path == "" {
+		return "", false
+	}
+	// Remove query params
+	if queryIndex := strings.Index(path, "?"); queryIndex != -1 {
+		path = path[:queryIndex]
+	}
+	segments := []string{"/v1/files/", "/v1/batches/"}
```
xiaolin593: same as above comment to leave `/v1/batches` from this PR

internal/extproc/server.go
```diff
@@ -57,20 +58,26 @@ type Server struct {
...
+
 // NewServer creates a new external processor server.
 func NewServer(logger *slog.Logger, enableRedaction bool) (*Server, error) {
 	debugLogEnabled := logger.Enabled(context.Background(), slog.LevelDebug)
 	srv := &Server{
 		logger:                   logger,
 		debugLogEnabled:          debugLogEnabled,
 		enableRedaction:          enableRedaction,
-		processorFactories:       make(map[string]ProcessorFactory),
+		processorFactories:       make([]*RouteProcessorMapper, 0, 20),
```
xiaolin593, line 80: any particular reason to set capacity 20?

internal/translator/openai_file.go
```diff
@@ -0,0 +1,289 @@
...
+type openAIToOpenAITranslatorV1RetrieveFile struct {
+	modelNameOverride internalapi.ModelNameOverride
+	// The path prefix of the retrieve file endpoint without file id to be used for the request. It is prefixed with the OpenAI path prefix.
+	pathPrefix string
+	// requestFileID stores the file id from the request so we can echo it back in the response.
+	requestFileID string
+}
+
+// RequestBody implements [OpenAIRetrieveFileTranslator.RequestBody].
+func (o *openAIToOpenAITranslatorV1RetrieveFile) RequestBody(reqHeaders map[string]string, original []byte, _ *struct{}, forceBodyMutation bool) (
```
xiaolin593, line 211: do we need to decode id for subsequent request routing to right backend when retrieve file?

xiaolin593 (commented):
Thank you so much for working on this! Overall looks good to me. left some comments. @envoyproxy/ai-gateway-maintainers please have a look.

internal/apischema/openai/openai.go
sivanantha321: Yes. It is specifically for gateway.

sivanantha321: will update the comment.

internal/apischema/openai/openai.go
sivanantha321: The limit may change for different providers. So I haven't added the check here.

internal/apischema/openai/openai.go
sivanantha321: This behaviour is taken from official OpenAI SDK. https://github.com/openai/openai-go/blob/cbf83a69541f646ec84071a0040f3ca524c2238f/internal/apiform/encoder.go#L366

internal/apischema/openai/openai.go
sivanantha321: sure

internal/apischema/openai/openai.go
sivanantha321: sure

internal/translator/openai_file.go
sivanantha321: Yes

internal/extproc/server.go
sivanantha321: no particular reason. It is to avoid repeated reallocation as the default capacity is usually low.

internal/apischema/openai/openai.go
sivanantha321: done

internal/apischema/openai/openai.go
sivanantha321: done

internal/extproc/processor_impl.go
sivanantha321: removed

internal/extproc/processor_impl.go
sivanantha321: removed

internal/apischema/openai/openai.go
sivanantha321: done

internal/apischema/openai/openai.go
sivanantha321: I have added it for compatibility reasons. I can remove it if you want.

internal/extproc/processor_impl.go
```diff
@@ -179,6 +185,98 @@ func (r *routerProcessor[ReqT, RespT, RespChunkT, EndpointSpecT]) ProcessRespons
...
+	if logger == nil {
+		logger = r.logger
+	}
+
+	fileID, ok := extractFileIDFromPath(r.requestHeaders[":path"])
+	if !ok {
+		logger.Error("failed to extract file_id from path", slog.String("path", r.requestHeaders[":path"]))
+		return createUserFacingErrorResponse(400, "BadRequest", "failed to extract file id / batch id from path"), nil
+	}
+	modelName, decodedID, err := translator.DecodeFileID(fileID)
```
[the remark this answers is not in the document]
sivanantha321: The decoding only fails for input-validation/format errors from the file_id string itself. So 400 is the suitable status code for this case.

internal/translator/openai_file.go
```diff
@@ -0,0 +1,289 @@
...
+func (o *openAIToOpenAITranslatorV1CreateFile) ResponseBody(_ map[string]string, body io.Reader, _ bool, _ tracingapi.CreateFileSpan) (
+	newHeaders []internalapi.Header, newBody []byte, tokenUsage metrics.TokenUsage, responseModel internalapi.ResponseModel, err error,
+) {
+	var resp openai.FileObject
+	bodyBytes, err := io.ReadAll(body)
+	if err != nil {
+		return nil, nil, tokenUsage, "", fmt.Errorf("failed to read response body: %w", err)
+	}
+	// Encode the file ID with model routing information. This allows us to route subsequent requests with the file ID to the correct model backend.
+	newBody, err = sjson.SetBytes(bodyBytes, "id", EncodeIDWithModel(gjson.GetBytes(bodyBytes, "id").String(), o.requestModel, "file"))
```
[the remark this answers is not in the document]
sivanantha321: create/upload file endpoint returns a `FileObject` (small metadata: id, filename, bytes, purpose, timestamps, etc.), not file contents. So the memory allocation impact is negligible.

internal/apischema/openai/openai.go
```diff
@@ -8381,3 +8385,249 @@ const (
...
+			extraBodyBytes, err := io.ReadAll(part)
+			if err != nil {
+				return err
+			}
+			extraBodyVal := any(extraBodyBytes)
+			if f.ExtraBody == nil {
+				f.ExtraBody = map[string]any{part.FormName(): extraBodyVal}
+			} else {
+				f.ExtraBody[part.FormName()] = extraBodyVal
+			}
```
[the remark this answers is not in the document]
sivanantha321: Read size is limited to 1MB for all fields except the file content by utilizing limitreader.

internal/translator/util.go
```diff
@@ -72,3 +78,98 @@ func serializeOpenAIChatCompletionChunk(chunk *openai.ChatCompletionResponseChun
...
+//
+// Returns:
+//
+//	Encoded ID starting with appropriate prefix and containing routing information
+//
+// Examples:
+//
+//	EncodeIDWithRouting("file-abc123", "gpt-4o-mini", "azure-openai", "file")
+//	-> "file-YWlndzpmaWxlLWFiYzEyMzttb2RlbDpncHQtNG8tbWluaTtiYWNrZW5kOmF6dXJlLW9wZW5haQ"
+func EncodeIDWithRouting(id, modelName, backendName, _ string) string {
```
nacx: nit: rename these methods to `EncodeFileIDWithRouting` or similar to make it explicit that this is about file IDs.

nacx (commented):
Thanks for keeping this PR updated!

This has a lot of changes. Overall looks good, but I have a few questions:

* With this change, we'll be generating one extra rule per backend in each HTTPRoute. Is this correct? I am concerned about how this may affect the current behaviour, as I think there is a limit in the gateway API of 16 rules per HTTPRoute: https://github.com/kubernetes-sigs/gateway-api/blob/ec7d5a2f6ff132f2b2465aff81dfcba862738a9c/apis/v1/httproute_types.go#L127. With this change, it may be very easy to hit that limit, and it could break existing setups on an upgrade, and we can't allow that to happen. Related: https://github.com/envoyproxy/ai-gateway/issues/2120
* Can you also add the corresponding documentation pages? I know this could be done in another PR, but given this is already a big one, it would help a lot to have in the same PR the user guide and usage examples for users to use these APIs.

sivanantha321:
> With this change, we'll be generating one extra rule per backend in each HTTPRoute. Is this correct? I am concerned about how this may affect the current behaviour, as I think there is a limit in the gateway API of 16 rules per HTTPRoute: https://github.com/kubernetes-sigs/gateway-api/blob/ec7d5a2f6ff132f2b2465aff81dfcba862738a9c/apis/v1/httproute_types.go#L127. With this change, it may be very easy to hit that limit, and it could break existing setups on an upgrade, and we can't allow that to happen. Related: https://github.com/envoyproxy/ai-gateway/issues/2120

@nacx The implementation now uses per backend httproute for sticky rules. So it shouldn't constrain the existing backend limit.

internal/translator/util.go
sivanantha321: Done. It can be reused for batch Ids as well, that's why i didn't mention the file part.

sivanantha321:
> Can you also add the corresponding documentation pages?

Done.

internal/controller/ai_gateway_route.go
```diff
@@ -458,3 +512,179 @@ func buildPriorityAnnotation(rules []aigv1b1.AIGatewayRouteRule) string {
...
+						Name:  internalapi.BackendNameHeaderKey,
+						Value: qualifiedBackendName,
+					},
+					// The model name header matching is required to ensure that this rule
+					// is more specific than the regular rule and takes priority.
+					{
+						Name:  internalapi.ModelNameHeaderKeyDefault,
+						Type:  ptr.To(gwapiv1.HeaderMatchRegularExpression),
+						Value: ".+",
+					},
```
yuzisun, line 629: The user created AIGatewayRoute may not necessarily have these header matching rules, this default sticky route generation should be put under a config.

sivanantha321:
Closing in favour of #2324.

PR #1988, opened 2026-03-25
fix: quick fixes on claude thinking

hustxiayang:
**Description**
  - Add missing OfAdaptive thinking config handling: `getThinkingConfigParamUnion` only handled `OfEnabled` and `OfDisabled`, silently dropping `{"type": "adaptive"}` thinking configs.
  https://github.com/envoyproxy/ai-gateway/pull/1842

- Fix `reasoning_effort` model support: The `reasoning_effort` → `output_config.effort` mapping was incorrectly using             
  `outputConfigAvailable`, which includes models that don't support the effort parameter (e.g., Claude Sonnet 4.5, Claude Haiku  4.5). Added a new effortAvailable function to use a list as suggested by @aabchoo so that correctly limits effort support to Claude Opus 4.6, Claude Sonnet 4.6, and Claude Opus 4.5. 

Referece: https://platform.claude.com/docs/en/build-with-claude/effort
https://github.com/envoyproxy/ai-gateway/pull/1892

### DIFF
--- a/internal/translator/anthropic_helper.go
+++ b/internal/translator/anthropic_helper.go
@@ -576,15 +576,18 @@ func getThinkingConfigParamUnion(tu *openai.ThinkingUnion) *anthropic.ThinkingCo
 
 	result := &anthropic.ThinkingConfigParamUnion{}
 
-	if tu.OfEnabled != nil {
+	switch {
+	case tu.OfEnabled != nil:
 		result.OfEnabled = &anthropic.ThinkingConfigEnabledParam{
 			BudgetTokens: tu.OfEnabled.BudgetTokens,
 			Type:         constant.Enabled(tu.OfEnabled.Type),
 		}
-	} else if tu.OfDisabled != nil {
+	case tu.OfDisabled != nil:
 		result.OfDisabled = &anthropic.ThinkingConfigDisabledParam{
 			Type: constant.Disabled(tu.OfDisabled.Type),
 		}
+	case tu.OfAdaptive != nil:
+		result.OfAdaptive = &anthropic.ThinkingConfigAdaptiveParam{}
 	}
 
 	return result
@@ -597,6 +600,17 @@ func outputConfigAvailable(model internalapi.RequestModel) bool {
 	modelLower := strings.ToLower(model)
 	return strings.Contains(modelLower, "4-5") ||
 		strings.Contains(modelLower, "4-6")
+}
+
+// effortAvailable checks if the model supports the output_config.effort parameter.
+// The effort parameter is supported by Claude Opus 4.6, Claude Sonnet 4.6, and Claude Opus 4.5.
+// See: https://docs.anthropic.com/en/docs/build-with-claude/extended-thinking#adaptive-thinking
+func effortAvailable(model internalapi.RequestModel) bool {
+	modelLower := strings.ToLower(model)
+	if strings.Contains(modelLower, "4-6") {
+		return true
+	}
+	return strings.Contains(modelLower, "4-5") && strings.Contains(modelLower, "opus")
 }
 
 // mapReasoningEffortToOutputConfigEffort converts OpenAI reasoning effort levels to Anthropic output config effort levels.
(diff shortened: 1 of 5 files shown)


internal/translator/anthropic_helper.go
```diff
@@ -599,6 +602,17 @@ func outputConfigAvailable(model internalapi.RequestModel) bool {
 		strings.Contains(modelLower, "4-6")
 }
 
+// effortAvailable checks if the model supports the output_config.effort parameter.
+// The effort parameter is supported by Claude Opus 4.6, Claude Sonnet 4.6, and Claude Opus 4.5.
+// See: https://docs.anthropic.com/en/docs/build-with-claude/extended-thinking#adaptive-thinking
+func effortAvailable(model internalapi.RequestModel) bool {
+	modelLower := strings.ToLower(model)
+	if strings.Contains(modelLower, "4-6") {
```
aabchoo: This seem error prone. Could we instead have a clear spot to list all the supported model?

internal/translator/anthropic_helper.go
```diff
@@ -576,15 +576,18 @@ func getThinkingConfigParamUnion(tu *openai.ThinkingUnion) *anthropic.ThinkingCo
...
 			BudgetTokens: tu.OfEnabled.BudgetTokens,
 			Type:         constant.Enabled(tu.OfEnabled.Type),
 		}
-	} else if tu.OfDisabled != nil {
+	case tu.OfDisabled != nil:
 		result.OfDisabled = &anthropic.ThinkingConfigDisabledParam{
 			Type: constant.Disabled(tu.OfDisabled.Type),
 		}
+	case tu.OfAdaptive != nil:
+		result.OfAdaptive = &anthropic.ThinkingConfigAdaptiveParam{}
```
aabchoo: No type?

hustxiayang: thanks! updated to make it consistent.

internal/translator/anthropic_helper.go
hustxiayang: Both need to be manually updated but adding a list should be easier to detect the difference/changes. I did not add this because I did not want to assume too much on how users set the model ids. For example, we set `gcp.claude-4-5-haiku` internally  but the model id of gcp  is  actually `claude-haiku-4-5`. The name was reversed somehow. We did not how different users set them, thus it might not be easy to cover all cases with a list.

But, I would follow it up to make it easier to be detected, probably use a list for `modelNameOverride` instead because currently only bedrock and gcp are relevant?

hustxiayang: Updated to use a list, use `modelNameOverride` but also use `model` in the request body as backup.

internal/translator/anthropic_helper.go
```diff
@@ -672,7 +713,7 @@ func buildAnthropicParams(openAIReq *openai.ChatCompletionRequest, apiSchema str
 	}
 
 	// Map OpenAI reasoning_effort to Anthropic output_config.effort.
-	if openAIReq.ReasoningEffort != "" && outputConfigAvailable(openAIReq.Model) {
+	if openAIReq.ReasoningEffort != "" && effortAvailable(featureCheckModel) {
```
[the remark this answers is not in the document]
hustxiayang: The comment was not accurate because at that time, reasoning effort was not added. I updated the comment.

internal/translator/anthropic_helper.go
```diff
@@ -576,27 +576,62 @@ func getThinkingConfigParamUnion(tu *openai.ThinkingUnion) *anthropic.ThinkingCo
...
+// outputConfigModels lists model identifiers that support structured outputs (OutputConfig).
+// Structured outputs are available on Claude Opus 4.6, Claude Sonnet 4.6, Claude Sonnet 4.5, Claude Opus 4.5, and Claude Haiku 4.5.
 // See: https://platform.claude.com/docs/en/build-with-claude/structured-outputs
+var outputConfigModels = []string{
+	"opus-4-5",   // Claude Opus 4.5
+	"sonnet-4-5", // Claude Sonnet 4.5
+	"haiku-4-5",  // Claude Haiku 4.5
+	"opus-4-6",   // Claude Opus 4.6
+	"sonnet-4-6", // Claude Sonnet 4.6
+}
```
nacx, line 618: nit: I would move the two hardcoded lists tot he top of the file so they're easy to find when it comes to update them.

nacx (approved):
Thanks! LGTM

PR #1994, opened 2026-03-29
cli: handle standalone Envoy startup failures

nacx:
**Description**

In standalone mode, handle EG infrastructure errors and exit gracefully instead of hanging.

Related to: https://github.com/envoyproxy/ai-gateway/issues/1304

N/A


cmd/aigw/run_test.go
```diff
@@ -288,6 +288,68 @@ func Test_newEnvoyMiddleware(t *testing.T) {
...
+			expectInStderr: true,
+		},
+		{
+			name:           "infrastructure runner triggers graceful shutdown",
+			runner:         "infrastructure",
+			expectCancel:   true,
+			expectInStderr: true,
+		},
+		{
+			name:         "gateway-api runner does not trigger shutdown",
```
aabchoo, line 311: Could you explain this? Doesn't infrastructure runner depend on Gateway API runner?

nacx: No, the gw api runner is the one that processes resources and config, and the infra runner the one that deploys the data plane. We only want to fail on infra-related ones (func-e failing to start), but keep the normal envoy error reporting for the rest.

nacx:
@aabchoo mind stamping if this looks good to you?

aabchoo (approved):

PR #1998, opened 2026-03-30
feat: add pdf for aws-bedrock

hustxiayang:
**Description**

Add pdf support for aws bedrock schema. Btw, also add other documents: schema https://docs.aws.amazon.com/boto3/latest/reference/services/bedrock-runtime/client/converse.html


internal/translator/openai_awsbedrock.go
```diff
@@ -259,27 +305,29 @@ func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) openAIMessageToBedrockMes
...
+					block = &awsbedrock.ContentBlock{
+						Document: &awsbedrock.DocumentBlock{
+							Format: docFormat,
+							Name:   "document",
+							Source: awsbedrock.DocumentSource{
+								Bytes: b,
+							},
 						},
-					},
+					}
```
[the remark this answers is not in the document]
hustxiayang: I do not want to write a placeholder on ai-gateway side. Instead, I think it's better to track it and report the error to users.

hustxiayang:
@nacx Do you mind to review this as well? Thanks a lot!

internal/translator/openai_awsbedrock.go
```diff
@@ -290,6 +341,9 @@ func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) openAIMessageToBedrockMes
 				}
 			}
 		}
+		if hasDocument && !hasText {
```
gavrissh, line 344: If AI GW is responsible for validating these conditions, the conditions can change in upstream AWS Bedrock, and we might need to maintain that. Ideally, upstream will throw an error and pass it back to the user?

gavrissh (commented):
minor comment, rest lgtm

internal/translator/openai_awsbedrock.go
hustxiayang: @gavrissh Thanks a lot for the comment! I thought it would be better to maintain this on ai-gateway, because it is used for different providers and backends, and they would provide different interfaces (error code). So I wanted to have a consistent interface when users using ai-gateway.

hustxiayang:
@nacx Do you mind to review this one as well? Thanks!

malachischrager:
I have a production use case that needs exactly this: PDFs sent through `/v1/chat/completions` and routed to Bedrock (Claude), with no client-side rasterization or text extraction. Happy to help test or iterate if that unblocks a review.

PR #2023, opened 2026-04-08
feat: add /v1/audio/transcriptions and /v1/audio/translations endpoint support

PatilHrushikesh:
**Description**

This commit adds full data-plane support for OpenAI's audio endpoints:
`/v1/audio/transcriptions` (Whisper transcription) and `/v1/audio/translations`
(Whisper translation). These endpoints accept `multipart/form-data` requests
containing an audio file. 

### Key changes

**API schema** (`internal/apischema/openai`)
- Added `TranscriptionRequest`, `TranscriptionResponse`, `TranslationRequest`,
  `TranslationResponse` types and related structs (`TranscriptionSegment`,
  `TranscriptionWord`) to represent the parsed form fields and JSON responses.

**Endpoint specification** (`internal/endpointspec`)
- Added `TranscriptionEndpointSpec` and `TranslationEndpointSpec` implementing
  the full `EndpointSpec` interface.
- Extended the `EndpointSpec` interface with a new `ParseMultipartBody` method
  for multipart/form-data parsing. All existing JSON-only endpoints return
  `errMultipartNotSupported`.
- Multipart parsing extracts `model`, `file`, `language`, `prompt`,
  `response_format`, `temperature`, `stream`, and `timestamp_granularities[]`
  fields, validates required fields (`model` and `file`), and captures file
  metadata (name, size) without buffering file bytes.

**Request processor** (`internal/extproc/processor_impl.go`)
- `ProcessRequestBody` now inspects the `Content-Type` header and dispatches to
  `ParseMultipartBody` for `multipart/form-data` requests, falling back to the
  existing `ParseBody` for JSON.
- When setting a backend translator, the processor now checks for the
  `ContentTypeSetter` interface and propagates the original `Content-Type` to
  translators that need the multipart boundary for model rewriting.

**Translator** (`internal/translator`)
- New `openai_transcription.go` and `openai_translation.go` translators
  handling OpenAI-to-OpenAI passthrough, path rewriting, and response model
  tracking.
- New `multipart_helper.go` with `rewriteMultipartModel` to re-encode the
  multipart body when a model name override is configured (replaces only the
  `model` field value, copies all other parts including the file verbatim).
- New `ContentTypeSetter` interface so translators can receive the original
  request's `Content-Type` header (needed for boundary extraction).

**Tracing** (`internal/tracing`)
- Added `TranscriptionRecorder` and `TranslationRecorder` in the OpenInference
  tracing package, emitting `Transcription` / `Translation` span names with
  `openinference.span.kind=LLM` and recording request metadata and response
  text.
- Wired new tracers/recorders through the tracing infrastructure
  (`tracingapi`, `tracer.go`, `tracing.go`, `span.go`).
- Added noop implementations for both new span/tracer types.

**Metrics** (`internal/metrics`)
- Added `GenAIOperationTranscription` and `GenAIOperationTranslation` operation
  constants.

**Server wiring** (`cmd/extproc/mainlib/main.go`)
- Registered `/v1/audio/transcriptions` and `/v1/audio/translations` routes
  with their respective metrics factories, tracers, and endpoint specs.

**Documentation** (`site/docs/capabilities/llm-integrations/supported-endpoints.md`)
- Added "Audio Transcriptions" and "Audio Translations" sections with feature
  lists, supported providers, and `curl` examples.

- Model name override triggers a full multipart re-encode via
  `rewriteMultipartModel`, which changes the boundary. The new `Content-Type`
  and `Content-Length` headers are emitted accordingly.
- Streaming transcription (`stream=true`) is parsed but not specially handled
  in the translator — the request is passed through as-is.
- Only `OpenAI`-native backends are supported for these endpoints currently.
  
  Written with help of Claude


cmd/extproc/mainlib/main.go
```diff
@@ -283,6 +283,8 @@ func Main(ctx context.Context, args []string, stderr io.Writer) (err error) {
 	imageGenerationMetricsFactory := metrics.NewMetricsFactory(meter, metricsRequestHeaderAttributes, metrics.GenAIOperationImageGeneration)
 	responsesMetricsFactory := metrics.NewMetricsFactory(meter, metricsRequestHeaderAttributes, metrics.GenAIOperationResponses)
 	speechMetricsFactory := metrics.NewMetricsFactory(meter, metricsRequestHeaderAttributes, metrics.GenAIOperationSpeech)
+	transcriptionMetricsFactory := metrics.NewMetricsFactory(meter, metricsRequestHeaderAttributes, metrics.GenAIOperationTranscription)
+	translationMetricsFactory := metrics.NewMetricsFactory(meter, metricsRequestHeaderAttributes, metrics.GenAIOperationTranslation)
```
nejch, line 287: Sorry for the drive-by review here!

We're currently adopting the AI gateway in our team and looking at missing endpoints, so would love to see this merged. It feels like this might be easier/faster to get in by splitting into separate parts though:

1. backend prep work (multipart, ..)
2. transcriptions endpoint
3. translations endpoint

Or something like that. Just an idea but maybe you folks are more lax about large PRs here :)

gavrissh: > We're currently adopting the AI gateway in our team and looking at missing endpoints

Interested to know, what endpoints are missing that you are looking for?

We will review this shortly and get this to closure

nejch: @gavrissh thanks a lot! We're currently on vllm-router routing to vllm nodes, and adding Envoy in front to offload to Bedrock.

Ideally, we'd want to support all the major inference endpoints on the frontend, regardless of whether the model supports it (responses, messages, chat completions). In vLLM that's usually done by translating everything to chat completions internally. I see support for messages already (https://github.com/envoyproxy/ai-gateway/pull/1878) but not responses.

Additionally we use embeddings, rerank and score endpoints via vLLM at the moment. I haven't checked if these are all available here. Will open dedicated issues if needed, sorry for the noise here!

gavrissh: There is support for v1/responses, v1/embedding, and v2/rerank APIs. I think the score API is very vLLM specific

internal/endpointspec/endpointspec.go
```diff
@@ -635,3 +698,227 @@ func (SpeechEndpointSpec) RedactSensitiveInfoFromRequest(req *openai.SpeechReque
 
 	return &redacted, nil
 }
+
+// --- TranscriptionEndpointSpec ---
```
gavrissh: nit, unnecessary?

internal/translator/openai_transcription.go
```diff
@@ -0,0 +1,105 @@
...
+	return
+}
+
+// ResponseHeaders implements [OpenAIAudioTranscriptionTranslator.ResponseHeaders].
+func (o *openAIToOpenAITranslatorV1Transcription) ResponseHeaders(_ map[string]string) (newHeaders []internalapi.Header, err error) {
+	return nil, nil
+}
+
+// ResponseBody implements [OpenAIAudioTranscriptionTranslator.ResponseBody].
+func (o *openAIToOpenAITranslatorV1Transcription) ResponseBody(_ map[string]string, body io.Reader, _ bool, span tracingapi.TranscriptionSpan) (
```
AyushSawant18588: need to handle streaming response as well?

internal/apischema/openai/openai.go
```diff
@@ -8397,3 +8397,66 @@ const (
...
+type TranscriptionRequest struct {
+	Model                  string   `json:"model"`
+	Language               string   `json:"language,omitempty"`
+	Prompt                 string   `json:"prompt,omitempty"`
+	ResponseFormat         string   `json:"response_format,omitempty"`
+	Temperature            *float64 `json:"temperature,omitempty"`
+	TimestampGranularities []string `json:"timestamp_granularities,omitempty"`
+	Stream                 bool     `json:"stream,omitempty"`
+	FileName               string   `json:"file_name,omitempty"`
+	FileSize               int64    `json:"file_size,omitempty"`
```
AyushSawant18588, line 8483: file, chunking_strategy, include, known_speaker_names and known_speaker_references are optional fields as per openai docs https://developers.openai.com/api/reference/resources/audio/subresources/transcriptions/methods/create should we add it?

internal/apischema/openai/openai.go
```diff
@@ -8397,3 +8397,66 @@ const (
...
+}
+
+// TranslationRequest represents parsed form fields from a /v1/audio/translations multipart request.
+type TranslationRequest struct {
+	Model          string   `json:"model"`
+	Prompt         string   `json:"prompt,omitempty"`
+	ResponseFormat string   `json:"response_format,omitempty"`
+	Temperature    *float64 `json:"temperature,omitempty"`
+	FileName       string   `json:"file_name,omitempty"`
+	FileSize       int64    `json:"file_size,omitempty"`
```
AyushSawant18588, line 8493: I see there is only file field in the openai docs, wanted to know why is file_name and file_size added?

PatilHrushikesh: we dont actually use file_name and file_size just used as attribute in span

internal/endpointspec/endpointspec.go
```diff
@@ -637,3 +700,227 @@ func (SpeechEndpointSpec) RedactSensitiveInfoFromRequest(req *openai.SpeechReque
...
+	}
+
+	if !hasModel {
+		return "", nil, false, nil, fmt.Errorf("%w: missing required field 'model'", internalapi.ErrMalformedRequest)
+	}
+	if !hasFile {
+		return "", nil, false, nil, fmt.Errorf("%w: missing required field 'file'", internalapi.ErrMalformedRequest)
+	}
+
+	return req.Model, &req, req.Stream, nil, nil
```
PatilHrushikesh, line 744: // Propagate req.Stream so the upstream filter can switch Envoy to STREAMED response mode.
// OpenAI silently ignores stream=true for whisper-1 and returns JSON; the gateway must follow
// suit by branching on the actual response Content-Type, not on this flag — see the translator.

internal/apischema/openai/openai.go
PatilHrushikesh: Those are optional fields and we dont need to do any processing on that, so skipped

internal/apischema/openai/openai.go
```diff
@@ -8403,3 +8403,86 @@ const (
...
+}
+
+// TranscriptionResponse represents the JSON response from /v1/audio/transcriptions.
+// Covers both the simple (json) and verbose (verbose_json) response formats.
```
gavrissh: nit do we need this comment

AyushSawant18588 (approved):

jaimeluengo:
Nice work on the multipart plumbing — wanted to flag a gap I don't see addressed here:

Transcription and translation requests currently produce no `cost_microdollars` metadata, because the existing `AIGatewayRoute.spec.llmRequestCosts` shape is built around `input_tokens` / `output_tokens` fields that don't exist for `/v1/audio/transcriptions`. The new `TranscriptionRecorder` captures `FileName` and `FileSize` as span attributes but neither is plumbed into the cost path or surfaced as Envoy stream metadata. Net effect for operators relying on `cost_microdollars` for the rate-limit bucket or for downstream cost recording: transcription requests bill at zero, regardless of audio length.

Whisper's actual pricing is per-minute of audio, for example `$0.006/min` at OpenAI, so the natural integration point would be:

1. Extract audio duration in the extproc, either from the response body, as some providers return duration in the verbose JSON response, or, where unavailable, from the request's `FileSize` plus a configured bitrate hint, or from ffprobe-style probing if added as a build-time dependency, though that is probably out of scope.
2. Expose the duration as a CEL-readable variable, alongside `input_tokens` and `output_tokens`, for example a new `input_audio_seconds` / `output_audio_seconds` pair.
3. Operators can then add a `llmRequestCosts` rule such as `cost: int(input_audio_seconds / 60.0 * 600)`, representing microdollars per minute.

PatilHrushikesh:
> Nice work on the multipart plumbing — wanted to flag a gap I don't see addressed here:
> 
> Transcription and translation requests currently produce no `cost_microdollars` metadata, because the existing `AIGatewayRoute.spec.llmRequestCosts` shape is built around `input_tokens` / `output_tokens` fields that don't exist for `/v1/audio/transcriptions`. The new `TranscriptionRecorder` captures `FileName` and `FileSize` as span attributes but neither is plumbed into the cost path or surfaced as Envoy stream metadata. Net effect for operators relying on `cost_microdollars` for the rate-limit bucket or for downstream cost recording: transcription requests bill at zero, regardless of audio length.
> 
> Whisper's actual pricing is per-minute of audio, for example `$0.006/min` at OpenAI, so the natural integration point would be:
> 
> 1. Extract audio duration in the extproc, either from the response body, as some providers return duration in the verbose JSON response, or, where unavailable, from the request's `FileSize` plus a configured bitrate hint, or from ffprobe-style probing if added as a build-time dependency, though that is probably out of scope.
> 2. Expose the duration as a CEL-readable variable, alongside `input_tokens` and `output_tokens`, for example a new `input_audio_seconds` / `output_audio_seconds` pair.
> 3. Operators can then add a `llmRequestCosts` rule such as `cost: int(input_audio_seconds / 60.0 * 600)`, representing microdollars per minute.

Thanks for the the feedback, we will implement this in subsequent PRs pls refer ticket here for further discussions: https://github.com/envoyproxy/ai-gateway/issues/2186

johnugeorge (approved):

PR #2043, opened 2026-04-12
docs: update AWS examples to use infrastructure.parametersRef

lpiob:
**Description**

The `gateway.envoyproxy.io/envoy-proxy` annotation is ignored in current Envoy Gateway version, causing AWS examples to ignore the custom EnvoyProxy configuration.

Changes:
- Replace the legacy annotation with `spec.infrastructure.parametersRef` in Gateway resources.
- Move `EnvoyProxy` resources from 'envoy-gateway-system' to the 'default' namespace to match the Gateway and simplify example deployment.
- Fixes AWS IRSA and Pod Identity examples which were previously non-functional.


nacx (approved):
Thanks!
~~Would you mind taking a look at the PR title and description so that it passes the commit message style check?~~

PR #2045, opened 2026-04-13
feat: add response model metadata

### LINKED ISSUE #2044: expose response model as dynamic metadata
_Description_:

The ext_proc already extracts the model field from the upstream response, but discards it before building dynamic metadata. Adding it will allow the proxy to log the actual serving LLM even if it has been replaced/upgraded. Should be a small change.

Use case: operators can swap the backing model by patching a single Backend resource without needing to update every users AIGatewayRoute just for metadata accuracy.

Happy to open a PR or provide more information.

william-rom:
**Description**

This PR adds the response model from the upstream response to the dynamic metadata. This gives the proxy access to the actual model serving the request, unaffected by mode name overrides or other backend routing changes.

Close #2044

### DIFF
--- a/examples/access-log/basic.yaml
+++ b/examples/access-log/basic.yaml
@@ -51,7 +51,7 @@ spec:
               # the ones defined in the AIGatewayRoute llmRequestCosts field or
               # header-mapped attributes via OTEL_*_REQUEST_HEADER_ATTRIBUTES.
               gen_ai.request.model: "%REQ(X-AI-EG-MODEL)%"
-              gen_ai.response.model: "%DYNAMIC_METADATA(io.envoy.ai_gateway:model_name_override)%"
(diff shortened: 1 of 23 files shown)


nacx (commented):
Thanks!

Can you also check all the files in the `examples/` directory and update them? I see there are several examples that use the model name override as the response model in the access logs.

nacx (approved):
Thanks!

nacx:
Can you make sure all commits pass the DCO ?

nacx:
Can you fix the failing tests?

william-rom:
I replaced the response model key in the autoconfig test files.
There was also a failure in the inference-extension-e2e-test caused by some RBAC issues, but I believe these are unrelated to my changes.

PR #2047, opened 2026-04-13
feat(mcp): add per-backend ForwardHeaders to MCPRouteBackendRef with rename support

mohitgurnani:
**Description**

Add per-backend `forwardHeaders` to `MCPRouteBackendRef`, enabling explicit header forwarding from client requests to individual backend MCP servers with optional header renaming. This supports per-user authentication passthrough (e.g., personal access tokens) without requiring OAuth configuration, and ensures headers are only forwarded to explicitly opted-in backends.

Ref #1966

**Motivation**

Many MCP backends (Atlassian, Glean, Sourcegraph, etc.) accept per-user authentication via custom HTTP headers. Currently, the only way to populate forwardHeaders in the internal filter config is through OAuth claimToHeaders, which requires a full OIDC flow. For simpler scenarios — enterprise users passing personal access tokens — a direct header passthrough is more ergonomic and deployment-friendly.

Per the discussion in #1966, maintainers preferred per-backend forwarding on MCPRouteBackendRef to avoid leaking tokens to unrelated backends during fan-out operations like tools/list.

**Changes**

| Layer | File | Change |
|-------|------|--------|
| CRD | api/v1alpha1/mcp_route.go | Add MCPHeaderForward type (name + optional backendHeader for rename). Add ForwardHeaders to MCPRouteBackendRef |
| CRD | api/v1alpha1/zz_generated.deepcopy.go | DeepCopy for MCPHeaderForward and updated MCPRouteBackendRef |
| Internal API | internal/filterapi/mcpconfig.go | Add MCPHeaderForward and ForwardHeaders to MCPBackend |
| Controller | internal/controller/gateway.go | Read per-backend ForwardHeaders from CRD and populate filterapi.MCPBackend |
| Proxy | internal/mcpproxy/handlers.go | Add extractPerBackendForwardHeaders() with rename support |
| Proxy | internal/mcpproxy/mcpproxy.go | Populate perBackendExtraHeaders in session creation and reconstruction |
| Proxy | internal/mcpproxy/session.go | Add perBackendExtraHeaders to session; apply per-backend in sendRequestPerBackend |
| Docs | site/docs/capabilities/mcp/index.md | Add "Header Forwarding" section with usage examples |
| Docs | site/versioned_docs/version-0.5/capabilities/mcp/index.md | Same doc update for v0.5 |

**Design**

Headers are forwarded at two levels:

1. **Route-level** (MCPRoute.ForwardHeaders in filterapi) — populated by OAuth claimToHeaders only. Applied to ALL backends in the route (backward compatible with existing behavior).
2. **Per-backend** (MCPRouteBackendRef.ForwardHeaders) — new. Applied only to the specific backend. Each entry specifies:
   - name: the header to extract from the client request
   - backendHeader (optional): the header name to use when forwarding (enables renaming, e.g., Authorization -> X-Backend-Token)

During fan-out operations (tools/list, resources/list), only explicitly opted-in backends receive the forwarded headers. This prevents token leakage to unrelated backends.

**Example Usage**

```yaml
kind: MCPRoute
metadata:
  name: all-route
spec:
  path: "/mcp"
  backendRefs:
    - name: atlassian-mcp
      port: 8080
      forwardHeaders:
        - name: X-Atlassian-Jira-Personal-Token
        - name: X-Atlassian-Jira-Url
        - name: Authorization
          backendHeader: X-Original-Auth  # rename example
    - name: sourcegraph-mcp
      port: 9090
      # No forwardHeaders - won't receive any client tokens
```

**Test plan**

- [x] Verified end-to-end on live deployment: client sends per-user Jira headers through aggregated /mcp route -> gateway forwards them ONLY to Atlassian backend -> comment posted as user's identity ("Mohit Gurnani"), not service account
- [x] Verified aggregated route with 10+ backends all initialize successfully (per-backend headers don't break other backends)
- [x] Unit tests for extractPerBackendForwardHeaders() with rename
- [x] Unit tests for mcpConfig() with per-backend ForwardHeaders
- [x] Integration test with mock MCP backend asserting per-backend header scoping

- The internal plumbing for route-level forwarding (forwardHeaders in mcpProxyConfigRoute, extractForwardHeaders()) is preserved for backward compatibility with OAuth claimToHeaders. Per-backend forwarding is additive.
- MCPHeaderForward uses BackendHeader *string in the CRD (optional pointer) and BackendHeader string in filterapi (empty string = use original name) to keep the internal API simple.
- The perBackendExtraHeaders map on session is re-extracted from each incoming request (not persisted in the session ID), consistent with how route-level extraHeaders already works.

> Note: I used Cursor (AI) as a coding assistant. I fully understand and own
> all changes in this PR.


internal/mcpproxy/mcpproxy.go
```diff
@@ -408,14 +429,28 @@ func (m *mcpRequestContext) invokeJSONRPCRequest(ctx context.Context, routeName
...
 	if routeConfig := m.routes[routeName]; routeConfig != nil {
 		for _, header := range routeConfig.forwardHeaders {
 			if value := m.requestHeaders.Get(header); value != "" {
 				req.Header.Set(header, value)
 			}
 		}
 	}
+	// Forward per-backend headers (from MCPRouteBackendRef.forwardHeaders) with optional renaming.
+	if routeConfig := m.routes[routeName]; routeConfig != nil {
+		if b, ok := routeConfig.backends[backend.Name]; ok {
```
nacx, line 441: Add this to the previous conditional instead of creating a new one.

internal/mcpproxy/handlers.go
```diff
@@ -1253,6 +1253,30 @@ func extractForwardHeaders(reqHeaders http.Header, headers []string) map[string]
...
+		return nil
+	}
+
+	result := make(map[string]string)
+	for _, m := range mappings {
+		if value := reqHeaders.Get(m.Name); value != "" {
+			dest := m.Name
+			if m.BackendHeader != "" {
+				dest = m.BackendHeader
+			}
```
nacx: nit: encapsulate this in a method like: `m.ForwardName()` so it can be reused in all the palces this is accessed

internal/mcpproxy/session.go
```diff
@@ -54,11 +54,14 @@ type session struct {
...
-	// These are derived from the route's configured forward headers and the current request's headers.
-	// The key is the HTTP header name, the value is the header value.
+	// extraHeaders contains header values extracted from the current HTTP request to be forwarded to ALL backends.
+	// These are derived from the route's configured forward headers (e.g., OAuth claimToHeaders) and the current request's headers.
 	// Note: extraHeaders is NOT encoded in the session ID. It is re-extracted from each incoming request.
 	extraHeaders map[string]string
+	// perBackendExtraHeaders contains per-backend header values extracted from the current HTTP request.
+	// Key is the backend name; value is a map of destination header name -> value.
+	// These are derived from each MCPRouteBackendRef's forwardHeaders config.
+	perBackendExtraHeaders map[filterapi.MCPBackendName]map[string]string
```
nacx, line 65: Add the note about this not being encoded into the session id.

nacx (changes requested):
There is no single test covering the changed functionality. Please add proper tests.

internal/mcpproxy/session.go
mohitgurnani: Done

internal/mcpproxy/handlers.go
mohitgurnani: Done

nacx:
Thanks for the PR and addressing the comments. However, please take your time to read the [contributing guidelines](https://github.com/envoyproxy/ai-gateway/blob/main/CONTRIBUTING.md#opening-a-pull-request), especially the section about tests and commit squashing. Doing so makes it much harder to do incremental reviews, forcing the reviewers to try to figure out what really changed since the last review.

internal/mcpproxy/mcpproxy.go
mohitgurnani: Done

mohitgurnani:
> Thanks for the PR and addressing the comments. However, please take your time to read the [contributing guidelines](https://github.com/envoyproxy/ai-gateway/blob/main/CONTRIBUTING.md#opening-a-pull-request), especially the section about tests and commit squashing. Doing so makes it much harder to do incremental reviews, forcing the reviewers to try to figure out what really changed since the last review.

Thanks for the review @nacx . Yes, I will check the guidelines and ensure I follow it to the T, going forward.

internal/mcpproxy/handlers.go
```diff
@@ -1253,6 +1253,26 @@ func extractForwardHeaders(reqHeaders http.Header, headers []string) map[string]
 	return result
 }
 
+// extractPerBackendForwardHeaders reads per-backend header forwarding config from the incoming request.
+// It supports header renaming: each entry maps a source header to an optional destination header name.
+func extractPerBackendForwardHeaders(reqHeaders http.Header, mappings []filterapi.MCPHeaderForward) map[string]string {
+	if len(mappings) == 0 {
+		return nil
+	}
```
Hritik003, line 1261: any reason to keep it in handlers than keeping it in mcp proxy?

api/v1alpha1/mcp_route.go
```diff
@@ -121,7 +121,33 @@ type MCPRouteBackendRef struct {
 	// +optional
 	SecurityPolicy *MCPBackendSecurityPolicy `json:"securityPolicy,omitempty"`
 
-	// TODO: add fancy per-MCP server config. For example, Rate Limit, etc.
+	// ForwardHeaders specifies HTTP headers to extract from the incoming client request
+	// and forward to this backend MCP server.
+	// This enables per-user authentication passthrough (e.g., personal access tokens)
+	// without requiring OAuth configuration.
+	// Each entry specifies a header name to extract and an optional rename for the backend.
```
Hritik003, line 128: one general question here, when we say "without requiring OAuth configuration", is user expected to patch the mcp route whenever the access token is expired?

internal/mcpproxy/handlers.go
mohitgurnani: extractPerBackendForwardHeaders is a pure utility function that reads HTTP headers and returns a map — it has no dependency on the proxy or session state. It sits in handlers.go alongside the existing extractForwardHeaders function (which does the same thing for route-level headers) so the two related extraction functions stay co-located. The proxy layer (mcpproxy.go) then calls both of them during session creation. If we moved it into mcpproxy.go, it would separate it from its sibling function for no real benefit.

api/v1alpha1/mcp_route.go
mohitgurnani: No, the MCPRoute itself is never patched — it only declares which headers to forward. The actual token values are sent by the client in each HTTP request as regular headers (e.g., X-Atlassian-Jira-Personal-Token: <token>). The gateway just reads them from the incoming request and passes them through to the specified backend.

So if a token expires, the client simply starts sending the new token in the next request — no Kubernetes resource changes are needed. This is the same pattern as a browser sending an Authorization header on every request. The MCPRoute config is static (declares what to forward); the token values are dynamic (carried per-request by the client).

For short-lived tokens that need automatic refresh, OAuth with claimToHeaders is the better path — this forwardHeaders feature targets long-lived credentials like personal access tokens (PATs) where the client manages the token lifecycle.

Hritik003: Got it, basically the APIKeyAuth which we already have hardcodes a static token in a secret, whereas here we are just forwarding the headers and not their values.

Hritik003 (commented):
lgtm

nacx (approved):

PR #2050, opened 2026-04-14
fix: propagate request context to Kubernetes client calls in PostTranslateModify

ashnaaseth2325-oss:
**Description**

## Summary

This PR ensures proper propagation of `context.Context` through the extension server request flow, allowing downstream Kubernetes API calls to respect cancellation and deadlines. The changes primarily affect `internal/extensionserver/post_translate_modify.go` and related test updates in `extensionserver_test.go`.

---

## Fix

```go
// BEFORE
func (s *Server) PostTranslateModify(_ context.Context, req *...) (*..., error) {
    err = s.k8sClient.Get(context.Background(), key, &aigwRoute)
}

// AFTER
func (s *Server) PostTranslateModify(ctx context.Context, req *...) (*..., error) {
    err = s.k8sClient.Get(ctx, key, &aigwRoute)
}
```

---

## Verification

All existing unit tests were updated to pass the correct arguments and executed successfully (`go test ./internal/extensionserver/...`). The change was verified by ensuring context flows correctly through the call chain without altering existing behavior, while enabling proper cancellation and timeout handling.

### DIFF
--- a/internal/extensionserver/extensionserver_test.go
+++ b/internal/extensionserver/extensionserver_test.go
@@ -155,7 +155,7 @@ func Test_maybeModifyCluster(t *testing.T) {
 			var buf bytes.Buffer
 			s, err := New(c, logr.FromSlogHandler(slog.NewTextHandler(&buf, &slog.HandlerOptions{})), udsPath, false, nil, nil)
 			require.NoError(t, err)
-			err = s.maybeModifyCluster(tc.c)
+			err = s.maybeModifyCluster(t.Context(), tc.c)
 			require.NoError(t, err)
 			t.Logf("buf: %s", buf.String())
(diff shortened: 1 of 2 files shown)


ashnaaseth2325-oss:
Hi! This PR passes the request context through to Kubernetes calls so they respect cancellation and deadlines instead of using `context.Background()`.
Precommit checks and tests pass locally. Happy to make any changes if needed.

ashnaaseth2325-oss:
The Data Plane MCP test and ci-required checks are failing in CI, but everything passes locally, including precommit.
Since this change only passes the existing context through, I wouldn’t expect it to impact those tests, but I could be missing something. Could someone please re-run the checks? Happy to take a closer look if it still fails.
Thanks!

nacx (approved):

PR #2053, opened 2026-04-15
fix: prevent panic in extractSubject for malformed bearer Authorization header

ashnaaseth2325-oss:
**Description**

## Summary

This PR fixes a potential index out of bounds panic in `extractSubject` when the `Authorization` header contains a `bearer` scheme without a token. It adds a defensive length check and a corresponding test case in `handlers_test.go`.

---

## Fix

```go
// BEFORE
parts := strings.SplitN(authzHeader, " ", 2)
if !strings.EqualFold(parts[0], "bearer") {
    return ""
}
_, _, _ = jwt.NewParser().ParseUnverified(parts[1], &claims)


// AFTER
parts := strings.SplitN(authzHeader, " ", 2)
if !strings.EqualFold(parts[0], "bearer") {
    return ""
}
if len(parts) < 2 {
    return ""
}
_, _, _ = jwt.NewParser().ParseUnverified(parts[1], &claims)
```

---

## Verification

All unit tests pass locally with `go test ./internal/mcpproxy/...`, including the newly added case for `Authorization: bearer`. The change was also validated by confirming that malformed headers are handled gracefully without causing a panic.

### DIFF
--- a/internal/mcpproxy/handlers.go
+++ b/internal/mcpproxy/handlers.go
@@ -1228,6 +1228,9 @@ func extractSubject(r *http.Request) string {
 	if !strings.EqualFold(parts[0], "bearer") {
 		return ""
 	}
+	if len(parts) < 2 {
+		return ""
+	}
 
 	var claims jwt.RegisteredClaims
 	_, _, _ = jwt.NewParser().ParseUnverified(parts[1], &claims)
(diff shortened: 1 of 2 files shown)


ashnaaseth2325-oss:
Hello @nacx 
This PR adds a small guard to handle `Authorization: bearer` safely and prevent a panic in `extractSubject`. It also includes a test for this case. All tests pass locally, happy to make any changes if needed!

nacx (approved):

PR #2063, opened 2026-04-16
fix: remove unsupported response_format validation for AWS Bedrock

nuthalapativarun:
**Description**

The AWS Bedrock Converse API does not support `response_format` with `json_schema` (structured outputs). A previous version of this PR returned a `ErrInvalidRequestBody` error to signal this, but that pattern is inconsistent with how the rest of the gateway handles unsupported fields.

Per reviewer feedback, this PR reverts that approach. The `response_format` field now passes through silently — if Bedrock doesn't support it, Bedrock will return its own error. This matches the existing behavior for other unsupported fields across providers.

**Changes**

- `internal/translator/openai_awsbedrock.go`: Remove the proactive `ResponseFormat.OfJSONSchema` validation check.
- `internal/translator/openai_awsbedrock_test.go`: Remove the test case that covered the now-removed validation.

**Testing**

`make precommit` passes with 0 issues.

**Related**

Addresses the silent-drop behavior noted in the structured output pass-through gap. Bedrock will surface its own error if the field is unsupported.

### DIFF
--- a/internal/translator/openai_awsbedrock.go
+++ b/internal/translator/openai_awsbedrock.go
@@ -134,6 +134,13 @@ func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) RequestBody(_ []byte, ope
 		bedrockReq.AdditionalModelRequestFields = getAwsBedrockThinkingMap(openAIReq.Thinking)
 	}
 
+	// response_format.json_schema is not supported by the Bedrock Converse API.
+	// Return a clear error rather than silently dropping the field, which would
+	// result in the model producing unstructured output against the caller's intent.
+	if openAIReq.ResponseFormat != nil && openAIReq.ResponseFormat.OfJSONSchema != nil {
+		return nil, nil, fmt.Errorf("%w: response_format json_schema is not supported for AWS Bedrock", internalapi.ErrInvalidRequestBody)
+	}
+
 	// Convert Chat Completion messages.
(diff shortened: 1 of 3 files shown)


internal/translator/openai_awsbedrock.go
```diff
@@ -134,6 +134,13 @@ func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) RequestBody(_ []byte, ope
 		bedrockReq.AdditionalModelRequestFields = getAwsBedrockThinkingMap(openAIReq.Thinking)
 	}
 
+	// response_format.json_schema is not supported by the Bedrock Converse API.
+	// Return a clear error rather than silently dropping the field, which would
+	// result in the model producing unstructured output against the caller's intent.
+	if openAIReq.ResponseFormat != nil && openAIReq.ResponseFormat.OfJSONSchema != nil {
```
gavrissh: I'm not sure we follow a blocking of unsupported field patterns as this behavior can change in the future, and we cannot keep track of them.

Do you see that we do this for any other case?

nuthalapativarun: Good point — I looked through the codebase and didn't find any other place where we proactively block unsupported fields like this. I've updated the PR to remove the validation and revert the doc changes. The field will now pass through silently, and Bedrock will return its own error if it doesn't support it.

gavrissh:
Do we still need this change?

PR #2064, opened 2026-04-16
fix: remove OTEL span attribute count limit for large context windows

### LINKED ISSUE #2051: OTEL traces stop working properly when context window is too big
_Description_:

We are exposing traces to OpenTelemetry via OTEL_EXPORTER_OTLP_ENDPOINT. When the context window becomes ~60k tokens, the gateway changes its behaviour of how traces are parsed and sent . 
2 issues we noticed:
- it stopped sending "output" 
- it stopped mapping headers to OTEL params (such as session.id/user.email etc).

Tested for 2 different otel collectors: langfuse and phoenix

_Repro steps_:
Set up OTEL exporter the same way as in docs.
Add  to helm deployment `-- set controller.requestHeaderAttributes='X-Claude-Code-Session-Id:session.id'`

In Claude code setup
``` 
export ANTHROPIC_BASE_URL="<your_domain>/anthropic"
export ANTHROPIC_AUTH_TOKEN="dummy"
export CLAUDE_CODE_USE_VERTEX="0"
export CLAUDE_CODE_SKIP_VERTEX_AUTH="1"
export CLAUDE_CODE_DISABLE_EXPERIMENTAL_BETAS="1"
claude
```

And then run it with a conversation longer than  60k tokens.

_Environment_:
AI Gatewaty version: 0.0.8b88b90b7d7dc070c4ea042fddf27682224f71c3
Envoy version: v1.6.4
Model: opus-4.6 (via VertexAI)

_Logs_:
No error logs were found.

edan-binshtok:
**Description**
The OTEL Go SDK defaults SpanLimits.AttributeCountLimit to 128. In LLM tracing, each conversation message generates multiple indexed span attributes (role, content, tool
  calls), so a conversation with 100+ messages easily exceeds this limit. The SDK silently drops excess attributes, and since header-mapped attributes (session IDs, user
  emails) are set last, they are always the first to be lost — appearing in production as "incomplete traces" with no error or log.

  Uses WithRawSpanLimits (not the deprecated WithSpanLimits, which silently replaces negative values with the default) to set AttributeCountLimit to -1 (unlimited), gated on
   OpenInference message capture being enabled via a new TraceConfig.CapturesMessages() helper. When operators opt out via OPENINFERENCE_HIDE_INPUTS / HIDE_OUTPUTS /
  HIDE_INPUT_MESSAGES / HIDE_OUTPUT_MESSAGES, the recorders emit no high-cardinality attributes and OTEL's default limit is preserved. NewSpanLimits() is used as the base so
   all other limits (value length, event count, etc.) keep their defaults and respect environment variable overrides.

  Closes #2051


nacx:
Hi! See my comment here: https://github.com/envoyproxy/ai-gateway/issues/2051#issuecomment-4261771493
Instead of hardcoding removing the default limits, can this just be addressed by configuring hte desired values in the existing env vars?

edan-binshtok:
> Hi! See my comment here: [#2051 (comment)](https://github.com/envoyproxy/ai-gateway/issues/2051#issuecomment-4261771493) Instead of hardcoding removing the default limits, can this just be addressed by configuring hte desired values in the existing env vars?

hi @nacx , please see my reasoning at https://github.com/envoyproxy/ai-gateway/issues/2051#issuecomment-4261964870

DanielChrn:
For a context increase of Span attributes worked, tnx. 
However, I am not sure about the solution.  I see it more like a symptom.
Checked in OT database, and I see that, for example, each tool is registered as a separate attribute. Which does not really make sense if we take the format of it into account (huge JSON payload with prompts, etc.).  

Since LLM call is not the most typical format for OpenTelemetry I would suggest to store atributes with lower granularity AKA, instead of storing each tool as a separate attribute - create a single attribute for tools and store an array there. It will also reduce  some parsing jobs from CPU during processing  (unless there is some performance concern)

nacx (approved):

nacx:
Thanks!

PR #2065, opened 2026-04-16
fix: remove hardcoded runAsUser/runAsGroup from extproc security context

### LINKED ISSUE #2046: extproc initContainer hardcodes runAsUser: 65532, breaks OpenShift restricted SCC
_Description_:

The `ai-gateway-controller` v0.5.0 webhook injects an initContainer `ai-gateway-extproc` with a hardcoded `runAsUser: 65532` when mutating Envoy Gateway pods. This prevents deployment on OpenShift clusters using the default `restricted` or `restricted-v2` Security Context Constraint (SCC), which requires UIDs within a namespace-allocated range (e.g., `1001540000-1001549999`).

The main `envoy` and `shutdown-manager` containers correctly respect the EnvoyProxy `securityContext: {}` spec (fixed in Envoy Gateway v1.3 via PR envoyproxy/gateway#4938). However, the `ai-gateway-extproc` initContainer ignores the EnvoyProxy configuration.

Pod spec after webhook mutation:
```yaml
initContainers:
  - name: ai-gateway-extproc
    image: docker.io/envoyproxy/ai-gateway-extproc:v0.5.0
    securityContext:
      runAsUser: 65532  # hardcoded, not configurable
containers:
  - name: envoy
    securityContext: {}  # respects EnvoyProxy spec ✓
  - name: shutdown-manager
    securityContext: {}  # respects EnvoyProxy spec ✓
```

_Repro steps_:

1. Deploy Envoy Gateway v1.7.1 on OpenShift (OKD 4.x)
2. Deploy AI Gateway Controller v0.5.0 with the mutating webhook
3. Create a Gateway referencing an EnvoyProxy with `securityContext: {}`
4. Create an AIGatewayRoute
5. Observe pod creation failure:

```
pods "foryouassist-ai-gateway-xxx-" is forbidden: unable to validate against any security context constraint:
  provider restricted-v2: .initContainers[0].runAsUser: Invalid value: 65532:
  must be in the ranges: [1001540000, 1001549999]
```

**Workaround**: Create a custom SCC with `runAsUser: type: RunAsAny` bound to the Envoy service account. This requires cluster-admin privileges.

**Expected behavior**: The `ai-gateway-extproc` initContainer should either:
1. Inherit the securityContext from the EnvoyProxy spec (like envoy and shutdown-manager)
2. Not set runAsUser, letting the platform assign a UID from the namespace range
3. Be configurable via the ai-gateway-helm chart values

_Environment_:

- Envoy Gateway: v1.7.1
- AI Gateway Controller: v0.5.0
- Platform: OpenShift OKD 4.x with default `restricted-v2` SCC
- Helm chart: ai-gateway-helm

_Logs_:

```
Warning  FailedCreate  replicaset/foryouassist-ai-gateway-xxx
  Error creating: pods "foryouassist-ai-gateway-xxx-" is forbidden:
  unable to validate against any security context constraint:
  [provider restricted-v2: .initContainers[0].runAsUser: Invalid value: 65532:
  must be in the ranges: [1001540000, 1001549999]]
```

Koushik-Salammagari:
**Description**

The extproc container's default `SecurityContext` hardcoded `runAsUser: 65532`
and `runAsGroup: 65532`. On OpenShift clusters using the default `restricted-v2`
SCC, pod UIDs must fall within a namespace-allocated range (e.g.
`1001540000–1001549999`), so any hardcoded UID outside that range causes the
pod to be rejected with:

```
unable to validate against any security context constraint:
  provider restricted-v2: .initContainers[0].runAsUser: Invalid value: 65532:
  must be in the ranges: [1001540000, 1001549999]
```

`RunAsNonRoot: true` and `AllowPrivilegeEscalation: false` are sufficient to
enforce safe execution without pinning a specific UID — the platform assigns
one from its allocated range automatically.

Users who need a fixed UID can still set `kubernetesExtProc.SecurityContext`
explicitly in their `GatewayConfig`, which continues to take full precedence
(unchanged behaviour).

Fixes #2046

The existing override path (`kubernetesExtProc.SecurityContext != nil`) is
untouched. Only the hardcoded default values are removed.

> Note: I used Claude (AI) as a coding assistant. I fully understand and own
> all changes in this PR.

### DIFF
--- a/internal/controller/gateway_mutator.go
+++ b/internal/controller/gateway_mutator.go
@@ -428,9 +428,7 @@ func (g *gatewayMutator) mutatePod(ctx context.Context, pod *corev1.Pod, gateway
 			Drop: []corev1.Capability{"ALL"},
 		},
 		Privileged:   ptr.To(false),
-		RunAsGroup:   ptr.To(int64(65532)),
 		RunAsNonRoot: ptr.To(true),
-		RunAsUser:    ptr.To(int64(65532)),
 		SeccompProfile: &corev1.SeccompProfile{
 			Type: corev1.SeccompProfileTypeRuntimeDefault,
 		},
--- a/internal/controller/gateway_mutator.go
+++ b/internal/controller/gateway_mutator.go
@@ -431,9 +431,7 @@ func (g *gatewayMutator) mutatePod(ctx context.Context, pod *corev1.Pod, gateway
 			Drop: []corev1.Capability{"ALL"},
 		},
 		Privileged:   ptr.To(false),
-		RunAsGroup:   ptr.To(int64(65532)),
 		RunAsNonRoot: ptr.To(true),
-		RunAsUser:    ptr.To(int64(65532)),
 		SeccompProfile: &corev1.SeccompProfile{
 			Type: corev1.SeccompProfileTypeRuntimeDefault,
 		},
--- a/internal/controller/gateway_mutator_test.go
+++ b/internal/controller/gateway_mutator_test.go
@@ -97,6 +97,18 @@ func TestGatewayMutator_mutatePod(t *testing.T) {
 			},
 		},
 		{
+			name: "default security context does not set RunAsUser or RunAsGroup",
+			extprocTest: func(t *testing.T, container corev1.Container) {
+				require.NotNil(t, container.SecurityContext)
+				require.Nil(t, container.SecurityContext.RunAsUser,
+					"RunAsUser must not be hardcoded so platforms like OpenShift can assign their own UID")
+				require.Nil(t, container.SecurityContext.RunAsGroup,
+					"RunAsGroup must not be hardcoded so platforms like OpenShift can assign their own GID")
+				require.Equal(t, ptr.To(true), container.SecurityContext.RunAsNonRoot)
+				require.Equal(t, ptr.To(false), container.SecurityContext.AllowPrivilegeEscalation)
+			},
+		},
+		{
 			name:    "basic extproc container with MCPRoute",
 			needMCP: true,
 			extprocTest: func(t *testing.T, container corev1.Container) {


nacx:
The value matches the default UID/GID of the container images, so it is the right default, I'd say.
I think we should keep these defaults, and keep letting users that need a different value to just change them by using hte available mechanisms (GatewayConfig).

Koushik-Salammagari:
Thanks for the context, @nacx — that's a fair point and I appreciate the review.

The core issue is that on OpenShift with the default `restricted-v2` SCC, the pod UID must fall within a namespace-allocated range (e.g. `1001540000–1001549999`). A hardcoded `65532` falls outside that range and causes the pod to be rejected at admission:

```
unable to validate against any security context constraint:
  provider restricted-v2: .initContainers[0].runAsUser: Invalid value: 65532:
  must be in the ranges: [1001540000, 1001549999]
```

The security goal — ensuring the process does not run as root — is fully preserved by `RunAsNonRoot: true`. The distroless `nonroot` image already has UID `65532` baked in as its default user, so when OpenShift assigns a UID from its allocated range, the container still runs as non-root; `RunAsNonRoot: true` enforces that at the kubelet level.

The current workaround requires users to set the **entire** `SecurityContext` block in `GatewayConfig`, which means they must also manually replicate `AllowPrivilegeEscalation: false`, `Capabilities.Drop: ALL`, `Privileged: false`, and `SeccompProfile`. That's easy to get wrong and leaves OpenShift users with a worse security posture if they miss a field.

I understand the concern about changing the default, but I'd argue `RunAsNonRoot: true` is the authoritative enforcement here, and the explicit UID is redundant (and actively harmful on platforms with UID allocation policies). Happy to discuss further or explore a middle ground if you have a different approach in mind.

nacx:
That's fair. Thanks for sharing your thoughts! LGTM
cc'ing @johnugeorge for a final thumbs-up

nacx:
Can you take a look at the failing tests?

Koushik-Salammagari:
@nacx Looked into the failures. All 8 E2E jobs failed simultaneously across every k8s version (v1.32–v1.35), Namespaced Controller, and Inference Extensions — that breadth points to CI infrastructure rather than this change. The failure mode is `timed out waiting for the condition on pods/envoy-gateway-ns-gw-*` (Envoy Gateway data plane pods, not extproc) and `array index out of bounds: index 0, length 0` when listing pods, suggesting the kind cluster itself had issues that run.

Unit tests, controller tests, data plane tests, and CRD CEL validation all passed.

Rebased onto current main. Re-triggering CI should confirm.

nacx:
Failures keep happening, and they're not because of CI. Please run the e2e locally

PR #2071, opened 2026-04-18
fix: ensure InferencePool metadata is assigned when metadata is nil

ashnaaseth2325-oss:
**Description**

## Summary

This PR fixes an issue where InferencePool metadata was not getting saved when the initial metadata was nil.
`buildEPPMetadata` was creating metadata, but since it wasn’t returned, the caller never stored it.

This change makes `buildEPPMetadata` return the metadata and ensures it is assigned back by the caller.

---

## Use Case

Clusters and routes created without existing metadata (which is the default case) were missing InferencePool metadata.

While debugging this, I found that `buildEPPMetadata` initializes metadata when it is nil, but the updated value wasn’t being returned. Because of that, the metadata was never actually set on the cluster/route.

As a result, the InferencePool reference was getting dropped and later code could not read it.

---

## Reproduction (Concrete Code Path)

This happens in the following flow:

1. A cluster/route is created with `Metadata = nil`
2. `buildEPPMetadataForCluster` / `buildEPPMetadataForRoute` is called
3. This calls `buildEPPMetadata(metadata, inferencePool)`
4. Inside it, metadata gets initialized and updated

Before this change:

* The updated metadata was not returned
* So the caller never assigned it back
* `Metadata` stayed nil

Because of this:

* `getInferencePoolByMetadata` could not find the InferencePool
* Downstream logic skipped processing

---

### Before

```go
func buildEPPMetadata(metadata *corev3.Metadata, inferencePool *gwaiev1.InferencePool) {
    if metadata == nil {
        metadata = &corev3.Metadata{}
    }
}
```

### After

```go
func buildEPPMetadata(metadata *corev3.Metadata, inferencePool *gwaiev1.InferencePool) *corev3.Metadata {
    if metadata == nil {
        metadata = &corev3.Metadata{}
    }
    return metadata
}
```

Callers now assign it back:

```go
cluster.Metadata = buildEPPMetadata(cluster.Metadata, inferencePool)
route.Metadata = buildEPPMetadata(route.Metadata, inferencePool)
```

---

## Verification

All unit tests pass locally using `go test ./internal/extensionserver/...` . The change was verified by ensuring no regressions in existing behavior and that metadata initialization no longer gets dropped when initially nil.

### DIFF
--- a/internal/extensionserver/inferencepool.go
+++ b/internal/extensionserver/inferencepool.go
@@ -133,17 +133,17 @@ func getInferencePoolByMetadata(meta *corev3.Metadata) *gwaiev1.InferencePool {
 // encoded as a string in the format: "namespace/name/serviceName/port".
 func buildEPPMetadataForCluster(cluster *clusterv3.Cluster, inferencePool *gwaiev1.InferencePool) {
 	// Initialize cluster metadata structure if not present.
-	buildEPPMetadata(cluster.Metadata, inferencePool)
+	cluster.Metadata = buildEPPMetadata(cluster.Metadata, inferencePool)
 }
 
 // buildMetadataForInferencePool adds InferencePool metadata to the route for reference by other components.
 func buildEPPMetadataForRoute(route *routev3.Route, inferencePool *gwaiev1.InferencePool) {
 	// Initialize route metadata structure if not present.
-	buildEPPMetadata(route.Metadata, inferencePool)
+	route.Metadata = buildEPPMetadata(route.Metadata, inferencePool)
 }
 
 // buildEPPMetadata adds InferencePool metadata to the given metadata structure.
-func buildEPPMetadata(metadata *corev3.Metadata, inferencePool *gwaiev1.InferencePool) {
+func buildEPPMetadata(metadata *corev3.Metadata, inferencePool *gwaiev1.InferencePool) *corev3.Metadata {
 	// Initialize cluster metadata structure if not present.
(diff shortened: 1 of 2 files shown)


ashnaaseth2325-oss:
Hello @nacx ,
This PR ensures InferencePool metadata is correctly assigned when it’s initially nil by returning and setting it back in the callers. All tests pass locally.
Thanks!

siddharth1036:
@ashnaaseth2325-oss can you add the use case where the issue was noticed to the description? Also it might be helpful to add a test for the same

ashnaaseth2325-oss:
Hello @siddharth1036,
Thanks for the suggestion!
Added the use case to the PR description and included a test for the nil metadata case.

mathetake:
@ashnaaseth2325-oss the question is exactly when you encountered this, not describing some imaginary situation

ashnaaseth2325-oss:
Hi @mathetake,
I’ve updated the use case to reflect the actual scenario where I encountered this during testing.

mathetake:
> While testing InferencePool integration, I noticed that clusters/routes created without pre-existing metadata (the default case) were missing the expected InferencePool metadata.

how? do you have a concrete step by step repro that actually hits that code path?

ashnaaseth2325-oss:
Hi @mathetake,
Thanks for the feedback! I’ve added a clear step by step repro with the exact code path and impact in the PR description. Could you please take another look?

PR #2078, opened 2026-04-22
feat: add anthropic count endpoint with anthropic/v1/messages/count_tokens

hustxiayang:
**Description**

Add support for the Anthropic `/v1/messages/count_tokens` endpoint.

**Motivation:** 
Claude Code frequently calls `/v1/messages/count_tokens` to count input tokens for context window management. Without this endpoint, requests fail and Claude Code falls back to invoking a smaller model (e.g., Haiku) with `max_tokens=1` just to read back `prompt_tokens` from the usage field. This workaround increases latency, costs more money, and produces noisy failures when the fallback model doesn't support the same features as the primary model (e.g., tool use schemas valid for Opus/Sonnet may be rejected by Haiku).

**What this PR adds:**

- New `AnthropicCountTokensTranslator` interface with implementations for all three Anthropic-compatible providers: direct Anthropic, GCP Vertex AI, and AWS Bedrock.
- Request/response translation for each provider's `count_tokens` API format.
- Metrics (`GenAIOperationCountTokens`) and tracing (`SpanKindTokenCounter`) support.
- Endpoint registration at `{prefix}/anthropic/v1/messages/count_tokens`.

**Provider-specific model name handling:**

The `modelNameOverride` configured on a backend is shared across all API types (`messages`, `chat`, `count_tokens`). However, the `count_tokens` endpoints on GCP and AWS have stricter model name requirements than their inference counterparts:

- **GCP Vertex AI:** The `count-tokens` endpoint does not accept the `@default` or `@latest` version aliases (e.g., `claude-sonnet-4-6@default` returns "not supported for token counting"). Explicit version tags like `@20250514` are accepted, as are bare model names. Since `@default` is required for inference endpoints that share the same backend config, the count_tokens translator strips `@default` and `@latest` suffixes only.
  - Ref: https://cloud.google.com/vertex-ai/generative-ai/docs/partner-models/claude/count-tokens

- **AWS Bedrock:** The CountTokens API (`POST /model/{modelId}/count-tokens`) does not support cross-region inference (CRIS) model IDs (e.g., `us.anthropic.claude-sonnet-4-6` returns "The provided model doesn't support counting tokens"). CRIS prefixes are required for inference endpoints (InvokeModel, Converse) that share the same backend config, so the count_tokens translator strips short region prefixes (e.g., `us.`, `eu.`) only.
  - Ref: https://docs.aws.amazon.com/bedrock/latest/userguide/count-tokens.html


internal/translator/counttokens_gcpanthropic.go
```diff
@@ -0,0 +1,98 @@
...
+	// is required for inference endpoints (messages, chat) which share the same
+	// modelNameOverride, so we strip it here for count-tokens only.
+	// See: https://docs.cloud.google.com/gemini-enterprise-agent-platform/models/partner-models/claude/count-tokens
+	if t.modelNameOverride != "" {
+		model := t.modelNameOverride
+		if strings.HasSuffix(model, "@default") || strings.HasSuffix(model, "@latest") {
+			model = model[:strings.LastIndexByte(model, '@')]
+		}
+		newBody, _ = sjson.SetBytesOptions(newBody, "model", model, sjsonOptions)
+	}
```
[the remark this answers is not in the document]
hustxiayang: updated, no harm to add this.

internal/apischema/anthropic/anthropic.go
```diff
@@ -1747,6 +1747,12 @@ func (m *MessagesStreamChunk) UnmarshalJSON(data []byte) error {
 	return nil
 }
 
+// CountTokensResponse represents a response from the Anthropic Count Tokens API.
+// https://docs.anthropic.com/en/api/counting-tokens
```
aabchoo: https://platform.claude.com/docs/en/api/messages/count_tokens?

internal/endpointspec/endpointspec.go
```diff
@@ -388,6 +390,48 @@ func (MessagesEndpointSpec) RedactSensitiveInfoFromRequest(req *anthropic.Messag
 	return req, nil
 }
 
+// ParseBody implements [EndpointSpec.ParseBody].
+func (MessagesCountTokensEndpointSpec) ParseBody(
+	body []byte,
+	_ bool,
+) (internalapi.OriginalModel, *anthropic.MessagesRequest, bool, []byte, error) {
+	var req anthropic.MessagesRequest
```
aabchoo: Why reuse the MessageRequest instead of having a dedicated CountTokensRequest?

internal/tracing/openinference/anthropic/count_tokens.go
```diff
@@ -0,0 +1,73 @@
...
+
+// StartParams implements the same method as defined in tracingapi.CountTokensRecorder.
+func (r *CountTokensRecorder) StartParams(*anthropic.MessagesRequest, []byte) (spanName string, opts []trace.SpanStartOption) {
+	return "CountTokens", startOpts
+}
+
+// RecordRequest implements the same method as defined in tracingapi.CountTokensRecorder.
+func (r *CountTokensRecorder) RecordRequest(span trace.Span, req *anthropic.MessagesRequest, body []byte) {
+	attrs := []attribute.KeyValue{
+		attribute.String(openinference.SpanKind, openinference.SpanKindTokenCounter),
```
aabchoo, line 46: Is this supposed to be added for https://github.com/envoyproxy/ai-gateway/blob/main/internal/tracing/openinference/openai/tokenize.go#L89-L92 as well?

internal/tracing/tracer.go
```diff
@@ -52,6 +54,7 @@ type (
 	transcriptionTracer   = requestTracerImpl[openai.TranscriptionRequest, openai.TranscriptionResponse, openai.TranscriptionStreamEvent]
 	translationTracer     = requestTracerImpl[openai.TranslationRequest, openai.TranslationResponse, struct{}]
 	rerankTracer          = requestTracerImpl[cohereschema.RerankV2Request, cohereschema.RerankV2Response, struct{}]
+	countTokensTracer     = requestTracerImpl[anthropicschema.MessagesRequest, anthropicschema.CountTokensResponse, struct{}]
```
aabchoo: side note: we need to add a tokenizer tracer

internal/tracing/tracer.go
```diff
@@ -40,6 +41,7 @@ var (
 	_ tracingapi.TranscriptionTracer   = (*transcriptionTracer)(nil)
 	_ tracingapi.TranslationTracer     = (*translationTracer)(nil)
 	_ tracingapi.RerankTracer          = (*rerankTracer)(nil)
+	_ tracingapi.CountTokensTracer     = (*countTokensTracer)(nil)
```
aabchoo: similar here with tokenizer

internal/translator/counttokens_anthropic.go
```diff
@@ -0,0 +1,81 @@
...
+		if err != nil {
+			return nil, nil, fmt.Errorf("failed to set model name: %w", err)
+		}
+	}
+
+	if forceBodyMutation && len(newBody) == 0 {
+		newBody = original
+	}
+
+	newHeaders = []internalapi.Header{{pathHeaderName, "/v1/messages/count_tokens"}}
```
aabchoo: use constants instead of magic strings

internal/translator/counttokens_awsanthropic.go
```diff
@@ -0,0 +1,142 @@
...
+	// AWS Bedrock's CountTokens API (POST /model/{modelId}/count-tokens) does not support
+	// cross-region inference (CRIS) model IDs (e.g. "us.anthropic.claude-sonnet-4-6" returns
+	// "The provided model doesn't support counting tokens"). The CRIS prefix is required for
+	// inference endpoints (InvokeModel, Converse) which share the same modelNameOverride, so we
+	// strip it here for count-tokens only. A CRIS ID prepends a geography prefix (e.g. "us.",
+	// "eu.", "apac.", "us-gov.") to the base model ID; anchor on the "anthropic." provider
+	// segment and drop anything before it, so every geography prefix is handled regardless of
+	// length. A bare base ID ("anthropic.claude-...") has the segment at index 0 and is left as-is.
+	// See: https://docs.aws.amazon.com/bedrock/latest/userguide/count-tokens.html
+	if i := strings.Index(model, "anthropic."); i > 0 {
```
aabchoo: this is repeated logic from another translator above. Let's make one function to simplify the code

internal/translator/counttokens_awsanthropic.go
```diff
@@ -0,0 +1,142 @@
...
+	encodedBody := base64.StdEncoding.EncodeToString(invokeBody)
+	newBody = []byte("{}")
+	newBody, err = sjson.SetBytesOptions(newBody, "input.invokeModel.body", encodedBody, sjsonOptions)
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to wrap body: %w", err)
+	}
+
+	// URL encode the model ID for the path.
+	encodedModelID := url.PathEscape(model)
+	path := fmt.Sprintf("/model/%s/count-tokens", encodedModelID)
```
aabchoo: is the path correct? also we should abstract this

internal/apischema/anthropic/anthropic.go
hustxiayang: Thanks a lot! This link might be outdated a while ago, would update it!

internal/endpointspec/endpointspec.go
hustxiayang: Thanks a lot! a dedicated CountTokensRequest should be cleaner, would switch to it!

internal/translator/counttokens_awsanthropic.go
hustxiayang: the path is correct, and verified locally

hustxiayang: would follow the suggestion to abstract it

internal/translator/counttokens_anthropic.go
hustxiayang: would follow the suggestion.

internal/translator/counttokens_awsanthropic.go
hustxiayang: would follow it

internal/tracing/openinference/anthropic/count_tokens.go
hustxiayang: Thanks! We need to add this. Just need to confirm whether we could add this directly in this pr or make it a follow-up to keep it clean.

internal/tracing/tracer.go
hustxiayang: same

internal/tracing/tracer.go
hustxiayang: same

aabchoo (approved):
Please avoid force pushing in the future to make reviewing a lot easier

PR #2086, opened 2026-04-26
feat: add Peyeeye PII redaction & rehydration processor

tim-peyeeye:
**Description**

Adds an extproc Processor that integrates the Peyeeye (https://peyeeye.ai) PII redaction & rehydration API into the AI Gateway request/response pipeline. On the request path it calls `POST /v1/redact` with the user's chat-completion / completion / embedding / responses / Anthropic-messages text and rewrites the body in place with redacted strings and placeholder tokens (`[EMAIL_1]`, `[PHONE_2]`, …). On the response path it calls `POST /v1/rehydrate` to swap placeholders back to the original PII before returning bytes to the caller.

The integration is wired as a decorator (`PEyeEyeProcessor` wraps `extproc.Processor`, `WrapFactory` wraps `extproc.ProcessorFactory`) so the existing generic factory plumbing is untouched. It's gated behind a new `-peyeeye` CLI flag in `cmd/extproc/mainlib/main.go`; when the flag is off the binary's behavior is byte-identical to today.

Behavioral invariants:
- **Length-guard**: if Peyeeye returns a different number of strings than we sent, the request fails closed — partial redaction is never forwarded upstream.
- **Fail-closed on redact**: any redact error (timeout, 5xx, decode) returns a `PEyeEyeProcessorError` and the request never reaches the model.
- **Fail-open on rehydrate**: rehydrate errors are logged and the model's raw response is returned. The model only ever sees placeholders, so this is safe.
- **Cleanup**: stateful sessions (`ses_…`) are best-effort `DELETE`d on `EndOfStream`; stateless (`skey_…`) is a no-op.
- Two session modes from the upstream API are supported: `stateful` (default; `ses_…` id) and `stateless` (sealed `skey_…` blob, nothing retained server-side).

Configuration knobs (`PEyeEyeConfig` in `internal/extproc/peyeeye/config.go`, mirrored as `PeyeEyeFilterConfig` in `internal/filterapi`):
- `apiKey` / `PEYEEYE_API_KEY`
- `apiBase` / `PEYEEYE_API_BASE` (default `https://api.peyeeye.ai`)
- `locale` (BCP-47, default `auto`)
- `entities` (optional allow-list, e.g. `["EMAIL", "PHONE_NUMBER"]`)
- `sessionMode` (`stateful` | `stateless`)

Streaming SSE responses are passed through unchanged in this PR — placeholders are inert in the model output, so leaving them in is safe; rehydration of streaming bodies is a follow-up.

Naming convention follows the upstream brand: Go identifiers use mixed-case `PEyeEye` (`PEyeEyeProcessor`, `PEyeEyeClient`, `PEyeEyeConfig`, `PEyeEyeProcessorError`); package directory, JSON keys, and env vars stay `peyeeye` / `PEYEEYE_*`. Lint stutter warnings on these types are suppressed with a `//nolint:revive` comment explaining the brand naming.

Files:
- `internal/extproc/peyeeye/config.go` — `PEyeEyeConfig`, `Resolve()`, `SessionMode`
- `internal/extproc/peyeeye/client.go` — `PEyeEyeClient` (Redact / Rehydrate / DeleteSession), typed `PEyeEyeProcessorError`, `Doer` interface for testability
- `internal/extproc/peyeeye/processor.go` — `PEyeEyeProcessor` decorator, request/response body extraction & sjson rewrite
- `internal/extproc/peyeeye/factory.go` — `WrapFactory` decorator
- `internal/extproc/peyeeye/processor_test.go` — 24 sub-tests across config, processor, and client layers
- `internal/filterapi/filterconfig.go` — adds `PeyeEye *PeyeEyeFilterConfig` to `Config`
- `cmd/extproc/mainlib/main.go` — `-peyeeye` flag + symmetric wrap of chat / completions / embeddings / responses / anthropic-messages registrations

N/A

Test plan:
- [x] `make precommit` — license headers, gofmt, golangci-lint, helm-lint all clean
- [x] `make test` — full unit-test suite passes; new `internal/extproc/peyeeye` package covered by 24 sub-tests
- [x] Table-driven coverage: redact length mismatch, redact 5xx fail-closed, multimodal `content: [{type:"text", …}]` shape, stateless skips DELETE, rehydrate error is non-fatal, no-texts pass-through, forwarding hooks (headers / SetBackend), HTTP client error wrapping, DeleteSession no-op for `skey_…`
- [ ] Integration with a real Peyeeye API key (manual; not in CI)


johnugeorge:
Since it is an extProc, why do we need it to integrate it within ai-gateway ? Can't we run it outside ai-gateway?

tim-peyeeye:
Quick clarification on the shape: it's not a separate ext_proc next to ai-gateway. It's a decorator inside the existing `cmd/extproc` binary that wraps the chat / completions / embeddings / responses / anthropic-messages processors, gated behind a `-peyeeye` flag (off by default, byte-identical to today).

The reason it lives in-tree is to piggyback on the body parsing those processors already do. A standalone ext_proc would re-parse every API surface, run as a second filter hop, and need its own config plane for what's a body-rewrite step. Running peyeeye externally is still totally fine; this is for operators who'd rather configure it like the rest of the gateway.

Happy to relocate the package if you'd prefer it elsewhere.

siddharth1036:
@tim-peyeeye would it be possible to make a generic interface for wrapper extproc processing and Peyeeye could be an implementation of that?

tim-peyeeye:
@siddharth1036 makes sense. I'll refactor this around a generic wrapper interface (request body in/out plus response body in/out, with per-request state threaded between them) and reposition the peyeeye code as the first implementation under it. Any preferences on where the interface should live in the tree, or shape constraints I should know about before I start?

tim-peyeeye:
@siddharth1036 pushed the refactor. The generic decorator now lives in `internal/extproc/wrapper/` (`BodyTransformer` + `Wrapper` interfaces, plus `WrapFactory` to compose it with any `extproc.ProcessorFactory`), and peyeeye is the first implementation under `internal/extproc/wrapper/peyeeye/`. Behaviour is unchanged. Let me know if the shape feels right or if you'd want anything reworked.

internal/controller/controller.go
siddharth1036, line 554: I think this is already removed in main, can you sync your branch with main once?

tim-peyeeye:
@siddharth1036 rebased on main, force-pushed. Should be clean now.

internal/extproc/wrapper/wrapper.go
```diff
@@ -0,0 +1,249 @@
...
+	if body != nil && len(body.Body) > 0 {
+		mutated, err := p.wrapper.OnRequestBody(ctx, body.Body)
+		if err != nil {
+			p.logger.Error("wrapper: OnRequestBody failed; failing closed", slog.Any("error", err))
+			return immediateInternalError("wrapper: failed to transform request body"), nil
+		}
+		body.Body = mutated
+	}
+	resp, err := p.inner.ProcessRequestBody(ctx, body)
+	if body != nil && body.EndOfStream {
```
Hritik003, line 75: why is `close` not invoked when EndOfStream observed only on request half?

aishwaryaraimule21:
> @siddharth1036 makes sense. I'll refactor this around a generic wrapper interface (request body in/out plus response body in/out, with per-request state threaded between them) and reposition the peyeeye code as the first implementation under it. Any preferences on where the interface should live in the tree, or shape constraints I should know about before I start?

@tim-peyeeye currently, the implementation still looks vendor specific as the flag itself is called `-peyeeye`. Can we rather use generic flag like `-pii-provider` which supports values like `peyeeye, presidio (or other PII transformers), none`. 
The filterconfig can also have a generic top level struct with different provider specific configs inside it.

cmd/extproc/mainlib/main.go
```diff
@@ -292,23 +301,42 @@ func Main(ctx context.Context, args []string, stderr io.Writer) (err error) {
...
+		client := peyeeye.NewClient(&cfg, nil)
+		l.Info("peyeeye PII redaction enabled",
+			slog.String("apiBase", cfg.APIBase),
+			slog.String("sessionMode", string(cfg.SessionMode)),
+		)
+		transformer := peyeeye.NewTransformer(client)
+		wrap = func(f extproc.ProcessorFactory) extproc.ProcessorFactory {
+			return wrapper.WrapFactory(f, transformer, l.With("component", "peyeeye"))
+		}
+	}
```
aishwaryaraimule21: can we use vendor-specific factory functions here to support more PII providers in future?

aishwaryaraimule21 (commented):
PII redaction would also be required for MCP requests. LLM and MCP requests have different workflows - MCP is handled by the `mcpproxy` package which has no dependency on `extproc`. The common parts (client, config, text extraction, redaction/rehydration logic) should be reusable across both flows.

Can we move the `peyeeye` package (and the vendor-neutral interfaces it depends on) out of `extproc/wrapper/` into a shared location like `internal/piiredaction`? The extproc-specific decorator (`wrapper.go`) can stay where it is and import the shared package, and the MCP proxy can do the same later.

tim-peyeeye:
@aishwaryaraimule21 both make sense, thanks. Pushed in b2e7db4.

**Generic provider selection.** The `-peyeeye` boolean is gone. There's now a `--pii-provider` string flag: empty disables redaction, `peyeeye` selects the Peyeeye backend, and `presidio` is recognised but reserved until there's an implementation behind it. On the config side `filterapi.Config` carries a provider-neutral `PIIRedactionConfig{ Provider string; Peyeeye *PeyeEyeFilterConfig }`, so adding a backend later is a new sub-field rather than a reshape of the top-level config. Providers register themselves via a small registry (`piiredaction.Register` / `piiredaction.New`), and an unknown or not-yet-implemented provider fails fast at startup listing what's available, instead of silently running with redaction off.

**Shared location.** I moved the vendor-neutral `BodyTransformer` / `Wrapper` interfaces and the Peyeeye implementation out of `internal/extproc/wrapper/` into `internal/piiredaction/` (Peyeeye is now `internal/piiredaction/peyeeye/`). That package has no extproc dependency. The extproc-specific decorator in `wrapper.go` (`WrappingProcessor`, `WrapFactory`) stays where it is and imports `piiredaction`, so the MCP proxy can build a transformer the same way and run it over its own body stream without pulling in extproc. Keeping the registry transport-independent is what makes that reuse possible.

The Peyeeye behaviour itself is unchanged and its tests moved with the code. Happy to adjust the package boundary or the flag/config shape if you'd put either somewhere else.

tim-peyeeye:
@siddharth1036 @aishwaryaraimule21 done.

PR #2092, opened 2026-04-29
mcp: initial implementation of OAuth Token Exchange for upstream auth

### LINKED ISSUE #2036: Feature/Architecture Proposal]: OAuth 2.0 Token Exchange (RFC 8693) Support for SaaS MCP Backends via Wasm/ExtensionPolicy

nacx:
**Description**

Initial implementation for MCP upstream auth with OAuth Token Exchange [1].
The implementation is based on dynamic modules and uses the Built On Envoy Token Exchange [2] extension to provide the data plane functionality.

It takes some of the plumbing and structure defined in the Dynamic Modules PR [3], with some additions to make the dynamic modules part of the `aigw` image and the `e2e` tests.
Since that PR uses an old version of the ABI, I think it is better to merge this one, then rebase the other PR on top of this.
With this structure, we have a nice way of growing the AIGW Dynamic Module with additional features (like token-exchange) that could be needed by the proxy.

Initial implementation of https://github.com/envoyproxy/ai-gateway/pull/2052
Fixes https://github.com/envoyproxy/ai-gateway/issues/2036

This PR should be merged after the MCP `v1beta1` API and not be part of `0.6`. I plan on rebasing hte PR once 0.6 is out.

@mathetake It would be great if you could have a look at this PR and the overall structure and direction, and give feedback. I think it sets a nice structure to make it easier to extend and enhance AIGW easily.
I'll leave comments inline with my thoughts on some of the changes.

1: https://builtonenvoy.io/extensions/token-exchange/
2: https://github.com/envoyproxy/ai-gateway/pull/2052
3: https://github.com/envoyproxy/ai-gateway/pull/1564


cmd/aigw/download_envoy.go
```diff
@@ -17,7 +17,7 @@ import (
 // This matches the version in the .envoy-version file at the repo root. That is ensured in tests.
 // The reason why we don't use the constant defined in the EG as a library is that when we depend
 // on the main branch, it will be a non-released development version that func-e may not have.
-const envoyVersion = "1.37.0"
+const envoyVersion = "1.37.1"
```
nacx, line 20: Starting in Envoy 1.37.1, we have the stable dynamic modules ABI, the token exchange dynamic module is built on top of.
This means that users who want to use this feature will need Envoy >= 1.37.1, but users who do not use token exchange should not have any issues.

cmd/dynamic_module/main.go
```diff
@@ -0,0 +1,13 @@
...
+// the root of the repo.
+
+package main
+
+import (
+	_ "github.com/envoyproxy/envoy/source/extensions/dynamic_modules/sdk/go/abi"
+	_ "github.com/tetratelabs/built-on-envoy/extensions/composer/token-exchange/embedded"
+)
+
+func main() {} // main is required to build as a C shared library.
```
nacx, line 13: This is the same directory as in the DYM pull request. The file is a placeholder that imports the ABI and the token exchange dynamic module code to create the `libaigateway.so`.
The idea is that this nicely aligns with the existing DYM PR.

cmd/dynamic_module/main.go
```diff
@@ -0,0 +1,13 @@
+// Copyright Envoy AI Gateway Authors
+// SPDX-License-Identifier: Apache-2.0
+// The full text of the Apache license is available in the LICENSE file at
+// the root of the repo.
+
+package main
+
+import (
+	_ "github.com/envoyproxy/envoy/source/extensions/dynamic_modules/sdk/go/abi"
+	_ "github.com/tetratelabs/built-on-envoy/extensions/composer/token-exchange/embedded"
```
nacx, line 10: Brings the token exchange features to the aigw DYM.

tests/e2e/testdata/mcp_route_oauth.yaml
```diff
@@ -132,16 +192,120 @@ spec:
...
     kubernetes:
       envoyDeployment:
+        # When creating the Kind cluster we mount the host path where the compiled
+        # dynamic module is located, so we can use it in the Envoy deployment.
+        pod:
+          volumes:
+            - name: dym
+              hostPath:
+                # This path is configured in the Kind configuration file
+                path: /var/dym/libaigateway.so
```
nacx: This is aligned with the updates Kind config in the e2elib setup, so that we can have the compiled dynamic module available for e2e tests. This will be useful for the DYM pull request as well.

Dockerfile.dym
```diff
@@ -0,0 +1,28 @@
+# Copyright Envoy AI Gateway Authors
```
nacx, line 1: cross-compiling with CGO is challenging, so we use this Dockerfile to compile the dynamic module for different platforms so we can easily use it in e2e tests, etc.

Makefile
```diff
@@ -303,6 +310,8 @@ build-e2e: ## Build the docker images for the controller, extproc and testupstre
 	@$(MAKE) docker-build.testupstream CMD_PATH_PREFIX=tests/internal/testupstreamlib DOCKER_BUILD_ARGS="--load"
 	@$(MAKE) docker-build.testmcpserver CMD_PATH_PREFIX=tests/internal/testmcp DOCKER_BUILD_ARGS="--load"
 	@$(MAKE) docker-build.testextauthserver CMD_PATH_PREFIX=tests/internal/testextauth DOCKER_BUILD_ARGS="--load"
+	@$(MAKE) docker-build.testtokenexchange CMD_PATH_PREFIX=tests/internal/testtokenexchangelib DOCKER_BUILD_ARGS="--load"
+	@$(MAKE) docker-build-dynamic-module
```
nacx, line 314: For e2e-tests we build in Docker to make sure we produce a `linux/$GOARCH)` library, no matter the host OS.

Makefile
```diff
@@ -335,12 +344,22 @@ endif
 docker-build.%: ## Build a docker image for a given command.
 	$(eval IMAGE_NAME := $(if $(filter aigw,$(*)),cli,$(*)))
 	$(eval VARIANT := $(if $(filter aigw,$(*)),base-nossl,static))
+	$(eval TARGET := $(if $(filter aigw,$(*)),aigw,default))
 	@$(MAKE) build.$(*) GOOS_LIST="linux" GOARCH_LIST="$(GOARCH_LIST)"
+	@if [ "$(*)" = "aigw" ]; then \
+		echo "-> Building the dynamic module for the Docker target platforms..."; \
+		$(MAKE) docker-build-dynamic-module PLATFORMS="$(PLATFORMS)"; \
```
nacx, line 351: We include the dynamic module in the `aigw` image, so we make sure it is built.

Dockerfile.dym
```diff
@@ -0,0 +1,28 @@
...
+# SPDX-License-Identifier: Apache-2.0
+# The full text of the Apache license is available in the LICENSE file at
+# the root of the repo.
+
+ARG GOLANG_BASE_IMAGE=golang:1.26
```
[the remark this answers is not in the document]
nacx: golang:1.26 is a valid one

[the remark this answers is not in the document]
nacx: It is a valid version

internal/extensionserver/tokenexchange.go
```diff
@@ -0,0 +1,490 @@
...
+
+	token := jwt.NewWithClaims(signingMethod, claims)
+	signed, err := token.SignedString(key)
+	if err != nil {
+		return "", "", fmt.Errorf("failed to sign client assertion JWT: %w", err)
+	}
+	return signed, defaultActorTokenType, nil
+}
+
+func loadSecret(ctx context.Context, k8sClient client.Client, ref *gwapiv1.SecretObjectReference, defaultNS string) (corev1.Secret, error) {
```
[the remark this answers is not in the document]
nacx: For the ApiKeys we are doing the same and it is fine for now, as we don't support reference grants for mcproutes yet. Adding this would be a part of a different PR.

internal/extensionserver/tokenexchange.go
```diff
@@ -0,0 +1,512 @@
...
+		Issuer:    actorTokenConfig.Issuer,
+		Subject:   actorTokenConfig.Subject,
+		IssuedAt:  jwt.NewNumericDate(now),
+		ExpiresAt: jwt.NewNumericDate(exp),
+	}
+
+	var signingMethod jwt.SigningMethod = jwt.SigningMethodRS256
+	if actorTokenConfig.SigningAlgorithm != nil {
+		signingMethod = jwt.GetSigningMethod(*actorTokenConfig.SigningAlgorithm)
+	}
```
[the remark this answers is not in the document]
nacx: There is a check a bit later in the code that handles returning an error if it is invalid.

sandangel:
This is great. I wonder if we can merge this soon.

mohitgurnani1 (commented):
Thanks for this — the API surface closely follows proposal 010 and staging the cache/`clientAssertionJWT` as reserved fields makes sense for an initial implementation.

One design-level question before this merges: the BOE token-exchange module performs a blocking STS callout per request with no token caching (`HttpCallout` + `StopAllAndBuffer` in the module's `plugin.go`). Since MCP traffic is chatty (initialize, tools/list, then every tools/call), every JSON-RPC POST pays a full STS round trip, the STS becomes a hard availability dependency on the hot path (502 on failure, no retry), and STS load scales linearly with request volume. Proposal 010 specified an LRU cache keyed on (subject_token, audience, scopes) as part of the design.

Should token caching land before this feature is usable, or is the intent to ship this as explicitly experimental with the latency caveat documented? Either seems reasonable — but as-is a user enabling this on a production route may be surprised by the added per-request latency and the STS dependency.

internal/extensionserver/tokenexchange.go
```diff
@@ -0,0 +1,512 @@
...
+// parameters sent by the filter to the STS endpoint.
+type tokenExchangePerRouteConfig struct {
+	// Cluster is the Envoy cluster name used to reach the STS endpoint.
+	Cluster string `json:"cluster"`
+	// TokenExchangeURL is the URL of the STS token exchange endpoint.
+	TokenExchangeURL string `json:"token_exchange_url"`
+	// ClientID is the OAuth 2.0 client_id for authenticating the gateway to the STS.
+	ClientID string `json:"client_id,omitempty"`
+	// ClientSecret is the OAuth 2.0 client_secret for authenticating the gateway to the STS.
+	ClientSecret string `json:"client_secret,omitempty"` // #nosec G101
```
mohitgurnani1, line 65: These fields embed the client secret and actor token in plaintext into the per-route config (`TypedPerFilterConfig`), which makes them visible in config dumps, the admin endpoint, and xDS snapshots. This looks like the same class of issue that #2134 fixed for MCP API keys (moving plaintext out of HTTPRoute into a managed Secret + credential injection).

Would it be feasible to resolve these at runtime instead — e.g. SDS, or having the module read a mounted Secret — rather than embedding them at translation time?

internal/extensionserver/tokenexchange.go
```diff
@@ -0,0 +1,512 @@
...
+			return "", "", fmt.Errorf("actor token key \"token\" not found in secret %s/%s", namespace, te.ActorToken.SecretRef.Name)
+		}
+		return string(val), defaultActorTokenType, nil // #nosec G101
+	}
+
+	// Otherwise, generate the token with the provided data
+
+	// If the last generated token is not expired, return it to avoid unnecessary regeneration and signing that
+	// could potentially trigger an unnecessary config change and RDS update in Envoy.
+	if m.token != "" && time.Until(m.expiry) > 30*time.Second {
```
mohitgurnani1, line 415: The comment above acknowledges the RDS-churn risk, but I think there's a second failure mode worth confirming: the actor JWT (default 300s lifetime) is minted at translation time and baked into the route config. What triggers a re-translation before it expires? If no reconcile event happens within the token's lifetime, the JWT in Envoy's config expires and every exchange fails until the next translation. And if something does re-translate on a timer, that's an RDS update roughly every ~4.5 minutes per route.

Minting the actor JWT inside the module (or a runtime provider) would avoid putting a TTL'd credential in xDS entirely.

internal/extensionserver/tokenexchange.go
```diff
@@ -0,0 +1,512 @@
...
+		return "", "", fmt.Errorf("failed to sign client assertion JWT: %w", err)
+	}
+
+	// Cache the generated token and its expiry time to avoid unnecessary regeneration on every request.
+	m.token = signed
+	m.expiry = exp
+	return signed, defaultActorTokenType, nil
+}
+
+func loadSecret(ctx context.Context, k8sClient client.Client, ref *gwapiv1.SecretObjectReference, defaultNS string) (corev1.Secret, error) {
```
mohitgurnani1, line 475: Cross-namespace secret refs are resolved here without a ReferenceGrant check, so any MCPRoute author can read Secrets from arbitrary namespaces via the gateway. Gateway API convention is to require a ReferenceGrant for cross-namespace `SecretObjectReference` — should that be enforced here?

internal/extensionserver/tokenexchange.go
```diff
@@ -0,0 +1,512 @@
...
+	// Scope lists the requested OAuth 2.0 scopes (space-separated) for the issued token.
+	Scope string `json:"scope,omitempty"`
+	// RequestedTokenType specifies the desired type URI for the issued token.
+	RequestedTokenType string `json:"requested_token_type,omitempty"`
+	// ActorToken is the static actor token value used for delegation mode.
+	ActorToken string `json:"actor_token,omitempty"` // #nosec G101
+	// ActorTokenType is the token type URI for the actor_token parameter.
+	ActorTokenType string `json:"actor_token_type,omitempty"`
+	// TimeoutMs is the request timeout in milliseconds for the STS callout.
+	TimeoutMs int `json:"timeout_ms,omitempty"`
```
mohitgurnani1, line 81: `TimeoutMs` is declared but never populated when building the per-route config, so the module falls back to its 5000ms default — a slow STS stalls every request on the route for up to 5s. Worth setting a tighter default here, and possibly exposing it in the `MCPBackendTokenExchange` API.

PR #2101, opened 2026-05-04
feat: add 'validate' command in AI Gateway CLI

Sakalya:
**Description**

Adds aigw validate <paths...> — a new subcommand that statically checks AI Gateway config YAML files for errors without requiring a running Kubernetes cluster or Envoy Gateway.

$ aigw validate config.yaml

  Validating 5 resource(s) from 1 file(s)...

  OK    AIGatewayRoute "default/my-route"
  OK    AIServiceBackend "default/openai-backend"
  ERROR BackendSecurityPolicy "default/my-policy":
    [spec.apiKey] required when type is APIKey
    [spec.targetRefs[0]] references AIServiceBackend "default/ghost" which is not defined in the provided config

  Validation failed: 2 error(s)

### DIFF
--- a/cmd/aigw/validate.go
+++ b/cmd/aigw/validate.go
@@ -1,0 +1,406 @@
+// Copyright Envoy AI Gateway Authors
+// SPDX-License-Identifier: Apache-2.0
+// The full text of the Apache license is available in the LICENSE file at
+// the root of the repo.
+
+package main
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"log/slog"
+
+	aigv1b1 "github.com/envoyproxy/ai-gateway/api/v1beta1"
+)
+
+// cmdValidate is the struct for the `aigw validate` command.
+type cmdValidate struct {
+	Paths []string `arg:"" name:"paths" help:"Path(s) to AI Gateway config YAML file(s) to validate." type:"path" optional:""`
+}
+
+// Validate is called by Kong after parsing to enforce that at least one path was given.
+func (c *cmdValidate) Validate() error {
+	if len(c.Paths) == 0 {
+		return fmt.Errorf("at least one path is required")
+	}
+	return nil
+}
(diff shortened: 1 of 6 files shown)


cmd/aigw/validate.go
```diff
@@ -0,0 +1,406 @@
...
+// validateCmd checks AI Gateway config YAML files for structural and cross-reference errors
+// without requiring a running Kubernetes cluster.
+func validateCmd(_ context.Context, c *cmdValidate, stdout, stderr io.Writer) error {
+	logger := slog.New(slog.NewTextHandler(stderr, &slog.HandlerOptions{}))
+	yamlContent, err := readYamlsAsString(c.Paths)
+	if err != nil {
+		return err
+	}
+
+	aigwRoutes, _, aigwBackends, backendSecurityPolicies, _, _, _, _, err := collectObjects(yamlContent, io.Discard, logger)
```
siddharth1036, line 41: nit, aigwBackends -> aiSvcBackends?

nacx (changes requested):
Thanks for the contribution!

I see the value in having a command that can eagerly validate the configuration, but I think the way it is implemented is going to be extremely hard to maintain and keep in sync with the API.

Every time a change is done to any API, now we'll not only have to update any validation or CEL expression in the API itself, but also update here the custom validation logic. This is IMHO too error-prone and too easy to get out of sync, and once that happens the command provides no value as it can't be trusted.

Instead, I'd suggest finding a way to do, as a starting point, validations that are based solely on the information in the API, without any additional validation code (like we do in the [crdcel tests](https://github.com/envoyproxy/ai-gateway/blob/main/tests/crdcel/main_test.go). That would provide validations that rely only on the API and will not impose any additional maintenance burden or cognitive load, and will always be in sync. Only when that is in place, we should think about extending the validations with additional business logic.

PR #2106, opened 2026-05-04
mcp: filter tools/list response by authorization rules

### LINKED ISSUE #2105: MCPRoute: filter tools/list response based on authorization rules
### Summary

`tools/list` on an MCPRoute currently returns every tool from every backend, filtered only by the route-level static `toolSelectors`. Authorization rules (JWT scopes / CEL) are evaluated only on `tools/call`. This means a caller sees tools they cannot invoke, and only discovers they're unauthorized when the model tries to call one and gets `access denied`.

### Current behavior

- `handleToolsListRequest` aggregates tools from all backends and applies `toolSelectors` during `mergeToolsList`.
- `handleToolCallRequest` evaluates authorization rules (`authorizeRequest`) against JWT claims / CEL, returning `-32099 access denied` on mismatch.
- Net effect: identical `tools/list` for every caller regardless of identity.

### Problem

1. **Poor UX / wasted turns** — LLM clients (Claude, Cursor, etc.) see tools they can't use, attempt them, fail, retry. Users see cryptic `access denied` mid-session instead of just not being offered the tool.
2. **Information disclosure** — the existence and description of restricted tools is visible to any authenticated caller.
3. **Blocks a consolidated `/mcp` endpoint pattern** — today operators have to split into multiple MCPRoutes (`/mcp/dev`, `/mcp/ops`, ...) per persona to get clean tool lists, which defeats one of the nicer aggregation properties of MCPRoute.

### Proposed behavior

Evaluate the same authorization rules (scope / CEL) used by `tools/call` against each tool entry during `mergeToolsList`, and omit entries the caller is not authorized to invoke. Rules that don't target a specific tool should be treated as they are today.

xadereq:
**Description**

`tools/list` previously returned all tools regardless of caller identity, leaving callers to discover authorization failures only at tools/call time. This causes wasted LLM turns and leaks tool names to unauthorized callers.

Apply the same authorization rules used by handleToolCallRequest during mergeToolsList, omitting tools the caller is not authorized to invoke.

Closes https://github.com/envoyproxy/ai-gateway/issues/2105


nacx (commented):
Nice!

When reading the configuration, we are checking if there are changes to the existing tools, and, if there is any change, we send a tools changed notification so that clients can list tools again: https://github.com/envoyproxy/ai-gateway/blob/main/internal/mcpproxy/config.go#L235-L237

We should probably do something similar now if the authorization rules change? That could change the visibility of tools, so probably we should check that as well?

xadereq:
@nacx good catch, just pushed fix with notification signal.

xadereq:
@nacx removed redundant wg.Add that caused wg.Wait() to hang. Unit tests should pass now ;)

nacx (approved):
LGTM. Thanks!

PR #2107, opened 2026-05-04
fix: quick fix finish reason chunk in gemini

hustxiayang:
**Description**
For gemini models/vertex api, when the `finish_reason` is `stop`, the `content` is not empty. For example 
```
  {"candidates": [{"content": {"role": "model", "parts": [{"text": "last word."}]}, "finishReason": "STOP"}]}   
```
However, when the `finish_reason` is something like `MAX_TOKENS`, the content would be empty. In this case, the `delta` would not be set in the chunk.

This is not consistent with openai api. In openai api, there is always a `finish_reason` chunk, and the `delta` always exist. Thus, this pr is to make it follow openai api.


hustxiayang:
@yuzisun  @nacx @aabchoo  please take a look. Thanks a lot!

gavrissh (approved):
lgtm

PR #2108, opened 2026-05-05
feat: respect prefix field for anthropic backend

ajac-zero:
**Description**

The `prefix` field on `VersionedAPISchema` is accepted by the CRD for any schema, but until now it was silently dropped for the Anthropic backend (the request path was hardcoded to `/v1/messages`). This wires `prefix` through `schemaToFilterAPI`, the filter API config, the endpoint spec, and the `anthropicToAnthropicTranslator` so users can route to Anthropic-compatible backends served under a custom prefix (e.g. `/gateway/v1/messages`).

Default behavior is unchanged: when `prefix` is unset the controller defaults it to `"v1"`, preserving today's `/v1/messages` request path.

`AWSAnthropic` and `GCPAnthropic` still ignore the field. The CRD doc comment is updated accordingly.

This change was developed with assistance from Claude Opus 4.7, with all code reviewed and tested by me.


siddharth1036:
@ajac-zero  could you mention the usecases where upstream providers have this requirement of prefix? Afaik, the prefix was introduced for openai providers because the upstream providers had different prefixes which did not fit in the context version field (examples mentioned in the field documentation).
However we can always use version field to make use of custom versions for other providers. The usecase mentioned can also be solved using the version field for anthropic backend.

ajac-zero:
@siddharth1036 Azure AI Foundry (the main service my team uses for LLMs) exposes Anthropic models in an Anthropic-compatible endpoint with this format:

`https://{AZ-RESOURCE}.services.ai.azure.com/anthropic/v1/messages`

Here, the `/anthropic` prefix is mandatory, and so aig doesn't forward it to the proper endpoint currently. Also, the same service exposes OpenAI models in an OpenAI-compatible endpoint with this format:

`https://{AZ-RESOURCE}.services.ai.azure.com/openai/v1/responses`

Which I successfully set up using the `prefix` field to add the `/openai` path, so it would be nice for the Anthropic backend to follow the same pattern, especially since no errors are raised during configuration, and the field is silently dropped at runtime, so I didn't know why it wasn't working for the Anthropic backend initially. As a user, it never occurred to me to try hacking it through the version field 😅.

siddharth1036:
> @siddharth1036 Azure AI Foundry (the main service my team uses for LLMs) exposes Anthropic models in an Anthropic-compatible endpoint with this format:
> 
> `https://{AZ-RESOURCE}.services.ai.azure.com/anthropic/v1/messages`
> 
> Here, the `/anthropic` prefix is mandatory, and so aig doesn't forward it to the proper endpoint currently. Also, the same service exposes OpenAI models in an OpenAI-compatible endpoint with this format:
> 
> `https://{AZ-RESOURCE}.services.ai.azure.com/openai/v1/responses`
> 
> Which I successfully set up using the `prefix` field to add the `/openai` path, so it would be nice for the Anthropic backend to follow the same pattern, especially since no errors are raised during configuration, and the field is silently dropped at runtime, so I didn't know why it wasn't working for the Anthropic backend initially. As a user, it never occurred to me to try hacking it through the version field 😅.

Oh I see, actually this would be a breaking change for anthropic backends as all use cases currently would be using version to deal with such situations. 

I agree that the documentation needs to mention that prefix field would only be effective for openai backends and for other providers version is to be used. I don't see an issue using the prefix field for anthropic but given we have already released beta version for CR so just wanted to check if its absolutely necessary. 

Also this change would need to maintain backward compatibility with version atleast for a release. A reference for this would be : https://github.com/envoyproxy/ai-gateway/pull/1674/changes

siddharth1036:
> > @siddharth1036 Azure AI Foundry (the main service my team uses for LLMs) exposes Anthropic models in an Anthropic-compatible endpoint with this format:
> > `https://{AZ-RESOURCE}.services.ai.azure.com/anthropic/v1/messages`
> > Here, the `/anthropic` prefix is mandatory, and so aig doesn't forward it to the proper endpoint currently. Also, the same service exposes OpenAI models in an OpenAI-compatible endpoint with this format:
> > `https://{AZ-RESOURCE}.services.ai.azure.com/openai/v1/responses`
> > Which I successfully set up using the `prefix` field to add the `/openai` path, so it would be nice for the Anthropic backend to follow the same pattern, especially since no errors are raised during configuration, and the field is silently dropped at runtime, so I didn't know why it wasn't working for the Anthropic backend initially. As a user, it never occurred to me to try hacking it through the version field 😅.
> 
> Oh I see, actually this would be a breaking change for anthropic backends as all use cases currently would be using version to deal with such situations.
> 
> I agree that the documentation needs to mention that prefix field would only be effective for openai backends and for other providers version is to be used. I don't see an issue using the prefix field for anthropic but given we have already released beta version for CR so just wanted to check if its absolutely necessary.
> 
> Also this change would need to maintain backward compatibility with version atleast for a release. A reference for this would be : https://github.com/envoyproxy/ai-gateway/pull/1674/changes

On another thought I checked again, neither prefix nor version was being respected for anthropic providers. So this change would actually be clean wrt prefix. So backward compatibility may not be required.

ajac-zero:
Yes, `version` was previously a no-op, so it shouldn't affect previous releases!
https://github.com/envoyproxy/ai-gateway/blob/dc7816e3957d562367dbdf0e6f02cbe0bcd51d56/internal/translator/anthropic_anthropic.go#L29-L35

sandangel:
hi, I'm hitting this issue as well. Currently have to patch by EnvoyPatchPolicy for each xDS cluster generated. Another way is to patch via Egress policy of service mesh and not enable TLS for Azure Backend. I wonder if we can merge this?

api/v1alpha1/shared_types.go
```diff
@@ -22,6 +22,7 @@ type VersionedAPISchema struct {
 	//
 	// When the name is set to AzureOpenAI, this version maps to "API Version" in the
 	// Azure OpenAI API documentation (https://learn.microsoft.com/en-us/azure/ai-services/openai/reference#rest-api-versioning).
+	// When the name is set to Anthropic, this field is ignored; use prefix to configure custom request paths.
```
siddharth1036: can you please check if any other schema which are no-op for version and document them as well?

ajac-zero: Added docs listing all the schemas where version is a no-op: OpenAI, AWSBedrock, GCPVertexAI, and Anthropic.

siddharth1036:
lgtm

PatilHrushikesh (approved):

ajac-zero:
Hi, any update on this @siddharth1036  @PatilHrushikesh ? Would really love to stop using my fork 😆

siddharth1036:
@ajac-zero can you take a look at the test failures?

ajac-zero:
@siddharth1036 The test pass locally now 👍

siddharth1036 (approved):

sandangel:
@johnugeorge hi we are missing one approval from write access person, could you kindly help us approve?

johnugeorge (approved):

PR #2116, opened 2026-05-07
fix: make streaming token usage emission resilient and absorb cumulative cache_creation from Anthropic message_delta

### LINKED ISSUE #2115: bug: streaming gen_ai.client.token.usage histogram silently drops requests when downstream context is cancelled
_Description_:
For streaming responses (observed on Anthropic-on-AWS-Bedrock with Claude
Sonnet, but the bug is generic to any streaming translator), the
`gen_ai.client.token.usage` Prometheus / OTLP histogram silently misses a
significant fraction of requests, while the Envoy access log
(`gen_ai.usage.input_tokens`, `gen_ai.usage.output_tokens` in dynamic
metadata) captures 100% of the same requests.
Both the histogram emission and the access-log dynamic-metadata population
are gated on the same `body.EndOfStream` chunk in
`internal/extproc/processor_impl.go` (currently lines ~520–545), so we know
EndOfStream IS firing — only the histogram observation is being lost.
Root cause analysis points to the OTEL `Histogram.Record` call being
invoked with the request `ctx`, which on long-running streams can be
cancelled by the downstream client between the last data delivery and the
EndOfStream chunk. Some OTEL exporters / readers can silently drop the
observation in that case.
Expected: every successful streaming request that reaches `body.EndOfStream`
must contribute to `gen_ai.client.token.usage` exactly once.
_Repro steps_:
1. Configure an AIGatewayRoute pointing to AWS Bedrock with Claude Sonnet
   (or any Anthropic streaming backend), with `stream: true`.
2. Send sustained streaming traffic (≥ 30 requests, mix of short and long
   streams). Long streams (≥ 60s) reproduce the drop most reliably.
3. From a downstream client that closes its half of the HTTP/2 connection
   shortly after receiving the final data chunk (e.g. Claude Code with its
   per-request budget), drive ~10 sequential streaming Sonnet requests.
4. Compare:
   - Sum of `gen_ai.usage.input_tokens` from Envoy access logs
     (dynamic metadata, populated on EndOfStream).
   - `sum(increase(gen_ai_client_token_usage_sum{gen_ai_token_type="input"}[10m]))`
     from Prometheus / OTLP for the same window.
In our staging cluster running the unmodified extproc binary, we see ~28%
of Sonnet streaming requests missing from the histogram while access logs
have 100% of them. Haiku (which has ~10× shorter average stream duration)
is unaffected (0% loss), supporting the long-stream / context-cancellation
hypothesis.
_Environment_:
- Envoy AI Gateway: `main` at commit `7687c2ed` (and any earlier release
  containing the streaming `RecordTokenUsage` gate, which has been the same
  since at least the metrics consolidation refactor in #1572 / #1583).
- Envoy Gateway: 1.x.
- Backend: AWS Bedrock — `global.anthropic.claude-sonnet-4-6` and
  `global.anthropic.claude-haiku-4-5`.
- OTEL readers: Prometheus reader (always-on) + autoexport OTLP reader
  (when `OTEL_EXPORTER_OTLP_ENDPOINT` is set).
_Logs_:
Per-request access-log proves EndOfStream IS firing for the affected
requests (10/10 Sonnet, 20/20 Haiku — `response_code=200`,
`gen_ai.usage.input_tokens` populated). Same window's Prom histogram only
contains 7/10 Sonnet requests.
_Suggested fix_:
Make streaming token-usage emission resilient by:
1. Detaching the request context with `context.WithoutCancel(ctx)` before
   calling `Histogram.Record`, so cancellation cannot cause silent drops.
2. Tracking emission with an idempotency flag on the upstream processor so
   multiple emission paths (EndOfStream + a deferred defensive fallback)
   cannot double-count.
3. Adding a deferred defensive fallback in `ProcessResponseBody` that emits
   accumulated usage on EndOfStream chunks where the happy-path emission
   was skipped (e.g. translator returned an error after we had already
   accumulated usage from earlier chunks).
Happy to send a PR with the fix and unit tests covering: cancelled-ctx
emission at EndOfStream, defensive fallback on translator error at
EndOfStream, mid-stream chunks never emitting, and idempotency across
duplicate EndOfStream chunks.

sc7565:
**Description**

For streaming responses (most visibly observed on Anthropic-on-AWS-Bedrock with Claude Sonnet, but the bug is generic to any streaming translator), the `gen_ai.client.token.usage` Prometheus / OTLP histogram silently dropped a meaningful fraction of requests, while the Envoy access log (`gen_ai.usage.input_tokens`, `gen_ai.usage.output_tokens` in dynamic metadata) captured 100% of the same requests. In our staging cluster we saw ~28% of Sonnet streaming requests missing from the histogram while access logs had 100% of them. Haiku (~10× shorter average stream duration) was unaffected.

After deploying the first part of the fix (resilient emission), the histogram was complete, but `gen_ai.client.token.usage{gen_ai_token_type="cache_creation_input"}` was still under-counted by ~37% on Sonnet vs. ground truth from the upstream `usage` block. Root cause: the Anthropic translator only absorbed `output_tokens` from `message_delta.usage`, ignoring the cumulative `cache_creation_input_tokens` / `cache_read_input_tokens` / `input_tokens` that arrive there for "establish-cache" requests where `message_start.usage.cache_creation_input_tokens=0`.

This PR contains two commits that together fix the streaming token-usage histogram so that every successful streaming request that reaches `body.EndOfStream` contributes to `gen_ai.client.token.usage` exactly once, with all four core token-type fields accurately populated.

**1. `internal/extproc/processor_impl.go` — resilient emission (commit 1)**

- Add an `emitTokenUsageOnce(ctx)` helper on `upstreamProcessor` that:
  - Detaches the request `ctx` with `context.WithoutCancel(ctx)` before calling `metrics.RecordTokenUsage`, so a cancelled downstream HTTP/2 stream cannot cause OTEL readers/exporters to silently drop the observation.
  - Tracks emission with a `tokenUsageRecorded bool` flag for idempotency, so multiple emission paths (EndOfStream + a deferred defensive fallback) cannot double-count.
- Call `emitTokenUsageOnce` from the happy-path:
  - Streaming: only on `body.EndOfStream`.
  - Non-streaming: unconditionally on the response.
- Add a deferred defensive fallback in `ProcessResponseBody` that emits accumulated usage on `body.EndOfStream` chunks where the happy-path emission was skipped (e.g. translator returned an error after we had already accumulated usage from earlier chunks).

**2. `internal/extproc/processor_impl.go` + `internal/translator/anthropic_anthropic.go` — accurate cumulative cache fields (commit 2)**

- `internal/translator/anthropic_anthropic.go`: in `reflectStreamingEvent`, when `message_delta.usage` carries any of `cache_creation_input_tokens`, `cache_read_input_tokens`, or `input_tokens` (`> 0`), absorb them as the new cumulative value while preserving anything already established by `message_start` (only fields with `> 0` are absorbed, so a delta that omits a field cannot clobber a valid `message_start` value). The composite `InputTokens` (gross) is recomputed as `regular + cache_read + cache_create`.
- `internal/extproc/processor_impl.go`: tighten the deferred defensive fallback so it only fires on `EndOfStream` when **all four** core token-type fields (`InputTokens`, `CacheCreationInputTokens`, `CachedInputTokens`, `OutputTokens`) are SET on `u.costs`, preventing emission of partially-populated rows from racy / mid-stream defensive paths.

**Tests added**

- `internal/extproc/processor_impl_test.go`:
  - `Test_chatCompletionProcessorUpstreamFilter_ProcessResponseBody/streaming_records_usage_with_cancelled_ctx_at_EndOfStream_(H2)` — reproduces the original bug; verifies `RecordTokenUsage` is invoked exactly once and the context handed to the metric recorder is **not** cancelled.
  - `.../streaming_defensive_fallback_emits_on_translator_error_at_EndOfStream_when_fields_complete` — covers the defensive fallback path with all four token fields set.
  - `.../streaming_defensive_fallback_does_NOT_emit_when_fields_incomplete` — verifies the fallback does not emit a partially-populated histogram row.
  - `.../streaming_mid-stream_chunks_never_emit_token_usage` — guards against premature emission.
  - `.../streaming_idempotent_across_duplicate_EndOfStream_chunks` — guards idempotency.
  - `.../non-streaming_records_usage_once_with_detached_ctx` — coverage for the non-streaming path.
- `internal/translator/anthropic_anthropic_test.go`:
  - `TestAnthropicToAnthropic_ResponseBody_streaming_establishCache` — reproduces the cache_creation undercount: `message_start.usage.cache_creation_input_tokens=0`, then `message_delta.usage.cache_creation_input_tokens=N`. Asserts the translator absorbs the cumulative value.
  - `TestAnthropicToAnthropic_ResponseBody_streaming_messageDelta_preservesMessageStartCache` — asserts that a `message_delta` which omits cache fields does **not** clobber valid values established by `message_start`.

`make format lint` is clean and all tests in `internal/translator/...`, `internal/extproc/...`, `internal/metrics/...` pass.

**Validation in staging**

After deploying both commits to our staging cluster:

- `sum(rate(gen_ai_client_token_usage_count[15m])) by (gen_ai_request_model)` matches the Envoy access-log request count for Sonnet streaming requests (no more silent histogram drops).
- `sum(rate(gen_ai_client_token_usage_sum{gen_ai_token_type="cache_creation_input"}[15m])) by (gen_ai_request_model)` for Sonnet matches the cumulative `cache_creation_input_tokens` from the upstream `usage` block (no more ~37% undercount).

Fixes #2115

- The two commits are kept separate intentionally because they fix two distinct (but related) bugs in the same end-to-end flow. Happy to squash on merge if preferred.
- The `context.WithoutCancel` change is intentionally applied to the OTEL metric path only, not to access-log dynamic-metadata population (which already works on the same `EndOfStream` chunk and is reliably captured for 100% of requests).
- Per the project's generative-AI policy: AI assistance was used while authoring this PR; the author reviewed, tested, and validated all changes against ground truth in a real Sonnet/Haiku streaming workload before submitting.

Signed-off-by: sc7565 <sc7565@users.noreply.github.com>

### DIFF
--- a/internal/extproc/mocks_test.go
+++ b/internal/extproc/mocks_test.go
@@ -182,6 +182,12 @@ type mockMetrics struct {
 	cachedInputTokenCount        int
 	cacheCreationInputTokenCount int
 	outputTokenCount             int
+	// recordTokenUsageCallCount tracks how many times RecordTokenUsage was invoked.
+	// Useful for asserting idempotent emission for streaming responses.
+	recordTokenUsageCallCount int
+	// recordTokenUsageCtxErrs captures ctx.Err() observed at each RecordTokenUsage
+	// invocation so tests can verify the context was detached from cancellation.
+	recordTokenUsageCtxErrs []error
 	// streamingOutputTokens tracks the cumulative output tokens recorded via RecordTokenLatency.
 	streamingOutputTokens int
 	timeToFirstToken      float64
@@ -212,7 +218,9 @@ func (m *mockMetrics) SetResponseModel(responseModel internalapi.ResponseModel) 
 func (m *mockMetrics) SetBackend(backend *filterapi.Backend) { m.backend = backend.Name }
 
 // RecordTokenUsage implements [metrics.Metrics].
-func (m *mockMetrics) RecordTokenUsage(_ context.Context, usage metrics.TokenUsage, _ map[string]string) {
+func (m *mockMetrics) RecordTokenUsage(ctx context.Context, usage metrics.TokenUsage, _ map[string]string) {
+	m.recordTokenUsageCallCount++
+	m.recordTokenUsageCtxErrs = append(m.recordTokenUsageCtxErrs, ctx.Err())
 	if input, ok := usage.InputTokens(); ok {
 		m.inputTokenCount += int(input)
 	}
--- a/internal/extproc/mocks_test.go
+++ b/internal/extproc/mocks_test.go
@@ -9,6 +9,7 @@ import (
 	"context"
 	"io"
 	"log/slog"
+	"sync"
 	"testing"
 	"time"
 
@@ -170,7 +171,16 @@ func (m *mockMetricsFactory) NewMetrics() metrics.Metrics {
 }
 
 // mockMetrics implements [metrics.Metrics] for testing.
+//
+// All mutable fields are guarded by `mu` so concurrent invocations from
+// tests that exercise the race-detector-friendly idempotency guards in
+// upstreamProcessor (e.g. concurrent ProcessResponseBody emissions) do not
+// themselves race on the mock's own internal bookkeeping. The production
+// metricsImpl uses thread-safe OTEL instruments, so this mutex models the
+// "what tests actually want to observe" semantics, not a production
+// hot-path constraint.
 type mockMetrics struct {
+	mu                           sync.Mutex
 	requestStart                 time.Time
 	originalModel                string
 	requestModel                 string
@@ -197,28 +207,44 @@ type mockMetrics struct {
 }
 
 // StartRequest implements [metrics.Metrics].
-func (m *mockMetrics) StartRequest(_ map[string]string) { m.requestStart = time.Now() }
+func (m *mockMetrics) StartRequest(_ map[string]string) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	m.requestStart = time.Now()
+}
 
 // SetOriginalModel implements [metrics.Metrics].
 func (m *mockMetrics) SetOriginalModel(originalModel internalapi.OriginalModel) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
 	m.originalModel = originalModel
 }
 
 // SetRequestModel implements [metrics.Metrics].
 func (m *mockMetrics) SetRequestModel(requestModel internalapi.RequestModel) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
 	m.requestModel = requestModel
 }
 
 // SetResponseModel implements [metrics.Metrics].
 func (m *mockMetrics) SetResponseModel(responseModel internalapi.ResponseModel) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
 	m.responseModel = responseModel
 }
 
 // SetBackend implements [metrics.Metrics].
-func (m *mockMetrics) SetBackend(backend *filterapi.Backend) { m.backend = backend.Name }
+func (m *mockMetrics) SetBackend(backend *filterapi.Backend) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	m.backend = backend.Name
+}
 
 // RecordTokenUsage implements [metrics.Metrics].
 func (m *mockMetrics) RecordTokenUsage(ctx context.Context, usage metrics.TokenUsage, _ map[string]string) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
 	m.recordTokenUsageCallCount++
 	m.recordTokenUsageCtxErrs = append(m.recordTokenUsageCtxErrs, ctx.Err())
 	if input, ok := usage.InputTokens(); ok {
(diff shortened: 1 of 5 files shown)


sc7565:
Pushing a third commit (`2c5f38e`) on this branch — please re-review.

After deploying commits 1 and 2 to our staging cluster, we observed a new failure mode: the OTEL `gen_ai_client_token_usage_sum` histogram was now over-counting requests. Empirically:

- Haiku non-streaming: clean **2.0×** the ground-truth count and sums (3 actual requests → 6 emissions).
- Sonnet streaming: **1.74-1.93×** depending on token type / 1.78× request count (32 actual requests → 57 emissions).
- Elasticsearch access-log dynamic-metadata captured 100% of requests at 1.00× — i.e. the access-log path was correct, the histogram path was double-counting.

Root cause: the same `upstreamProcessor` instance is reachable from two ext_proc gRPC streams in production (the router stream and the upstream-filter stream), which run on separate goroutines and can call `ProcessResponseBody` concurrently for the EndOfStream chunk. The previously-plain `bool` `tokenUsageRecorded` flag was racy: two goroutines could both observe `false`, both flip it to `true`, and both call `RecordTokenUsage`. Concurrent access to `u.costs` (translator-returned TokenUsage absorbed via `Override`) and the streaming gzip state (`u.compressedBuf`, `u.decompressedOffset`) was also unprotected.

`go test -race` reproduces both the data-race report on `u.costs.Override` and the double-emission count.

Commit `2c5f38e` ("fix(extproc): serialise ProcessResponseBody and use atomic.Bool for token-usage idempotency") fixes this by:

1. Adding `responseBodyMu sync.Mutex` on `upstreamProcessor` and locking it for the duration of every `ProcessResponseBody` call. This serialises concurrent invocations on the same processor and protects all per-request mutable state. The mutex is uncontended in the common case (one chunk at a time per processor) and only does work in the race scenarios.
2. Converting `tokenUsageRecorded bool` → `atomic.Bool` and using `CompareAndSwap` in both `emitTokenUsageOnce` and the new `emitTokenUsageBackstop`. Belt-and-suspenders alongside the mutex; also lets the helper be safely called from any future unsynchronised path.
3. Splitting emission into two helpers:
   - `emitTokenUsageOnce` — gated by the four-field completeness check from commit 2 (claim → check fields → release-claim-and-skip if incomplete, or record).
   - `emitTokenUsageBackstop` — bypasses the field gate, invoked only from the deferred close on `EndOfStream` as a last-resort backstop. Records whatever `u.costs` has so the request still appears in `gen_ai_client_token_usage_count`. We deliberately prefer "emit a partial observation" over "lose the request" on rare error paths.
4. Dropping the v2 narrower-than-needed gated check inside the defer; the new backstop fires for both streaming AND non-streaming on `EndOfStream`, which means OpenAI-style backends without cache fields (where the gated `emitTokenUsageOnce` skips because the field-completeness gate fails) still produce exactly one histogram observation via the backstop.

Three new sub-tests cover the race scenarios under `-race`:

- `concurrent_EndOfStream_emit_attempts_record_token_usage_exactly_once` — 32 goroutines hammer the same processor; passes only with the atomic flag + mutex.
- `non-streaming_without_cache_fields_emits_once_via_backstop` — Haiku/OpenAI-style payload (input + output, no cache fields) must still produce exactly one histogram observation via the EndOfStream backstop.
- `non-streaming_Haiku-like_duplicate_EndOfStream_concurrent_invocations_record_once` — 16 concurrent invocations on the same processor; reproduces the production "Haiku 2.0×" signature.

The previously-flipped v2 sub-test `streaming_defensive_fallback_does_NOT_emit_when_fields_incomplete` is renamed to `streaming_defensive_backstop_emits_partial_usage_on_translator_error_at_EndOfStream_when_fields_incomplete` with the assertion flipped (visibility-over-precision is the intended v3 posture).

`go test -race -count=10` on the concurrent sub-tests passes consistently. `make format lint` is clean on the changed packages.

I'd be happy to squash these three commits down to one ("fix: streaming token usage emission resilient and accurate") on merge if preferred.

PR #2122, opened 2026-05-10
feat: support Azure OpenAI Responses API

### LINKED ISSUE #2110: AzureOpenAI backend does not support Responses API (/v1/responses)
The OpenAI schema gained Responses API support in v0.5.0, but the AzureOpenAI
  schema still only routes to /chat/completions. Sending a request to /v1/responses                            
  with x-ai-eg-model pointing to an AzureOpenAI backend returns:                                               
                                                                                                               
    "unsupported API schema: backend={AzureOpenAI 2025-04-01-preview}"                                         
                                                                                                               
  Use case: web_search_preview tool requires the Responses API endpoint                                        
  (/openai/deployments/{deployment}/responses?api-version=...) which Azure supports
  since API version 2025-04-01-preview, but the gateway has no translation layer for it.                       
                                                                                                               
  Expected: AzureOpenAI backend translates /v1/responses to Azure's                                            
    /openai/deployments/{deployment}/responses?api-version={version}                                           
    with proper OAuth token injection, the same way it does for chat/completions.                              
                                                                                                               
  Gateway version: pre-v0.6.0 (commit 777421a), envoy-gateway v1.7.1

ajac-zero:
**Description**

This commit adds Azure OpenAI support for the OpenAI-compatible `/v1/responses` endpoint by routing Responses requests to Azure’s `/openai/responses?api-version=...` path while preserving existing OpenAI Responses request and response handling.

**Related Issues/PRs**

Close #2110 

A new translator was needed because the Azure OpenAI Responses-compatible endpoint requires specific query parameters, such as `api-version`, and I didn't want to modify the OpenAI translator to include Azure-specific implementation details.

This PR was prepared with assistance from gpt-5.5. I reviewed and take ownership of the changes.


internal/translator/openai_azureopenai.go
```diff
@@ -60,3 +71,22 @@ func (o *openAIToAzureOpenAITranslatorV1ChatCompletion) RequestBody(raw []byte,
...
+// RequestBody implements [OpenAIResponsesTranslator.RequestBody].
+func (o *openAIToAzureOpenAITranslatorV1Responses) RequestBody(raw []byte, req *openai.ResponseRequest, forceBodyMutation bool) (
+	newHeaders []internalapi.Header, newBody []byte, err error,
+) {
+	newHeaders, newBody, err = o.openAIToOpenAITranslatorV1Responses.RequestBody(raw, req, forceBodyMutation)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	newHeaders[0] = internalapi.Header{pathHeaderName, fmt.Sprintf("/openai/responses?api-version=%s", o.apiVersion)}
```
nacx: Should we just append the query param to the value and respect the configured path in the current header value?

nacx (commented):
Thanks! Overall LGTM. Just one small nit.

internal/translator/openai_azureopenai.go
ajac-zero: Good point, it might be relevant in some edge cases. I added support for just appending the query param to the configured path.

Caveat: the chat/completions and embeddings Azure translators currently don't respect the configured path either. I left them as-is, so this PR stays more focused. I would argue it would only be worth revisiting if someone actually has a need for it.

nacx (approved):

PR #2127, opened 2026-05-11
feat: add OpenAI to Anthropic translator

### LINKED ISSUE #1936: Support for OpenAI to Anthropic schema translator
When calling with openAI schema on `/v1/chat/completions` endpoint, gateway crashes with: `unsupported API schema: backend={Anthropic 2023-06-01}"`

I cannot find in the code a translator from OpenAI schema to Anthropic schema, would be nice to support this use case as well. Let me know if it is something that I can work on.

ajac-zero:
**Description**

This commit adds OpenAI-compatible Chat Completions support for the native Anthropic provider. It translates `/v1/chat/completions` requests to Anthropic's `/v1/messages` API, including non-streaming responses, streaming responses, and errors.

The native Anthropic implementation is a small adapter around the established GCP Anthropic translator. It reuses shared request conversion, response translation, streaming, tracing, redaction, and Anthropic JSON error handling while handling the direct provider's transport differences:

- placing the model in the request body
- targeting `/v1/messages`
- setting the `anthropic-version` header
- labeling non-JSON fallback errors for the native provider
- reporting the response model from native Anthropic streaming events

The change also adds focused translator and data-plane coverage and updates the supported-provider documentation and Anthropic Chat Completions example.

Fixes #1936

GCP Anthropic is used as the implementation base because both providers return Anthropic `Message` JSON and native Anthropic SSE. AWS Anthropic adds Bedrock-specific paths, error headers, and binary EventStream framing, so it is not a suitable base for the direct provider adapter.

A provider-neutral Anthropic core with separate native, GCP, and AWS transport adapters may be a useful future refactor, but it is intentionally outside this PR to avoid modifying established provider paths.

This change was developed with assistance from gpt-5.5 and gpt-5.6, with all code reviewed and tested by me.


PatilHrushikesh:
are you still working on this?

ajac-zero:
I confirm I am still working on this!

ajac-zero:
@PatilHrushikesh Ready for review when you get the chance!

ajac-zero:
I've been dogfooding this implementation for a bit in our deployment, and I found some rough edges that I think are worth mentioning here:

1. Omitted `max_tokens` defaulted to 0 in OpenAI clients -> Received empty responses:
This was a really confusing silent behavior. Since OpenAI clients set `max_tokens` to null by default, the gateway was defaulting that field to 0 before forwaring to Anthropic backends. A request with `max_tokens=0` is actually an official technique to populate the prompt cache, but it results in 200 responses with `content=None`. 
The proposed solution is to add a validation in the openai->anthropic path that rejects requests that omit `max_tokens`, essentially turning it into a required field within this path. This was prefered in order to keep behavior explicit, instead of defaulting to an arbitrary number like 1024 which could silently affect responses.

2. Missing mapping between `thinking` (Anthropic) and `reasoning_effort` (OpenAI):
Unfortunately the mapping is not very straightforward, because `thinking` requires setting a specific int budget instead of `low|medium|high` like other APIs. For now, it can still be done from openai->anthropic via the `extra_body` field in requests:

```python
from openai import OpenAI

client = OpenAI(
    base_url="http://<gateway-host>/v1",
    api_key="<gateway-api-key>",
)

response = client.chat.completions.create(
    model="claude-sonnet-4-6",
    messages=[
        {
            "role": "user",
            "content": "Explain why the sky is blue in one paragraph.",
        }
    ],
    max_tokens=1024,
    extra_body={
        "thinking": {
            "type": "enabled",
            "budget_tokens": 4096,
        }
    },
)

```

JAORMX:
# Panel review — OpenAI → first-party Anthropic translator

I ran a three-axis panel review (Spec / Standards / Domain) against this PR, with a parallel comparison PR (#2284) as a reference. Overall this is a well-executed translator that slots cleanly into the existing family — the redaction consolidation and streaming-model tracking are the right calls. Below are the findings, ordered by priority.

A kind note up front: please add `Closes #1936` to the PR body so the issue it fixes gets referenced and auto-closed — right now #1936 is untagged and hard to discover from the issue side.

---

## Spec — does this implement what was asked?

Issue #1936 asks to stop the `unsupported API schema: backend={Anthropic …}` crash and add an OpenAI→Anthropic translator. Both are shipped.

**Scope creep (all real, flagging for conscious sign-off):**
- **`max_tokens` made required across ALL Anthropic paths** — the check landed in the shared `buildAnthropicParams`, so it changes behavior for existing AWS + GCP Anthropic users, not just the new direct path. The issue only asked for the direct translator.
- **Anthropic `thinking` config forwarding removed** in the *reverse* direction (`openai_helper.go`) — unrelated to #1936.
- **Redaction refactored across 3 sibling translators** — reasonable, not asked for.
- **AWS error handling rewritten** (`openai_awsanthropic.go:110-142`) — not implied by the spec.

---

## Ship-blockers (1)

### [High] The `max_tokens` 422 fires *before* body mutation, breaking the documented workaround

`anthropic_helper.go` rejects an omitted `max_tokens` from `translator.RequestBody`. That runs at `processor_impl.go:338` and short-circuits to a 422 `ImmediateResponse` (`processor_impl.go:340-346`) **before** `applyBodyMutation` at `processor_impl.go:378`. So a route-level request-body mutation can no longer inject `max_tokens` to rescue such a request — yet `site/docs/getting-started/connect-providers/aws-bedrock.md:278` explicitly tells operators:

> *"if your clients do not include it, you must inject it via a request body mutation."*

An operator following the docs now gets the 422 regardless and has no thread to pull. **Fix:** either move the check to operate on the post-mutation body, or update the docs to say the client *must* send `max_tokens` and a body mutation cannot supply it.

---

## Cross-confirmed (1)

### [High] Behavior change with no disclosure: missing `max_tokens` now returns gateway 422 instead of backend 400, for GCP/AWS Anthropic backends

Previously a missing `max_tokens` against `gcp-anthropicai` / `aws-anthropic` returned HTTP 400 from the backend (the old data-plane tests documented this). After this PR the same request returns HTTP 422 from the gateway. This affects existing, already-supported backends (GCP/AWS Anthropic predate this PR). Operators with alerting keyed on 400, or clients retrying on backend 400, see a different signal with no release note. Worth a release-notes entry stating the status change and the new required field.

---

## Mechanical fixes (3)

- **[Medium] `max_tokens` validation is in the wrong layer.** `buildAnthropicParams` is a *translation* helper; reject-vs-forward is a *translator-level* policy. Embedding it in the shared path makes a loud cross-family behavior change silent — a fourth Anthropic-family translator added later will inherit the reject policy with no signal. Move the 3-line check into each translator's `RequestBody`, OR keep it in the helper with a one-line comment stating it's a deliberate cross-family policy + a release note.
- **[Medium] `redactAnthropicChatCompletionResponse` is mis-named and mis-located.** It operates purely on `*openai.ChatCompletionResponse` — nothing Anthropic-specific. "Anthropic" in the name is wrong; it belongs with the OpenAI redaction surface (`openai_openai.go`, where `redactReasoningContent` lives), not `anthropic_helper.go`. Rename to `redactChatCompletionResponse` and relocate. The Rule-of-Three was passed long ago (5+ sites); this PR stopped at 3.
- **[Medium] `openai_awsanthropic.go` is the missing fourth Anthropic translator** — it has no `RedactBody` / `SetRedactionConfig` and no debug-log block, unlike its three siblings the PR consolidated. Either wire it to the shared helper or leave a comment explaining the exclusion.

---

## Judgement calls (3)

- **[Medium] Unbounded read of untrusted backend bodies** (`io.ReadAll` with no `io.LimitReader`, `openai_anthropic.go:76-87`/`:104-113`, plus the AWS streaming path). CWE-400 / OWASP API4:2023. Requires a malicious upstream (semi-trusted per threat model), partially mitigated by Envoy's buffer limits + gRPC `maxRecvMsgSize`. This PR adds a *new instance* of a codebase-wide pattern. Fix is a one-line `io.LimitReader` wrapper at the boundary.
- **[Low] `params.Model = o.requestModel` reach-through** — the helper has an undocumented "leaves `Model` unset" contract. The PR adds a good comment at the reach-through site; add the matching contract to `buildAnthropicParams`'s doc comment. One line.
- **[Low] Constructor double-default on prefix.** `openai_anthropic.go` defaults `"" → "v1"`, but the only caller already passes `schema.AnthropicPrefix()` which itself defaults to `"v1"`. Redundant, and diverges from the OpenAI sibling where `""` yields no prefix. Drop the constructor-level default.

---

## Polish (4)

- **[Low] Docs contradiction worsens.** `supported-providers.md` says Anthropic = "Support only Native Anthropic messages endpoint" — now stale. `supported-endpoints.md:402` already claims Anthropic ✅ Chat Completions. The Chat Completions provider list at `supported-endpoints.md:37-42` omits Anthropic. Update both to match the new reality.
- **[Low] No runnable Anthropic Chat Completions example** with `max_tokens` set — the existing example uses `gpt-4o-mini` with no `max_tokens`, which now 422s against Anthropic.
- **[Low] `anthropic-version` is hardcoded** to `2023-06-01` with no override knob, unlike the GCP sibling which threads `apiVersion`. The Anthropic docs confirm the header is required and the SDK pins this version, so it's defensible — but PR #2284 shows the override is a one-liner.
- **[Info] Backend error messages pass through unredacted** to the OpenAI client. Intentional and tested across the family; noted so it isn't re-flagged.

---

## Test-coverage gaps (compared to the parallel #2284, which is being closed in favor of this one)

A few things #2284 did better that are worth lifting over:

- **Adversarial SSE test coverage.** #2284's test file has dedicated cases for: empty content, `tool_use` without text, **truncated/malformed SSE data lines**, **missing `\n\n` event boundaries with partial-buffer flush across calls**, and **mid-stream `error` event surfacing**. This PR has none of these — its streaming tests are happy-path only. The streaming parser is shared infrastructure; a malformed-SSE regression here breaks all four Anthropic translators. This is the single biggest test-coverage gap.
- **Data-plane integration tests for the new path.** #2284 adds 3 end-to-end cases (non-stream, stream, error) exercising `/v1/chat/completions` → Anthropic through the real extproc harness. This PR only retargets two existing "missing max_tokens" cases; it never proves the new translator works end-to-end through the harness.
- **Docs update.** #2284 adds Anthropic to the `supported-endpoints.md` provider list; this PR leaves the contradiction above.

---

## Summary

- **Spec:** 0 missing, 4 scope-creep items (the `max_tokens` one is the load-bearing concern), 0 wrong.
- **Standards:** 0 hard violations, 1 convention gap (`Closes #1936` missing from PR body).
- **Domain:** 1 ship-blocker (docs/workaround breakage), 1 cross-confirmed (undisclosed 400→422 behavior change), 3 mechanical, 3 judgement, 4 polish.

**Most important single issue:** the `max_tokens` 422 fires before body mutation, breaking the documented operator workaround at `aws-bedrock.md:278`. Fix the ordering or fix the docs before merge.

Great work overall — the design wins (configurable prefix, shared redaction, streaming response-model, `thinking` fix) are real and worth keeping. Happy to help land any of the mechanical fixes.

ajac-zero:
I substantially reworked this PR after recognizing that the existing OpenAI → GCP Anthropic translator already implements almost all the protocol behavior needed for first-party Anthropic.

The previous implementation introduced a separate, largely duplicated translator and accumulated several adjacent changes while the branch was open. This made the PR larger than necessary and increased review burden and regression risk.

Direct Anthropic is now a small adapter around the established GCP Anthropic translator:

- Request parameter conversion remains shared through `buildAnthropicParams`.
- Response translation, streaming, tracing, redaction, and Anthropic JSON error handling reuse the existing GCP path.
- The direct adapter only handles the actual provider differences:
  - model placement in the request body
  - `/v1/messages` path construction
  - `anthropic-version` header
  - direct-provider fallback error type
  - response-model tracking from native Anthropic streaming events

GCP is the practical base because its response body and streaming transport closely match the first-party API: both return Anthropic `Message` JSON and native Anthropic SSE. AWS Anthropic would be a poorer base because it adds Bedrock-specific paths, error headers, and binary EventStream framing.

Longer term, a provider-neutral Anthropic core with separate native, GCP, and AWS transport adapters would be the cleanest architecture. Extracting that abstraction here would increase scope and unnecessarily modify established paths. Reusing the proven GCP implementation is the smallest and safest way to add the requested capability.

This pivot also addresses the main concerns raised by @JAORMX:

- Removed the unrelated reverse Anthropic → OpenAI `thinking` change.
- Removed the shared redaction refactor and AWS error-handling rewrite.
- Avoided adding new unbounded response-body reads by reusing the existing response implementation.
- Preserved existing AWS and GCP Anthropic behavior.
- Added focused direct Anthropic data-plane coverage for non-streaming, streaming, and error responses.
- Updated provider documentation and the runnable Anthropic Chat Completions example.
- Documented the shared request builder’s model-placement contract.
- Removed redundant constructor-level prefix defaulting.

The effective PR diff is now limited to the direct-provider adapter, the small shared streaming change needed to report Anthropic’s response model, focused tests, endpoint registration, and documentation. Complex behavior remains delegated to established code already exercised by the GCP Anthropic test suite.

I believe this version is substantially easier to review and less likely to introduce regressions while still fully resolving #1936.

JAORMX (approved):

pmerrison:
I hit this gap independently and filed #2529 before finding this PR — closing that as a duplicate of #1936, but bringing over two things that aren't on either thread, because I think together they reframe how urgent this is.

### 1. The shipped Anthropic quickstart is broken on `main` today

`examples/basic/anthropic.yaml` declares the backend as:

```yaml
spec:
  schema:
    name: Anthropic
```

and step 4 of [Connect Anthropic](https://github.com/envoyproxy/ai-gateway/blob/main/site/docs/getting-started/connect-providers/anthropic.md) — "Test Chat Completions" — then curls `$GATEWAY_URL/v1/chat/completions` against exactly that backend. Since `ChatCompletionsEndpointSpec.GetTranslator` has no `case filterapi.APISchemaAnthropic`, that request cannot do anything but fall through to `unsupported API schema: backend=Anthropic`.

So a user following the Anthropic getting-started guide verbatim fails at the guide's own verification step. That makes this a docs-correctness bug as much as a missing feature, and I think it's the single strongest argument for landing this PR — note that this PR's one-line addition of `"max_tokens": 1024` to that same curl only makes sense *once the translator exists*.

### 2. The workaround users find instead is a silent cost regression

With the quickstart broken, the way to get Anthropic behind a chat-completions route today is to declare the backend as `APISchema.Name: OpenAI`, which selects the `openai→openai` passthrough (`openai_openai.go:56`). The body is forwarded byte-for-byte to Anthropic's OpenAI-compatibility endpoint, which ignores `cache_control`. The request **succeeds**, nothing is cached, full input price is paid every turn, and no error is surfaced anywhere.

Measured against `api.anthropic.com`, the same body sent twice with a ~22k-token stable prefix and `cache_control: {"type": "ephemeral"}` on both the system block and a tool definition:

```jsonc
// via /v1/chat/completions declared as APISchema OpenAI (the workaround) — both calls identical,
// and byte-identical to the same request with no cache_control at all:
{"completion_tokens": 4, "prompt_tokens": 22511, "total_tokens": 22515}

// the same content via /v1/messages, which does get a native translator:
call 1: {"cache_creation_input_tokens": 22427, "cache_read_input_tokens": 0,     "input_tokens": 84}
call 2: {"cache_creation_input_tokens": 0,     "cache_read_input_tokens": 22427, "input_tokens": 84}
```

That second call bills ~9.6x cheaper. Operators on this path are paying the difference today with no signal that anything is wrong.

### This PR already fixes both, at no extra cost

Because `RequestBody` here goes through `buildAnthropicParams`, it inherits the cache-breakpoint propagation that path already does — tools (`anthropic_helper.go:190`), text/image/document content parts, system blocks, and tool results, ~15 sites in total. That's the same helper `openai_awsanthropic.go:73` and `openai_gcpanthropic.go:59` call, which is exactly why AWSAnthropic and GCPAnthropic already carry `cache_control` correctly from chat completions and direct Anthropic is the only Anthropic-flavoured backend that doesn't. Nothing extra is needed here — the reuse gets it for free.

For what it's worth, having read the diff: reusing the GCP translator rather than duplicating it looks like the right call. Both backends return Anthropic `Message` JSON and native SSE, so the only real deltas are model placement, path, and the version header — which is precisely what the override covers.

### One small suggestion

`openai_anthropic_test.go` doesn't currently assert the caching behaviour. Given it's the practical payoff of the change and it arrives implicitly through a shared helper, a case that sets `cache_control` on a system content part and on `tools[i].function` and asserts both survive into the marshalled body would pin it down against future refactors of that helper. Happy to open a PR against your branch with it if that's useful.

---

CI is green across all 34 checks and the branch is `MERGEABLE` (just behind `main`). This has been sitting at `REVIEW_REQUIRED` since 20 July. Could a maintainer take a look?

internal/translator/openai_anthropic.go
```diff
@@ -0,0 +1,95 @@
...
+type openAIToAnthropicTranslatorV1ChatCompletion struct {
+	openAIToGCPAnthropicTranslatorV1ChatCompletion
+	path string
+}
+
+// RequestBody implements [OpenAIChatCompletionTranslator.RequestBody].
+func (o *openAIToAnthropicTranslatorV1ChatCompletion) RequestBody(_ []byte, openAIReq *openai.ChatCompletionRequest, _ bool) (
+	newHeaders []internalapi.Header, newBody []byte, err error,
+) {
+	params, err := buildAnthropicParams(openAIReq, "Anthropic", o.modelNameOverride)
```
kanurag94: I think this still leaves `max_tokens` to 0 when client omits it. Same thing that you commented about. WDYT about adding a line here itself:

```
if params.MaxTokens == 0 {
  return nil, nil, fmt.Errorf("%w: max_tokens is required by the Anthropic Messages API", internalapi.ErrInvalidRequestBody)
}
```

kanurag94: Also, if possible please add a UT for this so we don't miss it

internal/translator/anthropic_helper.go
```diff
@@ -681,13 +681,13 @@ func mapReasoningEffortToOutputConfigEffort(reasonEffort openai.ReasoningEffort)
...
-// buildAnthropicParams is a helper function that translates an OpenAI request
-// into the parameter struct required by the Anthropic SDK.
+// buildAnthropicParams translates an OpenAI request into Anthropic SDK parameters.
+// It leaves Model unset because cloud providers identify the model in the request path.
 // The apiSchema parameter indicates the backend API schema (e.g., "AWSAnthropic", "GCPAnthropic").
 func buildAnthropicParams(openAIReq *openai.ChatCompletionRequest, apiSchema string, modelNameOverride internalapi.ModelNameOverride) (params *anthropic.MessageNewParams, err error) {
 	// 1. Handle simple parameters.
 	// max_tokens is required by the Anthropic API but optional in the OpenAI API.
-	// If not set, pass 0 and let the Anthropic API reject the request.
+	// If not set, preserve the zero value for the provider to interpret.
```
kanurag94: Once we fix the 0 value, we need to revisit this comment.

internal/translator/openai_anthropic.go
```diff
@@ -0,0 +1,95 @@
...
+	return &openAIToAnthropicTranslatorV1ChatCompletion{
+		openAIToGCPAnthropicTranslatorV1ChatCompletion: openAIToGCPAnthropicTranslatorV1ChatCompletion{
+			modelNameOverride: modelNameOverride,
+		},
+		path: path.Join("/", prefix, "messages"),
+	}
+}
+
+// openAIToAnthropicTranslatorV1ChatCompletion reuses the GCP Anthropic
+// response translation; direct Anthropic only differs in request construction.
```
kanurag94: nit: this looks wrong. It has more differences.

internal/translator/anthropic_helper.go
```diff
@@ -681,13 +681,13 @@ func mapReasoningEffortToOutputConfigEffort(reasonEffort openai.ReasoningEffort)
 	}
 }
 
-// buildAnthropicParams is a helper function that translates an OpenAI request
-// into the parameter struct required by the Anthropic SDK.
+// buildAnthropicParams translates an OpenAI request into Anthropic SDK parameters.
+// It leaves Model unset because cloud providers identify the model in the request path.
```
kanurag94: nit: I think it is set by transport. This comment may be wrong.

kanurag94 (commented):
thanks for the great work!
Some small comments and hope we can merge this soon

pmerrison:
Heads-up: this went `CONFLICTING` since my comment yesterday — it was `MERGEABLE` (just behind) then. Flagging it with the details so you don't have to go digging, since it's small.

One file conflicts, `internal/translator/anthropic_helper.go`, in one hunk. #2383 (`2dce717`, structured output for Claude on GCP) retyped `buildAnthropicParams`'s second parameter from a plain `string` to `filterapi.APISchemaName`:

```go
// before
func buildAnthropicParams(openAIReq *openai.ChatCompletionRequest, apiSchema string, ...)
// now on main
func buildAnthropicParams(openAIReq *openai.ChatCompletionRequest, apiSchema filterapi.APISchemaName, ...)
```

It collides with this branch's doc-comment change on the same lines. The two existing call sites were updated to the typed constants, so `openai_anthropic.go` is the only caller still passing the old string form.

I resolved it locally against current `main` to check the size of it — it's two changes:

1. **`anthropic_helper.go`** — take `main`'s signature and comment, keeping this branch's `// It leaves Model unset because cloud providers identify the model in the request path.` line, which is the part worth preserving and is what makes the `params.Model` reach-through in `openai_anthropic.go` legible.

2. **`openai_anthropic.go`** — the call site becomes
   ```go
   params, err := buildAnthropicParams(openAIReq, filterapi.APISchemaAnthropic, o.modelNameOverride)
   ```
   plus the `internal/filterapi` import. This is a small improvement on the original: it drops the magic `"Anthropic"` string in favour of the same typed constant `openai_gcpanthropic.go` and `openai_awsanthropic.go` now use.

Nothing else conflicted — `endpointspec.go`, `endpointspec_test.go`, `testupstream_test.go` and the docs all auto-merged. With those two changes `go build ./...` is clean and `./internal/translator/...` and `./internal/endpointspec/...` both pass locally, though I've obviously not run the full CI matrix.

Happy to open a PR against your branch with the resolution if that's easier than doing it yourself — just say the word. Otherwise this is hopefully enough to make it a five-minute job.

Also worth noting for anyone reading: #2478 merged this morning, which fixed the mirror-image `cache_control` gap on the Anthropic → Bedrock path. This PR is the remaining hole in the caching matrix — chat completions → direct Anthropic is the last Anthropic-capable route that can't carry a cache breakpoint.

YeiSimon:
Hit this exact crash independently while load-testing a multi-protocol setup (OpenAI/Anthropic/Bedrock mocks behind Envoy AI Gateway v1.1.0) — filed our repro as [YeiSimon/sparkgate#140](https://github.com/YeiSimon/sparkgate/issues/140):

```
error processing request message: rpc error: code = Internal desc = cannot set backend:
failed to create translator for backend .../rule/12/ref/0: unsupported API schema: backend={Anthropic  v1}{via_upstream}
```

Same 500 on every request to a first-party Anthropic backend, 0 bytes reaching the upstream — matches what's described in #1936. Worth noting this bug is easy to mis-diagnose as a CRD/schema-validation problem rather than an ext_proc runtime gap, since the `AIServiceBackend` object itself applies and goes `Accepted` cleanly — we lost some time chasing that theory before finding the real cause in the ext_proc logs.

This PR looks done modulo the small nits already noted and the merge conflict — just adding another independent confirmation in case it helps move the review along. Happy to help test against a build of this branch if useful.

ajac-zero:
@kanurag94 Thanks for the review! I've implemented the improvements

PR #2132, opened 2026-05-12
feat: redact less data in logs for debugging

hustxiayang:
**Description**
### Motivation

The AI Gateway's debug logging supports redaction (`--enableRedaction` flag) to protect sensitive user data in log output. However, the previous redaction policy was too broad: it redacted developer-authored schema metadata alongside actual user content, making debug logs much less useful without providing any meaningful privacy benefit.

Specifically, the following categories were incorrectly treated as sensitive:

- **Tool definition `description` and `parameters`** — These are written by the
  application developer and describe the tool's interface (what it does, what arguments
  it accepts). They are no different from an API spec; they do not contain user data.

- **Tool call `function.name`** — This is the name the developer gave to the tool
  (e.g., `get_weather`). It matches the name in the tool definition and is not user data.

- **`response_format.json_schema`** (name, description, schema body) — This is the
  structured output schema the developer defines. It describes the shape of the response,
  not the content of any user interaction.

- **`guided_json`** — Same as above: a developer-authored JSON schema for constrained
  decoding.

The distinction is: redaction should protect **user-provided content** (message text,
uploaded images/audio/files, tool call arguments which can echo user data) and
**AI-generated content** (response text, reasoning chains). It should not obscure
**developer-authored configuration** that is already known to whoever deployed the
service.


nacx (approved):
LGTM

hustxiayang:
@nacx could you please help to merge this as well? Thanks a lot!

PR #2138, opened 2026-05-13
fix: handle per-backend httproute cluster names

### LINKED ISSUE #2135: bug: maybeModifyCluster skips per-backend clusters generated by Envoy Gateway v1.7.0 (7-part cluster names)
## Description

When using priority-based failover with backends of different schemas (e.g. OpenAI → AzureOpenAI), the `maybeModifyCluster()` function in `internal/extensionserver/post_translate_modify.go` skips all backend clusters. This prevents the upstream ext_proc filter from being injected, so on retry to the fallback backend, no per-backend path/body rewriting occurs. The fallback backend receives the primary backend's URL path and returns 404.

**Expected behavior:** The upstream ext_proc filter should be injected into every backend cluster regardless of whether Envoy Gateway generates 5-part or 7-part cluster names. On retry, the ext_proc should re-translate the request path and body to match the fallback backend's schema (e.g. `/v1/chat/completions` → `/openai/deployments/<model>/chat/completions?api-version=...`).

**Actual behavior:** `maybeModifyCluster()` checks `len(parts) != 5` on the cluster name, but Envoy Gateway v1.7.0 generates 7-part names (`httproute/<ns>/<name>/rule/<idx>/backend/<ref>`) when `priority` is set on `backendRefs`. All clusters are logged as `"non-ai-gateway cluster name"` and skipped. No upstream ext_proc, no endpoint metadata, and no header mutation filter are injected.

## Repro steps

1. Deploy AI Gateway v0.6.0 with Envoy Gateway v1.7.0
2. Create an AIGatewayRoute with two backends using different schemas and priorities:

```yaml
kind: AIGatewayRoute
metadata:
  name: example-failover
  namespace: envoy-models
spec:
  targetRefs:
      kind: Gateway
      name: my-gateway
  rules:
    - backendRefs:
        - name: openai-backend    # schema: OpenAI
          weight: 1
          priority: 0             # primary
        - name: azure-backend     # schema: AzureOpenAI
          weight: 1
          priority: 1             # fallback
      matches:
        - headers:
            - type: Exact
              name: x-ai-eg-model
              value: "*"
```

3. Configure a BackendTrafficPolicy with retry and `envoy.retry_priorities.previous_priorities`
4. Make the primary backend unreachable (e.g. point to a wrong port)
5. Send a chat completion request through the gateway
6. Observe the retry reaches the Azure fallback but gets 404 because the path is still `/v1/chat/completions`

## Environment

- **AI Gateway**: v0.6.0 (Helm chart `ai-gateway-helm-v0.6.0`). Also confirmed the same code exists on current `main` branch.
- **Envoy Gateway**: v1.7.0 (`docker.io/envoyproxy/gateway:v1.7.0`)
- **Envoy Proxy**: v1.37.0 (distroless)
- **ext_proc image**: `docker.io/envoyproxy/ai-gateway-extproc:v0.6.0`
- **Kubernetes**: v1.31

### AI Gateway Controller Logs

The controller repeatedly logs `"non-ai-gateway cluster name"` for every backend cluster on every reconciliation. No clusters are ever modified:

```
2026-05-13T12:47:42Z  INFO  envoy-gateway-extension-server  non-ai-gateway cluster name  {"cluster_name": "httproute/envoy-models/vllm-azure-fallback/rule/0/backend/0"}
2026-05-13T12:47:42Z  INFO  envoy-gateway-extension-server  non-ai-gateway cluster name  {"cluster_name": "httproute/envoy-models/vllm-azure-fallback/rule/0/backend/1"}
2026-05-13T12:47:42Z  INFO  envoy-gateway-extension-server  non-ai-gateway cluster name  {"cluster_name": "envoy-models/vllm-azure-fallback"}
2026-05-13T12:47:42Z  INFO  envoy-gateway-extension-server  Added extproc-uds cluster to the list of clusters
2026-05-13T12:49:43Z  INFO  envoy-gateway-extension-server  inserting AI Gateway extproc filter into listener  {"listener": "envoy-models/vllm-azure-fallback/http"}
2026-05-13T12:49:43Z  INFO  envoy-gateway-extension-server  non-ai-gateway cluster name  {"cluster_name": "httproute/envoy-models/vllm-azure-fallback/rule/0/backend/0"}
2026-05-13T12:49:43Z  INFO  envoy-gateway-extension-server  non-ai-gateway cluster name  {"cluster_name": "httproute/envoy-models/vllm-azure-fallback/rule/0/backend/1"}
```

### Envoy Access Logs (on retry to Azure fallback)

```
x-envoy-origin-path: /v1/chat/completions    ← NOT rewritten to Azure format
:authority: chatopsapi.openai.azure.com        ← correct (set by TLS/cluster config)
response code: 404                             ← Azure returns "Resource not found"
```

Direct curl to Azure with the correct path succeeds (HTTP 200), confirming the deployment exists and the API key works:
```
curl https://chatopsapi.openai.azure.com/openai/deployments/gpt-4.1-mini/chat/completions?api-version=2024-10-21
```

### Envoy Config Dump (relevant sections)

**Listener:** has the router-level ext_proc filter (disabled by default, enabled per-route) ✅
**Backend clusters:** have NO upstream HTTP filters in `TypedExtensionProtocolOptions` ❌

### Root cause in code

The code at [`post_translate_modify.go` — `maybeModifyCluster()`](https://github.com/envoyproxy/ai-gateway/blob/main/internal/extensionserver/post_translate_modify.go):

```go
// Expected format: "httproute/<namespace>/<name>/rule/<index_of_rule>".
parts := strings.Split(cluster.Name, "/")
if len(parts) != 5 || parts[0] != "httproute" {
    s.log.Info("non-ai-gateway cluster name", "cluster_name", cluster.Name)
    return nil
}
```

Envoy Gateway v1.7.0 generates per-backend clusters with 7-part names when `priority` is set:
```
httproute/envoy-models/vllm-azure-fallback/rule/0/backend/0   → 7 parts
httproute/envoy-models/vllm-azure-fallback/rule/0/backend/1   → 7 parts
```

The `len(parts) != 5` guard rejects them all.

### Suggested fix

Accept both 5-part and 7-part cluster name formats. When 7-part names are used (per-backend clusters), each cluster corresponds to a single backend ref rather than all backends in a rule, so the metadata/filter injection logic should use `parts[6]` as the backend ref index instead of iterating all `BackendRefs`:

```go
parts := strings.Split(cluster.Name, "/")
if (len(parts) != 5 && len(parts) != 7) || parts[0] != "httproute" {
    s.log.Info("non-ai-gateway cluster name", "cluster_name", cluster.Name)
    return nil
}
```

mturac:
**Description**

Fixes #2135.

This updates `maybeModifyCluster()` to accept both route-level cluster names (`httproute/<namespace>/<name>/rule/<index>`) and per-backend cluster names (`httproute/<namespace>/<name>/rule/<index>/backend/<index>`). For per-backend names, the backend index is parsed and metadata/filter injection is scoped to the matching backend reference. Existing 5-part cluster handling remains unchanged.

**Testing**

- `go test ./internal/extensionserver/...`
- Build: pass
- Tests: pass
- Lint: clean


PatilHrushikesh:
can you add tests for  7-part cluster names?

PatilHrushikesh:
it think it would be better if whole parsing of clusternames logic is moved to new function out of maybeModifyCluster

internal/extensionserver/post_translate_modify.go
```diff
@@ -179,6 +179,20 @@ func (s *Server) maybeModifyCluster(ctx context.Context, cluster *clusterv3.Clus
 			"cluster_name", cluster.Name, "rule_index", httpRouteRuleIndexStr)
 		return nil
 	}
+	backendRefIndexes := []int{-1}
+	if len(parts) == 7 {
```
Hritik003: Why create an array with one element instead of just using var backendRefIndex int = -1?

mturac:
Thanks, addressed the review points.

Changes since the earlier revision:
- added direct 7-part per-backend `maybeModifyCluster` coverage for both the `LoadAssignment` path and the nil-`LoadAssignment` cluster metadata fallback
- tightened cluster-name parsing so malformed `httproute/<ns>/<name>/<segment>/<rule>` names do not parse as AI Gateway rule clusters
- merged current `upstream/main` into the PR branch per the contributor guide

Local validation:
- `go test ./internal/extensionserver -run 'TestParseAIGatewayClusterName|TestMaybeModifyClusterPerBackendClusterName|Test_maybeModifyCluster' -count=1`\n- `go test ./internal/extensionserver -coverprofile=/tmp/aigw-extensionserver-cover-after.out -count=1` (`parseAIGatewayClusterName` 100.0%, `maybeModifyCluster` 89.0%, package 88.4%)\n- `make test`\n- `make precommit`\n\nThe PR workflows for Build and Test, Precommit Check, and CodeQL are currently in `action_required` pending maintainer approval for the fork run.

PR #2142, opened 2026-05-14
fix: preserve inferencepool metadata on extension hooks

immanuwell:
**Description**

PostClusterModify and PostRouteModify dropped InferencePool metadata when Envoy Gateway sent a cluster or route with nil Metadata. The helper allocated metadata, but only in a local pointer, so later InferencePool handling could not read it. Small fix, real footgun.

Repro:
1. Revert this patch, or apply only the added assertions.
2. Run `go test ./internal/extensionserver -run 'TestPost(Cluster|Route)Modify/with_InferencePool' -count=1`.
3. Both hook tests fail because getInferencePoolByMetadata returns nil.

This is reachable for normal InferencePool backendRefs; Metadata is optional in the Envoy proto and the repo example uses AIGatewayRoute to InferencePool. No cloud quota or provider API ceiling blocks this path.

Verification:
`go test ./internal/extensionserver -run 'TestPost(Cluster|Route)Modify/with_InferencePool' -count=1`
`go test ./internal/extensionserver`
`make test`
`make precommit`

Related PR: #2071

AI-assisted patch prep; I reviewed the code and own the change.

### DIFF
--- a/internal/extensionserver/inferencepool.go
+++ b/internal/extensionserver/inferencepool.go
@@ -130,24 +130,26 @@ func getInferencePoolByMetadata(meta *corev3.Metadata) *gwaiev1.InferencePool {
 }
 
 // buildMetadataForInferencePool adds InferencePool metadata to the cluster for reference by other components.
-// encoded as a string in the format: "namespace/name/serviceName/port".
+// encoded as a string in the format: "namespace/name/serviceName/port/bodyMode/allowModeOverride".
 func buildEPPMetadataForCluster(cluster *clusterv3.Cluster, inferencePool *gwaiev1.InferencePool) {
 	// Initialize cluster metadata structure if not present.
+	if cluster.Metadata == nil {
+		cluster.Metadata = &corev3.Metadata{}
+	}
 	buildEPPMetadata(cluster.Metadata, inferencePool)
 }
 
 // buildMetadataForInferencePool adds InferencePool metadata to the route for reference by other components.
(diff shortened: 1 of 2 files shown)


internal/extensionserver/inferencepool.go
```diff
@@ -130,15 +130,21 @@ func getInferencePoolByMetadata(meta *corev3.Metadata) *gwaiev1.InferencePool {
...
 	buildEPPMetadata(cluster.Metadata, inferencePool)
 }
 
 // buildMetadataForInferencePool adds InferencePool metadata to the route for reference by other components.
 func buildEPPMetadataForRoute(route *routev3.Route, inferencePool *gwaiev1.InferencePool) {
 	// Initialize route metadata structure if not present.
+	if route.Metadata == nil {
+		route.Metadata = &corev3.Metadata{}
+	}
 	buildEPPMetadata(route.Metadata, inferencePool)
```
nacx, line 148: Thanks!
Is it worth removing the metadata assignment inside `buildEPPMetadata` as it is misleading? We can error there if it is nil?

nacx: friendly ping. WDYT?

immanuwell: Addressed in e874c521. I removed the dead `metadata == nil` initialization from `buildEPPMetadata`, rebased the branch onto current `main`, and reran the focused extensionserver tests.

immanuwell:
@nacx all your suggestions are now addressed, PTAL

nacx (approved):

Issue #2149, opened 2026-05-20
SSE parser silently drops fields when colon is not followed by a space


sriyer:
**Component:** `internal/mcpproxy/sse.go` — `parseEvent`

**Version:** v0.6.0 (also present on `main`)

## Description

The SSE event parser in `parseEvent` uses `bytes.HasPrefix` with prefixes that include a mandatory space after the colon (`"data: "`, `"event: "`, `"id: "`). When a backend sends SSE fields without the optional space (e.g. `data:{"jsonrpc"...}`), none of the `switch` cases match and the lines are silently dropped. The function returns a valid `*sseEvent` with zero messages and no error.

Per the [SSE specification](https://html.spec.whatwg.org/multipage/server-sent-events.html#event-stream-interpretation), the space after the colon is optional:

> *"Collect the characters on the line after the first U+003A COLON character (:), and let value be that string. If value starts with a U+0020 SPACE character, remove it from value."*

Both `data:value` and `data: value` are valid and must produce the same result. The current parser only handles the latter.

## Impact

For aggregated requests (`tools/list`, `prompts/list`, `resources/list`), the empty event causes `sendToAllBackendsAndAggregateResponsesImpl` to skip the backend's response entirely (`len(event.messages) == 0`), resulting in an empty merged response (e.g. `{"tools":[]}`) returned to the client. No error is logged anywhere in this path.

## Reproduction

Any MCP backend that emits SSE without a space after the colon triggers this. Spring Boot's `SseEmitter` (used by Spring AI MCP Server) is a common example:

```
id:c882a621-975e-4429-8e21-ba7436f8a8c7
event:message
data:{"jsonrpc":"2.0","id":2,"result":{"tools":[...]}}
```

Hex dump confirming no space byte (`0x20`) after colon (`0x3a`):

```
64 61 74 61 3a 7b 22 6a 73 6f   data:{"jso
65 76 65 6e 74 3a 6d 65         event:me
69 64 3a 63 38 38 32           id:c882
```

## Affected code

- [Prefix definitions (L19-L22)](https://github.com/envoyproxy/ai-gateway/blob/v0.6.0/internal/mcpproxy/sse.go#L19-L22)
- [Prefix matching in `parseEvent` (L98-L109)](https://github.com/envoyproxy/ai-gateway/blob/v0.6.0/internal/mcpproxy/sse.go#L98-L109)

sriyer:
issue fixed and merged to main.

PR #2155, opened 2026-05-21
fix:  SSE parser silently drops fields - issue  2149 

sriyer:
**Description**

The SSE event parser in `parseEvent` uses `bytes.HasPrefix` with prefixes that include a mandatory space after the colon (`"data: "`, `"event: "`, `"id: "`). When a backend sends SSE fields without the optional space (e.g. `data:{"jsonrpc"...}`), none of the `switch` cases match and the lines are silently dropped. The function returns a valid `*sseEvent` with zero messages and no error.

Per the SSE specification, the space after the colon is optional:

> *"Collect the characters on the line after the first U+003A COLON character (:), and let value be that string. If value starts with a U+0020 SPACE character, remove it from value."*

Both `data:value` and `data: value` are valid and must produce the same result. The current parser only handles the latter.

**Component:** `internal/mcpproxy/sse.go` — `parseEvent`

**Version:** v0.6.0 (also present on `main`)

### DIFF
--- a/api/v1alpha1/shared_types.go
+++ b/api/v1alpha1/shared_types.go
@@ -22,6 +22,8 @@ type VersionedAPISchema struct {
 	//
 	// When the name is set to AzureOpenAI, this version maps to "API Version" in the
 	// Azure OpenAI API documentation (https://learn.microsoft.com/en-us/azure/ai-services/openai/reference#rest-api-versioning).
+	// This field is ignored for OpenAI, AWSBedrock, GCPVertexAI, and Anthropic.
+	// For OpenAI and Anthropic, use prefix to configure custom request paths.
 	//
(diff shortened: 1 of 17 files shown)


sriyer:
pull request for - https://github.com/envoyproxy/ai-gateway/issues/2149

nacx (commented):
Thanks! Overall LGTM.
Can you fix the DCO and other PR check failures?

Hritik003 (commented):
LGTM, great catch. Please fix the build failures, thanks!

sriyer:
thanks @nacx & @Hritik003 let me try to fix the failures

sriyer:
@nacx  @Hritik003  DCO and the other failures are good now.

Hritik003 (commented):
Lgtm

nacx (approved):

PR #2158, opened 2026-05-25
feat: configurable allowlist of extproc filters that run before AI Gateway

siddharth1036:
**Description**
Generalizes the ordering of extproc filters in the HCM in front of the AI Gateway extproc. Previously the placement was hardcoded for one specific upstream extproc. This PR replaces that with a controller-wide allowlist of exact Envoy filter names that are permitted to run before AI Gateway. When set, the AI Gateway extproc is inserted immediately after the last filter whose name matches; when unset, behavior is unchanged (AI Gateway runs before all other extprocs).

Wired through:

- Controller CLI flag: `--extProcBeforeFilterNames` (comma-separated exact names).
- Standalone `aigw` env var: `AIGW_EXTPROC_BEFORE_FILTER_NAMES`.
- Helm value: `controller.extProcBeforeFilterNames` (list).

Exact names (not prefixes) are used intentionally so operators must explicitly opt specific filters in, rather than allowing broad bypass.

N/A

- Default behavior is unchanged — flag/value unset → AI Gateway runs before all other extprocs.
- Allowlist entries should be the exact Envoy filter name as Envoy Gateway materializes them, e.g. for an `EnvoyExtensionPolicy`: `envoy.filters.http.ext_proc/envoyextensionpolicy/<ns>/<name>/extproc/<index>`.
- Tests cover: empty allowlist (fallback), single match, multiple matches (last wins), no match (fallback), non-extproc names, partial matches.


nacx (commented):
Thanks! overall LGTM

PR #2159, opened 2026-05-25
fix: avoid panic when LoadAssignment has fewer endpoints than active BackendRefs

mtparet:
**Description**

`maybeModifyCluster` in `internal/extensionserver/post_translate_modify.go` indexed `cluster.LoadAssignment.Endpoints[lbEndpointIndex]` inside the `BackendRefs` loop without bounds-checking. When envoy-gateway emits a `LoadAssignment` with fewer endpoint sets than the rule's active (non-zero-weight) backendRefs - a transient state during translation churn - the controller panicked with `index out of range` and crashed, eventually triggering CrashLoopBackOff on the affected replica.

This change adds a length guard that breaks out of the loop and logs a structured warning (cluster name, route namespace/name, endpoints length, lb endpoint index, backend ref index) so the underlying translation mismatch remains observable. A unit test constructs a 2-backendRef rule against a 1-entry `LoadAssignment` and asserts `maybeModifyCluster` returns without panicking; the test fails on `main` without the fix.

None upstream.

The guard uses `break` (rather than `continue`) because `lbEndpointIndex` is monotonically increasing across iterations; once it exceeds the available endpoint sets, no later backendRef in the same rule can be served either.

This PR was drafted with the help of an AI assistant; the change has been reviewed and is owned by the submitter.


mtparet:
Same test passed locally, could be a flaky test error ?

internal/extensionserver/post_translate_modify.go
```diff
@@ -224,6 +224,20 @@ func (s *Server) maybeModifyCluster(ctx context.Context, cluster *clusterv3.Clus
 				if backendRef.Weight != nil && *backendRef.Weight == 0 {
 					continue
 				}
+				if lbEndpointIndex >= len(cluster.LoadAssignment.Endpoints) {
+					// envoy-gateway emitted a LoadAssignment with fewer Endpoints entries
+					// than the rule has active (non-zero-weight) BackendRefs. This is a
+					// transient translation mismatch; skip the remaining backendRefs rather
```
nacx: Could you elaborate a bit more on what the "transient translation mismatch" is, and under what conditions it can appear?

mtparet: One of our custom controller was patching BackendSecurityPolicy / Backend every ~30 s forcing EG to re-translate continuously. One intermediate LoadAssignment had 1 endpoint set for a 2-backendRef rule → index out of range [1] with length 1 and causing a panic. I suppose it can happens in any system quickly automatically modifying gateway resources and causing this race condition.

nacx: Can you explain this and elaborate on the comment in the code itself, so that users and maintainers who get to this code in the future fully understand the conditions under which this can happen?

nacx: friendly ping

mtparet: Sorry for the delay, I just adjusted the comment!

PR #2167, opened 2026-05-28
feat: add EarlyTranslate to EndpointSpec so the EPP receives an OpenAI-format body for Anthropic requests

isztldav:
**Description**

### What

Adds `EarlyTranslate(req *ReqT) (headers []Header, body []byte, err error)` to the
`EndpointSpec` interface. All existing specs implement it as a no-op except
`MessagesEndpointSpec`, which uses `NewAnthropicToChatCompletionOpenAITranslator` to
produce an OpenAI chat-completions body and rewrites `:path` to `/v1/chat/completions`.

The router ext_proc (`routerProcessor.ProcessRequestBody`) calls `EarlyTranslate` after
capturing `originalRequestBodyRaw`. If a body is returned:
- The headers (`:path`, `content-length`) are appended to the header mutation.
- The body is included as a `BodyMutation` in the router-phase response.
- `forceBodyMutation` is set so the upstream translator always emits a full body,
  preventing the EPP-facing OpenAI body from being forwarded to the backend untouched.
- A `ModeOverride: NONE` is added to the response to pre-empt Envoy buffering the response
  body before the upstream filter re-enables it.

Also fixes `routerProcessor.ProcessResponseHeaders` (the no-upstream-filter path) to return
`ResponseBodyMode: NONE` instead of falling through to `passThroughProcessor`, and ensures
`upstreamProcessor.ProcessResponseHeaders` always emits a `ModeOverride` (BUFFERED or
STREAMED) to override the pre-emption when an upstream filter is present.

### Why

The Gateway API Inference Extension EndpointPicker (EPP) runs in the listener filter chain
before the AI Gateway upstream ext_proc. Without early translation, EPP always sees
untranslated Anthropic bodies and rejects them. This change enables Anthropic clients to
benefit from KV-cache-aware routing.

Depends on PR 1 (control-plane: `InferencePool` as a valid `backendRef` kind).

### Testing

- `TestMessagesEndpointSpec_EarlyTranslate`: verifies that a `MessagesRequest` produces a
  valid OpenAI chat-completions body with correct `:path` and `content-length` headers.
- `TestEarlyTranslate_NoOpSpecs`: verifies that all other endpoint specs return
  `nil, nil, nil`.
- `Test_messagesProcessorRouterFilter_ProcessRequestBody_EarlyTranslate`: integration-style
  unit test that runs `ProcessRequestBody` on a valid Anthropic body and asserts that the
  response includes the OpenAI body mutation, the `:path` rewrite, `forceBodyMutation =
  true`, and `ModeOverride = NONE`.
- `make precommit test` passes.

- Relates to PR 1 (#2166) and #902.

- AI assistant (Claude) was used for drafting and reviewing code; all logic has been
  reviewed and is understood by the author.


isztldav:
Update: as of v0.7.0 this patch is still needed.

aurianer:
@nacx @mathetake , this would be quite useful for us, any comments from your side on this?

isztldav:
## Re: "can this be solved by moving the EPP in the filter chain?"

Short answer: **reordering the EPP alone can't solve it** — because of *what* the EPP is for.

The EPP is a **downstream (listener) HTTP filter that must run before the router filter**: its whole job is to pick an endpoint in the `InferencePool` and steer Envoy's load-balancing decision. The AI Gateway's authoritative request translation is a **per-cluster upstream filter that runs after routing**. So the EPP is, by construction, always *earlier* than the real translation — you can't move it "after translation" without moving it after endpoint selection, which defeats its purpose.

### What the chain actually looks like

Downstream listener HCM chain (all before the router filter):

```
AI Gateway ext_proc (router-level)  →  EPP ext_proc  →  router  ──route──▶  cluster
```

Per-cluster upstream chain (`HttpProtocolOptions.HttpFilters`):

```
... → AI Gateway ext_proc (UPSTREAM)  →  upstream_codec  →  backend
```

Two consequences:

1. The EPP and the authoritative translation live on **different chains**. Translation happens at the **upstream** filter on purpose — it can access the original body in memory and re-runs per-endpoint on retries/failover, where different backends may need *different* translations. The EPP is downstream and always runs before any of that.
2. The router-level AIGW ext_proc is already inserted as the **first** ext_proc, i.e. before the EPP. That ordering is exactly why `EarlyTranslate` works: our router filter rewrites Anthropic → OpenAI, the EPP (running next) sees a parseable body, and the upstream filter still re-translates from the stored original.

### The reorder that *would* remove `EarlyTranslate`, and why it's a bigger change

The only reordering that eliminates `EarlyTranslate` is to **move the canonical translation onto the downstream router chain, ahead of the EPP**. That regresses the current design:

- Translation-at-upstream is what cleanly handles **retries and multi-provider failover** — each endpoint may need a different translation, decided *after* routing. Translating once, downstream, before routing loses that.
- The AIGW ext_proc filter is **shared across all routes**, not just InferencePool ones, so this isn't a localized change.

### Conclusion

`EarlyTranslate` is the **minimal, targeted fix**: it gives the EPP a body it can parse for model extraction / KV-cache-aware routing, while leaving the authoritative upstream translation (and retry/failover semantics) untouched. Moving the EPP in the chain doesn't achieve the same thing without a larger architectural change to where translation happens.

PR #2172, opened 2026-05-29
fix(openai): parse typeless assistant output messages

dzibma:
**Description**

Fix OpenAI Responses API request parsing for multi-turn inputs that include assistant output messages without an explicit `type: "message"` field.

Some clients, including OpenCode, send prior assistant turns as items like:

```json
{"role":"assistant","content":[{"type":"output_text","text":"..."}]}
```
Previously, typeless message-like items were always parsed as EasyInputMessageParam, whose content union only accepts input content types. That caused JSON unmarshalling to fail on output_text.

This change treats typeless assistant messages as output messages when they match the ResponseOutputMessage shape, while preserving the existing easy-message fallback for other typeless message inputs.

I used ai assistance to help draft the tests. I reviewed and own the final change.


nacx (approved):

nacx:
Thanks!

PR #2175, opened 2026-05-30
fix: source MCP session encryption seed from a Secret

### LINKED ISSUE #2145: bug: MCP session encryption seed is exposed as plaintext in controller Deployment args
_Description_:

> What issue is being seen? Describe what should be happening instead of
> the bug, for example: Envoy should not crash, the expected value isn't
> returned, etc.

The MCP session encryption seed (`mcpSessionEncryptionSeed`) is passed as a literal command-line argument in the controller `Deployment`, and subsequently forwarded as a literal argument to ext-proc containers by the gateway mutator. This means the seed value is visible in plaintext in:
- The controller Deployment spec (`kubectl get deploy -o yaml`)
- The ext-proc container args on mutated Gateway pods
- etcd (Deployment and Pod objects are not covered by Kubernetes Secret encryption at rest)
- Process listings (`/proc/<pid>/cmdline`, `ps aux`) on any node running the pods
- Kubernetes audit logs for Deployment/Pod reads

It feels like it should be treated as a credential where if compromised, an attacker could decrypt or forge MCP session tokens? 

The same issue applies to the fallback seed (`mcpFallbackSessionEncryptionSeed`).

**Expected behavior:** Users should be able to provide the encryption seed via a Kubernetes Secret reference so that the seed never appears in Deployment/Pod specs, etcd, or process arguments, and access is governed by standard Kubernetes Secret RBAC and encryption-at-rest.

_Repro steps_:

> required to reproduce the bug.

> **Note**: If there are privacy concerns, sanitize the data prior to
> sharing.

- Install the Helm chart with a custom seed:

```bash
   helm install ai-gateway oci://docker.io/envoyproxy/ai-gateway-helm \
     --set controller.mcp.sessionEncryption.seed="my-secret-seed-value"
```

- Inspect the controller Deployment:

```bash
kubectl get deploy -l app.kubernetes.io/name=ai-gateway-helm -o yaml
```

- Observe that `my-secret-seed-value` appears in plaintext in the container args:

```bash
args:
  - --mcpSessionEncryptionSeed=my-secret-seed-value
```

There is currently no way that I can see to source the seed from a Kubernetes Secret. The Helm chart only supports setting the value directly in `values.yaml` or via `--set`, and the controller/ext-proc binaries only read it from CLI flags I think

_Environment_:

> Include the environment like gateway version, envoy version and so on.

All versions. The relevant code paths:

- Helm chart: [deployment.yaml L103-L108](https://github.com/envoyproxy/ai-gateway/blob/main/manifests/charts/ai-gateway-helm/templates/deployment.yaml#L103-L108)
- Controller flag parsing: [cmd/controller/main.go L227-L234](https://github.com/envoyproxy/ai-gateway/blob/main/cmd/controller/main.go#L227-L234)
- Ext-proc flag parsing: [cmd/extproc/mainlib/main.go L131-L138](https://github.com/envoyproxy/ai-gateway/blob/main/cmd/extproc/mainlib/main.go#L131-L138)
- Gateway mutator forwarding seed as flag: [internal/controller/gateway_mutator.go L154-L164](https://github.com/envoyproxy/ai-gateway/blob/main/internal/controller/gateway_mutator.go#L154-L164)

_Logs_:

> Include the access logs and the Envoy logs.

N/A as this is a configuration/manifest exposure issue, not a runtime error.

**Potential approach:**

Add env var fallback support to both binaries and an `existingSecret` pattern in the Helm chart:

- Go (controller + ext-proc): If the `--mcpSessionEncryptionSeed` flag is empty/default, fall back to a `MCP_SESSION_ENCRYPTION_SEED` env var. Same for the fallback seed.
- Helm chart (`values.yaml`): Add `controller.mcp.sessionEncryption.existingSecret` with `name`, `seedKey`, and optional `fallbackSeedKey` fields.
- Helm chart (`deployment.yaml`): When `existingSecret` is set, inject `secretKeyRef` env vars and omit the seed from args.
- Gateway mutator: Pass the seed to ext-proc containers as an env var instead of a CLI flag.

I'd be happy to try put a PR together for this if folks think the direction makes sense here. Let me know if maybe I am missing something here in general but this looks like an issue to me!

walsm232:
**Description**

> [!NOTE]
> **AI assistance:** I used `Claude Opus 4.8` via Cursor to help with writing many parts of the code and tests. I've reviewed the changes myself, understand them and the possible other directions this could take, and take ownership of the PR. I'll address review feedback directly and revise as needed.

The MCP session encryption seed was exposed as plaintext in process arguments in two places: the `--mcpSessionEncryptionSeed` flag on the controller Deployment, and the `-mcpSessionEncryptionSeed` arg forwarded to every extproc container. Both are readable by anyone who can `kubectl get/describe` the Deployment or gateway Pods, and can also show up in node process listings.

This change removes the seed from controller and extproc process arguments when users configure a Kubernetes Secret-backed source:

- The controller can now read the seed, and optional fallback seed, from files via the new `--mcpSessionEncryptionSeedFilePath` / `--mcpFallbackSessionEncryptionSeedFilePath` flags. These take precedence over the inline flags.
- The Helm chart adds `controller.mcp.sessionEncryption.existingSecret` with `name`, `seedKey`, and optional `fallbackSeedKey`. When set, the chart mounts that Secret read-only on the controller and passes the file-path flags instead of the inline seed value.
- The controller no longer forwards the seed to extproc as a CLI arg. Instead, it writes the resolved seed and optional fallback into the per-gateway filter-config Secret it already generates, under `MCPConfig.SessionEncryption`.
- Extproc installs MCP session crypto from `MCPConfig.SessionEncryption` during `LoadConfig`, so gateway Pods do not need direct access to the user-provided Secret.

The inline flags and chart seed value are retained as the default path, so existing installations are unaffected.

**Security boundary / remaining Secret exposure**

This PR removes the seed from Deployment/Pod specs and process args for the Secret-backed Helm path, which is the primary issue described in #2145.

It does still copy the resolved seed into each generated per-gateway filter-config Secret so extproc can read it without requiring the user-provided Secret to exist in every gateway Pod namespace. That means the seed remains visible to actors with read access to those generated Secrets. I chose this tradeoff to avoid needing the Secret to be available in each of the namespaces (as I think this is a scenario that this would lead to?), but I would especially like reviewer input on whether the generated filter-config Secret is the right propagation boundary.

**Related Issues/PRs**

Fixes https://github.com/envoyproxy/ai-gateway/issues/2145

**Controller Secret source:** I think the cross-namespace concern only applies to extproc. For the controller, which runs in a single known namespace, the seed could also be sourced via a `secretKeyRef` env var instead of a mounted file. I chose a file mount because extproc already reads the generated filter-config Secret from a projected file watched by the config watcher, so this keeps both ends Secret-file-backed. Happy to switch the controller side to `secretKeyRef` if that is preferred.

**Validation behavior:** If a seed file path is configured but the file is empty, the controller currently falls back to the inline flag/default. I am leaning toward changing this to fail closed instead, since an empty Secret key probably indicates misconfiguration and otherwise can silently fall back to `default-insecure-seed` - I wasn't sure here though so let me know what you think.

**Controller wiring:** `GatewayController.SetMCPSessionEncryption` is a setter rather than a constructor parameter to avoid threading four more arguments through `NewGatewayController`'s existing call sites. Happy to fold it into the constructor if reviewers prefer that shape, let me know what you think.

**Testing**

`make precommit test` passes.

Added/updated unit tests in `cmd/controller`, `internal/controller`, and `internal/mcpproxy` covering:

- File-sourced primary and fallback seeds.
- Propagation into the per-gateway filter-config Secret.
- Absence of seed and fallback seed from extproc args.
- MCP crypto installation from `LoadConfig`.
- Fallback crypto behavior for seed rotation.

Also verified end-to-end on a local Kind cluster with `existingSecret` set:

- The controller args contain only the `...FilePath` flags, with no plaintext seed.
- The seed file is present in the controller Pod from the mounted Secret.
- The resolved seed/fallback are propagated into the per-gateway filter-config Secret.
- The extproc container args contain no seed or fallback seed.


mathetake (changes requested):
How can this be backward compatible

walsm232:
@mathetake can you share your thoughts / more details around how it is not? This is my first contribution here, so it's possible I'm missing something fundamental. I'll lay out my thinking here so hopefully you can point at where it is breaking down.

So my intent was for the original / current inline path to be unchanged:

- The inline Helm values (`sessionEncryption.seed` / `iterations` / `fallback.seed` / `fallback.iterations`) are all still there, and `existingSecret` is purely additive/opt-in.
- For an install that doesn't set `existingSecret`, the controller still resolves the same seed and the resolved value reaches extproc, just via the generated per-gateway filter-config Secret (`MCPConfig.SessionEncryption`) instead of the extproc CLI args.

The flow for the inline (default) case:

1. The chart renders `--mcpSessionEncryptionSeed=<seed>` / `--mcpSessionEncryptionIterations=<n>` on the controller exactly as before (the file-path flags + volume mount only appear when `existingSecret.name` + `seedKey` are set).
2. The controller resolves that seed and writes it into the per-gateway filter-config Secret it already generates, under `MCPConfig.SessionEncryption`.
3. extproc picks it up in `LoadConfig` and installs the session crypto (primary + optional fallback) from there.

So the seed value itself is unchanged on the default path. The main difference would be that it now travels controller to filter-config Secret to extproc rather than being passed as an extproc CLI arg. If that change in delivery is the compatibility concern, I'm happy to keep the CLI args for the inline path and only route through the Secret for the opt-in `existingSecret` path.

Keen to understand better where you see the compatibility breaking.

walsm232:
I think I was only looking at the seed value (which doesn't change on the inline path) and went over the fact that I'm changing how it actually gets to extproc. Right now it's in the args when extproc starts. With my changes, the injected extproc has no seed arg and just waits for the controller to put it in the filter-config Secret, which it reads later in `LoadConfig`.

The bit that I am thinking about now is that `-mcpSessionEncryptionSeed` is dropped from the args, but the extproc flag still defaults to `default-insecure-seed`. So during a rolling upgrade, if a new extproc pod starts before the controller has rewritten its Secret, it'd come up on the insecure default instead of the real seed which is out of sync with the old pods that still have it in their args (so decrypts fail), and silently insecure if the Secret never gets rewritten?

Is that what you were getting at? If so would it make sense to just leave the args alone on the inline path and only go through the Secret for the existingSecret case, and make it error out instead of falling back to the insecure default? I can look at making some changes if it sounds right.

PR #2179, opened 2026-06-01
docs: propose Gemini native client API support

CodePrometheus:
**Description**

Adds a proposal for Google native client API support.

Today, Gemini and Vertex AI users generally need to use an OpenAI-compatible entrypoint
when calling AI Gateway. This proposal describes native passthrough entrypoints for
Google-shaped requests under `/gemini` and `/vertex-ai`.

The proposal covers:

- Path-aware matching for Google native REST paths.
- Extracting model context from URL paths for route matching.
- Preserving native request, response, streaming, and error shapes.
- Forwarding Gemini requests with gateway-managed API keys.
- Forwarding Vertex AI requests with gateway-managed GCP auth.
- Reading usage metadata where Google responses expose it.

N/A

Feedback on the API surface and implementation approach would be appreciated.


CodePrometheus:
First proposal here, any guidance would be appreciated.

CodePrometheus:
Could you take a look at this proposal when you have time, thanks! friendly ping @nacx @gavrissh

PR #2180, opened 2026-06-02
fix: status of routes when Gateway is not found

Hritik003:
**Description**

When an MCPRoute/AIGW ROute references a non-existent Gateway in its spec.parentRefs, the controller still marks the route as "Accepted" with the message "Gateway Route reconciled successfully". The controller logs "Gateway not found" but does not propagate this as an error, so the reconciliation appears successful.

Root Cause:

In internal/controller/mcp_route.go, the syncGateway function was a void function that silently returned on errors — it logged "Gateway not found" but never returned an error to the caller. Since syncGateways (which calls syncGateway for each parentRef) always returned nil, the syncMCPRoute function completed successfully, and Reconcile marked the status as ConditionTypeAccepted.
- and same goes for ai gateway route controller 

Fix:

Changed syncGateway to return an error. When the referenced Gateway is not found, it now returns a descriptive error ("gateway <namespace>/<name> not found"). This error propagates through syncGateways → syncRoute → Reconcile, which then calls updateRouteStatus with ConditionTypeNotAccepted and the error message.

During Route deletion (via the finalizer path), the error is still logged but does not block cleanup — handleFinalizer already handles onDeletionFn errors non-fatally.


nacx (approved):
Thanks! LGTM

PR #2103, opened 2026-05-04
feat: add Anthropic /messages to AWS Bedrock translator

Killusions:
**Description**

Adds a translator from the Anthropic /messages API to the AWS Bedrock API (`anthropic_awsbedrock.go`). This is similar to how `openai_awsbedrock.go` translates OpenAI requests to Bedrock, it lets you use the Anthropic client format to reach any model on Bedrock, not just Anthropic models.

Handles messages, system prompts, tools, thinking config, images, streaming (EventStream unwrapping to SSE), and error translation.

Related:
- #2102

⚒️ with ❤️ by @siemens

PR #2189, opened 2026-06-03
fix: promote role:system messages from messages array to system param in Bedrock translator

### LINKED ISSUE #2206: Fix: Anthropic system messages dropped in Bedrock translator
Some clients (like Claude Code) send system prompts as `{"role": "system"}` entries in the messages array instead of the top-level `system` parameter when using the Anthropic `/messages` endpoint with the Bedrock translator. These hit the default case and error out. It would be nice to promote those entries to the top-level parameter, same as how the Anthropic API handles it natively.

Killusions:
**Description**

In the Anthropic-to-AWS-Bedrock translator, Claude Code with mid-conversation-system beta sends system prompts as {"role": "system"} in the messages array. The SDK has no system role constant so these hit the default error case. This promotes them to the top-level system param before conversion.

Fixes #2206

⚒️ with ❤️ by @siemens


Killusions:
@nacx This was an oversight in #2103 and now surfaced due to changes in Claude Code's behavior, could you maybe have a look?

nacx (approved):

PR #2190, opened 2026-06-03
feat: forward reasoning_effort as reasoning_config to AWS Bedrock

### LINKED ISSUE #2207: Feat: Forward reasoning_effort as reasoning_config to AWS Bedrock
In the OpenAI-to-Bedrock translator, `thinking` already gets forwarded as the Anthropic `thinking` key and works for Claude models. But `reasoning_effort` is silently dropped - it never reaches Bedrock at all. Models like GLM-5 and Nova on Bedrock use `reasoning_config` instead. It would be nice to also forward `reasoning_effort` as `reasoning_config` in the additional model request fields. Unknown keys in that passthrough map are ignored by models that don't support it. In the two-tiered setup (vllm-router to aigw), vllm-router already passes through unknown fields unchanged, so this would work end-to-end once the translator handles it.

Killusions:
**Description**

In the OpenAI-to-AWS-Bedrock translator, OpenAI clients send reasoning_effort (low/medium/high) in chat completion requests but it was silently dropped. This forwards it as reasoning_config in AdditionalModelRequestFields for Bedrock models like Nova and GLM that support it. Unknown keys in that passthrough map are ignored by models that dont support it.

Fixes #2207

⚒️ with ❤️ by @siemens


Killusions:
@mathetake Could you take a look here?

I would like to also translate it from the "universal" `thinking`, but then it might break for anthropic models. So let's keep this field separate, otherwise we'd need to differentiate between models.

nacx (approved):
LGTM. @aabchoo mind taking a look as well?

Killusions:
@aabchoo @nacx Thank you, what's the current state here? Are there any changes I should make? Should I rebase?

PR #2191, opened 2026-06-03
fix: skip empty-content assistant messages in Bedrock translator

### LINKED ISSUE #2208: Fix: Empty assistant messages crash Bedrock translation
Some clients (like Opencode) send assistant messages with empty content (no text, no tool calls) in the message history during multi-turn conversations. AWS Bedrock's Converse API rejects these. It would be nice to skip empty-content assistant messages during conversion instead of passing them through and getting a provider error.

Killusions:
**Description**

In the OpenAI-to-AWS-Bedrock message converter, Bedrock Converse rejects assistant messages with empty content arrays. Some clients send empty assistant messages. This skips them when they have no content blocks. Tool-call-only messages are preserved.

Fixes #2208

⚒️ with ❤️ by @siemens


internal/translator/openai_awsbedrock.go
```diff
@@ -521,7 +521,9 @@ func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) openAIMessageToBedrockMes
 			if err != nil {
 				return err
 			}
-			bedrockReq.Messages = append(bedrockReq.Messages, bedrockMessage)
+			if len(bedrockMessage.Content) > 0 {
```
nacx, line 526: Can you add a comment explaining why we just ignore empty messages?

Killusions: Done, it's because Converse rejects them with an error (and many other APIs don't).

nacx (approved):

PR #2193, opened 2026-06-04
fix: preserve per-block cache_control for anthropic

tuckervento:
**Description**

When a developer or system message contains multiple content parts (as []ChatCompletionContentPartTextParam), the previous implementation concatenated all part texts into a single string and kept only the last cache_control seen. This destroyed Anthropic prompt caching breakpoints: a caller who marked part 0 as ephemeral-cacheable (e.g., a static system prompt) would lose that annotation after translation.

This commit changes extractSystemPromptFromDeveloperMsg to return []anthropic.TextBlockParam — one block per content part — carrying each part's cache_control through verbatim. Both call sites in openAIToAnthropicMessages simplify to a single append(..., fn()...) as a result.


nacx (commented):
Thank you! Overall LGTM!

PR #2197, opened 2026-06-04
fix: make error processing consistent across ai gateway

hustxiayang:
**Description**
The Anthropic translator (`anthropic_helper.go`) detected several classes of invalid user input during
format conversion — invalid base64 image data, unsupported image media types, malformed tool definitions,
invalid tool_choice values, and unsupported content types — but returned them as plain Go errors. Because
these errors were not wrapped with `ErrInvalidRequestBody`, the extproc error handler couldn't distinguish
them from real internal failures and Envoy defaulted to HTTP 500.

This is the same category of fix as #1621, which established the `ErrInvalidRequestBody` → HTTP 422
framework. The Anthropic translator had several error paths that were missed in that PR.

## Changes

- **Image content**: wrap `parseDataURI` failure and unsupported media type errors with
  `ErrInvalidRequestBody` → 500 becomes 422, matching the Gemini and Bedrock translators
- **Tool definitions**: reject tools with a missing or non-`"function"` type and nil function definition
  instead of silently skipping them → 200 becomes 422
- **Tool parameters**: wrap the `map[string]any` type-cast failure with `ErrInvalidRequestBody` → 500
  becomes 422
- **Tool choice**: wrap unsupported value/type errors with `ErrInvalidRequestBody` → 500 becomes 422
- **Content types**: wrap unsupported audio and file content errors with `ErrInvalidRequestBody` → 500
  becomes 422

## Note

The silent acceptance of malformed tool definitions (missing `type` field) was a behavioral gap, not just
a missing error wrap — the translator was skipping invalid tools without error. The fix changes this to an
explicit rejection, making the Anthropic path consistent with how the Gemini translator handles the same
case.

### DIFF
--- a/internal/translator/anthropic_helper.go
+++ b/internal/translator/anthropic_helper.go
@@ -96,7 +96,7 @@ func translateAnthropicToolChoice(openAIToolChoice *openai.ChatCompletionToolCho
 			toolChoice = anthropic.ToolChoiceUnionParam{OfTool: &anthropic.ToolChoiceToolParam{Name: choice}}
 			toolChoice.OfTool.DisableParallelToolUse = disableParallelToolUse
 		default:
-			return anthropic.ToolChoiceUnionParam{}, fmt.Errorf("unsupported tool_choice value: %s", choice)
+			return anthropic.ToolChoiceUnionParam{}, fmt.Errorf("%w: unsupported tool_choice value: %s", internalapi.ErrInvalidRequestBody, choice)
 		}
 	case openai.ChatCompletionNamedToolChoice:
 		if choice.Type == openai.ToolTypeFunction && choice.Function.Name != "" {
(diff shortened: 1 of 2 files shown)


internal/translator/anthropic_helper.go
```diff
@@ -131,9 +132,13 @@ func translateOpenAItoAnthropicTools(openAITools []openai.Tool, openAIToolChoice
...
-			if openAITool.Type != openai.ToolTypeFunction || openAITool.Function == nil {
-				// Anthropic only supports 'function' tools, so we skip others.
-				continue
+			if openAITool.Type != openai.ToolTypeFunction {
+				err = fmt.Errorf("%w: tool type must be 'function', got %q", internalapi.ErrInvalidRequestBody, openAITool.Type)
+				return
+			}
+			if openAITool.Function == nil {
+				err = fmt.Errorf("%w: tool function definition is required", internalapi.ErrInvalidRequestBody)
+				return
```
nacx, line 141: This is a change of behaviour that can break existing clients. Can you explain why the change od behaviour is needed?

hustxiayang: Hi, sorry to reply late. The current status for ai-gateway is that it provides an unified api to use models for different model providers, but it does not provide a consistent interface for error handling yet. For the same error, the error code would be different (400, or 422 or something else). Thus, users need to test the error codes for each provider respectively, processing them respectively, and even the error code a particular provider could change and cause the error processing crash.  Thus, I think since we provide unified interface to use it, we should provide unified interface for error processing, accordingly?

nacx: In this case, we're trading a portable model where we skip unknown stuff for the provider, to fail the request directly.  How is this helping achieve greater cross-provider portability through the abstraction?

PR #2163, opened 2026-05-27
fix: buffer Responses streaming SSE events

CodePrometheus:
**Description**

This commit buffers streaming OpenAI Responses SSE data across ext_proc `ResponseBody` calls.

In `STREAMED` response body mode, ext_proc body chunks may split a single SSE event. The OpenAI Responses passthrough translator previously parsed each chunk independently, so
a split `response.completed` event could be skipped and usage would remain unset.

This change keeps incomplete SSE bytes in the translator and only parses complete `\n\n`-terminated SSE events.

**Related Issues/PRs (if applicable)**

Fixes #2162

**Special notes for reviewers (if applicable)**

Added tests covering a split `response.completed` event across two `ResponseBody` calls, and a chunk boundary that ends exactly after a complete SSE event.

Validation run:

```bash
go test ./internal/translator
go test ./internal/extproc
```

Issue #2162, opened 2026-05-27
OpenAI Responses passthrough translator drops usage when SSE event spans ext_proc body chunks


CodePrometheus:
_Description_:

When using the OpenAI Responses API passthrough translator with a streaming request, token usage can be lost if a `response.completed` SSE event is split across multiple ext_proc `ResponseBody` calls.

The Responses translator currently parses each `ResponseBody` chunk independently. In streaming response body mode, Envoy ext_proc does not guarantee that body chunks align with SSE event boundaries. If the final `response.completed` event is split before the terminating `\n\n`, the translator cannot parse the full JSON event and does not extract usage.

This results in silent usage loss: the request can complete successfully, but token usage remains unset.

Expected behavior:

The translator should preserve incomplete SSE bytes across `ResponseBody` calls and only parse complete SSE events.

Actual behavior:

The split `response.completed` event is skipped, so usage is not extracted.

_Repro steps_:

1. Configure an OpenAI backend using the Responses API passthrough path.
2. Send a streaming Responses request:

```bash
curl -N https://<gateway-host>/v1/responses \
  -H "content-type: application/json" \
  -H "authorization: Bearer <token>" \
  -d '{
    "model": "gpt-5.5",
    "stream": true,
    "input": "hello, CodePrometheus!"
  }'
```

Note: the end-to-end request depends on Envoy/ext_proc body chunking and may not reproduce deterministically. The deterministic repro is to split one SSE event across two translator `ResponseBody` calls.

3. Reproduce at translator level by splitting a single-line SSE event across two `ResponseBody` calls.

First `ResponseBody` chunk:

```text
data: {"type":"response.completed","response":{"model":"gpt-4o-2024-11-20","usage":{"input_tokens":10,
```

Second `ResponseBody` chunk:

```text
"input_tokens_details":{"cached_tokens":2},"output_tokens":5,"output_tokens_details":{"reasoning_tokens":0},"total_tokens":15}}}
```

Expected result after the second chunk:

```text
input_tokens=10
output_tokens=5
cached_tokens=2
total_tokens=15
```

Actual result:

```text
usage remains unset
```

This is more likely to happen for Responses streaming than for Chat Completions streaming because `response.completed` can contain a large final event, including response metadata and usage. Large SSE events can span multiple ext_proc body chunks.

_Environment_:

- Envoy AI Gateway: main branch
- API path: `/v1/responses`
- Backend schema: OpenAI Responses passthrough
[...]

PR #2198, opened 2026-06-05
fix: buffer non-streaming response body across multi-chunk ext_proc messages

mtparet:
**Description**

When the ext_proc filter is chained behind a `FULL_DUPLEX_STREAMED` ext_proc
filter in Envoy, a **non-streaming** response body can be delivered to
ai-gateway across multiple ext_proc `ResponseBody` messages — most commonly a
single full-body chunk (`EndOfStream=false`) followed by an empty trailing
chunk (`EndOfStream=true`), or a body split across two non-empty chunks.

`upstreamProcessor.ProcessResponseBody` invokes the translator on every chunk,
and the non-streaming translators decode unconditionally with
`json.NewDecoder(body).Decode(...)` (`internal/translator/openai_embeddings.go`,
`internal/translator/openai_openai.go`). The empty/partial chunk therefore
surfaces `io.EOF`, wrapped as `failed to unmarshal body: EOF`, which Envoy
converts into a synthesized **500 with `bytes_sent=0`**.

This is not hit by the default deployment: for non-streaming responses the
processor leaves the ext_proc response body mode at its default (`BUFFERED`) —
it only requests `STREAMED` for streaming `200` responses — so the body
normally arrives in a single chunk. The multi-chunk delivery only occurs when
ai-gateway's ext_proc is chained behind another `FULL_DUPLEX_STREAMED` ext_proc
filter.

This PR:

- Buffers raw response-body bytes for non-streaming responses in
  `ProcessResponseBody` until `EndOfStream=true`, then invokes the translator
  exactly once with the complete body. This is the primary fix and covers both
  the empty-trailing-chunk and split-body shapes for all non-streaming
  translators (chat completions, embeddings, etc.).
- Adds an empty-body guard to the OpenAI embeddings translator
  (`ResponseBody`) as defense in depth, so it is a no-op on an empty body
  rather than returning `io.EOF`.
- Adds processor- and translator-layer regression tests.

Related to the streaming-side counterpart fixed in #2163 (buffer Responses
streaming SSE events across ext_proc body chunks, #2162) — this PR addresses
the analogous problem on the **non-streaming** response path.

- `internal/translator/openai_openai.go` is intentionally **not** modified — the
  processor-level buffer fully covers the chat-completions non-streaming path,
  so no per-translator guard is needed there.
- Regression coverage:
  - `Test_chatCompletionProcessorUpstreamFilter_ProcessResponseBody_MultiChunkBug`
    (`empty_trailing_chunk`, `body_split_across_two_chunks`) — asserts the
    translator is invoked exactly once with the full buffered body.
  - `TestOpenAIToOpenAITranslatorV1EmbeddingResponseBody_MultiChunkBug`
    (`empty_body_end_of_stream`, `empty_body_not_end_of_stream`) — asserts the
    embeddings translator returns no error and zero `TokenUsage` on an empty
    body regardless of `endOfStream`.
  - Both were verified to fail with `failed to unmarshal body: EOF` when the
    processor buffer is removed.

Validation run:

```bash
go test ./internal/extproc ./internal/translator
```

### DIFF
--- a/internal/extproc/processor_impl.go
+++ b/internal/extproc/processor_impl.go
@@ -116,13 +116,20 @@ type (
 		responseEncoding   string
 		compressedBuf      []byte // accumulates raw compressed bytes across streaming chunks
 		decompressedOffset int    // tracks decompressed bytes already returned
-		translator         translator.Translator[ReqT, tracingapi.Span[RespT, RespChunkT]]
-		modelNameOverride  internalapi.ModelNameOverride
-		headerMutator      *headermutator.HeaderMutator
-		bodyMutator        *bodymutator.BodyMutator
-		backendName        string
-		routeName          string
-		handler            filterapi.BackendAuthHandler
+		// responseBodyBuf accumulates raw response-body bytes across HttpBody chunks
+		// for non-streaming responses. Envoy can deliver a non-streaming response across
+		// multiple ResponseBody messages (e.g. chained behind FULL_DUPLEX_STREAMED
+		// ext_proc filters, the final chunk is empty with EndOfStream=true). Buffering
+		// until EndOfStream hands the translator one complete body and avoids io.EOF on
+		// the empty tail.
+		responseBodyBuf   []byte
+		translator        translator.Translator[ReqT, tracingapi.Span[RespT, RespChunkT]]
+		modelNameOverride internalapi.ModelNameOverride
+		headerMutator     *headermutator.HeaderMutator
+		bodyMutator       *bodymutator.BodyMutator
+		backendName       string
+		routeName         string
+		handler           filterapi.BackendAuthHandler
 		// cost is the cost of the request that is accumulated during the processing of the response.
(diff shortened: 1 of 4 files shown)


mathetake (commented):
> When the ext_proc filter is chained behind a FULL_DUPLEX_STREAMED ext_proc
filter in Envoy, a non-streaming response body can be delivered to
ai-gateway across multiple ext_proc ResponseBody messages — most commonly a
single full-body chunk (EndOfStream=false) followed by an empty trailing
chunk (EndOfStream=true), or a body split across two non-empty chunks.

seems like a pure bug in envoy. could you elaborate more on exactly how it happens?

mtparet:
@mathetake I searched and good point, this is this bug https://github.com/envoyproxy/envoy/issues/41654 so we just need to bump the envoy version.

Thank you ! Sorry for surfacing here, I did not think it could be envoy related here..

PR #2205, opened 2026-06-08
feat: add mirror feature to aigatewayroute

### LINKED ISSUE #2137: Feature Request: shadow traffic mirroring for AIGatewayRoute with per-mirror model rewriting
# Summary

- What: Add a mirrors field to AIGatewayRouteRule that sends a fire-and-forget copy of each matched request to one or more shadow backends, with the AI Gateway's upstream ExtProc pipeline applied to the mirror leg (model rewrite,
header/body mutation).
- Why: Evaluate a new model, provider, or region against live production traffic without affecting client responses — a common ask for canary, regression, and A/B-eval workflows.
- Where: New field on AIGatewayRouteRule in api/v1beta1, plus follow-through in the controller and extension server.
- Impact: Backward compatible (opt-in field). Mirror leg billing is suppressed so shadow traffic does not double-count in LLMRequestCosts.

# Problem

Shadow-traffic evaluation is the standard way to qualify a new LLM backend before flipping production routing. Today, an operator can technically add an HTTPRequestMirrorFilter by editing the controller-generated HTTPRoute out-of-band,
but:

1. The change is overwritten on the next reconcile — AIGatewayRoute has no surface for it.
2. The mirror cluster does not receive the AI Gateway's upstream ExtProc filter chain, so model rewriting (x-ai-eg-model, body model field) and header/body mutations do not run — the shadow backend sees a request shaped for the
primary, not for itself.
3. LLMRequestCosts dynamic metadata is emitted for the mirror leg, double-counting tokens in access logs and downstream billing pipelines.

# Proposal

Add a mirrors field to AIGatewayRouteRule. Each entry wraps an AIGatewayRouteRuleBackendRef (so it inherits modelNameOverride, headerMutation, bodyMutation) plus an HTTPRequestMirrorFilter-style percent for sampling:

```
type AIGatewayRouteRule struct {
    // ...existing fields...
    // +optional
    // +kubebuilder:validation:MaxItems=16
    Mirrors []AIGatewayRouteRuleMirror `json:"mirrors,omitempty"`
}

type AIGatewayRouteRuleMirror struct {
    AIGatewayRouteRuleBackendRef `json:",inline"`
    // +optional
    Percent *gwapiv1.Fraction `json:"percent,omitempty"`
}
```

Semantics: responses from mirror backends are always discarded; only the primary backendRefs respond to the client. Each entry maps to one Gateway API HTTPRequestMirrorFilter.

# What this enables

- Side-by-side evaluation of a new model or provider on real traffic.
- Region-failover dry runs (mirror to a candidate region, compare error rates).
- Regression testing of modelNameOverride / translation changes against production payloads.
- Per-tenant audit of model behavior without paying for it twice.

# Open questions

1. Per-mirror translation shape: should mirrors carry full modelNameOverride / headerMutation / bodyMutation, or reference an AIServiceBackend only and inherit its config? Inheriting is cleaner but forces operators to create a dedicated AIServiceBackend per shadow target.
2. Cost emission policy: always-suppress LLMRequestCosts on mirrors, opt-in via a field on the mirror entry, or controlled at BackendSecurityPolicy / GatewayConfig level?
3. Mirror-cluster naming contract: the extension server needs to identify mirror clusters in the xDS push to install (or skip) the upstream filter chain. Envoy Gateway currently emits names of the form
httproute/<ns>/<name>/rule/<ruleIdx>-mirror-<mirrorIdx> with 1-based mirror indices. Is it acceptable to depend on that wire format, or should AI Gateway negotiate a stable contract / metadata-based marker with Envoy Gateway?
4. Failure handling: should mirror-leg ExtProc failures be silently absorbed (current Envoy mirror semantics) or surface as gen_ai.* error metrics tagged is_mirror=true?

# Related

- Gateway API HTTPRequestMirrorFilter spec.
- #2135 — maybeModifyCluster cluster-name parsing assumptions; overlap with question 4 above.

We have a working PoC on a fork with e2e coverage and would be happy to upstream it once the design questions above are resolved.

mtparet:
**Description**

Adds a `mirrors` field to `AIGatewayRouteRule` that sends a fire-and-forget copy of each matched request to one or more shadow backends, with the AI Gateway's upstream ExtProc pipeline applied to the mirror leg (model rewrite, header/body mutation). This is the standard way to qualify a new model, provider, or region against live production traffic without affecting client responses.

Each mirror entry embeds an `AIGatewayRouteRuleBackendRef`, so a mirror references an `AIServiceBackend` by name and inherits the same per-backend transformations as primary `backendRefs` — most importantly `modelNameOverride`, so the shadow backend can be addressed with a different model name. Each entry maps to one Gateway API `HTTPRequestMirror` filter and supports `percent`/`fraction` sampling.

Concretely:
- **api**: new `AIGatewayRouteRuleMirror` (embeds `AIGatewayRouteRuleBackendRef`, inherits its InferencePool CEL rule; `percent`/`fraction`).
- **controller**: `newHTTPRoute` emits a `RequestMirror` filter per mirror (rejecting InferencePool mirrors); `reconcileFilterConfigSecret` emits a `filterapi.Backend{IsMirror: true}` per mirror so ExtProc can resolve the shadow backend's overrides by cluster-metadata name lookup.
- **extension server**: `maybeModifyCluster` installs the same upstream ExtProc + header-mutation filters on mirror clusters as on primary clusters, and tags cluster/endpoint metadata with a mirror marker. Envoy Gateway names mirror clusters `<ruleIdx>-mirror-<mirrorIdx>` with **1-based** mirror indices, converted back to 0-based for slice access.
- **extproc**: mirror legs skip `LLMRequestCost` dynamic-metadata emission so shadow traffic does not double-bill tokens through the access-log / billing pipeline.

Fixes #2137

This addresses the open questions from #2137 along the lines @nacx suggested in the issue:
- *Per-mirror translation shape* — mirrors are ref-based (embed an `AIServiceBackend` ref) but also carry `modelNameOverride`, so operators don't need a dedicated `AIServiceBackend` per shadow target.
- *Cost emission* — this PR suppresses `LLMRequestCost` on the mirror leg (avoids double-billing). Surfacing shadow cost via an `is_mirror`/`shadow` metric attribute (your suggestion) is left to a **follow-up** so cost can be observed per-leg without double-counting.
- *Mirror-cluster naming contract* — this depends on Envoy Gateway's current `<ruleIdx>-mirror-<mirrorIdx>` wire naming. Proposing we start with that convention and open a separate issue to discuss a stable contract / metadata marker with the EG community (**follow-up**).

Not breaking for existing users — the `mirrors` field is new and opt-in.


internal/controller/gateway.go
```diff
@@ -481,6 +481,60 @@ func (c *GatewayController) reconcileFilterConfigSecret(
...
+				if _, exists := routeBackendNamesSet[b.Name]; !exists {
+					routeBackendNamesSet[b.Name] = struct{}{}
+					routeBackendNames = append(routeBackendNames, b.Name)
+				}
+			}
+			// Emit a filterapi.Backend entry for each mirror destination so that
+			// extproc, running on the upstream HTTP filter chain of the mirror
+			// cluster, can resolve the shadow backend's ModelNameOverride and
+			// header/body mutations by cluster-metadata backend name lookup.
+			for mirrorIndex := range rule.Mirrors {
```
siddharth1036, line 477: The code in this block seems more or less similar to the L421-L489, can we refactor it to avoid duplication?

api/v1beta1/ai_gateway_route.go
```diff
@@ -281,6 +281,49 @@ type AIGatewayRouteRule struct {
...
+	// Responses from mirror backends are always discarded; only the primary backendRefs
+	// respond to the client. Mirror traffic is cloned via the Gateway API HTTPRequestMirror
+	// filter, but each mirror's BackendRef supports the same AI-specific transformations
+	// (ModelNameOverride, HeaderMutation, BodyMutation) as primary backendRefs, so the
+	// shadow backend can be addressed with a different model name or modified request body.
+	// Each entry corresponds to one shadow backend; use Percent or Fraction to sample.
+	//
+	// +optional
+	// +kubebuilder:validation:MaxItems=16
+	Mirrors []AIGatewayRouteRuleMirror `json:"mirrors,omitempty"`
```
siddharth1036, line 295: should we add this to alpha CRD as well?

internal/controller/gateway.go
mtparet: Thanks, I just pushed this refactor.

api/v1beta1/ai_gateway_route.go
mtparet: Indeed, I missed that.

internal/extensionserver/post_translate_modify.go
```diff
@@ -213,7 +248,56 @@ func (s *Server) maybeModifyCluster(ctx context.Context, cluster *clusterv3.Clus
...
+			setClusterMetadataBackendName(cluster, aigwRoute.Namespace, mirror.BackendRef.Name,
+				aigwRoute.Name, httpRouteRuleIndex, mirrorIndex)
+			// Overwrite with the mirror-specific name (setClusterMetadataBackendName uses
+			// PerRouteRuleRefBackendName; we need the mirror variant).
+			cluster.Metadata.FilterMetadata[internalapi.InternalEndpointMetadataNamespace].
+				Fields[internalapi.InternalMetadataBackendNameKey] = structpb.NewStringValue(mirrorBackendName)
+			// Mark this cluster as a mirror leg so the cost emitter in extproc can
+			// skip LLMRequestCost emission and avoid double-billing.
+			cluster.Metadata.FilterMetadata[internalapi.InternalEndpointMetadataNamespace].
+				Fields[internalapi.InternalMetadataMirrorKey] = structpb.NewBoolValue(true)
```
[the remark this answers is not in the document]
mtparet: No panic possible here — setClusterMetadataBackendName one line above already does all the nil-init, so the chain was guaranteed non-nil. But the safety was non-local, so I refactored anyway: extracted ensureClusterInternalMetadata() which owns the nil-checks, and the mirror branch now writes through it directly (also removes the write-then-overwrite of the backend name).

PR #2222, opened 2026-06-11
ci: make publishing jobs configurable instead of assuming credentials

JAORMX:
**Description**

The Release and Build and Test workflows unconditionally run their publishing jobs (`docker_push`, `helm_push`, `release`) on tag and branch pushes, assuming DockerHub credentials are configured and that artifacts always go to `docker.io/envoyproxy`. On any repository where that isn't true — forks, downstream distributions, test mirrors — every tag or branch push produces a run that fails after ~35s on "Username and password required", polluting the checks UI and making it harder to spot real failures.

This makes publishing configuration-driven instead:

* The publishing jobs are gated on `vars.DOCKERHUB_USERNAME != ''` (the `vars` context is available in job-level `if` expressions, unlike `secrets`). Repositories without publishing credentials skip the jobs cleanly instead of failing; dependent jobs (`release`, `helm_push`) skip automatically via `needs`.

* An optional `OCI_REGISTRY` repository variable is introduced, defaulting to `docker.io/envoyproxy`. The Makefile already declares `OCI_REGISTRY ?=`, so the job-level env flows into the image tags (via `OCI_REPOSITORY_PREFIX`) and the `helm push` destination with no Makefile changes. The login steps derive the registry host from it (e.g. `ghcr.io/myorg` → `ghcr.io`); `docker/login-action` defaults to DockerHub when it is empty.

Behavior on `envoyproxy/ai-gateway` itself is unchanged: the existing `DOCKERHUB_USERNAME` variable is set, so the jobs run exactly as before and publish to the same registry. The benefit is for everyone else carrying a copy of these workflows — out of the box, publishing skips quietly, and a repository that does want to publish only needs to set `DOCKERHUB_USERNAME`, `DOCKERHUB_PASSWORD`, and optionally `OCI_REGISTRY`, rather than patching the workflows. This also avoids the common alternative of hardcoding `github.repository == 'envoyproxy/ai-gateway'` guards, which downstream copies then have to maintain or invert.

Verified with actionlint (no findings).

n/a

The `registry:` input passed to `docker/login-action` is the host portion of `OCI_REGISTRY`; for the default `docker.io/envoyproxy` this resolves to `docker.io`, which is identical to the action's default behavior.


mathetake (approved):
This is fine but note that the maintenance of this will be best-effort by nature

PR #2226, opened 2026-06-12
fix: prevent empty namespace from breaking RBAC name rendering in helm chart

ChrisJBurns:
**Description**

The controller and inference-pool `ClusterRole`/`ClusterRoleBinding` names are built with `printf "%s:%s" <base> .Release.Namespace` to keep them unique per-release (introduced in #1511). When `.Release.Namespace` is empty, this renders `name: <base>:` — an unquoted trailing colon that strict YAML parsers reject with `mapping values are not allowed in this context`. The whole render fails and zero resources are produced.

This doesn't trigger via `helm install` or `helm template` (the CLI always defaults the namespace), only where a renderer leaves it unset — e.g. the Helm SDK's `engine.Render` without `ReleaseOptions`, or Replicated's image-extraction render.

The fix adds two independent layers of defense across the three affected helpers (`_helpers.tpl`) and their six call sites (`serviceaccount.yaml`, `envoy_gateway_cluster_role_for_inference_pool.yaml`):

- Wrap the namespace in `default "default"` so the `printf` argument is never empty. No behavior change for real installs, where the namespace is always populated.
- Quote the rendered `name` fields so a trailing colon is always unambiguous.

Either layer fixes the failure on its own; both are included for robustness.

Verified with `helm lint` and `helm template` (default and explicit namespaces) — names render correctly as `<base>:<namespace>` with no change to existing behavior.

### DIFF
--- a/manifests/charts/ai-gateway-helm/templates/_helpers.tpl
+++ b/manifests/charts/ai-gateway-helm/templates/_helpers.tpl
@@ -76,7 +76,7 @@ Create the name of the cluster role to use
 {{- if $existing }}
 {{- (include "ai-gateway-helm.controller.serviceAccountName" .) }}
 {{- else }}
-{{- printf "%s:%s" (include "ai-gateway-helm.controller.serviceAccountName" .) .Release.Namespace }}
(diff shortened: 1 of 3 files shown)


nacx (approved):
Thanks!

PR #2227, opened 2026-06-12
fix: insert AI Gateway extproc after the buffer filter

ChrisJBurns:
**Description**

**Problem**

If a buffer filter (`envoy.filters.http.buffer`) is added to the chain, for example via an EnvoyPatchPolicy to raise the request buffer limit, the AI Gateway extproc could be placed ahead of it. The AI Gateway extproc runs with `RequestBodyMode: BUFFERED`, so when it runs before the buffer filter it buffers the body against Envoy Gateway's default 32 KiB downstream per-connection limit and returns `413 Payload Too Large` for larger bodies before the buffer filter's higher limit applies.

The buffer filter raises the stream's buffer limit to `max_request_bytes`, but only for filters that decode after it. This was hit in practice with large request bodies such as Gemini CLI tool definitions.

**Solution**

When a buffer filter sits at or after the computed ext_proc insertion point, insert the AI Gateway extproc immediately after the last buffer filter. Envoy Gateway already orders the buffer filter before ext_proc, so this honors that order. Behavior is unchanged when no buffer filter is present.

This works because EnvoyPatchPolicy patches are applied before the extension hook runs and Envoy Gateway does not re-sort filters afterward, so the buffer filter is visible to this code.

The extension server only moves its own filter, not the user's buffer filter, so this fixes the 413 for AI Gateway's extproc specifically and not for other co-resident BUFFERED ext_procs.

**Testing**

Added table tests for three cases: ext_proc before buffer (inserted after buffer), buffer before ext_proc (inserted after buffer), and no buffer filter (unchanged).

### DIFF
--- a/internal/extensionserver/post_translate_modify.go
+++ b/internal/extensionserver/post_translate_modify.go
@@ -854,9 +854,20 @@ func shouldAIGatewayExtProcBeInserted(filters []*httpconnectionmanagerv3.HttpFil
 
 // insertAIGatewayExtProcFilter inserts the AI Gateway extproc filter into the HTTP connection manager.
 //
-// The order is simple: make sure that the AI Gateway extproc filter is the very first extproc filter in the standard
-// Envoy Gateway order. See:
+// The order is mostly simple: make sure that the AI Gateway extproc filter is the first extproc filter in the
+// standard Envoy Gateway order. See:
 // https://github.com/envoyproxy/gateway/blob/f1e6dab770fabc70d175237380eedfc1f9b1a9e5/internal/xds/translator/httpfilters.go#L93
+//
+// There is one exception: the HTTP buffer filter (envoy.filters.http.buffer). In Envoy Gateway's canonical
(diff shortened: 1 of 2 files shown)


johnugeorge:
@siddharth1036

nacx (approved):

siddharth1036:
Hi @ChrisJBurns , just wanted to check I am working on a PR: https://github.com/envoyproxy/ai-gateway/pull/2158 where you will be able to configure exact (or prefixed, under discussion) filter name which would be allowed to run before ai-gateway-extproc, could it be useful to solve your usecase as well?

ChrisJBurns:
Thanks for flagging this @siddharth1036 - we're definitely in the same code path. I think they're complementary rather than overlapping though: your allowlist is opt-in, but the buffer filter ordering needs to apply by default since it's a correctness fix (large bodies get a 413 otherwise). Could we treat the buffer filter as an always-on default in #2158, or let the automatic buffer handling stack on top of your allowlist?

PR #2229, opened 2026-06-12
fix: handle unframed JSON body with text/event-stream content type in MCP proxy

### LINKED ISSUE #2218: MCP proxy drops tools/list payload when backend sends Content-Type: text/event-stream with an unframed JSON body
### Summary

The MCP proxy silently drops the JSON-RPC payload of an upstream response (e.g. `tools/list` returns an empty `{"tools":[]}`) when the upstream MCP server responds with **`Content-Type: text/event-stream` but a plain, unframed JSON body** (no SSE `event:`/`data:` lines). The proxy appears to parse the body strictly as SSE based on the advertised content-type, finds no `data:` events, yields zero JSON-RPC messages, and returns an empty result — without error. The session otherwise initializes fine, so the MCP server "connects" but exposes no tools.

### Version

- `ai-gateway-controller:v0.7.0` (Envoy Gateway v1.8.0)
- MCPRoute with an external HTTPS `Backend` + `securityPolicy.oauth`; the backend's per-request credential injected via `backendRefs[].securityPolicy.apiKey`.

### Repro

An upstream whose `tools/list` responds like this (real example: `api.refero.design`'s hosted MCP server):

```
HTTP/2 200
content-type: text/event-stream          <-- advertises SSE
content-length: 24668

{"jsonrpc":"2.0","id":2,"result":{"tools":[{"name":"...", ...}, ... 8 tools ...]}}
```

i.e. the **content-type is `text/event-stream` but the body is raw JSON** — there are **zero** `event:` / `data:` lines. (Its `initialize` response, by contrast, correctly uses `application/json` — so the same server is inconsistent, which is its own bug, but proxies see this in the wild.)

- **Through the AIEG MCP proxy:** `tools/list` → `{"tools":[]}` (empty). `initialize` succeeds, a session id is issued, no error is surfaced.
- **Hitting the same upstream directly** with the identical request sequence (same path `/mcp`, same `Authorization: Bearer`, same `initialize → notifications/initialized → tools/list`, same empty `mcp-session-id`): returns all **8 tools**. Verified the proxy's outbound request to the backend (Envoy debug logs) carries the correct `authorization`, path, and headers — and the backend cluster returns `upstream_rq_2xx`. So the proxy *receives* the 8-tool response and drops it.
- **Contrast — a backend that returns proper SSE framing works:** `mcp.firecrawl.dev` returns `tools/list` as `content-type: text/event-stream` **with** real framing (`event: message\nid: ...\ndata: {...}\n\n`), and the proxy returns all 23 tools correctly.

So the differentiator is purely: **SSE-framed body → works; unframed-JSON body mislabeled as `text/event-stream` → silently empty.**

### Expected

When the response is `text/event-stream` but the SSE parse produces **no events** (or the body is valid JSON), the proxy should fall back to parsing the body as a raw JSON-RPC message — the same JSON/SSE/decompress robustness cascade introduced in #1996/#1997 — rather than returning an empty result. At minimum it should not *silently* drop a non-empty payload.

### Impact

Any MCP server that mislabels a JSON body as `text/event-stream` (not uncommon among hosted/proxied servers) appears to the client as connected-but-toolless, with no error to indicate why. This is hard to diagnose because `initialize` succeeds and the HTTP status is 200.

### Notes

Happy to provide full config dumps / debug traces. The `initialize`-vs-`tools/list` content-type inconsistency is the upstream server's bug, but the proxy silently discarding a valid 200 JSON-RPC payload (rather than JSON-fallback) is the gateway-side robustness gap this issue is about.

RegondaChandan:
**Description**

The MCP proxy parses upstream response bodies strictly according to the advertised
Content-Type. When a backend responds with `Content-Type: text/event-stream` but a
plain, unframed JSON body (no `event:`/`data:` lines), the SSE parser finds no events
and the JSON-RPC payload is silently dropped: `tools/list` returns `{"tools":[]}`,
the session initializes fine, and no error is surfaced. Such backends exist in the
wild (e.g. api.refero.design's hosted MCP server).

This is the mirror image of the existing fallback for backends that send SSE data
despite `Content-Type: application/json` (e.g. Slack MCP). For responses advertising
SSE, the proxy now peeks at the first significant byte of the stream: a valid SSE
stream never starts with `{`, so in that case the body is read in full and decoded
as a single JSON-RPC message, the response Content-Type is corrected to
`application/json`, and the message goes through the same modification/write path
as the JSON branch (which is extracted into a shared helper). If the body turns out
not to be valid JSON-RPC, it still falls through to the SSE parser over the buffered
bytes.

Genuine SSE streaming is unaffected: detection only consumes leading whitespace via
a buffered peek, and the buffered reader is handed to the SSE parser as before.

Fixes #2218

The three new regression tests fail on main (the unframed JSON payload is dropped
and the recorded body is empty) and pass with this change.


internal/mcpproxy/handlers.go
```diff
@@ -782,13 +783,73 @@ func copyProxyHeaders(resp *http.Response, w http.ResponseWriter) {
...
+	return responseError
+}
+
+// startsWithJSONObject reports whether the stream begins with '{', consuming only
+// leading whitespace, which is insignificant to both SSE and JSON parsing.
+func startsWithJSONObject(br *bufio.Reader) bool {
+	for {
+		b, err := br.ReadByte()
+		if err != nil {
+			return false
```
Hritik003, line 847: Can use Peek here? Peek(n) returns a view of the upcoming bytes without advancing the stream.

internal/mcpproxy/handlers.go
```diff
@@ -782,13 +783,73 @@ func copyProxyHeaders(resp *http.Response, w http.ResponseWriter) {
...
+}
+
 func (m *mcpRequestContext) proxyResponseBody(ctx context.Context, s *session, w http.ResponseWriter, resp *http.Response,
 	req *jsonrpc.Request, backend filterapi.MCPBackend,
 ) error {
 	// Some backends (e.g. Slack MCP) send SSE data despite Content-Type: application/json.
 	// Try to decode as a single JSON-RPC message first; if that fails, fall through to the
 	// SSE parser using the already-read bytes.
-	var sseReader io.Reader = resp.Body
+	var sseReader io.Reader
```
Hritik003, line 865: can you explain this change?

Hritik003 (commented):
I guess this #1997 missed the case you are addressing in the PR, thanks for fixing it

Hritik003:
just thinking out loud, 

In`text/event-stream` path, it first calls **startsWithJSONObject**, which only recognizes ASCII whitespace and {. But when the first byte is 0xEF (the BOM), which is neither—it returns false and then the message is treated as SSE, handed to the SSE parser, and effectively dropped. It's a minor edge case (only matters for backends that both mislabel the content type and emit a BOM). Can add a one liner which trims the BOM prefix before reading

```go
func startsWithJSONObject(br *bufio.Reader) bool {
	buf, _ := br.Peek(br.Size())
	buf = bytes.TrimPrefix(buf, utf8BOM) < ---add this snippet?
	for _, b := range buf {
	.......
	}
	return false
}
```

internal/mcpproxy/handlers.go
RegondaChandan: good point, switched the BOM check over to Peek. i left the whitespace skip
as a byte scan on purpose though, since this branch also handles real SSE
streaming. Peek blocks until it has the full count buffered (or hits EOF), so
peeking a big window could hold back the first event from flushing. just
peeking the 3 BOM bytes keeps that path cheap. can make the whole thing a
single Peek if you'd rather.

internal/mcpproxy/handlers.go
RegondaChandan: sure. the `= resp.Body` was just a default for sseReader. once i moved the
json write path out into writeSingleJSONRPCMessage, both branches set sseReader
themselves now (json branch sets bytes.NewReader(body) or returns early, the
event-stream branch sets the bufio reader or returns early), so resp.Body never
gets read unwrapped anymore. that left the initial value dead so i dropped it.

RegondaChandan:
@Hritik003 nice catch, fixed in d56acce8. startsWithJSONObject skips a leading BOM now before scanning, so a BOM-prefixed body doesn't get handed to the SSE parser and dropped. tryDecodeJSONRPCMessage already strips the BOM on the decode side so it was really just the detection missing it. added a regression test too.

Hritik003 (commented):
lgtm

PR #2234, opened 2026-06-15
fix: namespace _meta.ui.resourceUri and Content URIs in tools/call results

### LINKED ISSUE #2228: `resources/read` fails for MCP tool `_meta.ui.resourceUri` in multiplexing configuration
_Description_:

When the Envoy AI Gateway is running in multiplexing mode and a backend MCP server returns tool metadata containing a `_meta.ui.resourceUri` (e.g. ui://prefab/tool/xxx/renderer.html), Envoy forwards that URI to the agent unchanged. When the agent subsequently issues a `resources/read` call back to Envoy using that URI, Envoy rejects it with:

```
invalid resource URI: ui://prefab/tool/xxx/renderer.html
```

Envoy AI Gateway should:
1. Rewrite the `resourceUri` in `_meta.ui` at tool-response time so it points to a valid Envoy-routable URI. f.e. `ui://backend/prefab/tool/xxx/renderer.html`.
2. Recognise and proxy `resources/read` calls for `ui://` URIs originating from a named upstream MCP backend.

_Repro steps_:

1. Start Envoy AI Gateway in multiplexing configuration with at least one upstream MCP backend (e.g. `backend`).
2. Have an agent call a multiplexed tool, e.g. `backend__mcp_app`.
3. Envoy forwards the `mcp_app` tool call to the `backend` MCP server.
4. The `backend` MCP server returns a tool result containing:

```json
{
   "_meta": {
     "ui": {
       "resourceUri": "ui://prefab/tool/xxx/renderer.html"
     }
   }
 }
```

5. Envoy forwards this metadata to the agent as-is (no URI rewriting).
6. Agent issues a resources/read call to Envoy with the URI `ui://prefab/tool/xxx/renderer.html`.
7. Envoy returns an error: `invalid resource URI: ui://prefab/tool/xxx/renderer.html`.

_Environment_:

* Envoy AI Gateway `v0.7.0`
* Envoy Gateway `v1.8.1`

_Logs_:

```
[envoy] {":authority":"localhost:3002","bytes_received":113,"bytes_sent":132,"connection_termination_details":null,"downstream_local_address":"127.0.0.1:3002","downstream_remote_address":"127.0.0.1:45176","duration":12,"method":"POST","protocol":"HTTP/1.1","requested_server_name":null,"response_code":400,"response_code_details":"via_upstream","response_flags":"-","route_name":"httproute/default/ai-eg-mcp-main-mcp-route/rule/0/match/0/*","start_time":"2026-06-12T08:57:36.193Z","upstream_cluster":"httproute/default/ai-eg-mcp-main-mcp-route/rule/0","upstream_host":"127.0.0.1:9856","upstream_local_address":"127.0.0.1:41408","upstream_transport_failure_reason":null,"user-agent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:151.0) Gecko/20100101 Firefox/151.0","x-envoy-origin-path":null,"x-envoy-upstream-service-time":null,"x-forwarded-for":"10.244.0.73","x-request-id":"931afadb-475e-4602-a185-37ea2368d10f","x_mcp_method":"resources/read","x_mcp_resource":"ui://prefab/tool/xxx/renderer.html","x_mcp_tool":null,"x_user_id":"ytuy2"}
```

adilfulara:
**Description**
In MCP multiplexing mode, `tools/call` results carrying `_meta.ui.resourceUri` (or `ResourceLink`/`EmbeddedResource` URIs in `Content[]`) passed through the gateway bare. The agent then called `resources/read` with the unnamespaced URI, `upstreamResourceURI` found no `+` separator, and the gateway returned HTTP 400.

The fix extends `maybeResponseModify` to handle `tools/call` in addition to the existing `resources/read` case. It adds two helpers — `rewriteMetaResourceURIs` (walks a declarative path table, includes an idempotency guard) and `rewriteToolResultURIs` (covers `_meta`, `ResourceLink`, and `EmbeddedResource` surfaces) — and applies the same rewrite to tool descriptors in `mergeToolsList`. The inverse path (`upstreamResourceURI` on the `resources/read` request) already worked correctly and required no change.

Unit tests cover all three rewrite surfaces, the idempotency guard, nil/wrong-type no-op cases, and a regression guard for `resources/read`. Three new data-plane round-trip tests prove the 400 is gone end-to-end for each surface.

Fixes #2228

> This PR was developed with AI assistance (Claude Code).

### DIFF
--- a/internal/mcpproxy/handlers.go
+++ b/internal/mcpproxy/handlers.go
@@ -945,21 +945,35 @@ func (m *mcpRequestContext) maybeUpdateProgressTokenMetadata(ctx context.Context
 	return true
 }
 
-// maybeResponseModify modifies the client->server response to include the backend name where needed.
+// maybeResponseModify modifies the server->client response to include the backend name where needed.
 func (m *mcpRequestContext) maybeResponseModify(_ context.Context, req *jsonrpc.Request, msg *jsonrpc.Response, backend filterapi.MCPBackendName) error {
-	if req.Method != "resources/read" || msg.Result == nil {
+	if msg.Result == nil {
 		return nil
 	}
-
-	result := &mcp.ReadResourceResult{Contents: []*mcp.ResourceContents{}}
-	if err := json.Unmarshal(msg.Result, result); err != nil {
-		return fmt.Errorf("failed to unmarshal resources/read result: %w", err)
+	switch req.Method {
+	case "resources/read":
+		// No idempotency guard needed here: handleResourceReadRequest strips the backend
+		// prefix from the request URI before forwarding, so the backend always returns a
+		// bare URI and a single downstreamResourceURI call is always correct.
+		result := &mcp.ReadResourceResult{Contents: []*mcp.ResourceContents{}}
+		if err := json.Unmarshal(msg.Result, result); err != nil {
+			return fmt.Errorf("failed to unmarshal resources/read result: %w", err)
+		}
+		for _, res := range result.Contents {
+			res.URI = downstreamResourceURI(res.URI, backend)
+		}
+		msg.Result, _ = json.Marshal(result) // Already decoded result, so ignore error.
+	case "tools/call":
+		result := &mcp.CallToolResult{}
+		if err := json.Unmarshal(msg.Result, result); err != nil {
+			// Non-standard result shape — pass through unchanged; the backend owns this payload.
+			m.l.Debug("tools/call result is not a standard CallToolResult, skipping URI rewrite", slog.String("error", err.Error()))
+			return nil
+		}
+		if rewriteToolResultURIs(result, backend) {
+			msg.Result, _ = json.Marshal(result) // Already decoded result, so ignore error.
+		}
 	}
-	for _, res := range result.Contents {
-		res.URI = downstreamResourceURI(res.URI, backend)
-	}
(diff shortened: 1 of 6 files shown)


jappievw:
Thanks for jumping on to this. I'm afraid it requires a more elaborate change. The MCP Apps specification requires the `ui://` protocol to be used specifically. Specification: https://github.com/modelcontextprotocol/ext-apps/blob/main/specification/2026-01-26/apps.mdx.

So, agents rendering the UI will fail when the protocol is changed to `backend+ui://`.

adilfulara:
Thank you for the review — the feedback is correct. Prefixing `ui://` with `backend+` breaks the scheme agents rely on for rendering.

One approach we considered: maintain a per-session `{uri → backend}` map, populated when `tools/call` results or `tools/list` descriptors carry `_meta.ui.resourceUri`. When `resources/read` arrives with a bare `ui://` URI, the gateway consults the session map to find the originating backend. This preserves the `ui://` scheme end-to-end.

This works within the existing statefulness model (the gateway already requires sticky sessions for `Mcp-Session-Id` forwarding), but we want to check before implementing: is this the direction you had in mind, or does the spec / your use case call for a different mechanism?

PR #2236, opened 2026-06-16
fix: prevent double counting of Anthropic streaming cache tokens

Aias00:
**Description**

The `message_delta` event handler in `anthropicStreamParser` was using `Add*` methods for input tokens, cached tokens, and cache creation tokens. However, the Anthropic API reports cumulative (not incremental) token counts in `message_delta`, and `message_start` already sets these values correctly via `ExtractTokenUsageFromExplicitCaching`. Using `Add` on top of the already-set cumulative values caused cache tokens to be counted twice.

### Bug Example

With `input_tokens=9` and `cache_read_input_tokens=1`:

| Step | Action | InputTokens | CachedInputTokens |
|------|--------|-------------|-------------------|
| `message_start` | `SetInputTokens(10)`, `SetCachedInputTokens(1)` | 10 ✅ | 1 ✅ |
| `message_delta` (before fix) | `AddInputTokens(1)`, `AddCachedInputTokens(1)` | 11 ❌ | 2 ❌ |
| `message_delta` (after fix) | Only `SetOutputTokens(16)` | 10 ✅ | 1 ✅ |

### Root Cause

`ExtractTokenUsageFromExplicitCaching` computes `InputTokens = input_tokens + cache_read + cache_creation`. In `message_start`, this is correctly set via `Set*` methods. In `message_delta`, the same cumulative values were being added via `Add*` methods, causing double counting.

### Fix

The `message_delta` handler now only updates output tokens using `SetOutputTokens`, since `message_start` typically reports `output_tokens=0` and `message_delta` provides the final count. This matches the pattern already used in `anthropic_anthropic.go`'s `reflectStreamingEvent`. The `MessageDeltaUsage` struct uses non-pointer `int64` fields that default to 0 when absent from JSON, making it impossible to distinguish "not provided" from "actually zero" — so we cannot safely `Set` override from delta either, as it would erase correct `message_start` values with zeros.

### Changes

- `internal/translator/anthropic_helper.go`: Removed `Add*` calls for input/cache tokens from `message_delta` handler; replaced `AddOutputTokens` with `SetOutputTokens`; removed unused `ExtractTokenUsageFromExplicitCaching` call; added `>= 0` guard for `ThinkingTokens`
- `internal/translator/anthropic_helper_test.go`: Added `TestAnthropicStreamParserTokenUsage_NoDoubleCounting` with 6 test cases

### Test Coverage

- Cache tokens in both `message_start` and `message_delta` (core double-count scenario)
- Cache creation tokens in both events
- Both cache_read and cache_creation in both events
- No cache tokens (baseline correctness)
- Cache tokens only in `message_start`, not in `message_delta`
- Cache tokens only in `message_delta` — verifies they are ignored (non-pointer int64 fields cannot distinguish absent from zero)


Aias00:
close: https://github.com/envoyproxy/ai-gateway/issues/2279

nacx (approved):
Overall LGTM.
@aabchoo mind taking a look?

PR #2238, opened 2026-06-16
mcp: limit request body size by default

nacx:
**Description**

Limit by default the size of MCP request payloads.

N/A

N/A

### DIFF
--- a/internal/mcpproxy/handlers.go
+++ b/internal/mcpproxy/handlers.go
@@ -276,8 +276,18 @@ func (m *mcpRequestContext) servePOST(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 
-	body, err := io.ReadAll(r.Body)
+	limit := m.maxRequestBodySize
+	if limit <= 0 {
+		limit = defaultMaxRequestBodySize
+	}
+	body, err := io.ReadAll(http.MaxBytesReader(w, r.Body, limit))
 	if err != nil {
+		var maxBytesErr *http.MaxBytesError
+		if errors.As(err, &maxBytesErr) {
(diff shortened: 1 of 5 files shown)


internal/mcpproxy/handlers.go
```diff
@@ -276,8 +276,18 @@ func (m *mcpRequestContext) servePOST(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 
-	body, err := io.ReadAll(r.Body)
+	limit := m.maxRequestBodySize
+	if limit <= 0 {
```
aabchoo, line 280: Should this validation step occur in [getMaxRequestBodySize](https://github.com/envoyproxy/ai-gateway/pull/2238/changes#diff-a2867b8ddb3037f0880146c86c9ed6a8bf2abd2d0cc6c89bdf7ad060bc7d0501R49-R57) instead?

aabchoo: + add tests for this case

nacx: This is already handled in that method; the value is only set if `> 0` otherwise defaults to the default value.
This condition should never be reached in production code. It is more intended for testing, where many of the tests manually create the objects and don't go via the constructor. We could update all the tests, but it's going to be more tedious to maintain and add it always to new tests, etc, so I opted to have this default here as well.
If you feel strongly about it, I can update all the tests as well (it's going to become a bit noisy).

>add tests for this case

All tests that manually create the ProxyConfig are already exercising this path.

aabchoo (approved):
looks fine just a nit

PR #2239, opened 2026-06-16
fix: double counting for the usage in stream mode for anthropic models

hustxiayang:
**Description**

`ExtractTokenUsageFromExplicitCaching`  is called in both start message and the delta message, as a result. The usage report is not correct.

Fix: remove the accumulation in the message delta; also the previous tests can pass because the test cases did not mock the real cases, update all related tests cases as well.


internal/translator/anthropic_helper.go
```diff
@@ -1061,16 +1061,12 @@ func (p *anthropicStreamParser) handleAnthropicStreamEvent(eventType []byte, dat
...
-			// Accumulate any additional cache tokens from delta
-			p.tokenUsage.AddCachedInputTokens(cached)
+			p.tokenUsage.SetCachedInputTokens(cached)
 		}
 		if cacheCreation, ok := usage.CacheCreationInputTokens(); ok {
-			p.tokenUsage.AddInputTokens(cacheCreation)
-			// Accumulate cache creation tokens
-			p.tokenUsage.AddCacheCreationInputTokens(cacheCreation)
+			p.tokenUsage.SetCacheCreationInputTokens(cacheCreation)
 		}
```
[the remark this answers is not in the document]
hustxiayang: updated, thanks!

internal/translator/anthropic_helper.go
```diff
@@ -1061,17 +1061,8 @@ func (p *anthropicStreamParser) handleAnthropicStreamEvent(eventType []byte, dat
...
-			p.tokenUsage.AddInputTokens(cached)
-			// Accumulate any additional cache tokens from delta
-			p.tokenUsage.AddCachedInputTokens(cached)
-		}
-		if cacheCreation, ok := usage.CacheCreationInputTokens(); ok {
-			p.tokenUsage.AddInputTokens(cacheCreation)
-			// Accumulate cache creation tokens
-			p.tokenUsage.AddCacheCreationInputTokens(cacheCreation)
-		}
+		// Cache token details are already set in message_start — don't touch them here
```
yuzisun: can you update the tests

hustxiayang: sure, thanks a lot

hustxiayang: done

hustxiayang:
@yuzisun Do you mind to review this? I have update the tests

internal/translator/anthropic_helper.go
```diff
@@ -1061,17 +1061,8 @@ func (p *anthropicStreamParser) handleAnthropicStreamEvent(eventType []byte, dat
 		if output, ok := usage.OutputTokens(); ok {
 			p.tokenUsage.AddOutputTokens(output)
 		}
-		// Update input tokens to include any cache tokens from delta
```
xiaolin593, line 1064: I missed this PR when I was working on https://github.com/envoyproxy/ai-gateway/pull/2241.

After discussing offline with @hustxiayang and @yuzisun, we agreed to remove the extra cache token counting from message_start and instead rely on message_delta, which is the provider's final source of truth in streaming mode.

This keeps the logic cleaner and avoids edge cases like the one here: https://github.com/envoyproxy/ai-gateway/pull/2241/changes#diff-aa8ea9f59fc3acd18cfd4a4c571d683817274a213b6a5fad0ab6085cfe0912dfL974-L980.

hustxiayang:
@yuzisun About whether to set input tokens in the start message or the delta message, the anthropic's definition is like 
```
class MessageDeltaUsage(BaseModel):
    cache_creation_input_tokens: Optional[int] = None  # cumulative
    cache_read_input_tokens: Optional[int] = None      # cumulative
    input_tokens: Optional[int] = None                 # cumulative
    output_tokens: int                                 # required, cumulative
```
and the start message's definition:
```
class Usage(BaseModel):
    cache_creation_input_tokens: Optional[int] = None
    cache_read_input_tokens: Optional[int] = None
    input_tokens: int        # required
    output_tokens: int       # required
```

The difference is that `input_tokens` are `required` in the start message, but optional in the delta message. Thus it should be safer to use start message as it's guanranted to be set, so it should be safer to get the value from the start message.

hustxiayang:
there is no difference when there is a disconnection or 529 error in the middle of the stream, because it's just to set the value into an intermediate structure (`p`). It only reports the usage (usage chunk and metrics) until `endOfStream=true`, which is after both start message and delta message. So there should be no difference.

yuzisun:
> @yuzisun About whether to set input tokens in the start message or the delta message, the anthropic's definition is like
> 
> ```
> class MessageDeltaUsage(BaseModel):
>     cache_creation_input_tokens: Optional[int] = None  # cumulative
>     cache_read_input_tokens: Optional[int] = None      # cumulative
>     input_tokens: Optional[int] = None                 # cumulative
>     output_tokens: int                                 # required, cumulative
> ```
> 
> and the start message's definition:
> 
> ```
> class Usage(BaseModel):
>     cache_creation_input_tokens: Optional[int] = None
>     cache_read_input_tokens: Optional[int] = None
>     input_tokens: int        # required
>     output_tokens: int       # required
> ```
> 
> The difference is that `input_tokens` are `required` in the start message, but optional in the delta message. Thus it should be safer to use start message as it's guanranted to be set, so it should be safer to get the value from the start message.

Then we should get all inputs including cache read/write tokens in the start message since they are set before token streaming.

hustxiayang:
> Then we should get all inputs including cache read/write tokens in the start message since they are set before token streaming.

@yuzisun I think the current implementation is set like this. You could check line 971-979, these values are set in the start message.

internal/translator/anthropic_helper.go
yuzisun: Can you move the comment above, looks odd for comments without code

hustxiayang: Thanks, I updated the comment.

yuzisun (approved):

hustxiayang:
@yuzisun could you please help to merge it as well. Thanks!

PR #2244, opened 2026-06-17
docs: per-tenant rate limits from dynamic metadata, with an e2e

### LINKED ISSUE #2223: Emit per-request rate-limit limits into dynamic metadata
## What this is about

We want each tenant to have a **different rate limit** while all tenants share **one** rate-limit policy — so we don't create per-tenant config objects.

The way that works: the tenant's limit travels **with the request** as Envoy dynamic metadata, and Envoy reads it per request as the rate-limit value. (Envoy Gateway exposing this is [envoyproxy/gateway#9216](https://github.com/envoyproxy/gateway/pull/9216); Envoy honoring it on the cost path is [envoyproxy/envoy#45613](https://github.com/envoyproxy/envoy/pull/45613).)

This issue covers the missing piece in the AI Gateway: **putting the per-tenant limit into that metadata.**

## Who does what

```
1. A trusted ext-authz   resolves tenant → quota, puts it on the request as a header
   (knows the tenant)      e.g.  x-aigw-limit-input: 100000/HOUR

2. AI Gateway (this issue)  copies that header into io.envoy.ai_gateway metadata
                            so Envoy can read it per request

3. Envoy / Envoy Gateway   reads the metadata as the per-tenant rate-limit value
   (#45613 / #9216)
```

The AI Gateway stays **stateless about tenants** — it just copies a header it's given into metadata.

## Proposed change

A new type `RateLimitFromHeader` with a single header field. The header value encodes both the count and the time unit as `<uint64>/<SECOND|MINUTE|HOUR|DAY>`. The ext-authz sets both in one header; the gateway parses and emits the struct Envoy reads.

```go
type RateLimitFromHeader struct {
    MetadataKey string `json:"metadataKey"`
    // Header is the trusted request header whose value is parsed as "<count>/<unit>",
    // e.g. "100000/HOUR". If the header is absent or malformed the key is omitted.
    Header string `json:"header"`
}

// on AIGatewayRouteSpec
RateLimitsFromHeaders []RateLimitFromHeader `json:"rateLimitsFromHeaders,omitempty"`

// on GatewayConfigSpec
GlobalRateLimitsFromHeaders []RateLimitFromHeader `json:"globalRateLimitsFromHeaders,omitempty"`
```

At end-of-stream the gateway parses the header and emits the struct `RateLimit.Override.DynamicMetadata` reads:
```json
{ "requests_per_unit": 100000, "unit": "HOUR" }
```

## Example

**GatewayConfig:**
```yaml
spec:
  globalRateLimitsFromHeaders:
    - metadataKey: llm_input_token_limit
      header: x-aigw-limit-input
    - metadataKey: llm_output_token_limit
      header: x-aigw-limit-output
```

**Incoming request** (ext-authz has resolved tenant → quota and time window):
```
POST /v1/chat/completions
x-aigw-limit-input: 100000/HOUR
x-aigw-limit-output: 50000/DAY
```

**What AI Gateway emits** into `io.envoy.ai_gateway` at end-of-stream (alongside the existing cost):
```json
{
  "llm_input_token_limit":  { "requests_per_unit": 100000, "unit": "HOUR" },
  "llm_output_token_limit": { "requests_per_unit": 50000,  "unit": "DAY"  },
  "llm_token_cost":         1234
}
```

**Envoy rate-limit policy** reads the override from metadata:
```yaml
rateLimits:
  - actions:
      - metadata:
          descriptorKey: tenant
          metadataKey: { path: [{ key: x-tenant-id }] }
    limit:
      dynamicMetadata:
        metadataKey:
          namespace: io.envoy.ai_gateway
          key: llm_input_token_limit
```

If the header is absent or malformed, the key is omitted and Envoy falls back to the static limit in the policy.

## Out of scope

- **Reading the metadata as a limit** lives in Envoy / Envoy Gateway (`RateLimit.Override`), not here — see #45613 and #9216.
- **Resolving tenant → quota and setting the header** is the ext-authz's job, not the gateway's.

## Related

- Cost-from-metadata producer (the existing pattern this mirrors): `buildDynamicMetadata`, `internal/extproc/processor_impl.go`
- Cost config + example: `LLMRequestCosts`, `api/v1beta1/ai_gateway_route.go`
- Namespace constant: `AIGatewayFilterMetadataNamespace = "io.envoy.ai_gateway"`, `api/v1alpha1/shared_types.go`

kanurag94:
**Description**

Give each tenant a different rate limit while every tenant shares one policy and one set of routes. The tenant's limit travels with the request as dynamic metadata, so there are no per-tenant config objects.

This is documentation and an e2e. There is no API change: Envoy Gateway's `limit.fromMetadata` can read the limit from any metadata namespace, so whichever filter knows the tenant can write it and the policy reads it directly. AI Gateway's part is the other half — turning the LLM's token usage into the cost charged against that limit, which already exists as `globalLLMRequestCosts`.

**How it works**

1. `ext_authz` resolves the tenant and returns its budget as dynamic metadata.
2. One `BackendTrafficPolicy` reads that budget with `limit.fromMetadata`, bucketed per tenant with `Distinct`.
3. AI Gateway emits the request's token usage, and the policy charges it against the budget.

**Usage**

The auth server returns the budget in its Check response. Envoy files it under `envoy.filters.http.ext_authz`. The value must be the struct Envoy's rate limit override reads:

```json
{
  "dynamic_metadata": {
    "fields": {
      "total_limit": {
        "struct_value": {
          "fields": {
            "requests_per_unit": { "number_value": 100000 },
            "unit": { "string_value": "HOUR" }
          }
        }
      }
    }
  }
}
```

`unit` is one of `SECOND`, `MINUTE`, `HOUR`, `DAY`, `MONTH`, `YEAR`. Dynamic metadata is used rather than a header because only filters can write it, so a tenant cannot raise its own limit.

`GatewayConfig` declares the cost to charge:

```yaml
kind: GatewayConfig
metadata:
  name: envoy-ai-gateway
  namespace: default
spec:
  globalLLMRequestCosts:
      type: TotalToken
```

One policy then serves every tenant:

```yaml
kind: BackendTrafficPolicy
metadata:
  name: tenant-token-budget
  namespace: default
spec:
  targetRefs:
      kind: Gateway
      name: envoy-ai-gateway
  rateLimit:
    type: Global
    global:
      rules:
        - clientSelectors:
            - headers:
                - name: x-tenant-id
                  type: Distinct
          limit:
            requests: 1000 # default when ext_authz says nothing
            unit: Hour
            fromMetadata:
              namespace: envoy.filters.http.ext_authz # where ext_authz writes
              key: total_limit
          cost:
            request: { from: Number, number: 0 } # charge tokens, not requests
            response:
              from: Metadata
              metadata:
                namespace: io.envoy.ai_gateway
                key: llm_total_token
```

For separate input, output and total budgets, declare one cost key per token kind (`InputToken`, `OutputToken`, `TotalToken`), have the auth server return one budget per kind, and add one rule per kind. A tenant stops at whichever budget it exhausts first.

**Behavior**

Request-time cost is 0 and tokens are charged from the response, so a budget of N against a cost of C allows N/C + 1 calls.

| Metadata value                        | Effect                                             |
| ------------------------------------- | -------------------------------------------------- |
| valid `{ requests_per_unit, unit }`   | becomes the limit for that request                 |
| absent                                | the static `requests`/`unit` on the rule applies   |
| `requests_per_unit: 0`                | a real limit of 0, which suspends the tenant       |

Two things to plan for:

- Missing metadata fails open to the static default, so keep that default a conservative ceiling rather than a placeholder.
- `0` is a real limit and does not fall back. A source with no opinion about a request must omit the field, not send `0` — an uninitialized field or a failed lookup will otherwise suspend the tenant.

**Why no API change**

An earlier revision of this PR added `GatewayConfig.spec.globalRateLimits`, which read the limit from one namespace and re-emitted it into `io.envoy.ai_gateway`. That was written when the source was a request **header** (#2223), which Envoy's override cannot read, so something had to bridge it. Review moved the source to dynamic metadata so clients cannot forge it — and that removed the need for the bridge, because the filter writing the metadata can write the final struct itself.

What remained was reshaping `"100000/HOUR"` into a struct. It cost a `metadata` request attribute on the router ext_proc filter, a change to how every request is classified router vs upstream, and a rolling-upgrade break needing its own workaround. Not worth it for a string reshape on the path of every request.

A source that cannot produce the struct — a `jwt_authn` claim, or a third-party auth sidecar returning a plain string — still needs something to transcode. The docs point at a Lua filter. If that case turns out to be common, an API is worth revisiting on its own merits.

**Testing**

`Test_DynamicMetadataRateLimit` runs the configuration above and asserts each tier is throttled at its own number:

| tier         | budget from ext_authz       | calls allowed |
| ------------ | --------------------------- | ------------- |
| `premium`    | 6 total tokens/HOUR         | 4             |
| `basic`      | 2 total tokens/HOUR         | 2             |
| `unknown`    | none — static `requests: 4` | 3             |
| `suspended`  | 0 total tokens/HOUR         | 1, then 429   |
| `tri-input`  | 6 input, others generous    | 3             |
| `tri-output` | 3 output, others generous   | 4             |
| `tri-total`  | 12 total, others generous   | 4             |

The triplet cases use a lopsided reply (3 in, 1 out, 4 total) so the three budgets produce different counts. With an even reply, input and output would be indistinguishable and a mix-up between them would go unnoticed.

The test skips unless Envoy Gateway has `limit.fromMetadata` (envoyproxy/gateway#9216), which is on main but not yet in a tagged release. There is a TODO on the version check to relax it once a release includes it.

**Related Issues/PRs**

Closes https://github.com/envoyproxy/ai-gateway/issues/2223, with documentation instead of the API that issue proposed — see above.

Consumes envoyproxy/gateway#9216 (`limit.fromMetadata` on `BackendTrafficPolicy`).


zhaohuabing:
Nice feature. One thing worth calling out: since the limit is read from a request header, a client could spoof `x-aigw-limit: 999999999/HOUR` on their own request to raise their own quota. 

So the feature is only safe if any client-supplied copy of the header is stripped before `ext_authz`/`ext_proc` see it. 
We can use `ClientTrafficPolicy.spec.headers.earlyRequestHeaders.remove` to strip this header before the HTTP filter chain processing:

```yaml
kind: ClientTrafficPolicy
spec:
  targetRefs:
      kind: Gateway
      name: my-gateway
  headers:
    earlyRequestHeaders:
      remove: ["x-aigw-limit"]
```

zhaohuabing:
I think the  AIGW could create the CTP directly with the configured rate limit headers?

Or is it possible to use a metadata to pass the per tenant/request ratelimit from ext auth to the ext proc?

kanurag94:
> I think the AIGW could create the CTP directly with the configured rate limit headers?
> 
> Or is it possible to use a metadata to pass the per tenant/request ratelimit from ext auth to the ext proc?

Thanks for the great suggestion. I have updated the API (see updated PR description):
1. It now supports dynamic metadata source today.
2. It can extend to use req headers in future and the controller can create CTP for this. But it is not needed today.

zhaohuabing:
This may be worth adding an e2e test for, since it significantly changes rate-limit behavior and requires multiple components to work together correctly.

An EnvoyPatchPolicy that inserts a [set metadata filter](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/set_metadata_filter) into the HCM filter chain could be used to simplify the e2e test.

zhaohuabing (commented):
If I understand this correctly, ext_proc should also be configured to receive source metadata from the namespace that emits the ratelimit limits. 

The new API reads metadata from configurable namespaces such as `envoy.filters.http.ext_authz`, so this PR would also need to configure  `MetadataOptions.ForwardingNamespaces`. 

https://github.com/envoyproxy/ai-gateway/blob/5a8584a4740e19611e07377162a283c4ff7a7fff/internal/extensionserver/post_translate_modify.go#L648-L651

kanurag94:
> If I understand this correctly, ext_proc should also be configured to receive source metadata from the namespace that emits the ratelimit limits.
> 
> The new API reads metadata from configurable namespaces such as `envoy.filters.http.ext_authz`, so this PR would also need to configure `MetadataOptions.ForwardingNamespaces`.
> 
> https://github.com/envoyproxy/ai-gateway/blob/5a8584a4740e19611e07377162a283c4ff7a7fff/internal/extensionserver/post_translate_modify.go#L648-L651

Thanks for this great analysis. The code now collects namespaces from GatewayConfigs and injects them in forwardingNamespaces.

kanurag94:
@gemini review

zhaohuabing:
Codex reported this, could you check if it's true?

[P1] Dynamic limits do not reliably activate on GatewayConfig updates. The GatewayConfig controller only notifies the AI Gateway controller to rebuild the ext_proc filter config secret ([gateway_config.go (line 80)](/Users/huabingzhao/workspace/ai-gateway/internal/controller/gateway_config.go:80), [gateway.go (line 142)](/Users/huabingzhao/workspace/ai-gateway/internal/controller/gateway.go:142)). But the namespaces Envoy must forward into ProcessingRequest.metadata_context are patched in PostTranslateModify, and the code explicitly notes Envoy Gateway does not retranslate on GatewayConfig changes ([post_translate_modify.go (line 634)](/Users/huabingzhao/workspace/ai-gateway/internal/extensionserver/post_translate_modify.go:634)). So adding globalRateLimits, or changing source.fromMetadata.namespace, on an existing Gateway updates the sidecar config but leaves ForwardingNamespaces stale or unset ([post_translate_modify.go (line 696)](/Users/huabingzhao/workspace/ai-gateway/internal/extensionserver/post_translate_modify.go:696)). In that state ext_proc cannot see the source metadata, emits no override, and the static BackendTrafficPolicy limit applies until some unrelated Gateway/Route change forces EG translation. This should be fixed by making referenced GatewayConfig changes trigger EG retranslation, or by avoiding a forwarding namespace list that depends on a non-watched resource.

kanurag94:
> Codex reported this, could you check if it's true?
> 
> [P1] Dynamic limits do not reliably activate on GatewayConfig updates. The GatewayConfig controller only notifies the AI Gateway controller to rebuild the ext_proc filter config secret ([gateway_config.go (line 80)](/Users/huabingzhao/workspace/ai-gateway/internal/controller/gateway_config.go:80), [gateway.go (line 142)](/Users/huabingzhao/workspace/ai-gateway/internal/controller/gateway.go:142)). But the namespaces Envoy must forward into ProcessingRequest.metadata_context are patched in PostTranslateModify, and the code explicitly notes Envoy Gateway does not retranslate on GatewayConfig changes ([post_translate_modify.go (line 634)](/Users/huabingzhao/workspace/ai-gateway/internal/extensionserver/post_translate_modify.go:634)). So adding globalRateLimits, or changing source.fromMetadata.namespace, on an existing Gateway updates the sidecar config but leaves ForwardingNamespaces stale or unset ([post_translate_modify.go (line 696)](/Users/huabingzhao/workspace/ai-gateway/internal/extensionserver/post_translate_modify.go:696)). In that state ext_proc cannot see the source metadata, emits no override, and the static BackendTrafficPolicy limit applies until some unrelated Gateway/Route change forces EG translation. This should be fixed by making referenced GatewayConfig changes trigger EG retranslation, or by avoiding a forwarding namespace list that depends on a non-watched resource.

This is right and currently a comment is put out for this. Adding forwardingNamespaces will mean we start watching GatewayConfig resource. I left it for a followup.
Pushed a commit to fix it.

zhaohuabing (approved):
LGTM. Deferring to @envoyproxy/ai-gateway-maintainers .

nacx (approved):

Issue #2076, opened 2026-04-22
Feature Request: Support per-request AWS credentials via headers for per-user cost attribution


sc7565:
## Summary

**What:** Allow AI Gateway's AWS SigV4 signer to use per-request AWS credentials from headers instead of only using the pod's IRSA credentials.

**Why:** Enable per-user cost attribution in AWS billing by allowing an ext-auth service to inject STS temporary credentials (from `AssumeRole` with session tags) into request headers.

**Where:** Modify `internal/backendauth/aws.go` to check for `x-aws-access-key-id`, `x-aws-secret-access-key`, and `x-aws-session-token` headers before falling back to IRSA.

**Impact:** 
- ✅ Enables AWS Scenario 4 from the [Bedrock cost attribution blog](https://aws.amazon.com/blogs/machine-learning/introducing-granular-cost-attribution-for-amazon-bedrock/)
- ✅ Backward compatible (IRSA remains the default)
- ✅ Unlocks per-user/per-team cost tracking in AWS Cost Explorer and CUR 2.0

### Use Case

We're implementing **AWS Scenario 4: Per-user tracking through an LLM gateway** from the [AWS Bedrock granular cost attribution blog](https://aws.amazon.com/blogs/machine-learning/introducing-granular-cost-attribution-for-amazon-bedrock/).

Our ext-auth service:
1. Calls `STS AssumeRole` with per-user session names and session tags
2. Caches the temporary credentials in Redis (50-min TTL)
3. Injects credentials via headers: `x-aws-access-key-id`, `x-aws-secret-access-key`, `x-aws-session-token`
4. Forwards these headers to the AI Gateway extproc

### Current Behavior

The `ai-gateway-extproc` container ignores these headers and uses the pod's IRSA (IAM Roles for Service Accounts) credentials configured via environment variables:

```yaml
env:
  - name: AWS_ROLE_ARN
    value: arn:aws:iam::123456789012:role/ai-gateway-role
  - name: AWS_WEB_IDENTITY_TOKEN_FILE
    value: /var/run/secrets/eks.amazonaws.com/serviceaccount/token
```

**Evidence from CloudTrail:**
All Bedrock `InvokeModel` calls appear as:
```
arn:aws:sts::123456789012:assumed-role/ai-gateway-role/1776782298841908997
```

(Pod's default role with random numeric session name)

### Expected Behavior

The `ai-gateway-extproc` should:

1. **Check for per-request AWS credential headers** on each request:
[...]

PR #2253, opened 2026-06-18
feat: per-request backend credentials from request headers and dynamic metadata

### LINKED ISSUE #2216: Feature Request: Per-request backend credentials from request headers
## Background

A `BackendSecurityPolicy` holds one static credential, loaded at startup and used for every request through the backend. A request can't carry its own upstream credential.

#2076 asks for this for AWS (so Bedrock usage can be billed per caller). This issue generalizes it to the single-secret auth types: `APIKey`, `AnthropicAPIKey`, `AzureAPIKey`, `AzureCredentials`, `GCPCredentials`.

AWS is deferred. It needs three header values plus SigV4 signing, not one credential value, so it doesn't fit the shape here. Separate follow-up.

## Why

A trusted ext-authz/ext_proc resolves the caller's credential, and the gateway authenticates upstream with it instead of using one shared static credential for everyone. That gives per-caller cost attribution and auditing at the provider, and lets many callers share a single `BackendSecurityPolicy` rather than needing one policy per credential.

## Current behavior

`BackendSecurityPolicySpec` picks one mechanism via `Type` plus the matching field. The controller projects it into `filterapi.BackendAuth`, and each handler in `internal/backendauth/` injects its one static value:

| Handler | Sets |
| --- | --- |
| `apiKeyHandler` | `Authorization: Bearer <key>` |
| `anthropicAPIKeyHandler` | `x-api-key: <key>` |
| `azureAPIKeyHandler` | `api-key: <key>` |
| `azureHandler` | `Authorization: Bearer <token>` |
| `gcpHandler` | `Authorization: Bearer <token>` (+ `:path` rewrite) |

The returned header is applied with `OVERWRITE_IF_EXISTS_OR_ADD` in `processor_impl.go`, so it overwrites whatever the client sent in that header.

Two facts shape the design:

- Handlers already receive the full request-header map, so seeing an injected header needs no new plumbing.
- The only sensitive-header list today, `sensitiveHeaderKeys = {"authorization", "x-api-key"}` in `internal/extproc/server.go`, is for **log redaction only**. Nothing strips credentials before the upstream call, so stripping the new input header is new work.

## Proposed change

Add an optional `credentialOverride` to `BackendSecurityPolicySpec`. When set, the credential comes from the request, via one of two sources:

- **`fromRequestHeaders`** — a request header. Simple, but client-spoofable.
- **`fromDynamicMetadata`** — Envoy dynamic metadata set by a trusted filter. Can't be forged by the client; preferred where such a filter already runs.

`fallbackToConfigured` (default `true`) decides what happens when the source is empty: fall back to the static credential, or reject with 401.

```go
// proposed additions, api/v1alpha1

type BackendSecurityPolicySpec struct {
	// ... existing fields ...

	// CredentialOverride sources the per-request credential from the request instead of
	// the static credential above. nil disables it (the default).
	// +optional
	CredentialOverride *BackendSecurityPolicyCredentialOverride `json:"credentialOverride,omitempty"`
}

// +kubebuilder:validation:MaxProperties=1
type BackendSecurityPolicyCredentialOverride struct {
	// +optional
	FromRequestHeaders *CredentialOverrideFromRequestHeaders `json:"fromRequestHeaders,omitempty"`
	// +optional
	FromDynamicMetadata *CredentialOverrideFromDynamicMetadata `json:"fromDynamicMetadata,omitempty"`
}

type CredentialOverrideFromRequestHeaders struct {
	// Header overrides the default header name (see table). +optional
	Header string `json:"header,omitempty"`
	// FallbackToConfigured: header absent -> true (default) uses the static credential,
	// false returns 401.
	// +kubebuilder:default=true
	// +optional
	FallbackToConfigured *bool `json:"fallbackToConfigured,omitempty"`
}

type CredentialOverrideFromDynamicMetadata struct {
	// Namespace populated by the trusted filter.
	Namespace string `json:"namespace"`
	// Key overrides the default metadata key name. +optional
	Key string `json:"key,omitempty"`
	// +kubebuilder:default=true
	// +optional
	FallbackToConfigured *bool `json:"fallbackToConfigured,omitempty"`
}
```

The controller resolves the default name and projects a `CredentialOverride` onto `filterapi.BackendAuth`.

Default header names use an `x-aigw-` prefix so they can't collide with the real provider header:

| Type | Default header | Value goes to |
| --- | --- | --- |
| `APIKey` | `x-aigw-api-key` | `Authorization: Bearer` |
| `AnthropicAPIKey` | `x-aigw-anthropic-api-key` | `x-api-key` |
| `AzureAPIKey` | `x-aigw-azure-api-key` | `api-key` |
| `AzureCredentials` | `x-aigw-azure-access-token` | `Authorization: Bearer` |
| `GCPCredentials` | `x-aigw-gcp-access-token` | `Authorization: Bearer` |

Header source (defaults):

```yaml
spec:
  type: APIKey
  apiKey:
    secretRef:
      name: fallback-openai-key   # only used when the header is absent
  credentialOverride:
    fromRequestHeaders: {}        # x-aigw-api-key, fall back to configured
# A trusted ingress filter sets x-aigw-api-key and strips any client-supplied copy.
```

Dynamic metadata source:

```yaml
spec:
  type: APIKey
  apiKey:
    secretRef:
      name: fallback-openai-key
  credentialOverride:
    fromDynamicMetadata:
      namespace: envoy.filters.http.ext_authz
      key: upstream_api_key
      fallbackToConfigured: false  # no metadata -> 401, never the shared fallback
```

## Implementation notes

- **CRD + filterapi:** add `credentialOverride`; controller fills the default name and projects a resolved struct onto `filterapi.BackendAuth`.
- **Handlers:** a small helper reads the configured source; each handler uses the per-request value if present, else the static credential, else 401 per `fallbackToConfigured`. Output header still set with `OVERWRITE_IF_EXISTS_OR_ADD`.
- **Strip the input (header source only):** controller adds the `x-aigw-*` header to the request remove list (`filterapi.HTTPHeaderMutation.Remove`) so Envoy drops it before upstream, and to `sensitiveHeaderKeys` for log redaction. Metadata needs neither.

## Performance

For `GCPCredentials`/`AzureCredentials` the per-request value is an already-exchanged short-lived token; the gateway doesn't mint or exchange (out of scope). The filter producing it does a token exchange per request unless it caches, adding latency. Caching (keyed on the caller, TTL under the token lifetime) is that filter's job.

## Security

The sources differ in whether the client can forge the value:

- **Header:** anyone can set a request header, so this is only correct if a trusted filter strips client-supplied copies at ingress and sets the real one. That invariant lives in a separate `SecurityPolicy`/ext-authz the `BackendSecurityPolicy` can't enforce, which is why it's fragile.
- **Dynamic metadata:** set by filters, not part of the client request, so it can't be forged. Prefer it where a trusted filter already runs.

Also:

- Off unless `credentialOverride` is set on the policy. No global switch.
- Partial or absent value with `fallbackToConfigured: false` returns 401, never a silent fallback to the shared identity (which would mis-attribute cost and hide a broken ingress filter).
- The `x-aigw-*` input is stripped before upstream, never echoed, and redacted in logs.

## Failure modes

| Situation | Behavior |
| --- | --- |
| Source present | Per-request credential |
| Source absent, `fallbackToConfigured: true` | Configured credential |
| Source absent, `fallbackToConfigured: false` | 401 |
| Provider rejects the credential | Provider's auth error returned to the client |
| `credentialOverride` not set | Unchanged |

## Out of scope

- `AWSCredentials` (three headers + SigV4) — separate follow-up.
- Minting/exchanging/caching credentials in the gateway — the upstream filter's job. MCP token exchange is covered by proposal 010.
- Client-to-gateway auth — unchanged.

## Related

Generalizes #2076 across the single-secret auth types. See also #185, #232, #161.

kanurag94:
**Description**

BackendSecurityPolicy today uses one static credential for every request. This adds an optional credentialOverride field so a trusted upstream filter can supply the credential per-request instead.

Two sources: fromRequestHeaders reads from a request header the filter injects (stripped before forwarding upstream), fromDynamicMetadata reads from Envoy filter metadata which the client cannot forge. fallbackToConfigured decides whether a missing source falls back to the static credential or returns 401.

Default header names use the x-aigw- prefix (e.g. x-aigw-api-key) so they don't collide with real provider headers and are redacted from logs automatically. AWSCredentials is not supported here — it needs three values plus SigV4 signing, that's a separate follow-up.

Example with header source, falls back to the secret when header is absent:

```yaml
spec:
  type: APIKey
  apiKey:
    secretRef:
      name: openai-key
  credentialOverride:
    fromRequestHeaders:
      fallbackToConfigured: true
```

Example with dynamic metadata, 401 when key is missing:

```yaml
  credentialOverride:
    fromDynamicMetadata:
      namespace: envoy.filters.http.ext_authz
      key: upstream_api_key
      fallbackToConfigured: false
```

Closes #2216
Related: #2076

The fromRequestHeaders source strips the input header from the request before it reaches the upstream backend. This is done by adding it to the backend's HeaderMutation.Remove list, which tells Envoy to drop it while keeping it visible in the local header map so the handler can still read it.


nacx (approved):
Thanks!
Overall LGTM. Cahn you tke a look at the CI job failures?

kanurag94:
Thanks for the review @nacx. I have fixed the CI

PR #2257, opened 2026-06-18
mcp: share one Backend per namespace instead of one per MCPRoute

### LINKED ISSUE #2150: MCPRoute controller mints one redundant mcp-proxy Backend per route — consider sharing one
## Problem

The `MCPRoute` controller creates one Envoy Gateway `Backend` per `MCPRoute` (`internal/controller/mcp_route.go`, `ensureMCPProxyBackend`):

- name: `{namespace}-{name}-mcp-proxy`
- endpoint: a fixed dummy IP `192.0.2.42` (`mcpProxyBackendDummyIP`)
- owned by the `MCPRoute` for garbage collection

These Backends only exist so Envoy Gateway emits a CDS cluster. The extension server then rewrites every one of them in `modifyMCPGatewayGeneratedCluster` (`internal/extensionserver/mcproute.go`), replacing the whole cluster struct with an identical static cluster pointing at the in-pod MCP proxy (`127.0.0.1:9856`) — keeping only the cluster name. The dummy IP never survives translation.

So all per-route Backends and their clusters are identical apart from name, and all stand in for the same single MCP proxy.

## Why it matters

Envoy Gateway keys one CDS cluster per (HTTPRoute, rule), so N `MCPRoute`s produce N identical placeholder `Backend` CRs and N identical Envoy clusters. On a deployment with 86 `MCPRoute`s: 86 `*-mcp-proxy` Backends, all endpoint `192.0.2.42`, plus 86 CDS clusters each with its own connection-pool / stats / config state. Linear control-plane and dataplane overhead for what is logically one upstream, growing with route count.

## Proposal

Have the `MCPRoute` controller reference a single shared MCP-proxy `Backend`, scoped per Gateway.

The only thing the per-route Backend provides today is GC via the controller-owner reference. A shared Backend would instead be owned by the Gateway, so it is still garbage-collected (when the Gateway is deleted) without being tied to any single `MCPRoute`.

Happy to contribute a PR if the direction is agreeable.

## Environment

Verified against current `main` (`7a303f93`): `mcpProxyBackendName`, `mcpProxyBackendDummyIP`, `ensureMCPProxyBackend`, `modifyMCPGatewayGeneratedCluster`.

kanurag94:
**Description**

The MCPRoute controller created one `egv1a1.Backend` per MCPRoute (`{ns}-{name}-mcp-proxy`),
all pointing at the same dummy IP that the extension server rewrites to `127.0.0.1:9856`. That
is N identical placeholder Backend CRs per namespace.

Switch to a single shared Backend per namespace (`ai-eg-mcp-proxy`) that every MCPRoute's main
HTTPRoute references. It lives in the MCPRoute's own namespace, so the reference is valid for any
number of parent Gateways (including cross-namespace ones). It has no owner reference and is
deleted when the last MCPRoute in the namespace is removed, via the finalizer.

This only collapses the redundant Backend CRs; it does not change the number of CDS clusters
(Envoy Gateway emits one per HTTPRoute rule, not per Backend). No extension server change is
needed - it matches clusters by the main HTTPRoute name prefix, not the Backend name.

Closes #2150


nacx (approved):
Thanks!

internal/controller/mcp_route.go
```diff
@@ -545,22 +566,89 @@ func (c *MCPRouteController) ensureMCPProxyBackend(ctx context.Context, mcpRoute
...
-
-	return nil
 }
 
-func mcpProxyBackendName(mcpRoute *aigv1b1.MCPRoute) string {
-	return fmt.Sprintf("%s-%s-mcp-proxy", mcpRoute.Namespace, mcpRoute.Name)
+// deleteOldMCPRouteBackend deletes the legacy per-MCPRoute Backend if it still exists and is
+// owned by this MCPRoute. This is a one-time migration cleanup from the old naming scheme
+// ({ns}-{routeName}-mcp-proxy) to the shared per-namespace Backend (ai-eg-mcp-proxy).
+func (c *MCPRouteController) deleteOldMCPRouteBackend(ctx context.Context, mcpRoute *aigv1b1.MCPRoute) {
```
nacx, line 634: Can we mark the method `Deprecated` or similar, so it is flagged and we remember to remove it in future versions?

PR #2260, opened 2026-06-18
fix: tolerate truncated gzip at end-of-stream in streaming decode

### LINKED ISSUE #1894: ext_proc fails to parse GCP Anthropic streaming responses (streamRawPredict unexpected EOF, 0 bytes to client)
## Summary

The AI Gateway ext_proc fails to process streaming responses from Vertex AI's GCP Anthropic `:streamRawPredict` endpoint. The ext_proc logs `failed to transform response: failed to read body: unexpected EOF` and sends **0 bytes** back to the client despite receiving a complete 200 response (~40-80KB) from upstream. The client receives an empty body and hangs.

Non-streaming requests to `:rawPredict` work correctly.

## Environment

| Component | Version |
|---|---|
| AI Gateway ext_proc | `docker.io/envoyproxy/ai-gateway-extproc:latest` |
| AI Gateway Helm chart | `v0.0.0-latest` |
| Envoy | `docker.io/envoyproxy/envoy:distroless-v1.37.0` |
| Envoy Gateway | `docker.io/envoyproxy/gateway:v1.7.0` |
| Backend | Vertex AI GCP Anthropic (`claude-sonnet-4-6` via `:streamRawPredict`) |
| Platform | GKE |

## Steps to Reproduce

1. Deploy AI Gateway with Helm chart (`v0.0.0-latest`) and a Vertex AI GCP Anthropic backend
2. Configure an AIGatewayRoute targeting a Claude model (e.g. `claude-sonnet-4-6`)
3. Send an Anthropic `/v1/messages` or OpenAI `/v1/chat/completions` request (streaming is the default)
4. AI Gateway routes to Vertex AI's `:streamRawPredict` endpoint
5. Vertex AI returns HTTP 200 with an SSE streaming response body
6. ext_proc fails to read the buffered response body — client gets 0 bytes

### AI Gateway ext_proc logs

Every streaming response produces this error:

```
level=ERROR msg="error processing request message"
  error="cannot process response body: failed to transform response:
  failed to read body: unexpected EOF"
```

### Envoy access logs

Streaming responses (`streamRawPredict`) consistently show `bytes_sent: 0` despite `bytes_received` being populated:

```json
{"response_code":200, "bytes_received":77492, "bytes_sent":0,
 "path":"...claude-sonnet-4-6:streamRawPredict", "duration":825}

{"response_code":200, "bytes_received":77858, "bytes_sent":0,
 "path":"...claude-sonnet-4-6:streamRawPredict", "duration":1510}

{"response_code":200, "bytes_received":44469, "bytes_sent":0,
 "path":"...claude-sonnet-4-6:streamRawPredict", "duration":1047}

{"response_code":200, "bytes_received":46577, "bytes_sent":0,
 "path":"...claude-sonnet-4-6:streamRawPredict", "duration":802}
```

Non-streaming (`rawPredict`) works correctly — bytes are forwarded to the client:

```json
{"response_code":200, "bytes_received":77844, "bytes_sent":1122,
 "path":"...claude-sonnet-4-6:rawPredict", "duration":3984}
```

### Envoy warning log

Envoy logs a corresponding gRPC error for each failed streaming response:

```
[warning][ext_proc] Received gRPC error on stream: 2,
  message error processing request message: cannot process response body:
  failed to transform response: failed to read body: unexpected EOF{via_upstream}
```

## Expected Behavior

The ext_proc should successfully parse the buffered SSE response from Vertex AI's `:streamRawPredict` endpoint and forward the translated response body to the client.

## Impact

- **100% failure rate** on streaming responses to GCP Anthropic models via Vertex AI
- Clients (e.g. Claude Code CLI) that default to streaming receive empty responses and hang
- Clients that support non-streaming fallback (`:rawPredict`) eventually recover, but with significant added latency from the failed streaming attempt + retry

## Filter Chain (minimal reproduction)

Reproduced with the default AI Gateway filter chain — no custom filters, sidecars, or patches:

```
[0] envoy.filters.http.jwt_authn
[1] envoy.filters.http.ext_proc/aigateway  (ai-gateway-extproc-uds)
[2] envoy.filters.http.rbac
[3] envoy.filters.http.header_to_metadata
[4] envoy.filters.http.router
```

The ext_proc `processing_mode` has `response_body_mode: BUFFERED`.

## What We Ruled Out

- **Client-side beta headers**: Tested with `anthropic-beta` headers removed — streaming still fails
- **Network/upstream issues**: Upstream returns 200 with 40-80KB of data (`bytes_received` is populated). Data arrives at Envoy but ext_proc cannot parse it

## Related Issues

- #1370 — Fixed missing `"stream": true` in request body to GCP Anthropic, which enabled Vertex AI to return SSE streams. This bug is the response-side counterpart: the ext_proc cannot parse the SSE response it now correctly receives.
- #527 — Full duplex ext_proc mode. The current `BUFFERED` response body mode may not handle SSE response framing correctly.

ChrisJBurns:
**Description**

For gzip-encoded streaming responses, `decodeStreamingContent` treats a missing gzip footer on the final chunk as fatal. When an upstream connection resets mid-stream (e.g. an ALB idle timeout during a long streaming response), the footer never arrives, so the entire response fails and the client receives an empty `200` instead of the data that did decode.

This relaxes the end-of-stream guard so a truncated stream (`io.ErrUnexpectedEOF`) returns the partial data instead of erroring. Genuine corruption still fails: a bad CRC or malformed header surfaces as `gzip.ErrChecksum` / `gzip.ErrHeader` rather than `ErrUnexpectedEOF`, so those remain fatal. The `endOfStream` parameter was only used by this guard, so it is removed.

Fixes #1894

Draft — pending a design decision. This change makes the gateway deliver a partial response on an upstream mid-stream reset rather than failing the request. There is a legitimate semantic question here: "is silent partial delivery acceptable, or would you prefer it gated (e.g. behind a flag) and/or made observable (log / metric / response header) so that truncation is visible to operators and clients?" I'm keeping this as a draft until that direction is settled, and I'm happy to adjust the approach based on the answer.

Note that this only affects *truncation* — genuine corruption (`gzip.ErrChecksum` / `gzip.ErrHeader`) remains fatal — so it does not mask corrupt responses.

Testing:

- `TestDecodeStreamingContent_TruncatedGzipAtEndOfStream` — a footer-less (reset) stream returns its partial data without error.
- `TestDecodeStreamingContent_CorruptGzipStillFails` — corrupt input still errors.


ChrisJBurns:
@yanavlasov Hey bud, @nacx pointed me your way to get insights on this PR in related to how upstream filters behave.

It's a draft at the moment, I wanted to get approval on the solution outcome before I put it up as ready for review and was wondering what your thoughts were?

baiyutang:
Thanks for picking this up — the truncated-vs-corrupt distinction here is exactly the right fix for the ALB idle-timeout case.

One thing I'd like to double check: is io.ErrUnexpectedEOF guaranteed to only surface when the gzip footer (CRC32+ISIZE) is missing, or can it also occur when the compressed body itself is cut off mid-deflate-block (e.g. the connection resets in the middle of the payload, not just before the footer)? If the latter can happen, allDecompressed could contain a truncated deflate block that decoded successfully up to that point but doesn't represent a complete, valid prefix of the original message (e.g. a half-written SSE event or a JSON object cut mid-key). That's still much better than today's empty-200 behavior, but it might be worth a code comment (or a test case with a body truncated mid-block rather than just at the footer boundary) so future readers know this is accepted as "best-effort partial data" rather than "guaranteed valid prefix."

Happy to help test this against a real truncated-mid-body case if useful.

ChrisJBurns:
@baiyutang If you could that would be a great help!

PR #2266, opened 2026-06-21
fix: lua filters from EnvoyExtensionPolicy can now run after ext_proc on the response path

### LINKED ISSUE #2247: Lua filter order prevents running after ext_proc on responses
_Description_:

Since https://github.com/envoyproxy/ai-gateway/pull/2054 Lua filters are always ran before ext_proc on responses. Any content mutation done in the filter is lost.

_Repro steps_:

In our case we want to do a simple string concatenation on the output of the model to prepend and append some constant values.

Define a ConfigMap with the script and assign it to the `HTTPRoute` generated for the `AIGatewayRoute`

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: models-wrap
  namespace: my-namespace
data:
  lua: |
    function envoy_on_response(response_handle)
      if response_handle:headers():get(":status") ~= "200" then return end
      -- Never rewrite an encoded body.
      local enc = response_handle:headers():get("content-encoding")
      if enc ~= nil and enc ~= "" and enc ~= "identity" then return end
      local b = response_handle:body()
      if b == nil then return end
      local s = b:getBytes(0, b:length())
      -- Here we expect the model output to be just an array of non-negative integers
      local new, n = s:gsub('("content"%s*:%s*)"(%s*%[[%d%s,]*%])"', function(key, arr)
        return key .. '"{\\"PREFIX\\":' .. arr .. '}"'
      end)
      if n > 0 then
        b:setBytes(new)
        response_handle:headers():replace("content-length", tostring(#new))
      end
    end
---
kind: EnvoyExtensionPolicy
metadata:
  name: models-wrap
  namespace: my-namespace
spec:
  targetRefs:
      kind: HTTPRoute
      name: models
  lua:
    - type: ValueRef
      valueRef:
        group: ""
        kind: ConfigMap
        name: models-wrap
```

_Environment_:

Gateway 0.7.0 - Envoy 1.8.0

_Logs_:

N/A we see the output is the model output unchanged, Envoy filterOrder shows the Lua is applied before the Gateway's `ext_proc`.

yuehaii:
**Description**
lua filters are always ran before ext_proc on responses. any content mutation done in the filter is lost. lua filter order prevents running after ext_proc on responses.

Fixes #2247

introduces a new annotation on EnvoyExtensionPolicy that lets users declare the desired filter ordering relative to the ext_proc. during XDS translation (PostTranslateModify), the extension server reads the annotation and repositions the named filters in each listener's HCM filter chain accordingly.

the new annotation "aigateway.envoyproxy.io/filter-order" value is a comma-separated ordered sequence of filter type tokens, with the special token ExtProc acting as the pivot representing the ext_proc filter.
  - tokens listed before ExtProc: filters placed before ext_proc on the request path (run after ext_proc on responses)
  - tokens listed after ExtProc: filters placed after ext_proc on the request path (run before ext_proc on responses)
  - Filters not listed: keep their default position, after ext_proc on request path.

if "aigateway.envoyproxy.io/default-filter-order" is set as the value of DefaultFilterOrderAnnotation. it will places all of the policy's extension filters before the ext_proc.

the k8s List call for EnvoyExtensionPolicy is served from the controller-runtime in-memory cache. it does not hit the k8s api server on every xDS translation.


yuehaii:
hi @nacx , good day. could you please help review this pull request when it is convenient? thanks.

nacx:
Thanks for raising.

I would like to generalize the problem statement a bit and put more thought into this. Today, the filter order is hardcoded in the code and cannot be modified, and that has been a challenge already for several users.

This PR tackles specifically hte problem for Lua, but I think we can generalize the solution to apply to the rest of the filters that can be configured with an `EnvoyExtensionPolicy, as Wasm or dynamic module users may be facing the same problems.

Several things to consider and discuss:

* We need to make sure the lists of EnvoyExtensionPolicies are cached, as we don't wanna go to the k8s api server every time we need to resolve XDS configs to send to the Envoy pods.
* We want to centralize the ordering logic in a single place. Most of the logic lives today in the `insertAIGatewayExtProcFilter` method, and having ordering logic spread across different places will make it very hard to maintain and reason about.
* When generalizing this to other types of extensions (to provide a consistent API experience to users with semantics shared across), we also need to reason about ext-proc extensions.

yuehaii:
> Thanks for raising.
> 
> I would like to generalize the problem statement a bit and put more thought into this. Today, the filter order is hardcoded in the code and cannot be modified, and that has been a challenge already for several users.
> 
> This PR tackles specifically hte problem for Lua, but I think we can generalize the solution to apply to the rest of the filters that can be configured with an `EnvoyExtensionPolicy, as Wasm or dynamic module users may be facing the same problems.
> 
> Several things to consider and discuss:
> 
> * We need to make sure the lists of EnvoyExtensionPolicies are cached, as we don't wanna go to the k8s api server every time we need to resolve XDS configs to send to the Envoy pods.
> * We want to centralize the ordering logic in a single place. Most of the logic lives today in the `insertAIGatewayExtProcFilter` method, and having ordering logic spread across different places will make it very hard to maintain and reason about.
> * When generalizing this to other types of extensions (to provide a consistent API experience to users with semantics shared across), we also need to reason about ext-proc extensions.

hi @nacx, good day. can you please review the pull request again? I introduced a new annotation on EnvoyExtensionPolicy that lets users declare the desired filter ordering relative to the ext_proc. 
during XDS translation (PostTranslateModify), the extension server reads the annotation and repositions the named filters in each listener's HCM filter chain accordingly.

the new annotation "aigateway.envoyproxy.io/filter-order" value is a comma-separated ordered sequence of filter type tokens, with the special token ExtProc acting as the pivot representing the ext_proc filter.
  - tokens listed before ExtProc: filters placed before ext_proc on the request path (run after ext_proc on responses)
  - tokens listed after ExtProc: filters placed after ext_proc on the request path (run before ext_proc on responses)
  - Filters not listed: keep their default position, after ext_proc on request path.

if "aigateway.envoyproxy.io/default-filter-order" is set as the value of DefaultFilterOrderAnnotation. it will places all of the policy's extension filters before the ext_proc.

the k8s List call for EnvoyExtensionPolicy is served from the controller-runtime in-memory cache. it does not hit the k8s api server on every xDS translation.

PR #2268, opened 2026-06-22
feat: support Envoy Gateway MergeBackends clusters

kanurag94:
**Description**

Envoy Gateway 1.9 adds `mergeBackends`: routes referencing the same backend share one cluster, named `<kind>/<ns>/<name>/<port>/<protocol>` instead of `httproute/<ns>/<name>/rule/<i>`. AI Gateway resolved a request's backend from the cluster name and per-endpoint metadata. Neither survives that.

- **Recognising one** — read the `envoy-gateway` cluster metadata, not the name; the format is EG's to change and moves again under `XDSNameSchemeV2`. Only kind `Backend` counts, which keeps the feature dormant when `mergeBackends` is off.
- **Carrying the backend name** — the route holds a `merged_backend_names` mapping, selected at request time by `xds.cluster_name`. Flat `cluster=backend;...` string, because Envoy delivers struct-valued metadata to ext_proc as the literal `CelMap value`. Matched to backendRefs by backend identity, not position: EG groups weighted entries by kind.
- **Failing closed** — a merged cluster an AIGatewayRoute reaches always gets the upstream filters, so an unresolved backend fails in the processor rather than reaching the provider with no credentials. A plain HTTPRoute sharing the cluster carries none of that state and is passed through untouched.

Quota and MCP proxy lookups move onto route names, which `mergeBackends` does not rename.

**Cross-namespace quota bypass, also fixed**

Not gated behind `mergeBackends`, so it affects every install. A QuotaPolicy's `targetRefs` are local, so a quota belongs to its backend's namespace — but the cost expressions, the request-time descriptor and the policy lookup were keyed on the *route's* namespace, so a route elsewhere was never counted against the backend's quota. All three now key on `<backendNamespace>/<name>`, matching the rate limit service.

Adds `filterapi.Backend.AIServiceBackendName`; when absent the data plane falls back to the old derivation, so a rolling upgrade degrades to today's behaviour in both directions. **Behaviour change:** a cross-namespace route that was silently unlimited is now charged.

**Known limitations**

- `backendRef.Priority` is not honoured — priority is endpoint priority within one cluster, and a merged rule is weighted clusters. Logged rather than silently promoting a standby.
- Two distinct AIServiceBackends on one `Backend`, or one `Backend` on two ports/protocols, are indistinguishable from cluster metadata. Left unmapped, fail closed.
- The forward proxy applies only when every route on the cluster resolves the same address (`http_11_proxy` is cluster-level). Disagreement, or an unresolvable route, leaves it unproxied.
- In `aigw run` there is no annotation to read, so a route is claimed only if it resolves to an AIGatewayRoute — claiming a plain HTTPRoute would stamp our route name on it and defeat the pass-through above.

**Related**

Depends on #2507 (`mergeBackends` does not exist in EG 1.8). #2506 generalises the per-route mapping and should replace it once it lands.

**Notes for reviewers**

Check first that nothing changes when `mergeBackends` is off — `TestApplyMergedBackendRouting/no_merged_clusters_is_a_no-op` pins it, and the kind `Backend` restriction is what makes it true in a real deployment.

Endpoint metadata is un-shared before stamping (`privatizeEndpointMetadata`): EG assigns one `Metadata` pointer to every endpoint in a TLS locality, so a per-endpoint value such as the SigV4 signing host would otherwise land on all of them.

Coverage: metadata classification; the routing matrix (weighted, mixed, repeated-ref, weight-0, stale rule index, plain HTTPRoute); endpoint labelling and signing host; fail-closed classification; the quota fix at all three links; and `TestWithTestUpstream/mergeBackends_...` against real Envoy for both single-cluster and weighted-cluster route actions.

Rebuilt on current main rather than rebased: the previous revision targeted a pre-release `mergeBackends` whose cluster name format and config shape both changed before release.


siddharth1036:
Hey @kanurag94, quick check, are you still working on this? Could you share an estimated timeline for when it'll be ready? We're aiming to upgrade to Envoy 1.9.0 ahead of the next ai-gateway release.

kanurag94:
> Hey @kanurag94, quick check, are you still working on this? Could you share an estimated timeline for when it'll be ready? We're aiming to upgrade to Envoy 1.9.0 ahead of the next ai-gateway release.

This is mostly ready. I will complete it once https://github.com/envoyproxy/ai-gateway/pull/2507 is merged.

PR #2280, opened 2026-06-23
fix: allow x-ai-eg-model in headerMutation

### LINKED ISSUE #1872: Header value of `x-ai-eg-model` is not changed when headerMutation is set

immanuwell:
Description

Right now `headerMutation.set` pretty much drops `x-ai-eg-model`, so the backend still sees the original model header.
This lets `x-ai-eg-model` pass through `shouldIgnoreHeader` and keeps the rest of `x-ai-eg-*` blocked.

Repro
1. Apply the route from #1872 with `headerMutation.set.name: x-ai-eg-model`.
2. Send a `/v1/chat/completions` request with model `some-model`.
3. Check the upstream request headers.
4. Before this patch: `x-ai-eg-model: some-model`
5. After this patch: `x-ai-eg-model: different-model`

Related Issues/PRs (if applicable)

Fixes #1872
Related PR: #1977

Special notes for reviewers (if applicable)

Added a small regression test in `internal/headermutator`.

### DIFF
--- a/internal/headermutator/header_mutator.go
+++ b/internal/headermutator/header_mutator.go
@@ -110,14 +110,19 @@ func (h *HeaderMutator) Mutate(headers map[string]string, onRetry bool) (sets []
 //
 // Skip Envoy AI Gateway headers since some of them are populated after the originalHeaders are captured.
 // This should be safe since these headers are managed by Envoy AI Gateway itself, not expected to be
-// modified by users via header mutation API.
+// modified by users via header mutation API. The exception is x-ai-eg-model: it is user-visible in
+// AIGatewayRoute matching and can be useful to rewrite before forwarding upstream.
 //
 // Also, skip Envoy pseudo-headers beginning with ':', such as ":method", ":path", etc.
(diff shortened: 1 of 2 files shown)


johnugeorge:
what is the use case for mutating the special header?

mathetake:
let's understand the use case in the original issue #1872 before moving with the change

PR #2292, opened 2026-06-25
fix(translator): merge input/cache token usage reported on Anthropic message_delta

### LINKED ISSUE #2290: Anthropic streaming usage drops input/cache tokens from message_delta
_Description_:

Anthropic Messages streaming usage extraction drops input/cache token fields when they are reported on `message_delta`.

Some Anthropic-compatible streaming backends report the final usage only on the `message_delta` event, for example:

```json
{
  "type": "message_delta",
  "usage": {
    "input_tokens": 4522,
    "cache_creation_input_tokens": 4511,
    "output_tokens": 5
  }
}
```

Current AI Gateway streaming handling reads `message_start.usage` for input/cache tokens, but for `message_delta.usage` it only updates `output_tokens`.

As a result, AI Gateway records usage like:

```text
input_tokens = 0
cache_creation_input_tokens = 0
cache_read_input_tokens = 0
output_tokens = 5
total_tokens = 5
```

The expected behavior is to treat `message_delta.usage` as a valid source of final usage fields and merge the explicit fields into token usage:

```text
raw input_tokens = 4522
cache_creation_input_tokens = 4511
output_tokens = 5
input total = 9033
total_tokens = 9038
```

This affects internal token usage accounting, metrics, and any downstream cost or quota logic that depends on AI Gateway token usage.

_Repro steps_:

Use an Anthropic Messages streaming response where usage is emitted on `message_delta`.

Minimal sanitized SSE response body:

```text
event: message_delta
data: {"type":"message_delta","delta":{"stop_reason":"end_turn","stop_sequence":null},"usage":
{"input_tokens":4522,"output_tokens":5,"cache_creation_input_tokens":4511}}

event: message_stop
data: {"type":"message_stop"}
```

A cache-read variant:

```text
event: message_delta
data: {"type":"message_delta","delta":{"stop_reason":"end_turn","stop_sequence":null},"usage":
{"input_tokens":4522,"output_tokens":5,"cache_read_input_tokens":4511}}

event: message_stop
data: {"type":"message_stop"}
```

Expected result:

```text
input/cache fields from message_delta.usage are included in TokenUsage.
```

Actual result:

```text
Only output_tokens from message_delta.usage is included.
input_tokens and cache_*_input_tokens are ignored.
```

_Environment_:

```text
AI Gateway: main branch
Envoy: not Envoy-specific
Request format: Anthropic Messages API, streaming response
```

_Logs_:

Sanitized response stream:

```text
event: message_delta
data: {"type":"message_delta","delta":{"stop_reason":"end_turn","stop_sequence":null},"usage":
{"input_tokens":4522,"output_tokens":5,"cache_creation_input_tokens":4511}}

event: message_stop
data: {"type":"message_stop"}
```

pjdurden:
**Description**

Anthropic Messages streaming usage extraction reads input and cache token counts from `message_start`, and on `message_delta` it only updates `output_tokens`. Some Anthropic-compatable backends report the final `input_tokens`, `cache_creation_input_tokens` and `cache_read_input_tokens` on the `message_delta` event instead. In that case the input and cache counts get dropped and recorded as zero, which corrupts usage metrics and any downstream cost or quota accounting that depends on them.

This runs the `message_delta` usage through the same `ExtractTokenUsageFromExplicitCaching` path already used for `message_start` and merges it with `Override`. The fields are only merged when non-zero, so a standard delta that carries `output_tokens` alone doesnt overwrite the `message_start` baseline with zeros.

Fixes #2290

Added two streaming subtests (cache-creation and cache-read variants) using the values from the issue: `input_tokens` 4522 + cache 4511 gives 9033 input total and 9038 total. They fail on main and pass with this change. The existing streaming and non-streaming usage tests are untouched.


CodePrometheus:
@pjdurden Hi, thanks for working on this, I originally planned to open a PR for this after filing the issue, but you got to it first. This addresses the main case. One thing I think we should be careful about is merging usage at field level rather than replacing the whole usage snapshot. In real environments, `message_delta.usage` may only include the input/cache fields that apply, so field presence matters when merging usage across events.

pjdurden:
Thanks, that makes sense and you are right that field presence matters here.

Pushed an update that merges per field instead of replacing the whole usage snapshot. It now keeps the latest known value for raw input, cache_read and cache_creation, and only overwrites the ones the message_delta actualy reports (non-zero), then recomputes the unified input_tokens sum. So a delta that carries only cache_read no longer zeros out the input_tokens already set on message_start. output_tokens is taken unconditionally from message_delta since that is always the final value there.

I cant distinguish an omitted field from an explicit zero, since the upstream usage fields are plain ints rather than pointers, so non-zero is the best proxy for presence. If you would rather gate this behind a backend flag instead of always-on merging, happy to switch to that.

Added a regression test where message_start sets input=1000 and the delta reports only cache_read=500, asserting the merged total stays 1500 instead of collapsing to 500.

nacx (approved):

nacx:
Thanks! @pjdurden can you fix the DCO?

pjdurden:
> Thanks! @pjdurden can you fix the DCO?

DCO fixed now. Thanks for looking into this - @nacx

PR #2293, opened 2026-06-25
fix(translator): return error instead of panicking on stream marshal failure

### LINKED ISSUE #2288: translator: streaming event marshal failure causes panic, crashes extproc
**Description**

  The OpenAI-to-AWS Bedrock translator panics when streaming event serialization fails, causing the extproc container to crash. Instead of propagating the error upstream for graceful handling, the code calls
  `panic()` at `internal/translator/openai_awsbedrock.go:740`.

  **Location**

  `internal/translator/openai_awsbedrock.go:740`

  ```go
  for i := range o.events {
      event := &o.events[i]
      oaiEvent, ok := o.convertEvent(event)
      if !ok {
          continue
      }
      err = serializeOpenAIChatCompletionChunk(oaiEvent, &newBody)
      if err != nil {
          panic(fmt.Errorf("failed to marshal event: %w", err))  // ❌ PANIC
      }
      if span != nil {
          span.RecordResponseChunk(oaiEvent)
      }
  }

  Root Cause

  serializeOpenAIChatCompletionChunk() wraps json.Marshal() and returns an error when marshaling fails:

  func serializeOpenAIChatCompletionChunk(chunk *openai.ChatCompletionResponseChunk, buf *[]byte) error {
      chunkBytes, err := json.Marshal(chunk)
      if err != nil {
          return fmt.Errorf("failed to marshal stream chunk: %w", err)
      }
      // ...
      return nil
  }

  When marshal fails (e.g., malformed chunk data, circular references, or edge cases in the converted Bedrock event), the streaming handler panics instead of returning an error.

  Impact

  - Critical: extproc crash disrupts all ongoing requests through the gateway
  - No graceful degradation: Clients receive connection errors, no meaningful error response
  - Hard to debug: Panic crashes the process without proper error logging
  - Affects: All streaming requests through AWS Bedrock backend

  Expected Behavior

  Return the error to the caller, allowing Envoy to handle the failure gracefully:

  err = serializeOpenAIChatCompletionChunk(oaiEvent, &newBody)
  if err != nil {
      return nil, nil, metrics.TokenUsage{}, "", fmt.Errorf("failed to marshal streaming event: %w", err)
  }

  This pattern is used correctly elsewhere in the same file for non-streaming responses:

  // Correct error handling at line 754-756
  if err = json.NewDecoder(body).Decode(&bedrockResp); err != nil {
      return nil, nil, metrics.TokenUsage{}, "", fmt.Errorf("failed to unmarshal body: %w", err)
  }

  Repro Steps

  While json.Marshal() on valid OpenAI chunks rarely fails, edge cases can trigger this:

  1. Bedrock returns an unusual event structure
  2. convertEvent() produces a chunk with data that cannot be marshaled
  3. json.Marshal() fails → serializeOpenAIChatCompletionChunk() returns error
  4. Streaming handler panics → extproc crashes

  Example Scenario

  - Bedrock guardrail intervention returns partial/corrupted event
  - Coral framework error with unexpected fields
  - Malformed content block during conversion

  Environment

  - Affects all versions with OpenAI-to-Bedrock streaming translator
  - Reproducible on current main branch

  Fix Recommendation

  Replace panic() with error return:

  // Before (BUG):
  err = serializeOpenAIChatCompletionChunk(oaiEvent, &newBody)
  if err != nil {
      panic(fmt.Errorf("failed to marshal event: %w", err))
  }

  // After (FIX):
  err = serializeOpenAIChatCompletionChunk(oaiEvent, &newBody)
  if err != nil {
      return nil, nil, metrics.TokenUsage{}, "", fmt.Errorf("failed to marshal streaming event: %w", err)
  }

  Related

  Similar panic patterns may exist in other streaming translators - should audit all panic() calls in production code paths.

  Logs

  panic: failed to marshal event: failed to marshal stream chunk: json: error calling MarshalJSON for type ...
  goroutine 42 [running]:
  github.com/envoyproxy/ai-gateway/internal/translator.(*openAIToAWSBedrockTranslatorV1ChatCompletion).ResponseBody.func1(...)
      /path/to/internal/translator/openai_awsbedrock.go:740 +0x...

  ---

  ## Labels
  bug, translator

  ---

pjdurden:
**Description**

The OpenAI to AWS Bedrock streaming translator calls `panic()` when a converted chunk fails to serialize (`internal/translator/openai_awsbedrock.go:740`). A panic on the response path crashes the extproc process and tears down every in-flight request through the gateway, instead of failing just the one request. This returns the error to the caller, matching the non-streaming decode path a few lines futher down in the same function.

In practice a valid chunk almost never fails to marshal (it is all string and int fields), so the panic guarded a path that is hard to hit with real data. To still cover it, a small `jsonMarshal` seam is added in `util.go` that the test swaps to force a marshal failure.

Fixes #2288

The `jsonMarshal` package var defaults to `json.Marshal`, so there is no behaviour change outside tests. The new test asserts `ResponseBody` returns the error and does not panic. The `internal/translator`, `internal/metrics` and `internal/extproc` packages pass, vet and gofmt are clean.

### DIFF
--- a/internal/translator/openai_awsbedrock.go
+++ b/internal/translator/openai_awsbedrock.go
@@ -737,7 +737,9 @@ func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) ResponseBody(_ map[string
 			}
 			err = serializeOpenAIChatCompletionChunk(oaiEvent, &newBody)
 			if err != nil {
-				panic(fmt.Errorf("failed to marshal event: %w", err))
+				// Return the error so Envoy can handle the failure gracefully instead of
+				// panicking, which would crash extproc and disrupt all in-flight requests.
+				return nil, nil, metrics.TokenUsage{}, "", fmt.Errorf("failed to marshal streaming event: %w", err)
 			}
 			if span != nil {
 				span.RecordResponseChunk(oaiEvent)
--- a/internal/translator/openai_awsbedrock_test.go
+++ b/internal/translator/openai_awsbedrock_test.go
@@ -1518,6 +1518,35 @@ func TestOpenAIToAWSBedrockTranslator_ResponseError(t *testing.T) {
 	}
 }
 
+// TestOpenAIToAWSBedrockTranslatorV1ChatCompletion_Streaming_MarshalError verifies that a failure
+// to serialize a converted streaming chunk is returned to the caller instead of panicking and
+// crashing extproc. See https://github.com/envoyproxy/ai-gateway/issues/2288.
+func TestOpenAIToAWSBedrockTranslatorV1ChatCompletion_Streaming_MarshalError(t *testing.T) {
+	// Valid chunks never fail to marshal, so inject a marshal failure on the serialization path.
+	original := jsonMarshal
+	jsonMarshal = func(any) ([]byte, error) { return nil, fmt.Errorf("injected marshal failure") }
+	t.Cleanup(func() { jsonMarshal = original })
+
+	// Encode a single content-block delta so convertEvent yields a chunk that is then serialized.
+	buf := bytes.NewBuffer(nil)
+	encoder := eventstream.NewEncoder()
+	payload, err := json.Marshal(awsbedrock.ConverseStreamEvent{
+		EventType: awsbedrock.ConverseStreamEventTypeContentBlockDelta.String(),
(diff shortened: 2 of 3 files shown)


johnugeorge:
Fix already merged in https://github.com/envoyproxy/ai-gateway/pull/2289

pjdurden:
Thanks, didn't realize #2289 already landed the same fix. Closing this as a duplicate.

PR #2296, opened 2026-06-26
fix(backendauth): resolve AWS signing host from `:authority` header

bannu8615:
**Description**

- The `awsHandler.Do` method previously hardcoded the request URL as `https://bedrock-runtime.<region>.amazonaws.com<path>`, ignoring the `:authority` HTTP/2 pseudo-header passed in by the ExtProc filter. 
- This prevented the handler from targeting custom endpoints such as AWS Private VPC endpoints, custom domain proxies, or any non-standard Bedrock host.

This commit makes the host selection dynamic:

- Reads the host from the `:authority` pseudo-header when present.
- Falls back to the `host` header when `:authority` is absent.
- Falls back to `bedrock-runtime.<region>.amazonaws.com` when both `host` and `:authority` are absent, preserving existing behavior.
- Updates the `Do` doc comment to describe the resolution logic, following the inclusive documentation style.
- Adds tests to cover the new `:authority` and `host` fallback branches.

**Related Issues/PRs**

Addresses the TODO left in: https://github.com/envoyproxy/ai-gateway/pull/61#discussion_r1900599982

> "I believe we always use `bedrock-runtime.*.amazonaws.com` so hardcoded — let's see if someone complains"

- The fallback for `:authority` and `host` is identical to the original hardcoded
  values, so there is no behavior change for existing deployments that do not set this
  header.
- The `:authority` pseudo-header in Envoy's ExtProc is set to the (potentially rewritten)
  upstream host, making it the correct source of truth for the target endpoint — not the
  static region string on the handler struct.
- The `region` field on `awsHandler` is still used in the default URL.

### DIFF
--- a/internal/backendauth/aws.go
+++ b/internal/backendauth/aws.go
@@ -80,9 +80,12 @@ func newAWSHandler(ctx context.Context, awsAuth *filterapi.AWSAuth) (filterapi.B
 
 // Do implements [Handler.Do].
 //
-// The method reads :method, :path, and :authority from requestHeaders (HTTP/2 pseudo-headers).
-// When :authority is present, the signed request targets that host; when absent, the host defaults to
-// the standard AWS Bedrock endpoint for the configured region (bedrock-runtime.<region>.amazonaws.com).
+// The method reads :method, :path, :authority, and host from requestHeaders. It resolves the host
+// for the signed request in the following order:
+//  1. the :authority pseudo-header, when present;
+//  2. the host header, when :authority is absent;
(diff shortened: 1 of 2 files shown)


mathetake (approved):

mathetake:
Reverted as it broke the tests on main branch:  https://github.com/envoyproxy/ai-gateway/actions/runs/28264368578/job/83747421662

mathetake:
Fork PRs don't run with credentials (obvious reasons for security)

PR #2298, opened 2026-06-26
Revert "fix(backendauth): resolve AWS signing host from `:authority` header"

mathetake:
Reverts envoyproxy/ai-gateway#2296


mathetake:
Broke the test: https://github.com/envoyproxy/ai-gateway/actions/runs/28264368578/job/83747421662

PR #2300, opened 2026-06-28
fix(controller): honor backendRef namespace for InferencePool cross-namespace refs

### LINKED ISSUE #2173: bug: AIGatewayRouteBackendRefs does not honor backend's namespace in cross-namespace configuration when the backend is InferencePool
_Description_:

> What issue is being seen? Describe what should be happening instead of
> the bug, for example: Envoy should not crash, the expected value isn't
> returned, etc.

As per the documentation, `AIGatewayRoute`'s `backendRefs` supports cross-namespace backends. This does not seem to be the case when the backend is an `InferencePool`. The resulting `HTTPRoute` refers to the `AIGatewayRoute`'s namespace, regardless of the backend's specified namespace with a `ReferenceGrant` allowing cross-namespace references. This is not reflected anywhere in the documentation.

_Repro steps_:

> required to reproduce the bug.

> **Note**: If there are privacy concerns, sanitize the data prior to
> sharing.

1. Create an `InferencePool` in namespace `foo`
2. Create an `AIGatewayRoute` in namespace `bar` with `backendRefs` referring to the `InferencePool` in `foo`
```yaml
kind: AIGatewayRoute
metadata:
  name: model-deployment-route
  namespace: bar
spec:
  parentRefs:
      kind: Gateway
      name: envoy-ai-gateway
  rules:
    - backendRefs:
        - group: inference.networking.k8s.io
          kind: InferencePool
          name: model-inference-pool
          namespace: foo
```

3. Observe that the resulting `HTTPRoute` has the backend's namespace specified as the same as `AIGatewayRoute`'s namespace.
```yaml
kind: HTTPRoute
metadata:
  name: model-deployment-route
  namespace: bar
  ownerReferences:
  - apiVersion: aigateway.envoyproxy.io/v1beta1
    blockOwnerDeletion: true
    controller: true
    kind: AIGatewayRoute
    name: model-deployment-route
spec:
  parentRefs:
    kind: Gateway
    name: envoy-ai-gateway
  rules:
  - backendRefs:
    - group: inference.networking.k8s.io
      kind: InferencePool
      name: model-inference-pool
      namespace: bar
```

_Environment_:

> Include the environment like gateway version, envoy version and so on.

- Envoy AI Gateway: v0.6.0
- Envoy Gateway: v1.7.2
- Kubernetes: v1.33.1

_Logs_:

> Include the access logs and the Envoy logs.

No logs needed. `AIGatewayRoute` is accepted as-is. No errors are logged.

## Root cause

I have narrowed down to the cause [here](https://github.com/envoyproxy/ai-gateway/blob/edef4c5687f96f4c34ed962b84bee13cc549c02e/internal/controller/ai_gateway_route.go#L264-L276). In this snippet the namespace is taken from `AIGatewayRoute` and not from the backend reference. I don't know whether this is intended or not, but regardless of intention it should be reflected in the documentation.

AruneshDwivedi:
**Description**

When an AIGatewayRoute specifies an InferencePool backendRef in a different namespace, the generated HTTPRoute was always using the AIGatewayRoute's own namespace for the backendRef. This caused ResolvedRefs: False even with a valid ReferenceGrant in place.

Uses the backendRef's explicit namespace when provided, falling back to the AIGatewayRoute namespace when unset. Adds a test for the cross-namespace case.

### DIFF
--- a/internal/translator/anthropic_helper.go
+++ b/internal/translator/anthropic_helper.go
@@ -829,6 +829,67 @@ func (p *anthropicStreamParser) writeChunk(eventBlock []byte, buf *[]byte) error
 	return nil
 }
 
+type messageDeltaUsageFields struct {
+	Usage *struct {
+		InputTokens              *int64 `json:"input_tokens"`
+		CacheReadInputTokens     *int64 `json:"cache_read_input_tokens"`
+		CacheCreationInputTokens *int64 `json:"cache_creation_input_tokens"`
+	} `json:"usage"`
+}
+
+func (p *anthropicStreamParser) updateInputUsageFromMessageDelta(data []byte) error {
+	// message_delta provides cumulative (not incremental) token counts.
+	// This function handles input_tokens, cache_read_input_tokens, and cache_creation_input_tokens
+	// from message_delta. We use Set (not Add) because these are cumulative totals.
+	// This prevents double-counting when both message_start and message_delta report the same
+	// cache tokens, and also handles cases where:
+	// - Cache tokens are reported only in message_delta (not in message_start)
+	// - message_delta provides corrected/updated cache token values that override message_start
+	var event messageDeltaUsageFields
+	if err := json.Unmarshal(data, &event); err != nil {
+		return fmt.Errorf("unmarshal message_delta usage fields: %w", err)
+	}
+	if event.Usage == nil {
+		return nil
+	}
+
+	u := event.Usage
+	inputPresent := u.InputTokens != nil && *u.InputTokens >= 0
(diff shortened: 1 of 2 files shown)


aabchoo (approved):
Could you please address the failing CI? ([PR Style Check](https://github.com/envoyproxy/ai-gateway/actions/runs/28600628374/job/84807583694?pr=2300))

nacx:
@AruneshDwivedi can you resolve the conflicts?

AruneshDwivedi:
The conflict was just fork pollution — the branch carried unrelated commits and the fix was already merged into main, so I reset the branch to upstream/main. The PR diff is now empty and can be closed.

PR #2306, opened 2026-06-30
fix(mcpproxy): parse JSON content types with parameters

immanuwell:
**Description**

`mcpproxy` only treated the exact header `application/json` as JSON. If a backend returns `application/json; charset=utf-8`, it kinda falls into the SSE path by mistake. That can skip JSON-RPC decoding, miss `resources/read` URI rewriting, and `proxyResponseBody` can panic when no session is present.

This parses the media type instead, so JSON responses with params still use the JSON path. Pretty small fix, same behavior for existing bare `application/json` and `text/event-stream` cases.

Related to #2218
Related PR: #2229

Repro:
1. Return `Content-Type: application/json; charset=utf-8` from an MCP backend with a valid JSON-RPC body.
2. Hit `resources/read`, or the GET notification path.
3. Before this patch, the proxy takes the SSE path. `resources/read` stops rewriting downstream URIs, and `proxyResponseBody` can panic if no session is present.
4. With this patch, the same responses decode as JSON like normal.

### DIFF
--- a/internal/mcpproxy/content_type.go
+++ b/internal/mcpproxy/content_type.go
@@ -1,0 +1,25 @@
+// Copyright Envoy AI Gateway Authors
+// SPDX-License-Identifier: Apache-2.0
+// The full text of the Apache license is available in the LICENSE file at
+// the root of the repo.
+
+package mcpproxy
+
+import (
+	"mime"
+	"net/http"
+	"strings"
+)
+
+func hasMediaType(header http.Header, want string) bool {
+	got, _, err := mime.ParseMediaType(header.Get("Content-Type"))
+	if err == nil {
+		return got == want
+	}
+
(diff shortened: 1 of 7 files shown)


internal/mcpproxy/content_type.go
```diff
@@ -0,0 +1,25 @@
...
+
+import (
+	"mime"
+	"net/http"
+	"strings"
+)
+
+func hasMediaType(header http.Header, want string) bool {
+	got, _, err := mime.ParseMediaType(header.Get("Content-Type"))
+	if err == nil {
```
Hritik003, line 16: nit: Can we add this helper function inside the handler file itself?

Hritik003 (commented):
lgtm, 1 minor comment

PR #2307, opened 2026-07-01
fix: ensure appropriate CORS headers for browser based MCP clients

jappievw:
**Description**

This PR ensures that the appropriate CORS headers are included both for the `.well-known` paths as well as the main MCP route. This allows MCP Inspector to connect to an MCP Server exposed by the AI Gateway.

The CORS headers implemented to support the MCP Inspector were not functioning properly.

For the `.well-known/oauth-protected-resource/mcp` path CORS headers were returned, see below:

```
access-control-allow-headers: mcp-protocol-version
access-control-allow-methods: GET
access-control-allow-origin: *
```

The `access-control-allow-headers` and `access-control-allow-methods` are not needed for the protected resources path.

The `OPTIONS` preflight request for the `/mcp` route returns an `HTTP 401` due to an authorization error, see response headers below:

```
HTTP/2 401 
content-length: 0
www-authenticate: Bearer error="invalid_token", error_description="The access token is missing or invalid", resource_metadata="https://mcp.host/.well-known/oauth-protected-resource/mcp", scope="scope1 scope2"
access-control-allow-origin: *
access-control-allow-methods: GET
access-control-allow-headers: mcp-protocol-version
```

There are two problems with this response:
* By design, the CORS preflight calls are unauthenticated. Failing due to auth should not be possible.
* The methods and headers are not sufficient.

The Envoy SecurityPolicy is now always included to serve the `OPTIONS` preflight requests and include CORS headers.

_This PR has been created with help from Claude (thank you)_


nacx (commented):
Thanks! Overall LGTM!

I'm cusious about:

>This allows MCP Inspector to connect to an MCP Server exposed by the AI Gateway.

Could you describe what did not work properly? While developing the MCP feature we used the MCP inspector extensively without issues, so I'm wondering what issues you found?

internal/controller/mcp_route_security_policy.go
```diff
@@ -107,8 +104,23 @@ func (c *MCPRouteController) ensureSecurityPolicy(ctx context.Context, mcpRoute
 
 	securityPolicySpec := egv1a1.SecurityPolicySpec{}
 
+	// CORS applies to every MCP route (browser MCP requests are always preflighted). Envoy Gateway orders
+	// the cors filter ahead of the auth filters, so a preflight OPTIONS is answered and short-circuited
+	// before jwt_authn/ext_authz — the preflight is never authenticated.
+	securityPolicySpec.CORS = &egv1a1.CORS{
```
Hritik003, line 110: does performance concern here by not setting `egv1a1.CORS.maxAge`( maxAge tells the browser how long it can cache the preflight response) ?

Hritik003 (commented):
Thanks for pointing it out, I remember facing this issue while trying out the PoC with MCP Route Oauth. Had to exclusively add the CORS policy for the well known HTTP Route, for the inspector to able to able to send requests.

LGTM, one minor comment

jappievw:
Hey @nacx, thanks for your fast review. It's very much appreciated.

> > This allows MCP Inspector to connect to an MCP Server exposed by the AI Gateway.
> 
> Could you describe what did not work properly? While developing the MCP feature we used the MCP inspector extensively without issues, so I'm wondering what issues you found?

There are two problems that prevent using MCP Inspector with the AI Gateway today:
1. The `OPTIONS` call for the MCP route also has the auth filter running. This is why the `OPTIONS` call responds with a `401` instead of a `2xx`.
2. The CORS headers are not sufficient for the MCP route itself:
    * It only allows for `GET`, whereas also `POST` and `DELETE` should be possible.
    * The list of headers that is allowed to be exposed to JS essentially is too limited. It doesn't allow to use the session id, the www-authenticate, etc.

The solution direction I propose with this PR differs between the `.well-known` paths and the main MCP route:
1. For `.well-known`: use Response Headers to allow all hosts. This is sufficient for the `GET` calls.
2. For the main MCP Route handle making sure that the processing pipeline stops for `OPTIONS` methods after the CORS filter. Secondly, making sure that the appropriate headers are returned.

Does this clarify.

internal/controller/mcp_route_security_policy.go
jappievw: Great question. I think it would make sense to add a `maxAge`.

The reason I didn't do it yet is that it wasn't set before. Also, when we're setting a value I think it should be configurable. If we do so, I'm unsure if that's worth it. What do you think?

Hritik003: I don't think any user would have a usecase around it. But it is mostly helpful around browser based MCP clients.

nacx (approved):

nacx:
Thanks for the explanation. LGTM!

johnugeorge:
Tests are failing

Issue #2309, opened 2026-07-02
mcp: proposal to mandate forward headers in MCP Route


Hritik003:
### Description:

Add an optional `required` boolean field to the `MCPHeaderForward` type (used in MCPRouteBackendRef.forwardHeaders) so operators can enforce that a per-user forwarded header must be present on the incoming client request. When `required: true` and the header is absent from the incoming request, the gateway should reject the request (e.g. return a JSON-RPC / HTTP error) instead of silently forwarding it to the backend.

#### Desired behavior:

```go
// MCPHeaderForward specifies a header to extract from the incoming request and forward to a backend.
type MCPHeaderForward struct {
	// Name is the header name to extract from the incoming client request.
	Name string `json:"name"`
	// BackendHeader is the header name to use when forwarding to the backend.
	// If not specified, the original header name is used.
	BackendHeader *string `json:"backendHeader,omitempty"`
	// Required indicates whether the header must be present on the incoming client request.
	// When true and the header is missing, the request is rejected rather than
	// forwarded to the backend. Defaults to false (current behavior: silently skip).
	Required *bool `json:"required,omitempty"`
}
```

- required: false (default) preserves today's behavior — see extractPerBackendForwardHeaders in internal/mcpproxy/handlers.go, which just skips a header when it isn't found on the request.
- required: true — if the configured header is missing, the gateway short-circuits and returns an error to the caller, so the request never reaches the upstream MCP backend without the per-user header.

#### Scenario this enables

An upstream MCP server can be configured to accept requests authenticated by either a shared credential (injected via the backend's securityPolicy, e.g. an API key / Authorization header) or a per-user custom header, and it returns different responses depending on which is present (auth+header vs. header-only).

When an admin configures both a backend securityPolicy and a forwardHeaders entry for a per-user header, and an end user calls the dataplane (e.g. tools/list, tools/call) without supplying their own header, the request today still succeeds — silently falling back to the shared security-policy credential. This means the end user unintentionally acts under the shared credential's identity/permissions instead of their own.

Marking the forwarded header as required closes that gap: requests missing the per-user header are rejected up front, guaranteeing the user's own header is always the credential in play and preventing accidental privilege fallback to the shared policy credential.

#### How will it be used?

```yaml
apiVersion: aigateway.envoyproxy.io/v1beta1
kind: MCPRoute
metadata:
  name: example-mcp-route
spec:
  backendRefs:
    - name: github-mcp
      kind: Backend
[...]

PR #2312, opened 2026-07-02
mcp: add support for required flag in MCP Header forwarding

Hritik003:
**Description**

This PR adds an optional required boolean field to MCPHeaderForward (used in MCPRoute's per-backend forwardHeaders), enabling operators to enforce that a per-user forwarded header must be present on the incoming client request.

Previously, when a backend was configured with both a securityPolicy (which injects a shared credential, e.g. an Authorization/API-key header) and a forwardHeaders entry for a per-user header, an end user who called the dataplane (e.g. tools/list, tools/call) without supplying their own header would still succeed — silently falling back to the shared security-policy credential and unintentionally acting under its identity/permissions instead of their own.

With required: true, if the configured header is missing from the incoming request, the gateway rejects the request with 400 Bad Request instead of forwarding it to the backend, guaranteeing the user's own header is always the credential in play. The default is false, which preserves the existing behavior of silently skipping the missing header.

Changes:

- Added Required *bool to MCPHeaderForward in both api/v1beta1 and api/v1alpha1 (defaults to false); regenerated deepcopy, CRDs, and API docs.
- Threaded the field through the internal filter config (filterapi.MCPHeaderForward) and the controller translation (internal/controller/gateway.go).
- Enforced it in the MCP proxy: extractPerBackendForwardHeaders now returns an error (errRequiredHeaderMissing) when a required header is absent; newSession and sessionFromID propagate it, and the initialize/POST handlers return 400 with a - - new missing_required_header metric error type.
Added/updated unit tests.

Example:

```yaml
kind: MCPRoute
metadata:
  name: test
  namespace: test-ns
spec:
  parentRefs:
    - name: nai-ingress-gateway
      kind: Gateway
      namespace: nai-system
  path: /mcp
  backendRefs:
    - name: test-ns
      kind: Backend
      group: gateway.envoyproxy.io
      namespace: test-ns
      path: /mcp
      toolSelector:
        include:
          - whoami
          - debug_headers
      forwardHeaders:
        - name: X-user-name
          required: True

```

- curl request to a tools/call without a required header

```
➜  ~ curl 'http://localhost:6277/mcp?url=https%3A%2F%2F10.122.11.125%3A30902%2Fenterpriseai%2Fgateway%2Fmcp%2Fheader-
........
  --data-raw '{"method":"tools/call","params":{"name":"header-test-local__debug_headers","arguments":{},"_meta":{"progressToken":3}},"jsonrpc":"2.0","id":3}'
event: message
data: {"jsonrpc":"2.0","id":3,"error":{"code":-32001,"message":"Error POSTing to endpoint (HTTP 400): backend \"header-test-local\": required forward header missing: \"X-user-name\"\n","data":{}}}
```

Closes - #2309


Hritik003:
Closing this PR since we are going with a different approach, thanks

PR #2305, opened 2026-06-30
chore: update kiwi mcp tool to pass the tests.

hustxiayang:
**Description**
CI tests failed because the tool name changed: https://github.com/envoyproxy/ai-gateway/actions/runs/28460601685/job/84347247072?pr=1650. Update the kiwi mcp tool name.

PR #2315, opened 2026-07-02
ci: improve fragile mcp tests to avoid ci tests failures

hustxiayang:
**Description**
These MCP integration tests talk to third-party public MCP servers (Kiwi, GitHub) that add and rename tools upstream without notice, which caused recurring CI churn — see the revert of #2305. 

This change discovers Kiwi tools dynamically by prefix instead of pinning exact tool names, and asserts the flight search returns a non-empty result set (using a future departure date) so the test still meaningfully exercises the backend. The tradeoff is intentional: we verify expected tools are present rather than asserting an exact tool set.


tests/data-plane-mcp/publicmcp_test.go
```diff
@@ -81,62 +81,70 @@ func TestPublicMCPServers(t *testing.T) {
...
 	t.Run("tool calls", func(t *testing.T) {
+		// Discover a kiwi flight search tool dynamically since Kiwi may rename tools upstream.
+		listResp, err := session.ListTools(t.Context(), &mcp.ListToolsParams{})
+		require.NoError(t, err)
+		var kiwiFlightTool string
+		for _, tool := range listResp.Tools {
+			if strings.HasPrefix(tool.Name, "kiwi__") && strings.Contains(tool.Name, "flight") {
+				kiwiFlightTool = tool.Name
+				break
+			}
```
Hritik003: extract this tool discovering into a helper function?

Hritik003 (commented):
thanks for the PR, was facing the same. Minor nit, else LGTM

tests/data-plane-mcp/publicmcp_test.go
hustxiayang: thanks a lot! updated.

kanurag94 (approved):

PR #1995, opened 2026-03-29
translator: gemini non-stream reasoning as string and thinking_blocks

Flgado:
**Description**  
Non-streaming Vertex / Gemini chat completions now follow the `LiteLLM-style` split already discussed in the community: `reasoning_content` is always a plain string (the visible thinking summary text), and optional `thinking_blocks` carry structured metadata (e.g. signatures) that OpenAI’s core schema does not define. This matches the direction agreed on in Slack [1] and avoids exposing a Bedrock-shaped nested object in reasoning_content, which breaks common OpenAI-compatible clients.

Implementation highlights:  
- Extend the OpenAI-shaped `ChatCompletionResponseChoiceMessage` schema with ThinkingBlock  

- In the Gemini helper, map thought parts to the string union for `reasoning_content` and populate `thinking_blocks`; when the model attaches a thought signature to the first function-call part (parallel tools) or only there, merge or attach that signature into `thinking_blocks` so clients can **round-trip history together with tool_calls** and assistant content parts of type thinking + signature.

- Add unit tests in internal/translator/gemini_helper_test.go for geminiCandidatesToOpenAIChoices and for signature extraction in extractTextAndThoughtSummaryFromGeminiParts.

This commit improves interoperability with LiteLLM, LangChain, and other clients that expect `reasoning_content` to be a string while still preserving provider-specific seals for advanced use cases.

Fixes #1974  

**Special notes for reviewers**  
- Request-side replay of thinking continues to use assistant content as structured parts (type: "thinking", text, signature) plus tool_calls where applicable; this PR focuses on response shape and mapping from Gemini parts.  
- Optional follow-up (not in this PR): emit or accept Google’s OpenAI-compat extra_content.google.thought_signature on tool_calls for clients that follow that dialect verbatim; the LiteLLM-style fields remain the primary contract.

**Testing**:  
- Manually verified chat completions against Vertex with gemini-2.5-flash and gemini-3.1-pro-preview (thinking enabled): reasoning_content is a plain string, thinking_blocks present when expected, and multi-turn replay with thinking + signature in assistant content.

---

1: https://envoyproxy.slack.com/archives/C07Q4N24VAA/p1774221731287089

Issue #1974, opened 2026-03-19
OpenAI-GCP Vertex AI translator: reasoning_content returned as nested dict instead of string (non-streaming)


MotzWanted:
**Description**

The non-streaming OpenAI-to-GCP Vertex AI translator returns `reasoning_content` as a nested dict instead of a plain string for Gemini 3 models. This breaks OpenAI-compatible clients (e.g. litellm pydantic validation rejects the response).

The streaming path already handles this correctly using `StreamReasoningContent`. Gemini 2.5 models are unaffected.

**Actual response:**
```json
"reasoning_content": {"reasoningContent": {"reasoningText": {"text": "...", "signature": "..."}}}
```

**Expected:**
```json
"reasoning_content": "the thinking text..."
```

**Root cause:** 
`geminiCandidatesToOpenAIChoices` in internal/translator/gemini_helper.go wraps the value in Bedrock types. 

**Suggested fix:**
set Value to the string directly (3 occurrences):
```
// before
Value: &openai.ReasoningContent{ReasoningContent: &awsbedrock.ReasoningContentBlock{...}}
// after
Value: thoughtSummary
```

**Error (client-side):**
```
pydantic_core._pydantic_core.ValidationError: 1 validation error for Message
reasoning_content
  Input should be a valid string [type=string_type, input_value={'reasoningContent': {'re...}}, input_type=dict]
```

PR #2319, opened 2026-07-03
translator: stream reasoning_content as plain string, signatures via thinking_blocks

### LINKED ISSUE #2318: Streaming reasoning_content is a JSON object, breaking OpenAI-compatible clients (streaming counterpart of #1974)
_Description_:

When streaming chat completions with reasoning models, `delta.reasoning_content` is always serialized as a JSON **object**, but the de-facto OpenAI-compatible convention (established by DeepSeek, followed by vLLM, LiteLLM, `@ai-sdk/openai-compatible`, etc.) is `reasoning_content: string`. Clients that validate the chunk schema hard-fail and abort the entire stream — no partial text, no tool result.

Reported shape via the GCP Vertex AI translator with a Gemini thinking model (`gemini-2.5-pro`, `gemini-3.5-flash`) + tool calling:

```json
"delta": { "tool_calls": [...], "reasoning_content": { "signature": "<base64>" } }
```

This is the streaming counterpart of #1974. That issue stated "the streaming path already handles this correctly using `StreamReasoningContent`", but `StreamReasoningContent` (`internal/apischema/openai/openai.go`) is a struct `{Text, Signature, RedactedContent}` with no custom `MarshalJSON`, so streaming `reasoning_content` is *always* an object. #1995 fixed only the non-streaming path (via `ReasoningContentUnion.MarshalJSON`).

Because the bug is in the shared wire type, it is not Vertex-specific:

- **GCP Vertex AI** (`internal/translator/openai_gcpvertexai.go:404-441`): Gemini thinking models attach a `thoughtSignature` to function-call parts → `{"signature": "..."}`; thought summaries → `{"text": "..."}`.
- **AWS Bedrock** (`internal/translator/openai_awsbedrock.go:968-986`): Claude extended-thinking deltas → `{"text": "..."}`, plus `{"signature": ...}` / `{"redactedContent": ...}` variants — fails the same string validation.

There is also a latent inverse bug: the OpenAI→Anthropic streaming path (`internal/translator/openai_helper.go`) unmarshals backend chunks into the same struct, so an OpenAI-compatible backend that emits the conventional *string* `reasoning_content` (vLLM, DeepSeek) fails to parse.

Expected: `delta.reasoning_content` is a plain string, mirroring #1995. I have a working patch that does this the same way #1995 did for non-streaming: `reasoning_content` marshals as a plain string (text only), and signatures/redacted content move to the LiteLLM-convention `thinking_blocks` field on the delta (already present on the non-streaming `Message`), with unmarshaling accepting both the string and legacy object forms. All existing translator/extproc tests pass with updated goldens. Happy to open a PR if maintainers agree with that shape.

One design note: Gemini 3 expects thought signatures echoed on subsequent function-calling turns, and the request-side translator (`internal/translator/gemini_helper.go`) reads them from assistant message parts. Whatever shape is chosen needs a documented client-side home for the signature — `thinking_blocks` gives it one; today's object shape crashes the clients that would need it anyway.

_Repro steps_:

1. Configure an `AIGatewayRoute` with an `AIServiceBackend` using `schema.name: GCPVertexAI` and a thinking-capable Gemini model (e.g. `gemini-2.5-pro`).
2. `POST /v1/chat/completions` with `"stream": true` and at least one tool defined, using a prompt that triggers a tool call.
3. The first function-call chunk contains `delta.reasoning_content` as an object.

Client-side result with `@ai-sdk/openai-compatible` (zod schema `reasoning_content: z.string().nullish()`):

```
AI_TypeValidationError: expected string, received object
  at choices.0.delta.reasoning_content
```

(LiteLLM's pydantic validation fails the same way, as in #1974.)

Non-thinking models (e.g. `gemini-2.5-flash-lite`) never attach a signature and are unaffected. Equivalent repro for Bedrock: any Claude model with extended thinking enabled, no tools required.

_Environment_:

Envoy AI Gateway v1.0.0. The relevant code is unchanged on `main` as of today.

_Logs_:

Sample crashing SSE chunk (sanitized):

```json
{"id":"...","object":"chat.completion.chunk","model":"gemini-2.5-pro","choices":[{"index":0,"delta":{"role":"assistant","tool_calls":[{"index":0,"id":"...","type":"function","function":{"name":"...","arguments":"..."}}],"reasoning_content":{"signature":"AY89a1..."}},"finish_reason":null}]}
```

dragon3:
**Description**

Fixes #2318 — the streaming counterpart of #1974 / #1995.

Streaming chunks serialized `delta.reasoning_content` as a JSON object because `StreamReasoningContent` had no custom `MarshalJSON`. OpenAI-compatible clients that validate the de-facto `reasoning_content: string` convention (LiteLLM pydantic, `@ai-sdk/openai-compatible` zod, see 1) reject the chunk and abort the entire stream. With Gemini thinking models this fires on the very first tool call (`{"signature": ...}`); with Claude extended thinking on Bedrock it fires on the first reasoning delta (`{"text": ...}`).

This mirrors the non-streaming fix from #1995:

- `StreamReasoningContent.MarshalJSON` emits a plain JSON string (text only).
- Provider metadata that cannot live in a plain string moves to the LiteLLM-convention `thinking_blocks` field, added to `ChatCompletionResponseChunkChoiceDelta` (it already exists on the non-streaming `Message`): thought/tool-call signatures become `{"type":"thinking","signature":...}`, Bedrock redacted content becomes `{"type":"redacted_thinking","data":"<base64>"}`.
- `StreamReasoningContent.UnmarshalJSON` accepts both the plain-string form emitted by OpenAI-compatible backends (vLLM, DeepSeek) and the legacy object form emitted by ai-gateway <= v1.0.0. Previously the OpenAI-to-Anthropic streaming path failed to parse the conventional string form at all.
- The OpenAI-to-Anthropic streaming path also consumes `delta.thinking_blocks`, so signatures keep round-tripping to Anthropic-protocol clients.
- `anthropic_helper.go` now passes the chunk delta by pointer: the added `thinking_blocks` slice pushed the struct over golangci-lint's `hugeParam` threshold.

Existing translator/extproc tests pass with goldens updated to the new wire shape, plus new marshal/unmarshal regression tests.

Note for reviewers: this changes where streaming signatures appear on the wire. Any client parsing the old `{"signature": ...}` object must read `thinking_blocks` instead — though such clients were rare, since schema-validating clients could not complete a stream at all.

Per the generative AI policy: this change was developed with an AI assistant (Claude); I take ownership of the code and will address review feedback.

1: https://github.com/vercel/ai/issues/7866


nacx:
Thanks! Overall LGTM.
@aabchoo @yuzisun can you take a look and chime in regarding the breaking change?

dragon3:
@nacx 
Thank you for reviewing! Is there anything I should do more to get this merged?

dragon3:
Hi @aabchoo @yuzisun, friendly ping 🙂 — @nacx asked for your input on the wire-format change here.

To save you a full re-read, the question is about this breaking change: streaming signatures move from the legacy `delta.reasoning_content` object form to `delta.thinking_blocks` (matching the LiteLLM convention already used in the non-streaming response since #1995). Clients parsing the old object form would need to read `thinking_blocks` instead — though in practice such clients were rare, since schema-validating clients couldn't complete a stream at all (that's the bug #2318 this fixes).

Happy to adjust the approach if you'd prefer something more backward-compatible (e.g. a transition flag). Thanks!

johnugeorge:
Can you fix the CI?

johnugeorge:
Also, check the coverage as well

dragon3:
@johnugeorge Thanks for taking a look! I've addressed both:

**CI:** The failing `Data Plane Test` was a stale golden. The branch had fallen behind `main`, and since CI runs the PR merged into `main`, the newer `streaming with thinking config` test case was still asserting the old object form for `delta.reasoning_content` — which is exactly the wire shape this PR changes to a plain string. I've merged `main` and updated that golden to the string form.

**Coverage:** Added `TestOpenAIStreamToAnthropicState_ProcessBuffer_ThinkingBlocks`, which drives a `delta.thinking_blocks` chunk through the OpenAI→Anthropic path and covers the signature round-trip that was previously untested in `openai_helper.go`.

Both the `tests/data-plane` and `internal/translator` suites pass locally. Could you approve the workflow runs so the checks can execute? 🙏

One note: the earlier `aigw CLI` E2E failure was `TestMCP_standalone`, which looks unrelated to this PR — its sibling `TestMCP_standalone_oauth` passed in the same run, and the MCP code is untouched here. Hopefully just a flake on re-run, but flagging in case it's a known one.

dragon3:
@johnugeorge 

Thanks for reviewing and running the tests.

The only failing check is a transient `503 Service Unavailable` while fetching the envtest binary (`envtest-v1.33.0-linux-amd64.tar.gz`) from GitHub during test setup — `TestStartControllers` never actually ran, and it's unrelated to this change. Everything else is green now. Could you re-run the `Controller Test (ubuntu-latest)` job? 🙏

johnugeorge (approved):

dragon3:
Thanks!

PR #2324, opened 2026-07-05
feat: support for OpenAI Files API

sivanantha321:
**Description**

This PR adds support for the OpenAI **Files API** (`/v1/files*`: upload, list, retrieve, download
content, delete) to the data plane.

Unlike chat/embeddings — which are stateless and routed per-request by model — a file is a **stateful
object that lives on one specific upstream backend**. A file uploaded to backend A does not exist on
backend B, so every id-bearing request (retrieve, content, delete) must be routed back to the exact
backend that created the file, and a single `list` must present one unified view stitched across all
of a route's backends. The gateway achieves this **without any server-side state** by encrypting the
serving backend's identity into the opaque file id it returns to the client, and decoding it again on
subsequent requests.

The implementation is delivered in five cooperating pieces:

- **ID codec (`internal/idcodec/`)** — encrypts `kind | namespace/name | native-id` with
  AES-GCM/PBKDF2 (reusing the audited MCP-session crypto), re-encoded into OpenAI's id charset with a
  kind prefix (`file-`, `batch-`, `flcur-`). Ids are tamper-resistant and unforgeable: a client
  cannot read the backend out of an id nor mint one targeting an arbitrary backend. Supports key
  rotation via an optional fallback seed.

- **Files processor (`internal/extproc/files_processor.go`)** — a prefix-registered ext_proc
  processor (new `Server.RegisterPrefix` for the variable `/v1/files/{id}` segment) that classifies
  the operation, routes it, rewrites client⇄backend-native ids in both directions, and pins
  id-bearing requests to the owning backend.

- **Cross-backend list pagination (`internal/extproc/files_list_walk.go`)** — a stateless "walk"
  that visits a route's backends in a deterministic cycle, one page at a time, carrying the walk
  position inside an encrypted `flcur-` pagination cursor returned as `last_id`. The client follows
  `last_id → after` exactly like normal OpenAI pagination.

- **Backend-sticky routing primitive (`internal/extensionserver/sticky_routing.go`)** — the xDS
  machinery, injected during `PostTranslateModify`, that lets a request be pinned to one backend's
  endpoints: (1) each endpoint is tagged with its backend identity under `envoy.lb`, (2) the
  cluster's LB policy is wrapped in a subset policy keyed on `selected_backend` (with an
  `ANY_ENDPOINT` fallback so non-Files traffic is unaffected), and (3) per-backend "sticky" routes
  are synthesized that match on the `selected_backend` dynamic metadata the processor emits.

- **Files translator (`internal/translator/files.go`, `openai_openai_files.go`)** — a
  per-operation translator selected by the backend schema. **OpenAI is the only supported provider
  today** (byte-for-byte pass-through); every other schema (GCP/AWS/Azure/Anthropic/Cohere) is
  rejected with `501`, so a misconfigured route fails closed rather than leaking a backend-native id.
  The per-op types are the seam where future provider support drops in without touching the id-codec,
  sticky routing, or list-walk.

Routing model is **required** for the two id-less operations:
- **Upload** routes by a `model` field supplied via the OpenAI client's `extra_body` (multipart form
  field), which is stripped before forwarding upstream. A missing model → `400`.
- **List** routes by a required `?model=` query parameter (an AI-Gateway extension), stripped from
  the upstream path on every page. A missing model → `400`.

Error behavior (all returned to the client, nothing forwarded upstream): `400` for a missing `model`
or an invalid `after` cursor; `404` for a file id that cannot be decoded (unknown, tampered, or
forged) or an unsupported method/path; `410` when the backend encoded in an id/cursor no longer
exists in the config; `501` when the target backend's schema is not OpenAI.

Configuration: new extproc flags `fileIDEncryptionSeed` / `fileIDEncryptionIterations` and optional
`fileIDFallbackEncryptionSeed` / `fileIDFallbackEncryptionIterations` for the id encryption key
(defaults are insecure and must be changed in production).

User-facing documentation is included under `site/docs`: a new Files API guide
(`capabilities/llm-integrations/files-api.md`) with per-endpoint `curl` and OpenAI Python SDK
examples, configuration, cross-backend pagination, the encryption flags, and caveats; a new `Files`
section in `supported-endpoints.md`; and links from the capabilities index pages.

## Per-operation data flow

Notation: **R** = router-level filter, **U** = upstream-level filter, **LB** = Envoy load balancer.
Headers of note: `x-ai-eg-model` (routing model), `x-ai-eg-original-path` (preserved original path),
`selected_backend` dynamic metadata (backend pin, namespace `io.envoy.ai_gateway`).

### Upload — `POST /v1/files`

Routing depends on the multipart body, so the decision is deferred to the body phase.

```
1. client → POST /v1/files  (multipart: file=..., purpose=..., model=M)
2. R.ProcessRequestHeaders: classify = upload → continue (await buffered body)
3. R.ProcessRequestBody:
     - parse multipart, read `model` (M) from the form field (OpenAI SDK sends it via extra_body)
     - if model == "" → 400 (required)
     - set  x-ai-eg-model: M          (so the route matches)
     - strip the `model` field from the body (OpenAI Files API has no model field)
     - set content-length, ClearRouteCache = true
4. Envoy re-matches route by x-ai-eg-model = M ; LB picks a backend endpoint
5. U.SetBackend(backend, routeProcessor=R): R.backend* := ns/name of the LB-selected backend
                                            R.resolveTranslator(schema)  (501 here if non-OpenAI)
6. upstream returns {"id":"<native>", ...}
7. R.ProcessResponseBody → reEncodeResponse:
     - (translate response, no-op for OpenAI)
     - gw = codec.Encode({ns,name,KindFile,native}) ; set id = gw ; fix content-length
8. client sees {"id":"file-<encrypted>", ...}
```

Backend selection is by **model routing**, not by an id (there is none yet). The response id is what
first captures the backend, learned via `SetBackend` on whichever endpoint the LB actually served
(so on a retry/fallback the _last_ attempt's backend wins — `backendFromDecode` is false for upload).

### Retrieve / Content / Delete — id-bearing ops

All three share `handleIDBearingRequestHeaders`; they differ only in method/path suffix and
response handling.

```
1. client → GET /v1/files/file-<enc>            (or /content, or DELETE)
2. R.ProcessRequestHeaders → handleIDBearingRequestHeaders:
     - decoded = codec.Decode(id)   ; if err or kind != file → 404 "No such File object"
     - R.backend* := decoded.ns/name ; backendFromDecode = true
     - schema = schemaForBackend(ns,name) ; if missing → 410 ; resolveTranslator → 501 if non-OpenAI
     - rewrite :path → .../file-<native>            (native id substituted)
     - set x-ai-eg-original-path: <original gateway path>   (for upstream processor resolution)
     - EMIT dynamic metadata  io.envoy.ai_gateway/selected_backend = "ns.name"
     - ClearRouteCache = true
3. Envoy re-matches. The synthesized sticky route "<route>/sticky/ns.name" matches (its
   DynamicMetadata matcher equals selected_backend) and is FIRST → chosen.
   Its RouteAction.MetadataMatch pins the subset LB to that backend's endpoint subset.
4. Request goes ONLY to the owning backend. It sees the native id.
5. Response:
     - retrieve / delete → R.reEncodeResponse: re-encode top-level "id" → gateway id
     - content           → raw bytes, no id envelope → passed through unchanged
```

Here the backend comes from **decoding the id** (`backendFromDecode = true`), so `SetBackend` must
**not** override it — the id is authoritative, the LB choice is irrelevant.

### List — `GET /v1/files` (cross-backend walk)

See §8 for the walk state machine. The request-side flow (`handleListRequestHeaders`):

```
First page (no `after`):
  - modelOverride = ?model=  ; if "" → 400 (required)
  - set x-ai-eg-model = modelOverride ; strip ?model= from upstream :path ; ClearRouteCache
  - (unpinned) → route matches by model ; LB picks the STARTING backend, captured via SetBackend

Subsequent page (`after=<cursor>`):
  - decode cursor:
      · KindListCursor (flcur-) → walk position {start, current, nativeAfter}
      · KindFile (file-)        → continue within that file's backend (stock SDK pagination)
      · anything else / undecodable → 400 "invalid after cursor"
  - PIN current: emit selected_backend = "current.ns.current.name" ; ClearRouteCache
  - rewrite upstream ?after= to the backend-NATIVE cursor ; strip any stray ?model=
```

Response-side (`buildListWalkResponse`) re-encodes every `data[].id`/`first_id`/`last_id` to gateway
ids, computes the next walk step, and sets `has_more` + `last_id` to the next `flcur-` cursor (or, on
the terminal page, the re-encoded native `last_id

## Backend-sticky routing internals

The processor can _decide_ the backend (by decoding an id), but Envoy must physically _send_ there —
and all of a route rule's backends are endpoints of **one Envoy cluster**, indistinguishable by
ordinary routing. The sticky primitive (`sticky_routing.go`, invoked from `PostTranslateModify`)
bridges that gap with three cooperating pieces, all keyed on
`internalapi.AIGatewaySelectedBackendMetadataKey = "selected_backend"` with value
`SelectedBackendMetadataValue(ns, name) = "ns.name"`.

**Piece 1 — tag endpoints** (`tagLbEndpointWithStickyBackend`, in `maybeModifyCluster`).
Every `LbEndpoint` gets, under the well-known `envoy.lb` namespace:
`Metadata["envoy.lb"]["selected_backend"] = "ns.name"`.

**Piece 2 — subset LB policy** (`wrapClusterLbPolicyWithStickySubset`).
The cluster's `load_balancing_policy` is wrapped in Envoy's typed **subset** policy, selector
`[selected_backend]`, child = the cluster's original policy (or round-robin). Crucially,
`fallback_policy: ANY_ENDPOINT` means a request carrying **no** `selected_backend` metadata
load-balances normally — so chat/embeddings, uploads, and first-page lists are unaffected. Idempotent
(a cluster already wrapped is left as is). Note: Envoy rejects the legacy `lb_subset_config` alongside
the modern typed `load_balancing_policy`, so the subset must be expressed via the typed extension.

**Piece 3 — synthesized sticky routes** (`synthesizeStickyBackendRoutes`, after the InferencePool
pass so clones inherit ext_proc per-filter config). For each backend of each AI-Gateway route, the
original route is cloned into `<name>/sticky/ns.name` that:

- **matches only** when the request's `selected_backend` **dynamic metadata** (namespace
  `io.envoy.ai_gateway`) equals that backend value,
- keeps the original **path** match but drops header/query matches,
- pins the subset LB via `RouteAction.MetadataMatch["envoy.lb"]["selected_backend"] = "ns.name"`.

Sticky routes are **prepended** so they win the re-match, and are synthesized **even for a single
backend** (an id-bearing request carries no `x-ai-eg-model`, so after the route cache is cleared it
would otherwise match no model-based route). `collectStickyBackends` builds the input map purely from
the Piece-1 endpoint tags, keeping the three pieces consistent.

PR #1982

Documentation in this PR: user guide at `site/docs/capabilities/llm-integrations/files-api.md`.

Scope and behavior notes:

- **OpenAI-only today; other providers fail closed with `501`.** The translator seam is deliberately
  provider-neutral so Bedrock/Azure/GCP/etc. can be added later; this PR ships only the OpenAI
  identity translators.

- **`model` is required and is a routing selector, not a content filter.** Files carry no model, so
  `?model=X` (list) / the upload `model` field selects the *route/backends* that serve X and lists
  *all* of their files — it cannot filter files by the model they were "created under". Upload and
  list without a model now return `400` (previously they best-effort fell back; see commit
  `0f3f34a2`, marked `feat!`).

- **Known limitation — same model on two AIGatewayRoutes.** Both upload and list route by the
  `x-ai-eg-model` header, so if two routes declare the same model, Envoy's first-match-wins picks
  one and the other route's files are not listable. The supported configs are one route with multiple
  `backendRefs`, or distinct models per route.

- **Batch-ready foundation, but no Batch code ships here.** The codec already supports `batch-` ids
  and the translator is the generic `Translator[ReqT, any]`; a future Batch API reuses both. No
  `/v1/batches*` handlers are included.

- **Content responses are raw bytes** and are never passed through the translator body mapping; the
  content endpoint has its own dedicated translator type (`openAIFileContentTranslator`) used only
  for the schema gate and future specialization.

- **`selected_backend` dynamic-metadata / subset-LB approach**: reviewers familiar with Envoy may
  want to focus on `internal/extensionserver/sticky_routing.go` and `post_translate_modify.go` — the
  subset policy is expressed via the modern typed `load_balancing_policy` (the legacy
  `lb_subset_config` cannot be combined with it).

Testing: unit tests cover the codec (round-trip, tamper, rotation, kind cross-check), the list-walk
state machine, the required-model behavior for both endpoints, the `?model=` override and
model-stripping, and the per-op translator schema gate (OpenAI identity vs. `501` for all other
providers). `make precommit` passes (build, unit tests, lint/format, no generated diff), and the
Docusaurus docs site builds cleanly (`npm run build`, no broken links). The end-to-end
`tests/data-plane` Files flow against two OpenAI backends is recommended as a follow-up verification.


internal/extensionserver/sticky_routing.go
```diff
@@ -0,0 +1,263 @@
...
+					Name:        roundRobinLbPolicyName,
+					TypedConfig: rrAny,
+				},
+			}},
+		}
+	}
+
+	subsetAny, err := toAny(&subsetv3.Subset{
+		// No selection metadata on a request => fall back to load balancing across all endpoints.
+		FallbackPolicy: subsetv3.Subset_ANY_ENDPOINT,
```
yuzisun, line 104: What LB algorithm is used with fallback, least request or round robin ?

internal/extensionserver/sticky_routing.go
```diff
@@ -0,0 +1,263 @@
...
+		},
+		LocalityWeightAware: true,
+		ScaleLocalityWeight: true,
+		SubsetLbPolicy:      childPolicy,
+	})
+	if err != nil {
+		return fmt.Errorf("failed to marshal Subset to Any: %w", err)
+	}
+
+	cluster.LoadBalancingPolicy = &clusterv3.LoadBalancingPolicy{
```
yuzisun, line 116: @nacx are you ok with this approach to set the load balancing policy to subset LB for sticky routing with the fallback policy to any point? we might need to add a check for doing this only for sticky routes.

internal/extensionserver/sticky_routing.go
sivanantha321: Subset_ANY_ENDPOINT means if no subset is matched (chat and other endpoints),  ignore the selector entirely and pick from the cluster's full endpoint set, using the child policy (SubsetLbPolicy, line 110). If there is already a Lb policy present, then it will be used. If not present it defaults to round robin. So the pick is: priority level selected first by Envoy's normal priority/health logic, then locality chosen by
  weight, then endpoint chosen within that locality by provided Lb policy or round robin.

sivanantha321: This is an example envoy config [envoy-sticky-config.json](https://github.com/user-attachments/files/30969098/envoy-sticky-config.json)

Issue #2327, opened 2026-07-05
Scale-from-zero / hibernate-aware routing: retry with wake trigger instead of immediate 503


mohitgurnani1:
_Description_:

GPU-backed model deployments are increasingly scaled to zero replicas (KEDA, cluster autoscalers) or hibernated (suspended GPU pools, VM hibernate/resume) when idle, since idle accelerator capacity dominates self-hosted inference cost. Today, when a request arrives for such a backend, the gateway returns an immediate error: a connect failure for a Service/endpoint with nothing listening, or a local `503 no_healthy_upstream` for a cluster with zero endpoints. Every client is left to implement its own retry-and-wake logic.

Proposed behavior: for an `AIServiceBackend` explicitly marked as scale-from-zero capable (e.g. a new optional `coldStart` field), the gateway should

1. keep the (already buffered) request in flight using an injected Envoy route retry policy (`connect-failure`, `reset`, `retriable-status-codes: 503`) sized to a configurable `maxWait`, delivering the request once the backend is up;
2. optionally fire a user-provided wake-up HTTP endpoint (VM hibernate-resume API, custom GPU-pool controller) asynchronously and deduplicated (singleflight + cool-down), so a burst of requests produces one wake call;
3. after `maxWait`, return an OpenAI-shaped error with `Retry-After` instead of the raw Envoy local reply.

No queueing subsystem in the gateway: this composes existing primitives (Envoy retries, the retry-safe upstream ext_proc filter which already re-runs translation/auth per attempt, and endpoint-priority failover to warm backends). For `InferencePool` backends this is out of scope, since the Gateway API Inference Extension's Endpoint Picker Flow Control layer already queues requests during scale-from-zero; this targets the `AIServiceBackend` path which has no equivalent today.

A local data-plane PoC (real Envoy + extproc + fake upstream) validates feasibility: with such a retry policy, a request issued while nothing was listening on the backend port was delivered with `200` (4.0s end to end) when a listener appeared 3 seconds later; with no wake, the request failed with `503` only after the 9.4s retry budget, with the raw Envoy local reply body ("upstream connect error or disconnect/reset before headers...") that the terminal-error shaping would replace.

Happy to submit a design doc under `docs/proposals/` with the full API sketch, architecture, and phased implementation plan if this direction sounds reasonable.

_Relevant Links_:

[...]

PR #2328, opened 2026-07-05
docs: proposal for scale-from-zero aware routing

mohitgurnani1:
**Description**

This PR adds a design proposal for scale-from-zero / hibernate-aware routing: when an AIServiceBackend's upstream is scaled to zero replicas or hibernated, the gateway keeps the request in flight with a derived Envoy retry policy bounded by a configurable maxWait, optionally fires a deduplicated wake-up HTTP endpoint, and returns an OpenAI-shaped error with Retry-After only after the budget is exhausted, instead of surfacing an immediate connect failure or local 503.

The design deliberately composes existing primitives (Envoy route retry policies, the retry-safe upstream ext_proc filter, endpoint-priority failover) and introduces no request queue in the gateway. InferencePool backends are out of scope since the Gateway API Inference Extension Endpoint Picker Flow Control layer already queues requests during scale-from-zero; this targets the AIServiceBackend path which has no equivalent today.

Feasibility was validated with a local data-plane PoC (real Envoy + extproc + fake upstream, both ext_proc filters in path): a request issued while nothing was listening on the backend port was delivered with 200 (4.0s end to end) when a listener appeared 3 seconds later, and a never-waking backend produced a 503 only after the 9.4s retry budget, with the raw Envoy local reply body that the proposed terminal-error shaping replaces.

The proposal includes the API sketch (a new optional coldStart field on AIServiceBackendSpec), architecture, key design decisions, a phased implementation plan, out-of-scope items, and open questions for discussion.

Disclosure: this proposal was drafted with the assistance of generative AI (Claude); I have reviewed and own the content.

Related Issue: #2327

This is a docs-only PR (a single new file under docs/proposals/). Feedback is sought primarily on the open questions section: retry-policy injection vs documented BackendTrafficPolicy, the 503 retry trigger scope, wake firing heuristics, and naming.


docs/proposals/013-scale-from-zero-routing/proposal.md
```diff
@@ -0,0 +1,284 @@
...
+  retry counts, and budget exhaustion, plus access-log fields.
+
+## Open Questions for Community Discussion
+
+1. Should the extension server inject the retry policy, or should the feature only document a
+   required user-attached `BackendTrafficPolicy` (status quo pattern)? Proposed: inject, with user
+   policy precedence — is that precedence rule acceptable?
+2. Retrying on 503 cannot distinguish a local `no_healthy_upstream` 503 from a genuine upstream 503.
+   For a backend explicitly opted into `coldStart` this seems acceptable (and bounded by `maxWait`);
+   should we instead scope 503 retries with response-flag/header-based gating?
```
nacx: IMO, opting into `coldStart` is acceptable.

docs/proposals/013-scale-from-zero-routing/proposal.md
```diff
@@ -0,0 +1,284 @@
...
+3. Wake firing: fire on every request arrival for a cold-start backend (deduplicated by `coolDown`),
+   or only after a failed first attempt? The former is proposed for simplicity; the latter avoids
+   wake calls when the backend is already warm at the cost of detecting failure first.
```
nacx: I think we should avoid wake-up calls when the backend is already up.

docs/proposals/013-scale-from-zero-routing/proposal.md
```diff
@@ -0,0 +1,284 @@
...
+4. Naming: `coldStart` vs `scaleFromZero` vs `wakePolicy`?
+5. Is a per-route-rule override of `maxWait` needed in v1, or is per-backend sufficient?
```
nacx: I think per-backend is the right place to start.

docs/proposals/013-scale-from-zero-routing/proposal.md
```diff
@@ -0,0 +1,284 @@
...
+6. Standalone `aigw`: inject the retry policy identically (proposed), or gate the feature?
```
nacx: `aigw` should behave the same as k8s based to the extent that it is possible.

docs/proposals/013-scale-from-zero-routing/proposal.md
```diff
@@ -0,0 +1,284 @@
...
+  upstream ext_proc message timeout for cold-start backends.
+- **`wake.scaleRef`**: native Kubernetes scale-subresource wake as an alternative to the HTTP
+  endpoint.
+- **Cold-start observability**: metrics for wake latency (wake call → first successful attempt),
+  retry counts, and budget exhaustion, plus access-log fields.
+
+## Open Questions for Community Discussion
+
+1. Should the extension server inject the retry policy, or should the feature only document a
+   required user-attached `BackendTrafficPolicy` (status quo pattern)? Proposed: inject, with user
```
nacx: This is the main discussion point, I think.
If we take other features, such as provider fallback, we ask users to create the BTP that configures the retries themselves. To keep API consistency across features, I'd say we could start by not injecting the BTP and asking users to configure it.

nacx (commented):
Super interesting feature.
Thanks for the proposal! I left some comments for discussion

docs/proposals/013-scale-from-zero-routing/proposal.md
mohitgurnani1: Agreed — this is the one that matters most. I'll switch to a user-configured `BackendTrafficPolicy` (matching the provider-fallback precedent) instead of extension-server injection, for API consistency. Pushed: `maxWait` is now advisory only — it sizes the recommended BTP retry budget in the docs and the terminal error's `Retry-After` value, but the gateway no longer derives or injects any retry policy itself.

docs/proposals/013-scale-from-zero-routing/proposal.md
mohitgurnani1: Agreed, keeping `coldStart`.

docs/proposals/013-scale-from-zero-routing/proposal.md
mohitgurnani1: Agreed — pushed a change so the wake call fires only after the first attempt against a cold-start backend fails, not on every request arrival. Avoids waking an already-warm backend.

docs/proposals/013-scale-from-zero-routing/proposal.md
mohitgurnani1: Agreed, per-backend it is — no per-route-rule override in v1.

docs/proposals/013-scale-from-zero-routing/proposal.md
mohitgurnani1: Agreed. Also worth noting: since retry injection is moving out of the extension server (see the reply on the main discussion point above), there's no k8s-only xDS injection path to replicate for `aigw` anymore, so parity here gets simpler.

mohitgurnani1:
Thanks for the review — replied inline on each point, and agree across the board. Pushed an update to the proposal:

- Retry-hold is now user-configured (a `BackendTrafficPolicy` you attach yourself, same pattern as provider fallback) instead of extension-server-injected. `maxWait` is advisory only now — sizes the recommended BTP shape in the docs and the terminal error's `Retry-After`.
- Wake call now fires only after the first attempt against a cold-start backend fails, not on every request arrival.
- Naming (`coldStart`), per-backend scope, and `aigw` parity all kept as originally proposed, per your comments.

One open question from the original list didn't get a comment and is still genuinely unresolved: retrying on any `503` can't distinguish a local `no_healthy_upstream` from a genuine upstream error. For a backend explicitly opted into `coldStart` I think that's acceptable (bounded by the retry budget either way), but flagging in case you'd rather see it scoped with response-flag/header-based gating instead. Would appreciate a read on that one when you have a chance.

PR #2334, opened 2026-07-07
fix: name AIGatewayRoute HTTPRoute rules

### LINKED ISSUE #1474: Multiple AiGatewayRoute with each API Key
_Description_:

Hello, I'm trying to use multiple AIGatewayRoute (each for one model), with different api key for each of httproute.

An exemple (repeat for each model, mistral/ollama/etc... - change the secret value for each httproute):
```
kind: AIGatewayRoute
metadata:
  name: mistral
spec:
  parentRefs:
    - name: aigwroute
      kind: Gateway
  rules:
    - matches:
        - headers:
            - type: Exact
              name: x-ai-eg-model
              value: mistral-3.2
      backendRefs:
        - group: inference.networking.k8s.io
          kind: InferencePool
          name: mistral
          modelNameOverride: RedHatAI/Mistral-Small-3.2-24B-Instruct-2506-FP8
---
apiVersion: v1
kind: Secret
type: Opaque
metadata:
  name: apikey-secret-mistral
stringData:
  client1: mistral
---
kind: SecurityPolicy
metadata:
  name: apikey-auth-mistral
spec:
  targetRefs:
    - kind: HTTPRoute
      name: mistral
  apiKeyAuth:
    credentialRefs:
      - group: ""
        kind: Secret
        name: apikey-secret-mistral
    extractFrom:
      - headers:
          - x-api-key
```

For testing the access, I'm using the following curl command:
```
curl -H "Content-Type: application/json" \
     -H "x-user-id: user123" \
     -H "x-api-key: <MY API KEY>" \
     -d '{
        "model": "<MY MODEL>",
        "messages": [
            {
                "role": "user",
                "content": "Hi. Say this is a test. Respond me with ahoy captain"
            }
        ]
    }'   http://$GATEWAY_IP/v1/chat/completions -v
```

For each model (ollama/mistral/...) I have to use the same apikey and not the API key associated with the model.

Maybe add a name in the generated HTTPRoute, so  I can use property **sectionName** in  SecurityPolicy.spec.targetRefs ?

_Environment_:

gateway-helm: 1.5.4
ai-gateway-helm: v0.0.0-latest

Aias00:
**Description**

Name generated HTTPRoute rules for AIGatewayRoute resources so Envoy Gateway policies can target a specific rule with targetRefs.sectionName. Single-rule routes use the AIGatewayRoute name as the section name, while multi-rule routes use a stable <route>-rule-<index> name and preserve the existing route-not-found catch-all rule name.

Fixes #1474.

Validation:
- go test ./internal/controller
- go test ./internal/controller -run 'Test_newHTTPRoute|TestAIGatewayRouteHTTPRouteRuleName' -coverprofile=/tmp/ai-gateway-1474-controller-targeted.cover\n- go tool cover -func=/tmp/ai-gateway-1474-controller-targeted.cover | rg 'aiGatewayRouteHTTPRouteRuleName|newHTTPRoute|total:'\n- make lint\n\n`go test ./...` was started but interrupted after several minutes while waiting on the crdcel package to run Helm CRD retrieval; no test failure output was produced before interruption.

### DIFF
--- a/cmd/aigw/testdata/translate_basic.out.yaml
+++ b/cmd/aigw/testdata/translate_basic.out.yaml
@@ -82,7 +82,8 @@ spec:
       group: gateway.networking.k8s.io
       kind: Gateway
   rules:
-    - backendRefs:
+    - name: envoy-ai-gateway-basic-rule-0
+      backendRefs:
         - group: gateway.envoyproxy.io
           kind: Backend
           name: envoy-ai-gateway-basic-openai
@@ -101,7 +102,8 @@ spec:
             value: "/"
       timeouts:
         request: 60s
-    - backendRefs:
+    - name: envoy-ai-gateway-basic-rule-1
+      backendRefs:
         - group: gateway.envoyproxy.io
           kind: Backend
           name: envoy-ai-gateway-basic-aws
@@ -120,7 +122,8 @@ spec:
             value: "/"
       timeouts:
         request: 60s
-    - backendRefs:
+    - name: envoy-ai-gateway-basic-rule-2
+      backendRefs:
         - kind: Service
           name: envoy-ai-gateway-basic-testupstream
           port: 80
--- a/cmd/aigw/testdata/translate_basic.out.yaml
+++ b/cmd/aigw/testdata/translate_basic.out.yaml
@@ -82,8 +82,7 @@ spec:
       group: gateway.networking.k8s.io
       kind: Gateway
   rules:
-    - name: envoy-ai-gateway-basic-rule-0
-      backendRefs:
+    - backendRefs:
         - group: gateway.envoyproxy.io
           kind: Backend
           name: envoy-ai-gateway-basic-openai
@@ -102,8 +101,7 @@ spec:
             value: "/"
       timeouts:
         request: 60s
-    - name: envoy-ai-gateway-basic-rule-1
-      backendRefs:
+    - backendRefs:
         - group: gateway.envoyproxy.io
           kind: Backend
           name: envoy-ai-gateway-basic-aws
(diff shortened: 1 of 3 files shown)


siddharth1036:
@Aias00 I might be a little out of context here, but doesn't the `AIGatewayRouteRule.Name` solve the usecase mentioned in the issue? Just TOL since AIGatewayRouteRule.Name already lets users assign a stable sectionName for policy targeting, auto-naming unnamed rules adds little and introduces foot-guns: the generated names flip on 1→2 rule transitions and shift on reorder/insert, silently detaching any policy that targeted them. It can also collide with a user-set name (e.g. user sets foo-rule-1, generator emits the same for index 1), producing a duplicate sectionName and an invalid HTTPRoute. I'd prefer we steer users to set rule.Name explicitly for targeting rather than rely on unstable defaults.

Aias00:
> @Aias00 I might be a little out of context here, but doesn't the `AIGatewayRouteRule.Name` solve the usecase mentioned in the issue? Just TOL since AIGatewayRouteRule.Name already lets users assign a stable sectionName for policy targeting, auto-naming unnamed rules adds little and introduces foot-guns: the generated names flip on 1→2 rule transitions and shift on reorder/insert, silently detaching any policy that targeted them. It can also collide with a user-set name (e.g. user sets foo-rule-1, generator emits the same for index 1), producing a duplicate sectionName and an invalid HTTPRoute. I'd prefer we steer users to set rule.Name explicitly for targeting rather than rely on unstable defaults.

Good point, thanks. I updated the PR to avoid auto-naming unnamed rules. The generated HTTPRoute now only copies AIGatewayRouteRule.Name when the user sets it explicitly; unnamed rules stay unnamed, and the
  route-not-found catch-all keeps its reserved name. This avoids sectionName drift on insert/reorder and avoids collisions with user-provided names.

Aias00:
Good point, thanks. I updated the PR to avoid auto-naming unnamed rules. The generated HTTPRoute now only copies AIGatewayRouteRule.Name when the user sets it explicitly; unnamed rules stay unnamed, and the route-not-found catch-all keeps its reserved name. This avoids sectionName drift on insert/reorder and avoids collisions with user-provided names.

PR #2338, opened 2026-07-07
feat(helm): add controller security context defaults

walsm232:
**Description**

This commit updates the AI Gateway Helm chart to use hardened pod and container security contexts for the controller Deployment by default.

The controller image already runs as the distroless `nonroot` user, and the controller does not seem to require privileged Linux capabilities or privilege escalation. This change makes those runtime expectations explicit in the chart defaults by configuring the controller to run as UID/GID `65532`, require non-root execution, drop all capabilities, disable privilege escalation, and use the runtime default seccomp profile.

This brings the controller Deployment closer to the security posture used by Envoy Gateway's Helm chart and aligns it with the secure defaults that seem to already be used for the AI Gateway extProc container injected by the controller: https://github.com/envoyproxy/ai-gateway/blob/d63a020f166b16d821a3bbe7fab9492ace6c97f5/internal/controller/gateway_mutator.go#L427-L439

The existing `controller.podSecurityContext` and `controller.securityContext` values remain configurable, so users can override these defaults if needed.

N/A - I didn't open an issue for this but if it needs one let me know!

This intentionally does not set `readOnlyRootFilesystem: true`. I wasn't 100% sure if that would be a safe change to make here but if folks think it'd be ok then I'm happy to add it in this PR. If any of these other settings may potentially cause issues do let me know and I can adjust this, I did some light testing in a Kind cluster (see below).

**Testing**

  ```console
make helm-lint
make helm-test
```

Also installed the local chart into a Kind cluster with Envoy Gateway installed. Verified that the controller pod reached Running/Ready, the webhook and metrics servers started successfully, leader election completed, and the live pod had the expected pod/container security contexts.

Verified the controller metrics endpoint still responds:

```console
kc port-forward deploy/ai-gateway-controller -n envoy-ai-gateway-system 8080:8080
curl localhost:8080/metrics
```


nacx:
Thanks! Can you fix the conflicts?

nacx (approved):

walsm232:
@nacx thanks for the review, fixed now

PR #2347, opened 2026-07-10
docs: clarify QuotaPolicy and token-based rate limiting

### LINKED ISSUE #2340: The difference between QuotaPolicy and token-based rate limiting is not clear.
_Description_:

The comparison information section compares QuotaPolicy by tokens with Rate Limiting **by requests.**. But how is QuotaPolicy different from Rate Limiting by tokens? Another question, why does QuotaPolicy allow specifying the time period of maximum 1 day. Does token-based rate limiting has the same limitation? Can you provide some use case in the documentation where QuotaPolicy should be used instead of rate limiting?

> QuotaPolicy vs. Rate Limiting
QuotaPolicy manages total consumption budgets (for example, 100,000 tokens per hour). This is distinct from [usage-based rate limiting](https://aigateway.envoyproxy.io/docs/capabilities/traffic/usage-based-ratelimiting), which controls request velocity (for example, requests per second). Use QuotaPolicy when you need to cap cumulative token spend across a time window.

0YHR0:
**Description**

- explain the intent split between the platform's upstream provider/model budget and each client's route allowance
- compare configuration, scope, token accounting, client budgets, enforcement semantics, and time windows
- document the dedicated AI Gateway quota rate limit service, Redis dependency, and fail-open default
- clarify the current one-day `QuotaPolicy` limit and the longer windows available through usage-based rate limiting
- keep the cross-reference in the usage-based rate limiting guide consistent

## Why

The existing comparison incorrectly described usage-based rate limiting as request velocity even though it can charge response token metadata against cumulative limits. Both features can count tokens and return 429, but they answer different operational questions and enforce matched limits differently.

The updated guide frames `QuotaPolicy` as a platform budget attached to an `AIServiceBackend`, and usage-based rate limiting as a consumer allowance attached to route traffic.

## User impact

Users can choose the policy that matches who owns the budget, where it applies, how multiple matched limits are enforced, and which time windows are required. Operators also get explicit deployment and failure-mode prerequisites for quota enforcement.

## Validation

- `git diff --check`
- `npm ci`
- `npm run build`
- repository precommit checks on Ubuntu and macOS

## AI assistance

This documentation update was prepared with OpenAI Codex assistance. The statements were checked against the `QuotaPolicy` API and implementation in this repository and the Envoy Gateway API version used by the project.

Fixes #2340

### DIFF
--- a/site/docs/capabilities/traffic/quota-policy.md
+++ b/site/docs/capabilities/traffic/quota-policy.md
@@ -9,11 +9,10 @@ sidebar_position: 6
 `QuotaPolicy` enables token-based quota management for AI inference services in Envoy AI Gateway.
 When all related backend's quota are exceeded, requests are rejected with a `429 Too Many Requests` status code.
 
-:::note QuotaPolicy vs. Rate Limiting
-QuotaPolicy manages **total consumption budgets** (for example, 100,000 tokens per hour). This is
-distinct from [usage-based rate limiting](./usage-based-ratelimiting.md), which controls **request
-velocity** (for example, requests per second). Use QuotaPolicy when you need to cap cumulative token
-spend across a time window.
+:::note QuotaPolicy vs. token-based rate limiting
+Both `QuotaPolicy` and [usage-based rate limiting](./usage-based-ratelimiting.md) can cap cumulative
+token usage over a time window. The difference is the API and scope used to configure that budget,
+not whether tokens or requests are counted. See [Choosing a policy](#choosing-a-policy) for details.
 :::
 
 ## Overview
@@ -25,6 +24,30 @@ Key features of QuotaPolicy:
   computing how much a request burns down a quota.
 - **Client-selector bucket rules** — carve out per-tenant or per-header quotas using request attributes.
 - **Shadow mode** — evaluate quota rules without enforcing them, for safe rollout.
+
+## Choosing a Policy
+
+`QuotaPolicy` and usage-based rate limiting use the same rate limit infrastructure, and both reject
+requests with `429 Too Many Requests` when the applicable token budget is exhausted. Choose between
+them based on where and how you want to express the policy:
+
+| | `QuotaPolicy` | Usage-based rate limiting |
+| --- | --- | --- |
+| Configuration | One AI Gateway policy containing the token cost and quota buckets | `llmRequestCosts` on an `AIGatewayRoute`, plus an Envoy Gateway `BackendTrafficPolicy` |
+| Scope | Models served by one or more `AIServiceBackend` resources | Envoy Gateway rate limit rules and their client selectors |
+| Token accounting | One default or custom CEL cost per model | Multiple metadata keys can track and limit input, output, total, or custom token costs separately |
+| Client budgets | Default and header-selected buckets, with optional shadow mode | Envoy Gateway client selectors and rate limit rules |
+| Time windows | Exactly one second, minute, hour, or day | One second, minute, hour, day, month, or year |
+
+Use `QuotaPolicy` when the budget belongs naturally to a backend and model. For example, a platform
+team can give every tenant a separate daily budget for an expensive model, keep a shared backend
+budget as a fallback, and evaluate a new tenant rule in shadow mode. The policy stays with the
+`AIServiceBackend` even when multiple routes send traffic to it.
+
+Use usage-based rate limiting when you need lower-level Envoy Gateway controls, route-specific token
+cost metadata, separate limits for input and output tokens, or a monthly or yearly window. It is also
+the natural choice when token budgets need to be managed alongside existing Envoy Gateway rate limit
+rules.
 
 ## How It Works
 
@@ -259,6 +282,11 @@ The `duration` field selects the sliding-window size. It must be exactly one of 
 
 The window is fixed-size — arbitrary multiples such as `"30s"` or `"15m"` are **not** valid and will
 be rejected by the CRD schema. Choose the `limit` to express your budget within one of these windows.
+
+The one-day maximum is a current `QuotaPolicy` API restriction, not an inherent limitation of token
+accounting. Token-based rate limiting configured through an Envoy Gateway `BackendTrafficPolicy`
+also supports `Month` and `Year` units. Use [usage-based rate limiting](./usage-based-ratelimiting.md)
+when a token budget needs either of those longer windows.
 
 ## Service-Wide Quota
 
--- a/site/docs/capabilities/traffic/quota-policy.md
+++ b/site/docs/capabilities/traffic/quota-policy.md
@@ -9,10 +9,13 @@ sidebar_position: 6
 `QuotaPolicy` enables token-based quota management for AI inference services in Envoy AI Gateway.
 When all related backend's quota are exceeded, requests are rejected with a `429 Too Many Requests` status code.
 
-:::note QuotaPolicy vs. token-based rate limiting
-Both `QuotaPolicy` and [usage-based rate limiting](./usage-based-ratelimiting.md) can cap cumulative
-token usage over a time window. The difference is the API and scope used to configure that budget,
-not whether tokens or requests are counted. See [Choosing a policy](#choosing-a-policy) for details.
+:::note QuotaPolicy vs. usage-based rate limiting
+`QuotaPolicy` manages the platform's upstream consumption budget: how many tokens the gateway may
+spend against a provider backend and model across all routes and consumers. [Usage-based rate
+limiting](./usage-based-ratelimiting.md) manages consumer usage: how much a client may consume on a
+route. Both can count tokens and return `429 Too Many Requests`, but they answer different questions:
+whether the platform has exhausted its budget for a provider/model, or whether a client has exceeded
+its allowance. See [Choosing a policy](#choosing-a-policy) for details.
 :::
 
 ## Overview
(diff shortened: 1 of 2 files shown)


missBerg:
Thanks for taking this on — the old note was genuinely misleading (usage-based rate limiting does count tokens, not just "request velocity"), so removing that framing is a correct fix. 🙏

That said, I think the new text overcorrects. Saying the difference is "the API and scope used to configure that budget, not whether tokens or requests are counted" reduces the two features to configuration mechanics, and that's not what the implementation says. The two APIs have different *intents*:

- **`QuotaPolicy` is a platform concern**: it manages the gateway's **upstream consumption budget** — how many tokens the platform may spend against a provider backend/model, shared across all routes and consumers. It attaches to `AIServiceBackend`, is enforced after backend selection, and in `Shared` mode a request is denied only when **all** applicable buckets are exhausted. The API is also explicitly designed toward quota-aware routing (routing away from backends with exhausted budgets — see the TODO in `api/v1alpha1/quota_policy.go`).
- **Usage-based rate limiting is a consumer concern**: it governs how much each **client** may consume on a route, keyed by client descriptors, and denies as soon as **any** matched limit is exceeded.

Both count tokens and both return 429, but they answer different questions: "has *the platform* exhausted its budget for this provider/model?" vs. "has *this client* exceeded its allowance?"

I'd love to keep your comparison table (it's useful and the Month/Year point is accurate), but lead with the intent distinction and add the enforcement-semantics difference. I'll leave some inline suggestions.

site/docs/capabilities/traffic/quota-policy.md
```diff
@@ -9,11 +9,10 @@ sidebar_position: 6
 `QuotaPolicy` enables token-based quota management for AI inference services in Envoy AI Gateway.
 When all related backend's quota are exceeded, requests are rejected with a `429 Too Many Requests` status code.
 
-:::note QuotaPolicy vs. Rate Limiting
-QuotaPolicy manages **total consumption budgets** (for example, 100,000 tokens per hour). This is
-distinct from [usage-based rate limiting](./usage-based-ratelimiting.md), which controls **request
-velocity** (for example, requests per second). Use QuotaPolicy when you need to cap cumulative token
-spend across a time window.
+:::note QuotaPolicy vs. token-based rate limiting
```
missBerg: This is the part I'd most like to see reframed. "The difference is the API and scope used to configure that budget" is true mechanically but hides the intent. 

Suggestion to highlight intent more:

    :::note QuotaPolicy vs. usage-based rate limiting
    `QuotaPolicy` manages the **platform's upstream consumption budget** — how many tokens
    the gateway may spend against a provider backend and model, shared across all routes and
    consumers. [Usage-based rate limiting](./usage-based-ratelimiting.md) governs **consumer
    usage** — how much each client may consume on a route. Both count tokens and both return
    `429`, but they answer different questions: "has *the platform* exhausted its budget for
    this provider/model?" versus "has *this client* exceeded its allowance?" See
    [Choosing a policy](#choosing-a-policy) for details.
    :::

site/docs/capabilities/traffic/quota-policy.md
```diff
@@ -26,6 +25,30 @@ Key features of QuotaPolicy:
 - **Client-selector bucket rules** — carve out per-tenant or per-header quotas using request attributes.
 - **Shadow mode** — evaluate quota rules without enforcing them, for safe rollout.
 
+## Choosing a Policy
+
+`QuotaPolicy` and usage-based rate limiting use the same rate limit infrastructure, and both reject
```
missBerg: "the same rate limit infrastructure" isn't quite accurate. Both build on the same *kind* of infrastructure (envoyproxy/ratelimit + Redis), but `QuotaPolicy` is enforced by a **dedicated AI Gateway rate limit service** — its own domain (`ai-gateway-quota`), its own HTTP filter injected in the upstream filter chain, and a separate deployment (`controller.quotaRateLimitServiceAddr` in the helm chart) — while usage-based rate limiting uses Envoy Gateway's global rate limit service. Different deployments, different operational footprint.

Suggest: "Both build on Envoy's global rate limiting (envoyproxy/ratelimit backed by Redis), but `QuotaPolicy` is enforced by a dedicated AI Gateway rate limit service, separate from Envoy Gateway's."

site/docs/capabilities/traffic/quota-policy.md
```diff
@@ -26,6 +25,30 @@ Key features of QuotaPolicy:
...
+requests with `429 Too Many Requests` when the applicable token budget is exhausted. Choose between
+them based on where and how you want to express the policy:
+
+| | `QuotaPolicy` | Usage-based rate limiting |
```
missBerg: Two suggestions for the table:

1. **Add an "Enforcement" row** — this is the most decision-relevant difference and it's currently missing:

```
| Enforcement | Denied only when **all** applicable quota buckets are exhausted (`Shared` mode); checked per selected backend, after routing | Denied as soon as **any** matched limit is exceeded; checked at the route, before backend selection |
```

2. The **Scope** row currently compares topology on one side ("Models served by one or more `AIServiceBackend` resources") with config mechanics on the other ("Envoy Gateway rate limit rules and their client selectors"). Suggest making both sides topological:

```
| Scope | Backend- and model-scoped — the budget follows the `AIServiceBackend` across every route that sends traffic to it | Route/Gateway-scoped — budgets are keyed by client descriptors on a route |
```

site/docs/capabilities/traffic/usage-based-ratelimiting.md
```diff
@@ -9,8 +9,11 @@ import TabItem from '@theme/TabItem';
 
 This guide focuses on AI Gateway's specific capabilities for token-based rate limiting in LLM requests. For general rate limiting concepts and configurations, refer to [Envoy Gateway's Rate Limiting documentation](https://gateway.envoyproxy.io/docs/tasks/traffic/global-rate-limit/).
 
-:::info Quota Policy vs. Rate Limiting
-AI Gateway also provides [Quota Policy](./quota-policy.md) for managing **total consumption budgets** (for example, 100,000 tokens per hour). Use QuotaPolicy when you need to cap cumulative token spend, and usage-based rate limiting (this page) when you need to control **request velocity**.
+:::info Quota Policy vs. token-based rate limiting
```
missBerg: nit: the new titles say "token-based rate limiting" but the feature's doc page is titled "Usage-Based Rate Limiting" and the body text uses "usage-based rate limiting". Let's stick to one term ("usage-based rate limiting") to avoid readers thinking these are three different features.

missBerg (changes requested):

0YHR0:
Thanks for the detailed review. I revised the comparison to lead with the platform-budget versus client-allowance distinction, added the enforcement semantics, clarified the separate quota service, and standardized the usage-based rate limiting terminology. The updates are in 25f1f953.

site/docs/capabilities/traffic/quota-policy.md
0YHR0: Updated the note to lead with the platform's upstream budget versus a client's route allowance. It now states that both can count tokens and return 429 while answering those distinct questions.

site/docs/capabilities/traffic/quota-policy.md
0YHR0: Updated this wording to distinguish the dedicated AI Gateway quota rate limit service from Envoy Gateway's global rate limit service, while retaining the shared Redis-backed rate-limit protocol context.

site/docs/capabilities/traffic/quota-policy.md
0YHR0: Added separate Scope and Enforcement rows. The table now describes backend/model scope versus route/Gateway scope and documents Shared-mode exhaustion semantics alongside matched usage-based limits.

site/docs/capabilities/traffic/usage-based-ratelimiting.md
0YHR0: Standardized the notes to use usage-based rate limiting so the terminology matches the feature page.

missBerg:
@0YHR0 let me take a look at the new version :)

0YHR0:
@missBerg pls take a look :)

0YHR0:
@missBerg All of the earlier suggestions are addressed in 25f1f953, including intent, enforcement semantics, service separation, and terminology. Could you please re-review the current version when you have a moment?

site/docs/capabilities/traffic/quota-policy.md
```diff
@@ -27,27 +30,27 @@ Key features of QuotaPolicy:
 
 ## Choosing a Policy
 
-`QuotaPolicy` and usage-based rate limiting use the same rate limit infrastructure, and both reject
-requests with `429 Too Many Requests` when the applicable token budget is exhausted. Choose between
-them based on where and how you want to express the policy:
+Both policies use Envoy's rate limit protocol and Redis-backed counters, but `QuotaPolicy` is enforced
+by a dedicated AI Gateway rate limit service. Usage-based rate limiting uses Envoy Gateway's global
```
1ek, line 34: This sentence is the missing piece I burned a day on, and I think it needs a pointer — nothing in the docs tells you that "dedicated" means you have to deploy it yourself.

On main (@ddad0ab), QuotaPolicy enforcement expects a envoy-ai-gateway-ratelimit Deployment/Service that no chart creates:

- The extension server injects its own envoy.filters.http.ratelimit/ai-gateway-quota filter → cluster ai_gateway_ratelimit_cluster → envoy-ai-gateway-ratelimit.envoy-gateway-system:8081, domain ai-gateway-quota. The code comment is explicit that the suffix exists "to avoid conflicting with Envoy Gateway's own rate limit filter" (quota_ratelimit.go#L41-L47 (https://github.com/envoyproxy/ai-gateway/blob/main/internal/extensionserver/quota_ratelimit.go#L41-L47)) — so this really is a second RLS process, as the sentence says.
- That address is the default of the controller's --quotaRateLimitServiceAddr (cmd/controller/main.go#L257 (https://github.com/envoyproxy/ai-gateway/blob/main/cmd/controller/main.go#L257)), and the service pulls its config over xDS from the controller on :18002 with node ID envoy-ai-gateway-ratelimit (ratelimit/runner/runner.go (https://github.com/envoyproxy/ai-gateway/blob/main/internal/ratelimit/runner/runner.go)).
- ai-gateway-helm opens port 18002 on the controller Service but has no template for the RLS itself. The only place it exists in the repo is the e2e fixture tests/e2e/testdata/backend_quota_ratelimit.yaml#L160 (https://github.com/envoyproxy/ai-gateway/blob/main/tests/e2e/testdata/backend_quota_ratelimit.yaml#L160). #2214 looks like the chart template for it, but it's still draft.

Meanwhile the Prerequisites block below (unchanged by this PR) still says quota "uses the same infrastructure as usage-based rate limiting" and links to the Envoy Gateway rateLimit addon — which deploys EG's envoy-ratelimit, and as far as I can tell that never sees an ai-gateway-quota request. So you follow the docs, every pod is healthy, and nothing is enforced.

What makes it brutal to debug: --quotaRateLimitFailureModeDeny defaults to false, so with the RLS absent the filter fails open — no 429, no 5xx, no error log, just 200s and an unenforced quota.

Could this PR add a sentence here linking to a deploy step for that service (or land alongside #2214), and fix the Prerequisites block to say Redis is the shared piece but the RLS is separate? Two things I couldn't determine from the code:

1. For a QuotaPolicy-only setup, is the Envoy Gateway rateLimit addon needed at all, or is Redis the only genuinely shared prerequisite?
2. Is the --quotaRateLimitServiceAddr arg on the ratelimit container in that e2e fixture intentional, or leftover from the controller flags?

0YHR0: Addressed in ed3ed602. The prerequisites now state that the Helm chart does not deploy the dedicated QuotaPolicy RLS, link to the E2E deployment example and #2214, clarify that Envoy Gateway's rate-limit addon is not required for a QuotaPolicy-only setup, and document the default fail-open behavior. The --quotaRateLimitServiceAddr argument in the E2E ratelimit container appears to be leftover: it is a controller flag, not an envoyproxy/ratelimit flag, so I did not document it as part of the deployment contract.

0YHR0:
@missBerg @1ek I rebased this PR onto current main and addressed the deployment gap in ed3ed602. The prerequisites now explain the dedicated QuotaPolicy RLS, link to a working manifest and #2214, distinguish it from Envoy Gateway's addon, and call out the default fail-open behavior. The site production build passes. Could you please re-review the updated version?

0YHR0:
@missBerg I have now incorporated the intent and enforcement distinctions, documented the dedicated quota service and fail-open behavior, resolved the three addressed review threads, and updated the PR description to match the final text. All checks are green. Could you please re-review to clear the stale changes-requested state?

0YHR0:
@missBerg All requested wording changes are incorporated and every review thread is resolved, including the deployment prerequisite and fail-open clarification from @1ek. I also checked the recently merged #2536: it updates the quota-aware-routing proposal, while this PR updates the user-facing capability guides, so the work is complementary rather than superseded. Could you please re-review and clear the stale changes-requested state if the current text looks good?

PR #2351, opened 2026-07-12
fix: add per-request backend fan-out subset 

PraveenKumarInjam:
**Description**

Adds an optional spec.backendSelector to MCPRoute that restricts which of a route's backends a given client session may fan out to.

Today: initialize fans out to every backend on the route, for every caller. With backendSelector set: the route can restrict fan-out to a subset of backends per caller — evaluated once per candidate backend when the session starts, before any backend session is opened. It is never re-evaluated for tools/call.

spec:
  backendSelector:                # optional; absent = fan out to all backends (today's behavior)
    defaultAction: Deny
    rules:
      - cel: 'request.mcp.backend in request.auth.jwt.claims.mcp_backends'
        action: Allow

BackendSelector reuses the same CEL authorization engine and rule shape as spec.securityPolicy.authorization (MCPRouteAuthorizationRule.CEL) — same struct, same evaluator — just applied at a different point: once per candidate backend at session-initialize, instead of per-request.

Two ways to write the rule - Same API both times — the difference is only what the CEL expression references.

1. Direct — reference a request attribute you already have (JWT claims, headers, etc.): No extra plumbing. If the route already validates a JWT, this works immediately.
rules:
  - cel: 'request.mcp.backend in request.auth.jwt.claims.mcp_backends'
    action: Allow

2. Shim-backed — reference the trusted x-ai-eg-mcp-backend-subset header:
rules:
  - cel: '("," + request.headers["x-ai-eg-mcp-backend-subset"] + ",").contains("," + request.mcp.backend + ",")'
    action: Allow
    
For deployments where an externally-deployed ext_proc filter ("the shim" — not part of this PR) decides the subset out-of-band and writes it into mcp_backend_subset dynamic metadata. Envoy renders that metadata onto this header on the MCP proxy route via %DYNAMIC_METADATA% with OVERWRITE_IF_EXISTS_OR_ADD, so any client-supplied copy is always discarded — making it safe to reference in a rule. The gateway only renders the header; it never interprets it itself.

Motivation: 
On a large, shared MCPRoute (many backends, multi-tenant), initialize contacts every backend even when a caller only needs a few. That costs latency, load on backends the caller isn't entitled to, and shows up as unnecessary 403s in those backends' own access logs. backendSelector lets the route declare that decision up front — from a JWT claim or a shim-supplied header — without requiring every deployment to stand up an out-of-band shim just to get this.

Changes

- Pls check the Files changes, avoid here to save the space

Testing
go test ./internal/internalapi/... ./internal/mcpproxy/... ./internal/extensionserver/... -count=1 → all ok; go vet ./... and gofmt -l clean.

https://github.com/envoyproxy/ai-gateway/issues/2432

- Backward compatible: with backendSelector unset, every backend is a candidate and fan-out is unchanged from today.
- Evaluated once, not per-call: the selector runs once per candidate backend inside newSession, before any session to that backend is opened — never re-evaluated for the life of the session, and never touches tools/call.
- Deny-by-default, explicit 403: DefaultAction defaults to Deny; if the selector matches zero backends, initialize now returns 403 Forbidden (an authorization decision) rather than 500 (a system failure).
- Header is inert by design: x-ai-eg-mcp-backend-subset is rendered by Envoy from ext-proc-set dynamic metadata with OVERWRITE_IF_EXISTS_OR_ADD, so a client-supplied copy is discarded — but the gateway doesn't interpret it; it's only meaningful if a route's own backendSelector CEL rule chooses to reference it.
- Tool selection intentionally out of scope here — happy to open a follow-up doc/PR showing the ext_authz CEL pattern above for tools/list/tools/call filtering if that'd be useful, but no gateway code change is proposed for it.


internal/extensionserver/mcproute.go
```diff
@@ -186,6 +189,32 @@ func (s *Server) maybeUpdateMCPRoutes(routes []*routev3.RouteConfiguration) {
...
+	}
+}
+
+// dynamicMetadataHeader builds a request header whose value is sourced from the dynamic
+// metadata key under internalapi.InternalEndpointMetadataNamespace using Envoy's
+// %DYNAMIC_METADATA% command operator. OVERWRITE ensures any client-supplied copy is
+// replaced by the trusted value.
+func dynamicMetadataHeader(header, metadataKey string) *corev3.HeaderValueOption {
+	return &corev3.HeaderValueOption{
+		AppendAction: corev3.HeaderValueOption_OVERWRITE_IF_EXISTS_OR_ADD,
```
kanurag94: keep_empty_value defaults to false here, so when the metadata is absent Envoy renders an empty value and then skips the mutation entirely - the OVERWRITE never runs and a client-supplied header survives. It's the `if (!value.empty() || add_if_empty_)` guard in header_parser.cc. And since nothing sets these metadata keys in-tree yet, absent is the only case today, so right now the header is fully client-controlled.

For backend-subset that's harmless (a client can only narrow its own fanout). But tool-subset lets a client override the static selector and see tools it's supposed to be hidden from. Setting keep_empty_value: true on both makes the empty render overwrite to "", and the proxy falls back to static. Worth an e2e too - nothing here actually exercises the render/omit path through a real Envoy.

internal/mcpproxy/handlers.go
```diff
@@ -1682,10 +1685,19 @@ func (m *mcpRequestContext) mergeToolsList(s *session, responses []broadCastResp
...
+	// Dynamic per-request tool filter (x-ai-eg-mcp-tool-subset), supplied by the trusted
+	// shim. When present it takes precedence over the static per-backend toolSelector;
+	// when absent we fall back to the static selector.
+	dynamicTools := toolSubset(m.requestHeaders)
 	for _, r := range responses {
 		selector := route.toolSelectors[r.backendName]
 		for _, tool := range r.res.Tools {
-			if selector != nil && !selector.allows(tool.Name) {
+			full := downstreamResourceName(tool.Name, r.backendName)
+			if dynamicTools != nil {
```
kanurag94: When the header is present this drops the static selector entirely instead of intersecting with it. So a tool that's in the dynamic set but not allowed by the static selector still shows up in tools/list, then 400s on tools/call - the StaticSelectorOnly test kind of documents that mismatch. It also means the dynamic set can surface tools the static selector was hiding.

Could we intersect instead (selector allows it AND it's in the dynamic set)? That way we never list a tool the caller can't actually call, and the dynamic header can only narrow, never widen.

internal/mcpproxy/mcpproxy.go
```diff
@@ -163,9 +200,22 @@ func (m *mcpRequestContext) newSession(ctx context.Context, p *mcp.InitializePar
 
 	forwardHeaders := extractForwardHeaders(m.requestHeaders, backends.forwardHeaders)
 
+	// Optional per-request fan-out subset. x-ai-eg-mcp-backend-subset header, initialize sessions ONLY to that subset of the route's backends.
+	// If the header is absent, fall back to ALL backends on the route (original behavior — this change is backward compatible).
+	selectedBackends := backends.backends
+	if subset := backendSubset(m.requestHeaders); subset != nil {
```
kanurag94: If the subset only contains names that aren't on the route, selectedBackends ends up empty and newSession fails with "failed to create MCP session to any backend". So a stale or wrong subset from the shim takes the request down instead of just degrading. Intended? If so a more specific error would help; otherwise maybe log the unknown names and fall back to all.

kanurag94 (commented):
Had a proper look, mostly at the metadata -> header injection side. The keep_empty_value one on mcproute.go is the only thing I'd want sorted before this merges; the other two are smaller.

internal/extensionserver/mcproute.go
PraveenKumarInjam: Agreed on this part, we should set  keep_empty_value: true in all cases, so that if ext proc doesn't set header that will still over write to "" and doesn't use the client modified ones

internal/mcpproxy/mcpproxy.go
PraveenKumarInjam: We can log them and keep error more specific, the reason is in case of dynamic MCP config - we will be having 1 MCP route and 1000 MCP backends, this might result in hitting un intended MCP's or Eager fanout, instead if we throw the error, their is chance that users identify and correct the config or data that is fed to custom ext proc. Also the assumption is dynamic auth forward headers is sent along with MCP backend subset, if non of the backends are true then there is no auth to connect to rest of backends

internal/mcpproxy/handlers.go
PraveenKumarInjam: Yeah, we need to have a intersection of both like cloud providers ACL's implement, if there is a deny at any layer it should be deny only.

PraveenKumarInjam: This is what I am updating to

 
image

internal/extensionserver/mcproute.go
PraveenKumarInjam: Fixed now

internal/mcpproxy/handlers.go
PraveenKumarInjam: Fixed now

internal/mcpproxy/mcpproxy.go
PraveenKumarInjam: Fixed now

internal/mcpproxy/handlers.go
kanurag94: I think the purpose here is to let the shim hand each tenant their own slice of a shared route's tools, basically a subset of what the route already exposes.

what's happening right now is more like an override though: the dynamic list can surface a tool that's not in the static `include`, but tools/call still only checks the static selector. so that tool shows up in the list and then 400s when you actually call it (and the tool you *can* call ends up hidden if it's not in the dynamic set).

IMO the dynamic subset should just narrow within what static already allows (`allows(t) && inDynamic`), and we run the same check on both list and call. then you can always call what you see, static stays the ceiling, and if someone wants the shim to fully own the list they just leave `include` empty. it's literally called `subset` so it kinda should act like one.

PraveenKumarInjam: Updated the details. It is now a intersection

kanurag94 (approved):
thanks for this!
I think it'd be worth having a e2e for this in future, given that we setup infra for injecting dynamic metadata.

api/v1alpha1/mcp_route.go
```diff
@@ -102,6 +102,19 @@ type MCPRouteSpec struct {
...
+
+	// BackendSelector restricts which of this route's backends a given request may fan
+	// out to, evaluated once per candidate backend when a client session is initialized.
+	// If unspecified, all backends on the route are considered.
+	//
+	// Why: on a route with many backends serving different callers, fanning out to
+	// every backend on every session wastes connections and is also unnecessarily cause
+	// 403 in MCP server acess logs. BackendSelector lets a route declare that decision
+	// with a CEL rule instead of requiring an ext-proc shim to do it out of band.
+	//
```
kanurag94:
```suggestion
	// BackendSelector restricts which of this route's backends a given request may fan
	// out to, evaluated once per candidate backend when a client session is initialized.
	// If unspecified, all backends on the route are considered.
	//
```

api/v1alpha1/mcp_route.go
```diff
@@ -393,6 +406,56 @@ type MCPRouteAuthorizationRule struct {
...
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default:=Deny
+	// +optional
+	DefaultAction *egv1a1.AuthorizationAction `json:"defaultAction,omitempty"`
+
+	// Rules defines a list of backend selection rules.
+	// These rules are evaluated in order, the first matching rule will be applied,
+	// and the rest will be skipped.
+	//
+	// If no rules are defined, the default action will be applied to all backends.
```
kanurag94: Can you please add what happens if a rule doesn't match

api/v1alpha1/mcp_route.go
```diff
@@ -393,6 +406,56 @@ type MCPRouteAuthorizationRule struct {
...
+	//
+	// +kubebuilder:validation:MaxItems=32
+	// +optional
+	Rules []MCPBackendSelectorRule `json:"rules,omitempty"`
+}
+
+// MCPBackendSelectorRule defines a single backend selection rule or single CEL rule.
+type MCPBackendSelectorRule struct {
+	// CEL Evaluated once per candidate backend.  The expression must return a boolean;
```
kanurag94:
```suggestion
	// CEL Evaluated once per candidate backend. The expression must return a boolean;
```

api/v1beta1/mcp_route.go
```diff
@@ -102,6 +102,19 @@ type MCPRouteSpec struct {
...
+
+	// BackendSelector restricts which of this route's backends a given request may fan
+	// out to, evaluated once per candidate backend when a client session is initialized.
+	// If unspecified, all backends on the route are considered.
+	//
+	// Why: on a route with many backends serving different callers, fanning out to
+	// every backend on every session wastes connections and is also unnecessarily cause
+	// 403 in MCP server acess logs. BackendSelector lets a route declare that decision
+	// with a CEL rule instead of requiring an ext-proc shim to do it out of band
+	//
```
kanurag94:
```suggestion
	// BackendSelector restricts which of this route's backends a given request may fan
	// out to, evaluated once per candidate backend when a client session is initialized.
	// If unspecified, all backends on the route are considered.
	//
```

internal/mcpproxy/mcpproxy.go
```diff
@@ -163,9 +165,30 @@ func (m *mcpRequestContext) newSession(ctx context.Context, p *mcp.InitializePar
...
+		filtered := make(map[filterapi.MCPBackendName]filterapi.MCPBackend, len(backends.backends))
+		for name, backend := range backends.backends {
+			allowed, _ := m.authorizeRequest(backends.backendSelector, &authorizationRequest{
+				Headers: m.requestHeaders,
+				Backend: name,
+			})
+			if allowed {
+				filtered[name] = backend
+			}
+		}
```
kanurag94, line 186: We can reuse the optimisation similar to `mergeToolsList` inside `handlers.go`

Currently the cost is added at two parts:
1.  bearerToken + jwt.NewParser().ParseUnverified + extractScopes — base64 decode plus JSON unmarshal of the JWT, per backend.
2. buildCELActivation copying every header into two fresh maps per backend

Only request.mcp.backend really varies per iteration.

So what if we create a small struct:

```
type authzContext struct {
      claims     jwt.MapClaims
      scopes     sets.Set[string]
      activation map[string]any
      mcp        map[string]any // alias of activation["request"]["mcp"]
}
```

and fill it up per request

```
func (m *mcpRequestContext) newAuthzContext(req *authorizationRequest) *authzContext {
      scopes := sets.New[string]()
      claims := jwt.MapClaims{}

      token, err := bearerToken(req.Headers.Get("Authorization"))
      // Envoy verifies the JWT before it reaches here; this is only a sanity check.
      if err != nil {
              m.l.Info("missing or invalid bearer token", slog.String("error", err.Error()))
      } else if _, _, err := jwt.NewParser().ParseUnverified(token, claims); err != nil {
              m.l.Info("failed to parse JWT token", slog.String("error", err.Error()))
      } else {
              scopes = sets.New(extractScopes(claims)...)
              delete(claims, "scope")
              delete(claims, "scp")
      }
      return &authzContext{claims: claims, scopes: scopes}
}

// activationFor builds the request-invariant activation on first use, then rewrites only
// request.mcp on later calls.
func (a *authzContext) activationFor(req *authorizationRequest) map[string]any {
      if a.mcp == nil {
              a.activation, a.mcp = buildCELActivation(req, a.claims, a.scopes) // now returns both
              return a.activation
      }
      a.mcp["method"] = req.MCPMethod
      a.mcp["backend"] = req.Backend
      a.mcp["tool"] = req.Tool
      a.mcp["params"] = normalizeParams(req.Params)
      return a.activation
}
```

authorizeRequest keeps its signature and becomes a wrapper, so the two existing call sites are untouched:

```
func (m *mcpRequestContext) authorizeRequest(authorization *compiledAuthorization, req *authorizationRequest) (bool, []string) {
      if authorization == nil {
              return true, nil
      }
      if len(authorization.Rules) == 0 {
              return authorization.DefaultAction == filterapi.AuthorizationActionAllow, nil
      }
      return m.authorizeRequestWith(authorization, req, m.newAuthzContext(req))
}
```

authorizeRequestWith — keep the existing var celActivation map[string]any laziness so it's still created at most once per invocation:

```
      if celActivation == nil {
              celActivation = ac.activationFor(req)   // was: buildCELActivation(req, claims, scopeSet
      a.mcp["backend"] = req.Backend
      a.mcp["tool"] = req.Tool
      a.mcp["params"] = normalizeParams(req.Params)
      return a.activation
```

Then in newSession:

```
      authzCtx := m.newAuthzContext(&authorizationRequest{Headers: m.requestHeaders})
      for name, backend := range backends.backends {
              allowed, _ := m.authorizeRequestWith(backends.backendSelector, &authorizationRequest{
                      Headers: m.requestHeaders,
                      Backend: name,
              }, authzCtx)
              if allowed {
                      filtered[name] = backend
              }
      }

```

kanurag94 (changes requested):
This is in very good shape now. Thanks for working on this 

There are three things:
1. `make apidoc` is missing.
2. An optimisation I have left inline comment for.
3. A new regression because previously depended on assumption that every backend has a session attached to it.

Regarding 3:

getBackendForRoute() succeeds but now it doesn't guarantee getCompositeSessionEntry() will be nil or not.

So with nil backend currently the places that handle it return 400 (we can check if it is easy to return 403 here instead so agent doesn't retry).
And places which do not handle it currently will panic. These are `handleCompletionComplete` and `handleClientToServerNotificationsProgress` when tracing is enabled. We need to handle this.

kanurag94 (approved):
LGTM, can you please update PR description for other reviewers to easily understand the API being added.

johnugeorge (approved):

internal/controller/gateway_test.go
```diff
@@ -3005,6 +3005,68 @@ func Test_mcpConfig_ForwardHeaders(t *testing.T) {
...
+						BackendObjectReference: gwapiv1.BackendObjectReference{Name: gwapiv1.ObjectName("backend")},
+					}},
+					BackendSelector: &aigv1b1.MCPBackendSelector{},
+				},
+			},
+		}
+
+		mc, _ := mcpConfig(mcpRoutes)
+		require.Equal(t, filterapi.AuthorizationActionDeny, mc.Routes[0].BackendSelector.DefaultAction)
+		require.Empty(t, mc.Routes[0].BackendSelector.Rules)
```
aishwaryaraimule21, line 3069: @PraveenKumarInjam can you please fix this?

aishwaryaraimule21 (approved):
LGTM. I prefer this approach over https://github.com/envoyproxy/ai-gateway/pull/2415 for the following reasons:

1. It's simpler and keeps the concerns i.e. backend selection and other tool level authorization rules separate. 
2. Ambiguities problem in PR 2415: The ambiguity handling is defeating the purpose of the PR. The backend selection in PR 2415 will only work if there are backend-only authz rules. 
3. The partial evaluation behaviour is not intuitive for the user.

PR #2358, opened 2026-07-13
fix(translator): preserve tool result cache control

johnpanos:
**Description**

This commit preserves message-level `cache_control` fields on OpenAI tool result messages when translating requests to Anthropic-compatible backends. Previously, the OpenAI schema dropped the field before translation, so AWS Anthropic requests could not place a cache breakpoint on an outer `tool_result` block.

The translator now prefers the message-level marker and falls back to the existing content-part marker. Unit tests cover string and multipart results, consecutive tool messages, and the compatibility path for content-part markers.

N/A

I raised this PR because prompt caching worked for plain messages, tool definitions, and `tool_use` blocks, but missed when the breakpoint landed on a tool result through the OpenAI-compatible endpoint. The native Anthropic endpoint already preserved this shape.

The E2E matrix checks all five prompt shapes against the in-cluster test upstream and verifies the exact AWS Anthropic request body. It does not call a live provider.

I used Cursor to help investigate and implement this change, and I reviewed and verified the code and tests.

### DIFF
--- a/internal/apischema/openai/openai.go
+++ b/internal/apischema/openai/openai.go
@@ -587,6 +587,8 @@ type ChatCompletionToolMessageParam struct {
 	Role string `json:"role"`
 	// Tool call that this message is responding to.
 	ToolCallID string `json:"tool_call_id"`
+	// Anthropic-compatible cache breakpoint for the translated tool_result block.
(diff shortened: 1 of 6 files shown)


nacx (approved):
Thanks! LGTM

PR #2363, opened 2026-07-14
fix(extensionserver): validate rate limit service address

### LINKED ISSUE #2362: fix(extensionserver): reject malformed quota rate limit service addresses
**Description**
`quotaRateLimitServiceAddr` accepts malformed host:port values when `net.SplitHostPort` fails. The current parser treats every parsing error as a missing port and returns the original string with the default port, which lets an invalid address reach generated Envoy configuration.

**Expected behavior**
The flag should use the default port only for a host without a port (including an IPv6 literal). Malformed host:port values should fail controller startup with a clear error.

**Examples**
- Accepted with the default port: `envoy-ai-gateway-ratelimit`, `2001:db8::1`
- Rejected: `[2001:db8::1`, `ratelimit:8080:9090`

0YHR0:
**Description**
Reject malformed `quotaRateLimitServiceAddr` values instead of treating every `net.SplitHostPort` error as a missing port. Hostnames and bare IPv6 literals continue to use the default port, while malformed addresses now fail fast during extension server startup.

Fixes #2362

**Testing**
- go test ./internal/extensionserver -count=1


0YHR0:
@missBerg pls take a look, thx :)

nacx (approved):

PR #2314, opened 2026-07-02
docs: propose router-phase request mutation via a generic external service

siddharth1036:
**Description**

Adds a design proposal (docs/proposals/012-router-phase-request-mutation/) for calling a generic, schema-agnostic external service during the router ext_proc phase to mutate the request body/headers before the AI Gateway routes.

**Related Issues/PRs (if applicable)**
NA

**Special notes for reviewers (if applicable)**

PR #2364, opened 2026-07-14
docs: propose AIGatewayExtensionPolicy CRD

siddharth1036:
**Description**

This PR adds design proposal 012-aigatewayroutefilter for a new AIGatewayExtensionPolicy CRD that runs a router-phase ext_proc before the AI Gateway makes its routing decision. The policy attaches to AIGatewayRoute(s) via targetRefs and carries a headers gate. At PostTranslateModify the extension server inserts a disabled envoy.filters.http.composite filter ahead of the AI Gateway ext_proc and enables it per-route via a FilterConfig-wrapped CompositePerRoute (gated on the policy's headers), on the targeted route's rules and all catch-all routes. This is a declarative, validated, name-based replacement for the EnvoyPatchPolicy composite-wrap workaround, shrinking the blast radius from "all catch-all traffic" to "only traffic carrying the policy headers".

The proposal documents the end-to-end CRD→xDS flow, the per-route enablement mechanism (why CompositePerRoute, not ExtensionWithMatcher), and the Envoy version requirement: CompositePerRoute over RDS needs Envoy ≥ 1.39 / a build containing PR #43996 (envoyproxy/envoy:distroless-dev); ≤ 1.38.x NACKs the whole RouteConfiguration.

Follow up on: #2314 


siddharth1036:
cc: @johnugeorge @nacx @yanavlasov

PR #2366, opened 2026-07-14
translator: accept thinking_blocks signatures and fall back to skip_thought_signature_validator for Gemini function calls

### LINKED ISSUE #2365: OpenAI→GCPVertexAI: Gemini 3.x rejects tool-call follow-ups (400 missing thought_signature) for standard OpenAI clients
_Description_:

Multi-turn tool calling with Gemini 3.x models through the OpenAI → GCPVertexAI translator is broken for standard OpenAI-schema clients: the first request (which returns `tool_calls`) succeeds, but the follow-up request carrying the tool result is rejected by Vertex with:

```
HTTP 400
{"type":"error","error":{"type":"INVALID_ARGUMENT","code":"400","message":"Function call is missing a thought_signature in functionCall parts. This is required for tools to work correctly, and missing thought_signature may lead to degraded model performance. Additional data, function call `default_api:search_web` , position 2. Please refer to https://ai.google.dev/gemini-api/docs/thought-signatures for more details."}}
```

Gemini 3 requires the [thought signature](https://ai.google.dev/gemini-api/docs/thought-signatures) returned with a `functionCall` to be echoed back on that `functionCall` part in the conversation history. The translator's handling of this is currently asymmetric:

- **Responses** surface the signature as `message.thinking_blocks[].signature` (added in #1995).
- **Requests** (`assistantMsgToGeminiParts`) only read a signature from assistant `content` parts of `"type": "thinking"` (added in #1726). A `thinking_blocks` field on the assistant message is not read — so even echoing the gateway's own response message back verbatim still fails.
- **No fallback**: when no signature is present, the `functionCall` part is sent without `thoughtSignature` and Gemini 3 rejects the whole request.

Standard OpenAI clients (Open WebUI, anything built on the stock OpenAI SDK types) resend the assistant turn as `{"role": "assistant", "tool_calls": [...]}` — the standard schema has no signature field to echo. Net effect: **every tool round-trip with Gemini 3.x fails** (web search, MCP tools, function calling) unless the client is specially adapted to this gateway.

Expected behavior: the round-trip should succeed. Google documents a compatibility escape for exactly this situation — sending the dummy value `skip_thought_signature_validator` skips validation for that part (this is also what LiteLLM does).

_Repro steps_:

Route with `schema: {name: GCPVertexAI}` to Vertex, model `gemini-3.5-flash` (any Gemini 3.x).

Request 1:

```json
{
  "model": "gemini-3.5-flash",
  "messages": [{"role": "user", "content": "What is the latest Angular version? Use search_web."}],
  "tools": [{"type": "function", "function": {"name": "search_web", "description": "Search the web",
    "parameters": {"type": "object", "properties": {"query": {"type": "string"}}, "required": ["query"]}}}]
}
```

Response 200 (note the signature only in `thinking_blocks`):

```json
{"role": "assistant",
 "tool_calls": [{"id": "…", "type": "function", "function": {"name": "search_web", "arguments": "{\"query\":\"latest Angular version\"}"}}],
 "thinking_blocks": [{"type": "thinking", "signature": "AY89a19oMFJ2oM9KpHFs…"}]}
```

Request 2 — append the assistant turn the way any OpenAI-schema client does, plus the tool result:

```json
{"role": "assistant", "tool_calls": [ …verbatim from response… ]},
{"role": "tool", "tool_call_id": "…", "content": "[{\"title\": \"Angular v21\", …}]"}
```

→ HTTP 400 as above. Echoing the full response message including `thinking_blocks` fails identically. Only rewriting the assistant turn into `content: [{"type": "thinking", "signature": "…"}]` — a shape no standard client produces — succeeds.

_Environment_:

Reproduced on `main` (f91bb5b8) via `aigw run`, and with current release images on k8s (Envoy Gateway v1.8.2). Backend: Vertex AI, GCPVertexAI schema, `gemini-3.5-flash` / `gemini-3.1-flash-lite`. Claude models through the same gateway are unaffected.

_Logs_:

400 body above is passed through from Vertex to the client; no extproc errors logged.

Fix proposal (PR incoming): on the request side, (a) also accept `thinking_blocks[].signature` on the assistant message — making the contract symmetric with what the gateway emits, and (b) when no signature is present at all, attach `skip_thought_signature_validator` to the first `functionCall` part.

a-nych:
**Description**

Gemini 3.x requires the thought signature (see [1]) returned with a `functionCall` to be echoed back on that part in subsequent requests; otherwise Vertex rejects the request with `400 INVALID_ARGUMENT` ("Function call is missing a thought_signature in functionCall parts"). The OpenAI → GCPVertexAI translator emits the signature in responses as `message.thinking_blocks[].signature` (#1995) but on the request side only reads it from assistant `content` parts of type `"thinking"` (#1726), and sends nothing when absent. Standard OpenAI-schema clients have no signature field to echo, so every tool round-trip with Gemini 3.x fails.

This PR changes `assistantMsgToGeminiParts` to:

- also accept a signature echoed via the assistant message's `thinking_blocks[].signature` — the same shape the gateway itself emits, so round-tripping the response message now works. Content-part signatures keep priority.
- fall back to Google's documented compatibility value `skip_thought_signature_validator` on the first `functionCall` part when the client echoed no signature at all (matching LiteLLM's behavior). Attached to the first part only, consistent with how Gemini attaches signatures for parallel calls.

`ChatCompletionAssistantMessageParam` gains the `thinking_blocks` field to make the request/response contract symmetric.

Fixes #2365
Related PRs: #1726, #1995

- Four pre-existing fixtures in `gemini_helper_test.go` / `TestOpenAIMessagesToGeminiContents` exercised tool calls without a signature and now expect the dummy value — that's the intended new behavior, not weakened assertions.
- Verified end-to-end with `aigw run` against real Vertex (`gemini-3.5-flash`, `eu`): the signature-less round-trip returns 200 with this patch and 400 on `main`; `thinking_blocks` echo returns 200 and carries the real signature; streaming unaffected. Vertex accepts the dummy value (documented compatibility path, so no hard dependency on Google keeping undocumented behavior).

1: https://ai.google.dev/gemini-api/docs/thought-signatures


sqaisar:
This is an real issue, and we use gemini3 heavily, and always need workarounds for this. Can someone review this?

internal/translator/gemini_helper.go
```diff
@@ -289,8 +309,16 @@ func assistantMsgToGeminiParts(msg *openai.ChatCompletionAssistantMessageParam)
...
-			funcCallPart.ThoughtSignature = thoughtSignature
+		if i == 0 {
+			if thoughtSignature != nil {
+				funcCallPart.ThoughtSignature = thoughtSignature
+			} else {
+				// No signature was echoed back by the client at all (e.g. an OpenAI-schema client
+				// that doesn't round-trip thinking_blocks). Gemini 3.x rejects function calls in
+				// multi-turn requests that are missing a thought_signature, so fall back to
+				// Google's documented compatibility escape.
+				funcCallPart.ThoughtSignature = []byte(dummyThoughtSignature)
```
mathetake: this causes allocation. can you declare `var whatever = []byte("skip_thought_signature_validator")` as a global?

a-nych: sure, addressed in latest commit, also rebased my fork

mathetake (approved):

PR #2371, opened 2026-07-14
feat: add /tokenize endpoint for openai-compatible backend

hustxiayang:
**Description**
This is split from https://github.com/envoyproxy/ai-gateway/pull/1650. Specifically this is to support tokenization for openai-compatible backend.


internal/apischema/openai/tokenize/tokenize.go
```diff
@@ -0,0 +1,184 @@
...
+	}
+	if r.CompletionRequest == nil && r.ChatRequest == nil {
+		return errors.New("one request type must be set")
+	}
+	if (r.CompletionRequest != nil && r.CompletionRequest.Model == "") ||
+		(r.ChatRequest != nil && r.ChatRequest.Model == "") {
+		return errors.New("model is required")
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
hustxiayang: this could not be nil from the codes

internal/tracing/openinference/openai/tokenize.go
```diff
@@ -0,0 +1,168 @@
...
+		addGenerationPrompt := true
+		if req.AddGenerationPrompt != nil {
+			addGenerationPrompt = *req.AddGenerationPrompt
+		}
+		attrs = append(attrs,
+			attribute.String("tokenize.request_type", "chat"),
+			attribute.Bool("tokenize.add_generation_prompt", addGenerationPrompt),
+			attribute.Bool("tokenize.continue_final_message", req.ContinueFinalMessage),
+			attribute.Bool("tokenize.add_special_tokens", req.ChatRequest.AddSpecialTokens),
+		)
```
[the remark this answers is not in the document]
hustxiayang: These fields exist only on `ChatRequest`, so the qualified form is redundant and this repo's staticcheck config  flags it as QF1008 ("could remove embedded field ChatRequest from selector"), which fails make lint.

The reason `AddSpecialTokens` and `ReturnTokenStrs` are accessed as `req.ChatRequest.X` is that they exist on both `CompletionRequest` and `ChatRequest` — the promoted form would be ambiguous and wouldn't compile, so QF1008 doesn't apply to them.

internal/translator/tokenize.go
```diff
@@ -0,0 +1,140 @@
...
+}
+
+// ResponseError implements [TokenizeTranslator.ResponseError].
+// For OpenAI-based backends we return the OpenAI error type as is.
+// If connection fails the error body is translated to OpenAI error type for events such as HTTP 503 or 504.
+func (o *ToOpenAITokenize) ResponseError(respHeaders map[string]string, body io.Reader) (
+	newHeaders []internalapi.Header, newBody []byte, err error,
+) {
+	statusCode := respHeaders[statusHeaderName]
+	if v, ok := respHeaders[contentTypeHeaderName]; ok && !strings.Contains(v, jsonContentType) {
```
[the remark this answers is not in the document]
hustxiayang: followed the suggestion

internal/apischema/openai/tokenize/tokenize.go
```diff
@@ -0,0 +1,184 @@
...
+	MediaResolution genai.MediaResolution `json:"media_resolution,omitempty"`
+}
+
+// Validate checks that the request is valid.
+func (r *ChatRequest) Validate() error {
+	if r.ContinueFinalMessage && r.AddGenerationPrompt != nil && *r.AddGenerationPrompt {
+		return errors.New("cannot set both continue_final_message and add_generation_prompt to true")
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
hustxiayang: "if r == nil {" could not be nil, other than this, followed the suggestion

internal/tracing/openinference/openai/tokenize.go
```diff
@@ -0,0 +1,168 @@
...
+		if req.ChatRequest.ReturnTokenStrs != nil {
+			attrs = append(attrs, attribute.Bool("tokenize.return_token_strs", *req.ChatRequest.ReturnTokenStrs))
+		}
+		if len(req.Messages) > 0 {
+			attrs = append(attrs, attribute.Int("tokenize.message_count", len(req.Messages)))
+		}
+	}
```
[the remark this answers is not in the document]
hustxiayang: This was discussed in the last run.

internal/translator/tokenize.go
```diff
@@ -0,0 +1,140 @@
...
+		return nil, nil, tokenUsage, responseModel, fmt.Errorf("failed to unmarshal body: %w", err)
+	}
+
+	// Fallback to request model for test or non-compliant OpenAI backends
+	responseModel = o.requestModel
+	if span != nil {
+		span.RecordResponse(resp)
+	}
+	return
+}
```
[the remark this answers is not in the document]
hustxiayang: `/tokenize` is metered per request (RPM) rather than per token, so we intentionally don't emit `gen_ai `token-usage metrics here. `/tokenize` performs no inference, so recording its count as token usage would overload that metric. Updated the docs to drop the "token counts" claim so the feature list matches the implementation (request duration only).

internal/apischema/openai/tokenize/tokenize.go
```diff
@@ -0,0 +1,190 @@
...
+	}
+	if r.CompletionRequest == nil && r.ChatRequest == nil {
+		return errors.New("one request type must be set")
+	}
+	if (r.CompletionRequest != nil && r.CompletionRequest.Model == "") ||
+		(r.ChatRequest != nil && r.ChatRequest.Model == "") {
+		return errors.New("model is required")
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
hustxiayang: thanks, would follow it.

internal/endpointspec/endpointspec.go
```diff
@@ -418,6 +421,69 @@ func (RerankEndpointSpec) RedactSensitiveInfoFromRequest(req *cohereschema.Reran
...
+
+	var req tokenize.RequestUnion
+	var model string
+
+	// Check if this is a chat tokenize request (has "messages" field)
+	if _, hasMessages := rawRequest["messages"]; hasMessages {
+		var chatReq tokenize.ChatRequest
+		if err := json.Unmarshal(body, &chatReq); err != nil {
+			return "", nil, false, nil, fmt.Errorf("%w: failed to parse chat tokenize request: %w", internalapi.ErrMalformedRequest, err)
+		}
```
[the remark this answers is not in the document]
hustxiayang: Thanks for the comment! Instead of a separate detector struct, I moved the discrimination into `RequestUnion.UnmarshalJSON`: it uses `gjson` to check whether the payload has a `messages` field — if present, it decodes into `ChatRequest`; otherwise into `CompletionRequest` (which carries prompt). A payload containing both is rejected as ambiguous. `ParseBody` is now a single `json.Unmarshal`, dropping the generic-map parse,  following the existing EmbeddingRequest convention in this package.

internal/tracing/openinference/openai/tokenize.go
```diff
@@ -0,0 +1,168 @@
...
+
+	// Set span kind to LLM since tokenization is an LLM operation
+	attrs = append(attrs,
+		attribute.String(openinference.SpanKind, openinference.SpanKindLLM),
+		attribute.String(openinference.LLMSystem, openinference.LLMSystemOpenAI),
+	)
+
+	// Extract model name from the union
+	var model string
+	if req.CompletionRequest != nil {
```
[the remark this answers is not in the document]
hustxiayang: this is impossible from the codes

internal/tracing/openinference/openai/tokenize.go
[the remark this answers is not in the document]
hustxiayang: this is for lint check as discussed in the previous rounds

internal/translator/tokenize.go
```diff
@@ -0,0 +1,140 @@
...
+// ResponseBody implements [TokenizeTranslator.ResponseBody].
+// OpenAI tokenize responses are passed through unchanged. The response does not contain
+// a model field, so we fallback to the request model for metrics and tracing consistency.
+func (o *ToOpenAITokenize) ResponseBody(_ map[string]string, body io.Reader, _ bool, span tracingapi.TokenizeSpan) (
+	newHeaders []internalapi.Header, newBody []byte, tokenUsage metrics.TokenUsage, responseModel string, err error,
+) {
+	resp := &tokenize.Response{}
+	if err := json.NewDecoder(body).Decode(resp); err != nil {
+		return nil, nil, tokenUsage, responseModel, fmt.Errorf("failed to unmarshal body: %w", err)
+	}
```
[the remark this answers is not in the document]
hustxiayang: this is impossible.

internal/translator/tokenize.go
```diff
@@ -0,0 +1,140 @@
...
+func (o *ToOpenAITokenize) ResponseError(respHeaders map[string]string, body io.Reader) (
+	newHeaders []internalapi.Header, newBody []byte, err error,
+) {
+	statusCode := respHeaders[statusHeaderName]
+	if v, ok := respHeaders[contentTypeHeaderName]; !ok || !strings.Contains(v, jsonContentType) {
+		var openaiError openai.Error
+		buf, err := io.ReadAll(body)
+		if err != nil {
+			return nil, nil, fmt.Errorf("failed to read error body: %w", err)
+		}
```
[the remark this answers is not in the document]
hustxiayang: impossible from the codes

internal/tracing/openinference/openai/tokenize.go
```diff
@@ -0,0 +1,168 @@
...
+
+	return attrs
+}
+
+// buildTokenizeResponseAttributes builds OpenInference attributes from a tokenize response.
+func buildTokenizeResponseAttributes(resp *tokenize.Response, _ *openinference.TraceConfig) []attribute.KeyValue {
+	var attrs []attribute.KeyValue
+
+	// Add tokenization results
+	attrs = append(attrs, attribute.Int("tokenize.token_count", resp.Count))
```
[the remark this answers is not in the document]
hustxiayang: impossible from the codes

internal/apischema/openai/tokenize/tokenize.go
```diff
@@ -0,0 +1,223 @@
...
+
+// CompletionRequest represents a request to tokenize a completion prompt.
+type CompletionRequest struct {
+	// Model is the model to use for tokenization.
+	Model string `json:"model"`
+	// Prompt is the text prompt to tokenize.
+	Prompt string `json:"prompt"`
+
+	// AddSpecialTokens indicates if special tokens (e.g. BOS) will be added to the prompt.
+	// Default is true.
```
aabchoo: Could you explain the thought process behind having CompletionRequest AddSpecialTokens default to `true` whereas in ChatRequest defaults to `false`? It may be a bit confusion for when defining the resource.

internal/apischema/openai/tokenize/tokenize.go
```diff
@@ -0,0 +1,223 @@
...
+// in TokenizerInfoResponse (equivalent to Pydantic's ConfigDict(extra="allow")).
+func (r *TokenizerInfoResponse) UnmarshalJSON(data []byte) error {
+	// First unmarshal into a map to capture all fields
+	var allFields map[string]interface{}
+	if err := json.Unmarshal(data, &allFields); err != nil {
+		return err
+	}
+
+	// Extract known fields
+	if tokenizerClass, ok := allFields["tokenizer_class"]; ok {
```
aabchoo: define a variable instead of using a string

internal/apischema/openai/tokenize/tokenize_test.go
```diff
@@ -0,0 +1,537 @@
...
+func TestCompletionRequest_JSON(t *testing.T) {
+	tests := []struct {
+		name     string
+		input    string
+		expected CompletionRequest
+		wantErr  bool
+	}{
+		{
+			name:  "basic completion request",
+			input: `{"prompt": "Hello world", "model": "gpt-4"}`,
```
aabchoo, line 27: default for CompletionRequest's add_special_tokens is true. is the unmarshall supposed to handle setting the value?

internal/apischema/openai/tokenize/tokenize_test.go
```diff
@@ -0,0 +1,537 @@
...
+				Prompt:           "Hello",
+				Model:            "gpt-4",
+				AddSpecialTokens: boolPtr(true),
+				ReturnTokenStrs:  boolPtr(true),
+			},
+		},
+		{
+			name:  "completion request with defaults",
+			input: `{"prompt": "Hello"}`,
+			expected: CompletionRequest{
```
aabchoo, line 46: Shouldn't this be an error because all Tokenize requests requires a model

https://github.com/envoyproxy/ai-gateway/pull/2371/changes#diff-63f76e4db2a9bcc3965cff4925837489b24446471c37ecf12f1bb33bdb778a98R19

aabchoo: just trying to understand why some validation happens in unmarshal while others in code

internal/apischema/openai/tokenize/tokenize_test.go
```diff
@@ -0,0 +1,537 @@
...
+		})
+	}
+}
+
+func TestRequestUnion_JSON(t *testing.T) {
+	tests := []struct {
+		name         string
+		input        string
+		isChat       bool
+		isCompletion bool
```
aabchoo, line 208: these two are not used?

internal/apischema/openai/tokenize/tokenize_test.go
```diff
@@ -0,0 +1,537 @@
...
+		name     string
+		response Response
+		wantJSON string
+	}{
+		{
+			name: "basic response",
+			response: Response{
+				Count:       10,
+				MaxModelLen: 4096,
+				Tokens:      []int{1, 2, 3, 4, 5},
```
aabchoo, line 257: this example seems wrong? Len of tokens is 5 but count is 10?

internal/apischema/openai/tokenize/tokenize.go
```diff
@@ -0,0 +1,223 @@
...
+		if str, ok := tokenizerClass.(string); ok {
```
aabchoo: is this a required class?

internal/apischema/openai/tokenize/tokenize_test.go
```diff
@@ -0,0 +1,537 @@
...
+func TestTokenizeRequestValidation(t *testing.T) {
+	t.Run("empty prompt should be caught by validation", func(_ *testing.T) {
+		// This highlights the need for validation in CompletionRequest
+		req := CompletionRequest{
+			Prompt: "",
+			Model:  "gpt-4",
+		}
+
+		// Currently no validation exists - this is an area for improvement
+		// TODO: Add validation for empty prompt
```
aabchoo: Could you create a ticket to track these TODOs?

aabchoo: specifically to finish validation + add redaction?

internal/apischema/openai/tokenize/tokenize.go
hustxiayang: a `completion` request tokenizes a raw prompt string, so you normally want the model's special tokens (e.g. `BOS`) added. A `chat` request is rendered through the model's chat template, the special tokens were rendered with it, so it's false. I would add comments on it

internal/apischema/openai/tokenize/tokenize.go
hustxiayang: I do not think we do not need to provide the endpoint of `GET /tokenizer_info`, so I think it's better to just remove it.

internal/apischema/openai/tokenize/tokenize.go
hustxiayang: same as the above. Feel it's better to remove it

internal/apischema/openai/tokenize/tokenize_test.go
hustxiayang: `unmarshal` intentionally leaves it `nil` when the field is absent, which is what this test asserts (expected doesn't set it). `AddSpecialTokens` is a `*bool`, so `nil = "unspecified"` and the backend applies its own default (`true` for completion), while an explicit `true`/`false` is preserved. Forcing `true` at `unmarshal` would lose the `unspecified`-vs-`explicit` distinction

internal/apischema/openai/tokenize/tokenize_test.go
hustxiayang: thanks, would update it!

internal/apischema/openai/tokenize/tokenize_test.go
hustxiayang: Thanks! would use them.

internal/apischema/openai/tokenize/tokenize_test.go
hustxiayang: "some validation happens in unmarshal "this is because we extract the `messages/prompt` to determine the tokenization type. The purpose was not for validation. So it's fine missing model here, but may be better to add it in tests.

internal/apischema/openai/tokenize/tokenize_test.go
hustxiayang: I removed the empty validation tests because it seems that vllm can accept empty messages/prompt.

hustxiayang: `redaction` is an issue not specific for tokenization, i opened an issue for this: https://github.com/envoyproxy/ai-gateway/issues/2387, and would share a ticket offline.

aabchoo (approved):
Overall looks good to me!

PR #2373, opened 2026-07-15
feat: add `name` field to AIGatewayRoute rules

### LINKED ISSUE #2372: Feature Request: Add `name` field to AIGatewayRouteRule
### Description

This PR introduces a `name` field to `AIGatewayRouteRule`.

### Background

Currently, `AIGatewayRoute` does not support custom names for individual rules. As a
result, the generated `HTTPRoute` rules do not have explicit names, which causes Envoy
Gateway to fall back to index-based “magic numbers” such as `httproute/ns/gw/rule/0` for
metrics labels like `envoy_cluster_name` in Envoy Proxy.

This makes it difficult to associate metrics with specific routing rules, especially in
production environments with multiple rules.

### Proposed Changes

- Add an optional `name` field to the `AIGatewayRouteRule` API definition.
- Propagate this `name` field to the generated `HTTPRoute` rule’s `name` field.

By allowing users to customize this field, they can map human-readable names to routing
rules and improve metrics labels with `EnvoyProxy`'s `clusterStatName` field.

### Example

With this change, users can define explicit names for their rules:

```yaml
# Example configuration snippet
spec:
rules:
    - name: chat-completions-rule
    # ... other fields
```

### Why Is a New Field Needed?

One possible approach would be to generate the HTTPRoute rule name from the corresponding
AIGatewayRoute rule.

However, doing so could introduce compatibility issues for existing alerts or PromQL
queries that rely on the old metrics labels containing rule index numbers.

### Relevant Links

- https://gateway.envoyproxy.io/docs/api/extension_types/#proxymetrics

- https://github.com/envoyproxy/ai-gateway/pull/2334

haruyama480:
**Description**

Adds an optional `name` field to `AIGatewayRouteRule` and propagates it to the generated
`HTTPRoute` rule name.

This lets users assign stable, human-readable names to generated HTTPRoute rules instead of
relying on index-based rule labels in downstream metrics.

**Changes**

- Add `spec.rules[].name` to `AIGatewayRoute` v1alpha1 and v1beta1.
- Propagate the rule name to generated `HTTPRoute.spec.rules[].name`.
- Reserve `route-not-found` for the controller-injected fallback rule.
- Validate that rule names are unique within an `AIGatewayRoute`.
- Add CRD CEL coverage for valid, duplicate, and reserved rule names.
- Regenerate CRDs, deepcopy code, and API docs.

**Testing**

- `go test ./internal/controller`
- `go test ./tests/crdcel`

Fixes #2372


aabchoo (approved):
Thanks for the contribution!

PR #2377, opened 2026-07-16
fix(mcpproxy): validate float response IDs

### LINKED ISSUE #2376: fix(mcpproxy): reject malformed float response IDs
## Summary

Malformed floating-point response IDs can panic the MCP proxy while it reconstructs the original JSON-RPC ID.

## Root cause

handleClientToServerResponse hex-decodes the float payload and passes it to inary.LittleEndian.Uint64 without verifying that the decoded value contains eight bytes. A client can supply a syntactically valid ID such as

0YHR0:
**Description**

Reject malformed floating-point MCP response IDs with an HTTP 400 response instead of allowing an out-of-bounds read.

**Root cause**

The decoded float payload was passed to binary.LittleEndian.Uint64 without verifying that it contained eight bytes.

**Related Issue**

Fixes #2376

**Testing**

- go test ./internal/mcpproxy -run '^TestMCPProxy_handleClientToServerResponse$' -count=1


mathetake (approved):

0YHR0:
The only remaining failure is Unit Test (macos-latest). It fails in cmd/aigw before the MCP proxy tests, with port 18002 already in use followed by a watchable.Map closed-map panic. The same unit-test suite passes on Ubuntu, and the changed internal/mcpproxy package passes on both platforms. I do not have permission to rerun the upstream workflow. Could a maintainer please rerun the failed macOS job?

0YHR0:
> The only remaining failure is Unit Test (macos-latest). It fails in cmd/aigw before the MCP proxy tests, with port 18002 already in use followed by a watchable.Map closed-map panic. The same unit-test suite passes on Ubuntu, and the changed internal/mcpproxy package passes on both platforms. I do not have permission to rerun the upstream workflow. Could a maintainer please rerun the failed macOS job?

@mathetake

PR #2378, opened 2026-07-16
feat: add Azure Managed Identity token provider support (revival of #1183)

Sarthak160:
**Description**

This PR adds Azure Managed Identity support to `BackendSecurityPolicy`, so Azure OpenAI backends can be accessed with a managed identity available to the Envoy AI Gateway controller — no client secret to provision or rotate:

```yaml
spec:
  type: AzureCredentials
  azureCredentials:
    # system-assigned managed identity / Azure Workload Identity / default credential chain
    managedIdentity: {}
    # OR a user-assigned managed identity selected by client ID
    # managedIdentity:
    #   clientID: <uami-client-id>
```

How it works:

- A new `azureManagedIdentityTokenProvider` in `internal/controller/tokenprovider` obtains Azure access tokens using the identity in the controller's environment:
  - `managedIdentity: {}` (no client ID) uses `azidentity.DefaultAzureCredential`, which resolves Azure Workload Identity (federated service account tokens injected on AKS), the system-assigned managed identity via IMDS, environment variables, and Azure CLI credentials for local development.
  - `managedIdentity.clientID` selects a user-assigned managed identity, preferring `WorkloadIdentityCredential` when the federated token environment is present and falling back to IMDS-based `ManagedIdentityCredential` (chained), so it works both with AKS Workload Identity and with identities attached to the cluster VMs.
- The token flows through the existing `AzureTokenRotator` machinery unchanged: the controller rotates the token into the generated secret, and the data plane picks it up exactly like the client-secret and OIDC flows. No extproc/filterapi changes.
- API/CRD: `azureCredentials.managedIdentity` is added to both `v1alpha1` and `v1beta1`; `clientID`/`tenantID` become optional, with CEL validation requiring exactly one of `clientSecretRef` / `oidcExchangeToken` / `managedIdentity`, and requiring `clientID`+`tenantID` for the two existing methods while forbidding them for `managedIdentity` (they are unused there). Existing resources are unaffected.
- Also updates the `AI_GATEWAY_AZURE_PROXY_URL` proxy support to apply to the new credential types, adds an example manifest (`examples/basic/azure_openai_managed_identity.yaml`), CEL validation tests, controller tests, and provider unit tests.

**Related Issues/PRs (if applicable)**

Fixes #1180
Related PR: #1183

[...]

PR #2384, opened 2026-07-17
feat: support Azure Workload Identity (keyless) for AzureCredentials

ahmedsiddiqq:
## Description

Adds an ambient **Azure Workload Identity (keyless)** auth mode to `BackendSecurityPolicy` `type: AzureCredentials`.

When neither `clientSecretRef` nor `oidcExchangeToken` is set, the data-plane ext_proc mints an Entra token via `azidentity.WorkloadIdentityCredential` (the projected service-account token), with **no secret to store and no controller-side rotation**. This mirrors the two ambient patterns already in the repo:

- GCP Application Default Credentials (ADC) when neither `credentialsFile` nor `workloadIdentityFederationConfig` is set.
- AWS default credential chain when neither `credentialsFile` nor `oidcExchangeToken` is set.

The token is requested with the Azure Cognitive Services data-plane scope `https://cognitiveservices.azure.com/.default` (the same audience the existing Azure OIDC/client-secret rotator uses), so it authorizes Azure OpenAI inference calls.

This is offered as an alternative shape to #2378 (which adds an explicit `managedIdentity` field and mints the token controller-side with an ARM scope). See the discussion on that PR for the design trade-offs; the two substantive differences are data-plane vs controller-side minting and the token audience.

## Changes

- **CRD** (`api/v1beta1`): relax the `AzureCredentials` validation from "exactly one of clientSecretRef/oidcExchangeToken" to "at most one"; neither-set now selects ambient Workload Identity. Doc-comment updated.
- **filterapi** `AzureAuth`: add optional `clientID`/`tenantID` to pin the managed identity; `accessToken` now optional.
- **Data-plane** `internal/backendauth/azure.go`: keep the static-token path; add the ambient path using `WorkloadIdentityCredential`, honoring `AI_GATEWAY_AZURE_PROXY_URL` for the proxy transport. Adds an unexported seam for injecting a fake `azcore.TokenCredential` in tests.
- **Controller**: skip credential rotation for the ambient case; project `clientID`/`tenantID` into the filter config (no rotated secret).
- Regenerated CRD YAML and API docs; unit tests and CEL validation tests added/updated.

## How to keyless-enable

The pod's ServiceAccount must be federated to the managed identity (annotated with `azure.workload.identity/use`). Then:

```yaml
kind: BackendSecurityPolicy
spec:
  type: AzureCredentials
  azureCredentials:
    clientID: "<managed-identity-client-id>"
    tenantID: "<entra-tenant-id>"
    # no clientSecretRef, no oidcExchangeToken
```

## Testing

- `go build ./...`, `go vet` clean.
- `go test ./internal/backendauth/... ./internal/filterapi/... ./internal/controller/... ./tests/crdcel/...` all pass, including a new ambient-path handler test (fake credential asserts `Authorization: Bearer <token>`), a new gateway mapping case, and a new CEL `azure_workload_identity.yaml` valid fixture.

### DIFF
--- a/api/v1beta1/backendsecurity_policy.go
+++ b/api/v1beta1/backendsecurity_policy.go
@@ -245,10 +245,14 @@ type BackendSecurityPolicyGCPCredentials struct {
 }
 
 // BackendSecurityPolicyAzureCredentials contains the supported authentication mechanisms to access Azure.
-// Only one of ClientSecretRef or OIDCExchangeToken must be specified. Credentials will not be generated if
-// neither are set.
 //
-// +kubebuilder:validation:XValidation:rule="(has(self.clientSecretRef) && !has(self.oidcExchangeToken)) || (!has(self.clientSecretRef) && has(self.oidcExchangeToken))",message="Exactly one of clientSecretRef or oidcExchangeToken must be specified"
+// When neither ClientSecretRef nor OIDCExchangeToken is specified, ambient Azure Workload Identity is
+// used: the data-plane ext_proc mints an Entra token via WorkloadIdentityCredential using the projected
+// service-account token, with no secret and no controller-side rotation. This requires the ServiceAccount
+// to be federated to the managed identity (annotated with azure.workload.identity/use). At most one of
+// ClientSecretRef or OIDCExchangeToken may be set.
+//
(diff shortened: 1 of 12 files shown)


ahmedsiddiqq:
Withdrawing this — closing as a duplicate of #2378, which already covers Azure managed/workload identity for AzureCredentials. Apologies for the noise. Deferring to #2378.

PR #2386, opened 2026-07-17
feat(extproc): serve /anthropic/v1/models in Anthropic model list format

a-nych:
**Description**

Anthropic-native clients (e.g. Claude Code configured with a base URL that includes the /anthropic prefix, matching /anthropic/v1/messages) call /anthropic/v1/models to list models. Only the OpenAI-prefixed /v1/models endpoint was registered as a processor, so this path had no handler at all.

This adds a new Anthropic models processor that reuses the existing per-host model selection logic but serializes the response using Anthropic's model list schema (data/has_more/first_id/last_id, with type/id/display_name/created_at per model) instead of the OpenAI schema. It is registered under the Anthropic endpoint prefix, alongside the existing /anthropic/v1/messages route.

This change was made with the help of Claude Code. I reviewed the generated code, ran make precommit and the full unit test suite locally, and take ownership of the change.

N/A

N/A


johnugeorge:
LGTM
Thanks

johnugeorge (approved):

PR #2392, opened 2026-07-18
fix: strip content-length when switching to streamed mode

### LINKED ISSUE #2391: ext_proc does not strip content-length when switching from BUFFERED to STREAMED mode, truncating SSE responses
### What happened?

When the ai-gateway ext_proc switches its `response_body_mode` from `BUFFERED` to `STREAMED` (via `ModeOverride`) for streaming SSE responses (e.g. `/v1/responses` with `stream: true`), it does not remove the `content-length` header from the response. If the upstream includes `content-length` (e.g. computed by a prior HTTP/2 hop), Envoy truncates the response to the first body chunk, cutting off the SSE stream after the first few events.

### What did you expect to happen?

The ext_proc should strip `content-length` when switching to STREAMED mode, since `content-length` is incompatible with chunked/streamed HTTP responses.

### How to reproduce

1. Set up a two-hop topology: an Envoy gateway with the EPP from `gateway-api-inference-extension` as ext_proc, fronted by a public Envoy gateway running the ai-gateway ext_proc
2. The EPP copies `transfer-encoding: chunked` from the upstream vLLM response as a `SetHeader` mutation
3. The inference cluster's Envoy computes `content-length` from the first body chunk on the HTTP/2 hop to the public gateway
4. Send a streaming request to `/v1/responses` with `stream: true`
5. The client receives only the first 2 SSE events (`response.created` + `response.in_progress`), the rest is truncated

### Environment

- ai-gateway: v1.0.0
- gateway-api-inference-extension: v1.5.0
- Envoy: v1.37.1

### Root cause

In `internal/extproc/processor_impl.go`, `ProcessResponseHeaders` switches to STREAMED mode via `ModeOverride` but does not remove `content-length`:

```go
var mode *extprocv3http.ProcessingMode
if u.parent.stream && u.responseHeaders[":status"] == "200" {
    mode = &extprocv3http.ProcessingMode{ResponseBodyMode: extprocv3http.ProcessingMode_STREAMED}
}
headerMutation, _ := mutationsFromTranslationResult(newHeaders, nil)
// ← no RemoveHeaders here
```

### Suggested fix

```go
if mode != nil {
    headerMutation.RemoveHeaders = append(headerMutation.RemoveHeaders, "content-length")
}
```

### Additional context

This only manifests when the upstream sends `content-length` for a streaming response. Most ai-gateway users connect directly to OpenAI/Azure, where SSE responses don't include `content-length`. It requires a prior hop (e.g. an EPP ext_proc that copies `transfer-encoding: chunked`) to trigger the issue.

pyhp2017:
**Description**

Strips `content-length` from the response header mutation when the ext_proc switches from BUFFERED to STREAMED mode (via `ModeOverride`). Without this, a `content-length` header from the upstream causes Envoy to truncate the streaming response to the first body chunk.

### Security Impacts

- [x] No, this change does NOT impact security

---

### What does this PR do?

Strips `content-length` from the response header mutation when the ext_proc switches from BUFFERED to STREAMED mode (via `ModeOverride`). Without this, a `content-length` header from the upstream causes Envoy to truncate the streaming response to the first body chunk.

### Context

Fixes #2391

When the ai-gateway ext_proc switches its `response_body_mode` from `BUFFERED` to `STREAMED` for streaming SSE responses (e.g. `/v1/responses` with `stream: true`), it does not remove the `content-length` header. If the upstream includes `content-length` (e.g. computed by a prior HTTP/2 hop when an EPP ext_proc copies `transfer-encoding: chunked`), Envoy truncates the response to the first body chunk, cutting off the SSE stream after the first few events.

### Changes

- `internal/extproc/processor_impl.go`: append `content-length` to `RemoveHeaders` when `mode != nil` (i.e. switching to STREAMED mode)
- `internal/extproc/processor_impl_test.go`: update `Test_chatCompletionProcessorUpstreamFilter_ProcessResponseHeaders/ok/streaming` to verify `content-length` is in `RemoveHeaders`

### Testing

- All existing tests pass (`go test ./internal/extproc/... -count=1`)
- Verified on a dev cluster with a two-hop topology (EPP → ai-gateway ext_proc → client): `/v1/responses` streaming now returns all SSE events instead of being truncated to the first 2


nacx (approved):

nacx:
@kanurag94 mind taking a look?

kanurag94 (approved):

pyhp2017:
@nacx @kanurag94 Thank you for your review! 
is it possible to have this in the next release?

Issue #2204, opened 2026-06-08
Support provider-native API schemas as client input (AzureOpenAI, AWSBedrock, GCPVertexAI)


PraveenKumarInjam:
## Summary
Enable AIGatewayRoute to accept requests in provider-native formats (Azure OpenAI, AWS Bedrock, GCP Gemini/Vertex AI) as client input — not just OpenAI/Anthropic/Cohere. 

  ## Motivation

Enterprises operate a large-scale  proxy serving hundreds of application instances. Callers send requests in **provider-native format** (Azure OpenAI, AWS Bedrock, GCP Gemini) — this is baked into client SDKs and cannot be changed.

We want to adopt Envoy AI Gateway for its credential management (`BackendSecurityPolicy`), token-based rate limiting, failover, and GenAI observability metrics. However, the current requirement that all client input must be OpenAI/Anthropic/Cohere format is a **migration blocker** for us and likely for other enterprises with established provider-specific integrations.

### Why not just convert callers to OpenAI format?

  - Hundreds of application teams would need coordinated changes
  - Multi-month cross-organization migration
  - Provider-specific features (Azure `api-version`, Bedrock inference profiles, Gemini safety settings)
    don't always have OpenAI equivalents
  - During migration, we'd need to run parallel infrastructure

### The value proposition

With native input support, organizations can adopt Envoy AI Gateway **incrementally**:
  1. Day 1: Route existing provider-native traffic through AI Gateway — get credential automation,
     metrics, and failover immediately
  2. Later: Optionally migrate callers to OpenAI format for cross-provider failover

  Without it, adoption requires a big-bang migration of all callers first.

  ## Related Issues

  - #1960 — "Support Gemini native client endpoints" (open, no maintainer response)
  - #2165 — "native Gemini `/v1beta/models/<m>:generateContent` client schema" (open, no maintainer response)
  - #847 — "Support Native Anthropic API" (closed/completed — the precedent)
  - #1648 — PR adding `AWSAnthropic` output schema (merged — demonstrates the wiring pattern)

FYI - I generated some of the content using the chat I had with claude, but cross verified my self.

PraveenKumarInjam:
cc: @mathetake , @yuzisun _-( Got your details from other issues.)_ 
I understand this is a different approach from current Envoy AI gateway (Has one single kind of request format), so want to know whether this is any where your radar for future releases

johnugeorge:
[...]

Issue #2165, opened 2026-05-28
Feature request: native Gemini /v1beta/models/<m>:generateContent client schema


jaimeluengo:
### Summary

  Add a `GoogleGenAI` (or similarly-named) `APISchema` so that Envoy AI Gateway can accept client requests in Gemini's native shape:

  ```
  POST /v1beta/models/{model}:generateContent
  POST /v1beta/models/{model}:streamGenerateContent
  ```

  …and translate them to whichever backend schema the matching `AIServiceBackend` is configured for (OpenAI, GCPVertexAI, etc.) — same
  pattern that already exists for `Anthropic` → any backend.

  ### Why

  The `@google/genai` SDK family — which includes the official `gemini-cli`, `langchain-google`, several IDE plugins, and most "switch
  your AI provider" tutorials — only speaks the Gemini-native wire protocol. Today these clients can't talk to an Envoy AI Gateway at all,
   even when the gateway routes to AI Studio or Vertex AI under the hood (which it already supports as a backend via `GCPVertexAI` and the
   in-extproc OpenAI→Gemini translation).

  We verified by reading [`cmd/extproc/mainlib/main.go`](https://github.com/envoyproxy/ai-gateway/blob/v0.6.0/cmd/extproc/mainlib/main.go)
   — the registered client paths are limited to OpenAI (`/v1/chat/completions`, `/v1/completions`, `/v1/embeddings`, `/v1/responses`,
  `/v1/audio/speech`, `/v1/images/generations`, `/v1/models`), Anthropic (`/v1/messages`), and Cohere (`/v2/rerank`). No Gemini-native
  client surface.

  ### What "good" looks like

  | Client request | Already works | After this FR |
  |---|---|---|
  | `POST /v1/chat/completions {model: gemini-...}` | ✓ via OpenAI client schema → AI Studio | ✓ unchanged |
  | `POST /v1/messages {model: gemini-...}` | ✓ via Anthropic client schema → AI Studio | ✓ unchanged |
  | `POST /v1beta/models/gemini-2.5-flash:generateContent` | ✗ HTTP 404 "unsupported path" | ✓ via GoogleGenAI client schema → any backend
   |

  ### Scope sketch

  A complete implementation would need:

  - New `APISchemaGoogleGenAI` constant in `internal/apischema/...`
  - Request body translator: Gemini-native `{contents, generationConfig, systemInstruction, tools[functionDeclarations]}` → internal IR →
  backend-specific shape
[...]

PR #2396, opened 2026-07-20
feat: add Google AI Studio backend with image generation support

kanurag94:
**Description**

Adds Google AI Studio (the Gemini Developer API on `generativelanguage.googleapis.com`)  as a native backend for OpenAI /v1/images/generations.

Google AI Studio can already be reached with the OpenAI schema today: set `prefix: /v1beta/openai` on the AIServiceBackend and use the existing `APIKey` policy. That works end to end for `/v1/images/generations`.

The problem is the response metadata. The OpenAI-compatible endpoint returns only `data` and `model`. No `usage`, and no way to ask for it — its request schema rejects both `stream_options` and `usage` as unknown fields. So every image request meters as zero tokens: no token-based rate limiting, no cost attribution, nothing useful in the access log.

However, the native `generateContent` endpoint does return it:

```json
"usageMetadata": {
  "promptTokenCount": 6, "candidatesTokenCount": 1301, "totalTokenCount": 1307,
  "promptTokensDetails": [{"modality": "TEXT", "tokenCount": 6}],
  "candidatesTokensDetails": [{"modality": "IMAGE", "tokenCount": 1290}]
}
```

**Example**

```yaml
kind: AIServiceBackend
metadata:
  name: googleai
spec:
  schema:
    name: GoogleAIStudio
    version: v1beta
  backendRef:
    name: googleai # a Backend pointing at generativelanguage.googleapis.com:443
    kind: Backend
    group: gateway.envoyproxy.io
---
kind: BackendSecurityPolicy
metadata:
  name: googleai-apikey
spec:
  targetRefs:
    - group: aigateway.envoyproxy.io
      kind: AIServiceBackend
      name: googleai
  type: GoogleAIKey
  googleAIKey:
    secretRef:
      name: googleai-apikey
```

Client request:

```shell
curl -H "Content-Type: application/json" -d '{
    "model": "gemini-2.5-flash-image",
    "prompt": "A watercolor painting of a lighthouse at sunrise"
  }' \
  $GATEWAY_URL/v1/images/generations
```

What the gateway sends upstream:

```http
POST /v1beta/models/gemini-2.5-flash-image:generateContent
x-goog-api-key: <key from the secret>

{"contents":[{"parts":[{"text":"A watercolor painting of a lighthouse at sunrise"}],"role":"user"}],"tools":null,"generationConfig":{"responseModalities":["IMAGE","TEXT"]}}
```

What the client gets back, after `inlineData` is encoded and `usageMetadata` is mapped:

```json
{"created":1785838962,"data":[{"b64_json":"iVBORw0KGgoAAAANSUhEUg..."}],"usage":{"total_tokens":1300,"input_tokens":8,"output_tokens":1292}}
```

This is the backend side only. Accepting native Gemini requests from clients (#1960, #2165, #2204, #2179) is a separate concern and unaffected.

#1960 and #1961 cover native Gemini on Vertex AI. This is the AI Studio / API-key path instead. #1746 (OpenAI image edits) would extend the same translator.

- Scope is image generation only. Chat/completions to native Gemini is not in this PR.
- CodeQL logging fails: To get CI green, that line now logs `set_header_count` instead of the redacted response.


api/v1alpha1/backendsecurity_policy.go
```diff
@@ -98,6 +100,13 @@ type BackendSecurityPolicySpec struct {
 	//
 	// +optional
 	AnthropicAPIKey *BackendSecurityPolicyAnthropicAPIKey `json:"anthropicAPIKey,omitempty"`
+
+	// GoogleAIKey is a mechanism to access Google AI Studio (generativelanguage.googleapis.com) backend(s).
+	// The API key will be injected into the "x-goog-api-key" header.
+	// https://ai.google.dev/api/rest
+	//
+	// +optional
+	GoogleAIKey *BackendSecurityPolicyGoogleAIKey `json:"googleAIKey,omitempty"`
```
gavrissh: As Google have different Auth Mechanism, it is better to clearly call out here like `GoogleAIStudioKey` ?

Just Google AI might confuse some end users

kanurag94: Thanks!
I think I should have named it `GoogleAPIKey`. This injects a `x-goog-api-key` which is a general Google API key convention. It also covers Vertex and is generic enough but I understand calling it `GoogleAI` feels specific to a mechanism

PR #1797, opened 2026-01-21
aigw: split access logs and map request headers

codefromthecrypt:
**Description**

Split stdout/file access logs by request type using CEL over request headers (x-ai-eg-model for LLM, x-ai-eg-mcp-backend for MCP) so MCP-only fields never appear on LLM logs and vice‑versa. This avoids relying on `/mcp` paths, which are not present on backend‑listener requests.

Add OTEL request-header mapping env vars for users: `OTEL_AIGW_REQUEST_HEADER_ATTRIBUTES` (base mapping) and `OTEL_AIGW_LOG_REQUEST_HEADER_ATTRIBUTES` (log override). These are merged and wired through ext_proc and the Envoy Gateway extension server so access logs can include `session.id` without per‑request app changes. MCP uses JSON‑RPC `params._meta` for POST requests, but GET streams have no JSON‑RPC payload, so compose examples also send HTTP headers for access-log mapping.

Rename access‑log keys to the OTEL‑style `gen_ai.*` and `mcp.provider.name` while preserving gateway‑specific metadata fields, add `request.path` to the common access‑log fields, and update examples, fixtures, and docs accordingly. Compose examples pass user/session IDs via compose args so log output can be verified end‑to‑end (append `--debug` to the `aigw` command for verbose logs).

Ensure original downstream paths are preserved in access logs by setting `x-envoy-original-path` and `x-ai-eg-original-path` from the incoming request path across LLM and MCP flows.

**Related Issues/PRs (if applicable)**

Related: #1303

**Special notes for reviewers (if applicable)**

Access log samples from docker compose runs (Envoy stdout):

```
[...]

PR #2398, opened 2026-07-20
fix: always set original-path internal headers in extproc

kanurag94:
**Description**

The extproc router filter writes two "original path" internal headers from the incoming request path: `x-ai-eg-original-path` and `x-envoy-original-path`. The first is set unconditionally, but the second was only set when it was empty. Because of that, a client-supplied or otherwise pre-existing `x-envoy-original-path` survives and shadows the gateway's own value, and downstream logic that keys off the original path (for example the upstream-filter processor lookup on retry) then operates on the wrong path.

This change makes extproc the authoritative writer of the internal headers it owns. `x-envoy-original-path` is now set unconditionally, matching the sibling `x-ai-eg-original-path`. The header is already treated as gateway-owned elsewhere in extproc (the header mutator ignores it so users can't change it through the header mutation API), so writing it unconditionally simply aligns the two headers' behavior.

A unit test is added asserting that pre-existing values on both original-path headers are overwritten with the request path.

The MCP proxy sets both original-path headers with set-if-missing semantics when forwarding to its own upstream; that is a separate hop and is not affected by this change. Since the router filter already wrote `x-ai-eg-original-path` unconditionally, this only brings `x-envoy-original-path` in line with existing behavior rather than introducing new behavior.


kanurag94:
@gemini review

johnugeorge:
Wondering why empty case was added in the first place ?

kanurag94:
> Wondering why empty case was added in the first place ?

I think it was added in #1797 for both `x-ai-eg-original-path` and  `x-envoy-original-path` while before that it used to be always set unconditionally. Then it was removed from `x-ai-eg-original-path` when introducing `setHeaderIfMissing` in mcp proxy but this one was left stale.
I don't see a reason to have it guarded.

nacx (approved):

PR #2401, opened 2026-07-21
feat: add Hostnames to MCPRoute

kanurag94:
**Description**

MCPRoute currently generates host-less HTTPRoutes, so they land in the Gateway's wildcard virtual host. When the same Gateway also has a host-scoped inference route (`AIGatewayRoute.Hostnames` set), that host-scoped virtual host shadows the host-less MCP route and `/mcp` returns 404.

This adds a `Hostnames` field to `MCPRouteSpec`, mirroring the existing `AIGatewayRoute.Hostnames`, and carries it onto every HTTPRoute the MCPRoute controller generates (both the main mcp-proxy route and the per-backend routes). When set, the generated HTTPRoute includes these hostnames so the MCP route lands in the correct per-host virtual host instead of the wildcard vhost.

The field is optional and additive, so existing MCPRoutes are unaffected.

- Tests: a dedicated set/not-set unit test on `newMainHTTPRoute`, plus carry-through assertions on both the main and per-backend HTTPRoutes in the reconcile test.


kanurag94:
@gemini review

nacx (approved):

PR #2297, opened 2026-06-26
feat: add support of response input token endpoint 

hustxiayang:
**Description**
Implements the OpenAI `/v1/responses/input_tokens` endpoint — a token counting API that accepts the same request body as the Responses API and returns `{"input_tokens": <int>}`. 

### Motivation

Codex and other tools need a way to count input tokens before sending a full Responses API request, enabling clients to validate context window fit and estimate cost without incurring an inference call.

Issue #747, opened 2025-06-23
Support more OTEL gen-ai events & spans


ion-elgreco:
*Description*:
I would like to integrate with other observability tools that map to all the OTEL gen-ai conventions. There is already support for 4 metrics, but would like to see spans and events as well.

[optional *Relevant Links*:]
[>Any extra documentation required to understand the issue.](https://opentelemetry.io/docs/specs/semconv/gen-ai/)

mathetake:
i think this is reasonable. cc @codefromthecrypt as a subject matter expert

codefromthecrypt:
ack. I think there are a couple things to consider.

1. spans envoy would make by default if tracing was enabled (this is easy, but not about genai)
2. spans that would be custom for envoy but representative of the gen-ai specs (includes contentious things like log events)

personally, I think the goal is not spans/events, rather a specific tool working that uses them. For example, [Arize Phoenix](https://github.com/Arize-ai/phoenix) (OSS) has an otel span format [OpenInference](https://github.com/Arize-ai/openinference) (also OSS), which allows you to do evals. OpenInference also has practice around MCP which would likely become a feature here soon.

Conversely, the otel semantics are not widely adopted, and the log events api is in flux. For example, most eval solutions don't use any published otel semantic for this reason. [Log events is being changed to logs](https://github.com/open-telemetry/opentelemetry-python/pull/4651) (maybe). Some use older specs of span events, but that api is also in the [process of being deprecated](https://github.com/open-telemetry/opentelemetry-python/issues/4561)!

TL;DR; beyond normal envoy spans, I would suggest using openinference formatting as that solves one of the biggest obs issues in genai: evals.

ion-elgreco:
@codefromthecrypt Interesting, wasn't aware of OpenInference, I see also some other servers like LangFuse supporting it. To me this makes sense, especially since you mention that OTEL is not as defined as I initially thought ^^

So that means we would have to support:

- spans
- traces

using these https://github.com/Arize-ai/openinference/blob/main/spec/semantic_conventions.md

codefromthecrypt:
@ion-elgreco exactly right. I mean technically we would need something to emit the traces, so some plugin possibly wasm. It would need to treat buffering the req/resp as attributes carefully, for example with a hard cut-off size-wise in case someone puts the entire internet in their LLM context ;)

ion-elgreco:
[...]

PR #2405, opened 2026-07-21
feat: add OpenTelemetry GenAI semantic convention for tracing

dergraf:
**Description**

Spans are currently recorded using the OpenInference semantic convention, which predates
OpenTelemetry having its own convention for LLM calls. Users standardizing on OTel-native
tooling cannot consume our spans without translating them, and the gateway is internally
inconsistent: metrics already speak OTel GenAI (`gen_ai.request.model`) while spans speak
OpenInference (`llm.model_name`).

This adds the OTel GenAI conventions as a second, **opt-in** convention selected by
`AI_GATEWAY_TRACING_SEMCONV`. OpenInference remains the default and its output is unchanged
when the variable is unset.

**Why this is proposed now, given issue #747**

Issue #747 asked for OTel GenAI spans and was closed in favor of OpenInference. That call was
made in August 2025 and rested on two things that have since changed: the GenAI span
conventions were unstable, and the log-events API they depended on was in flux. The span
conventions have since settled on attributes rather than events, and the attribute set this
PR emits is pinned against semconv v1.41.0 in tests.

The reasoning in #747 in favor of OpenInference for evals still holds, which is why this does
not replace it. OpenInference stays the default; GenAI is something operators opt into when
their backend expects it.

**Semantic convention selection** (`internal/tracing/semconv.go`)

- Conventions are chosen through a declarative registry rather than being hardcoded in
  `NewTracingFromEnv`, so adding a third is an append.
- `recorderSet` bundles all eleven endpoint recorders, so a partially wired convention is a
  compile error rather than a nil panic at request time.
- An unrecognized value fails startup rather than silently falling back. A gateway quietly
  emitting a convention nobody watches is worse than one that refuses to start.

**GenAI recorders** (`internal/tracing/otelgenai`)

- The eleven endpoints share one generic recorder and differ only in a handful of hooks, so
  span name, span kind, error handling and status are identical by construction.
- Endpoints whose streaming chunks reconstruct a response supply `foldChunks` and reuse the
  unary path, so streamed and non-streamed requests produce the same span structurally rather
  than via two mappings that have to agree.
- `coverage_test.go` asserts per-endpoint hook wiring in a table, so wiring or dropping a hook
  has to be acknowledged there.

**Two behaviors that differ from OpenInference by design**

- Message content, system instructions and tool definitions are opt-in via
  `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT`, per the conventions. This also closes
  a leak on the new path: provider error bodies routinely echo the prompt, so they are no
  longer copied into the span status when capture is off. Reasoning blocks and non-text
  modalities are recorded by part type with no content for the same reason.
- The attribute count limit is now scoped per convention rather than being a property of
  tracing as a whole. GenAI encodes messages as one JSON attribute per direction, so it does
  not need the limit lifted. **The OpenInference predicate is unchanged**, so the behavior
  added in #2064 is preserved exactly, and `TestNewTracingFromEnv_CaptureDisabled` still
  guards it unmodified.

**Provider attribute** (`internal/tracing/tracingapi`, `internal/extproc`)

`gen_ai.provider.name` is required but only knowable after routing, so it is recorded through
an optional `BackendSpan` interface invoked from `SetBackend`. Making it optional rather than
part of `Span` keeps the eight existing mock span implementations compiling untouched.

`server.address` is deliberately not recorded: Envoy resolves the upstream cluster after
ExtProc runs, so the only host visible here is the gateway's own listener.

**Attribute naming**

Attribute names are hardcoded rather than imported from a versioned semconv package, matching
`internal/metrics/genai.go`, so a dependency bump cannot silently rename what users'
dashboards query. They are pinned against semconv v1.41.0 in tests instead, where an upstream
change surfaces as a failure that names it. No new `gen_ai.*` name is invented: endpoints the
conventions do not cover record only the core set.

Operation values are deliberately not shared with `internal/metrics`: two of them differ from
the registry values (`completion` vs `text_completion`, `messages` vs `chat`), and changing
the metrics values would break existing dashboards. Provider values *are* shared, and
`TestGenAIProviders_matchTracing` in `internal/metrics` pins them against the real constants
so spans and metrics stay joinable on `gen_ai.provider.name`.

**One fix to existing code falls out of this**

The Anthropic stream fold is moved from the OpenInference recorder to
`internal/apischema/anthropic` so both conventions share it, and no longer aliases the
caller's content blocks. It appended deltas in place, so folding the same chunks twice
previously returned doubled text. It also now skips nil chunks rather than panicking.
Regression tests: `TestMessagesResponseFromStream_doesNotMutateInput` and
`TestMessageRecorder_RecordResponseChunks_boundaries`, both of which fail before the fix.

**MCP tracing: OpenTelemetry semantic conventions, behind the same opt-in**

This PR also brings the MCP span path onto the OTel MCP conventions, and routes it through the
same selector as the LLM endpoints. `recorderSet` gains an `mcp` field, so a convention that
does not supply an MCP vocabulary is a compile error, and the choice is made in one place.

|                | default (unset / `openinference`)      | `gen_ai`                                       |
| -------------- | -------------------------------------- | ---------------------------------------------- |
| Span name      | `CallTool`, `ListTools`, ...           | `tools/call {tool}`, `tools/list`, ...         |
| Tool name      | `mcp.tool.name`                        | `gen_ai.tool.name` + `gen_ai.operation.name`   |
| Prompt name    | `mcp.prompt.name`                      | `gen_ai.prompt.name`                           |
| Request ID     | `mcp.request.id`                       | `jsonrpc.request.id`                           |
| Transport      | `mcp.transport`                        | `network.transport`, `network.protocol.*`      |
| Errors         | `exception` event                      | `error.type`, `rpc.response.status_code`       |
| Session        | per-backend event only                 | also `mcp.session.id` on the span              |
| List sizes     | not recorded                           | `mcp.tools.count`, `mcp.resources.count`, ...  |
| Tool call I/O  | not recorded                           | `gen_ai.tool.call.arguments`/`.result`, opt-in |

**The default output is unchanged.** `tests/data-plane-mcp/mcp_test.go` is byte-identical to
`main` and passes; `requireMCPSpan` compares the whole attribute map, so a leaked attribute
fails it. The signals added in this PR (list counts, the aggregation timeline, tool call
content) exist only under `gen_ai`, so opting in is the only way to change what a dashboard
sees. The gateway-specific `mcp.*` attributes are marked deprecated in the docs, with the
default to flip in a later release after an announced deprecation period.

The two vocabularies are declarative structs of function fields rather than branches threaded
through the tracer: `mcp_legacy.go` and `mcp_otel.go` hold everything convention-specific and
are meant to be read side by side, while propagation, sampling and custom attribute mappings
stay shared in `mcp.go`. A convention that does not record a signal supplies a no-op, so the
MCP proxy calls every method unconditionally.

Two MCP fixes ride along, both confined to the opt-in path:

- `mcp.session.id` recorded the *gateway-to-backend* session. `initialize` and the list
  aggregations call `RecordRouteToBackend` once per backend from parallel goroutines
  (`session.go`, `mcpproxy.go`), so as a span attribute it was last-writer-wins across
  backends. The span now records the client-facing session; the per-backend one stays on the
  per-backend event, where it is correctly scoped.
- `server.address`/`server.port` are dropped. They resolved to the local Envoy MCP listener,
  a hardcoded `127.0.0.1:10088`, so they were the same constant on every span regardless of
  which backend served the request. The real upstream address is not available to the proxy:
  `filterapi.MCPBackend` carries no address and Envoy resolves the backend by name one hop
  later.

Still open, tracked for a follow-up rather than done here: the gateway emits one `CLIENT` span
that plays both proxy roles. A `SERVER` span for the inbound client request, with per-backend
`CLIENT` children, is the correct shape and is where `client.address`/`client.port` would
belong. Doing it now would be a second change to span shape, so it waits.

**Rebase note: `/v1/responses/input_tokens`**

#2297 landed on main while this was open, adding the `/v1/responses/input_tokens` endpoint with
an OpenInference recorder. Because `recorderSet` requires every convention to supply every
endpoint, rebasing would not compile until the GenAI side existed, so this adds
`otelgenai.NewResponsesInputTokensRecorder`. It reports the existing `tokenize` operation rather
than inventing a `gen_ai.operation.name` value the conventions do not define - the endpoint
counts what a request would consume without running inference - and maps the result to
`gen_ai.usage.input_tokens`. Sampling parameters and output messages do not apply, which is
recorded as a row in `otelgenai/coverage_test.go`.

That endpoint therefore emits GenAI spans under `AI_GATEWAY_TRACING_SEMCONV=gen_ai`, which it
could not before. Its OpenInference output is untouched.

**Known gaps**

- Responses API: input messages and streaming chunks are not mapped yet.
- Tokenize: reuses the chat and completion mapping once those land.
- Six endpoints record only the core set because the conventions define nothing further for
  them. This is tracked distinctly from work not yet done in `coverage_test.go`.

**Related Issues/PRs**

Related: https://github.com/envoyproxy/ai-gateway/issues/747
Related: https://github.com/envoyproxy/ai-gateway/pull/2064

**Testing**

- `make precommit` passes.
- `go test ./internal/tracing/... ./internal/metrics/... ./internal/apischema/... ./internal/extproc/... -race -count=1`
- Per-file coverage for the new package is above the 70% floor in `.testcoverage.yml`.

End to end, through real Envoy + ExtProc:

| Suite | Covers | Result |
| ----- | ------ | ------ |
| `go test ./tests/data-plane-mcp/ -count=1` | every MCP method on the **default** convention | `ok` (158s) |
| `go test ./tests/data-plane-mcp/ -run TestMCPTracing_OTelSemConv -count=1` | the same data plane with `AI_GATEWAY_TRACING_SEMCONV=gen_ai` | `ok` (7.5s) |
| `go test ./tests/data-plane/vcr/... -run TestOtelGenAI -count=1` | the LLM ExtProc path, `gen_ai.provider.name` and the content opt-in in both states | `ok` |

The first row is the no-regression assertion: that file is byte-identical to `main`
(`git diff main -- tests/data-plane-mcp/mcp_test.go` is empty), and `requireMCPSpan` compares
the entire attribute map per span, so an extra or renamed attribute on the default path fails
it. The only change to `tests/data-plane-mcp/env.go` is additive - an ExtProc env hook, and
splitting the initialize span assertion out of session setup so the opt-in suite can assert
its own.

The second row asserts, on real spans: OTel span names and keys, absence of the legacy keys,
`mcp.tools.count` matching the aggregated list length, the `aggregation begin`/`end` timeline,
`mcp.session.id` equal to the session ID the client holds and different from every per-backend
session on the events, and the absence of `server.address`/`server.port`.

Unit side: `mcp_legacy_test.go` pins the frozen default vocabulary (span names, attribute
keys, and that the newer signals are no-ops under it), `mcp_semconv_test.go` pins the OTel
names against semconv v1.41.0, and `semconv_test.go` pins which convention selects which
vocabulary plus that the GenAI content opt-in cannot leak MCP content into the default.

The default path is meant to be byte-identical, for both the LLM endpoints and MCP. The
OpenInference diffs are a mechanical extraction of `getBoolEnv`/`getIntEnv` into
`internal/tracing/tracingenv` and of the HTTP status to error-type mapping into
`internal/tracing/httperr`; I diffed the latter element-by-element against the switch it
replaces. For MCP, the check is that the pre-existing end-to-end suite passes unmodified.

Thanks to @Hritik003 and @aishwaryaraimule21 for the reviews that shaped the MCP half of this,
and to @nacx for the call on gating rather than breaking. The MCP realignment was originally
unconditional; it is now opt-in with the legacy attributes deprecated, so existing dashboards
keep working and users migrate when they choose to.

Generative AI assistance was used in preparing this change, including the initial
implementation and a review pass over the diff. I have reviewed and understand all of it, and
can revise any part on request.


Hritik003:
hi @dergraf , appreciate for taking this big change!
I did some digging to help you out on MCP conventions that we currently follow and to refactor the same if required:


| OTel Convention Attr       | Current Implementation         | Requirement Level  | Status       | Notes                                                    |
|----------------------------|--------------------------------|--------------------|--------------|----------------------------------------------------------|
| `mcp.method.name`          | `mcp.method.name`              | Required           | OK           | Correctly set from req.Method                            |
| `gen_ai.tool.name`         | `mcp.tool.name`                | Cond. Required     | Wrong Key    | Should be `gen_ai.tool.name`                             |
| `gen_ai.prompt.name`       | `mcp.prompt.name`              | Cond. Required     | Wrong Key    | Should be `gen_ai.prompt.name`                           |
| `jsonrpc.request.id`       | `mcp.request.id`               | Cond. Required     | Wrong Key    | Should be `jsonrpc.request.id`                           |
| `error.type`               | `exception.type` (event only)  | Cond. Required     | Missing      | Only in event, not as span attribute                     |
| `rpc.response.status_code` | —                              | Cond. Required     | Missing      | JSON-RPC error code not recorded                         |
| `mcp.resource.uri`         | `mcp.resource.uri`             | Cond. Required     | OK           | Correctly set for resource operations                    |
| `gen_ai.operation.name`    | —                              | Recommended        | Missing      | Should be `"execute_tool"` for `tools/call`              |
| `mcp.protocol.version`     | `mcp.protocol.version`         | Recommended        | OK           | Correctly set                                            |
| `mcp.session.id`           | `mcp.session.id` (event only)  | Recommended        | Wrong Scope  | Only in RecordRouteToBackend event, not on span          |
| `network.transport`        | `mcp.transport = "http"`       | Recommended        | Wrong Key    | Wrong key AND value. Should be `network.transport = "tcp"` |
| `network.protocol.name`    | —                              | Recommended        | Missing      | Should be `"http"` for streamable HTTP                   |
| `network.protocol.version` | —                              | Recommended        | Missing      | Should be `"1.1"` or `"2"`                               |
| `server.address`           | —                              | Recommended        | Missing      | Backend server address for client spans                  |
| `server.port`              | —                              | Recommended        | Missing      | Backend server port                                      |

Span Naming Convention

| Method           | Current Name     | Expected Name                  | Notes                          |
|------------------|------------------|--------------------------------|--------------------------------|
| `tools/call`     | `CallTool`       | `tools/call {tool_name}`       | Should include tool name       |
| `tools/list`     | `ListTools`      | `tools/list`                   | Use raw method name            |
| `initialize`     | `Initialize`     | `initialize`                   | Use raw method (lowercase)     |
| `prompts/get`    | `GetPrompt`      | `prompts/get {prompt_name}`    | Include prompt name            |
| `resources/read` | `ReadResource`   | `resources/read`               | URI opt-in (high cardinality)  |
| `logging/setLevel` | `SetLoggingLevel` | `logging/setLevel`          | Use raw method name            |

Custom Attributes (Not in OTel Convention)

| Attribute                        | Method                 | Verdict          | Notes                     |
|----------------------------------|------------------------|------------------|---------------------------|
| `mcp.client.name / title / version` | `initialize`       | Custom           | Useful but non-standard   |
| `mcp.logging.level`             | `logging/setLevel`     | Custom           | Useful but non-standard   |
| `mcp.notifications.progress.*`  | `notifications/progress` | Custom         | Three custom sub-attrs    |
| `mcp.complete.argument.*`       | `completion/complete`  | Custom           | Not in convention         |
| `mcp.backend.name`              | (event)                | Gateway-specific | Routing detail, keep as-is |
| `mcp.session.new`               | (event)                | Gateway-specific | Routing detail, keep as-is |

dergraf:
Thanks for the thorough audit @Hritik003 — this was exactly the push needed. I verified every point against `internal/tracing/mcp.go` and they all checked out, so I've aligned the MCP tracing to the OTel conventions in this PR. Full alignment, including the `server.*` attributes that needed plumbing the backend listener address through.

**Attribute keys**
- `mcp.tool.name` → `gen_ai.tool.name` (+ `gen_ai.operation.name="execute_tool"` on `tools/call`)
- `mcp.prompt.name` → `gen_ai.prompt.name`
- `mcp.request.id` → `jsonrpc.request.id`
- `mcp.transport="http"` → `network.transport="tcp"` + `network.protocol.name="http"` + `network.protocol.version="1.1"`

**Errors** — `error.type` is now a span attribute, and JSON-RPC failures also record `rpc.response.status_code` (extracted from the typed `*jsonrpc.Error`).

**Scope/session** — `mcp.session.id` is now on the span (not just the `route to backend` event), and `server.address` / `server.port` are set from the resolved backend listener.

**Span names** — now the raw method (`tools/list`, `initialize`, `logging/setLevel`, …), with the target appended for the two high-value ops: `tools/call {tool_name}` and `prompts/get {prompt_name}`. `resources/read` keeps the URI out of the name to bound cardinality.

**Kept as-is** per your table (correct or gateway-specific): `mcp.method.name`, `mcp.protocol.version`, `mcp.resource.uri`, `mcp.logging.level`, `mcp.notifications.progress.*`, `mcp.complete.argument.*`, `mcp.client.*`, and the `route to backend` event's `mcp.backend.name` / `mcp.session.new`.

Covered by unit tests in `mcp_test.go` and the end-to-end `TestMCP` data-plane suite. Note this is a **breaking change** for existing MCP dashboards — called out in the PR description. Thanks again!

internal/tracing/mcp.go
```diff
@@ -134,9 +161,12 @@ func getMCPParamsAsAttributes(p mcp.Params) []attribute.KeyValue {
 			)
 		}
 	case *mcp.CallToolParams:
```
Hritik003: can we also set gen_ai attributes to all the mcp methods? else the `kind`of the trace will be empty( pfa image below). For tools/call we are able to see, because you have set gen_ai attributes for it.

Image


I know that MCP protocol operations (like tools/list, initialize) are not AI operations and therefore, they shouldn't have GenAI-specific attributes set. But for our tracing pov, we should add them to help user to identify the kind of the trace

internal/tracing/mcp.go
Hritik003: can we have a common function for RecordListOperations() which adds extra custom attributes depending on the operation like `list/resources` `list/tools` etc;

```go
// RecordListResponse implements recording of list operation results.
// For tools/list: records tool count and optionally a hash of tool names
// For resources/list: records resource count
// For prompts/list: records prompt count
func (s mcpSpan) RecordListResponse(result any) {
	switch v := result.(type) {
	case *mcp.ListToolsResult:
		attrs := []attribute.KeyValue{
			attribute.Int("mcp.tools.count", len(v.Tools)),
		}
		// Optional: Add a hash of tool names for correlation without high cardinality
		if len(v.Tools) > 0 {
			var names []string
			for _, tool := range v.Tools {
				names = append(names, tool.Name)
			}
			sort.Strings(names)
			hash := fnv.New32a()
			hash.Write([]byte(strings.Join(names, ",")))
			attrs = append(attrs, attribute.Uint64("mcp.tools.name_hash", hash.Sum32()))
		}
		s.span.SetAttributes(attrs...)
	case *mcp.ListResourcesResult:
		s.span.SetAttributes(attribute.Int("mcp.resources.count", len(v.Resources)))
	case *mcp.ListPromptsResult:
		s.span.SetAttributes(attribute.Int("mcp.prompts.count", len(v.Prompts)))
	}
}
```

Having tools_count, tools_hash (fingerprint of the current set of available tools, created by hashing a sorted, comma‑separated list of tool names.) resources_count inside a trace would be helpful

internal/mcpproxy/handlers.go
Hritik003, line 1820: can we make use of otel events ( formally called annotations in otel ) in the mcp methods like;

```
func (h *mcpHandler) HandleListTools(...) {
    span.AddEvent("fetching tools list begin")
    tools, err := h.service.ListTools(ctx)
    span.AddEvent("fetching tools list end", trace.WithAttributes(
....
```

and thus one can view the trace in this way

Span: tools/list
  Events:
    [0.0ms] mcp operation started
    [2.1ms] fetching tools list begin
    [5.7ms] fetching tools list end (tool.count=12)
    [6.0ms] mcp operation completed
  Attributes:
    mcp.method.name: tools/list
    mcp.tools.count: 12
    mcp.tools.name_hash: 0xAB12CD34

internal/tracing/mcp.go
```diff
@@ -33,7 +34,20 @@ type mcpSpan struct {
...
+	// The resolved backend, session, and server peer are only known after
+	// routing. Record them as span attributes for the OTel conventions
+	// (mcp.session.id, server.address, server.port) in addition to the
+	// gateway-specific "route to backend" event.
+	attrs := []attribute.KeyValue{attribute.String("mcp.session.id", sessionID)}
+	if serverAddr != "" {
+		attrs = append(attrs, attribute.String("server.address", serverAddr))
+	}
+	if serverPort != 0 {
+		attrs = append(attrs, attribute.Int("server.port", serverPort))
```
Hritik003: how about we include attribute on tools/call arguments, response? I see thats a keen part that we include in the trace
something like

```
{
 ...
    "gen_ai.tool.call.arguments": "{\"service\":\"payment-api\",\"duration\":\"1h\"}",
    "gen_ai.tool.call.result": "{\"status\":200,\"errors_found\":12}",
  },
...
```

Hritik003 (commented):
Reviewed the MCP part of the code and  have done research on how mcp traces are being handled across other gateways and I think we should include events in the traces ( formally annotations ) for example;

```
{
  "name": "tools/list",
  "kind": "SPAN_KIND_SERVER",
  "attributes": {
    "rpc.system": "mcp",
    "mcp.method.name": "tools/list",
    "mcp.tools.count": 4,
  },
  "events": [
    {
      "name": "mcp.tools_discovered",
      "attributes": {
        "mcp.tool.names": ["query_db", "fetch_logs", "restart_pod", "get_metrics"]
      }
    }
  ]
}
```

```
{
  "name": "initialize",
  "kind": "SPAN_KIND_SERVER",
  "attributes": {
    "rpc.system": "mcp",
    "mcp.method.name": "initialize",
    "mcp.protocol.version": "2024-11-05",
    "mcp.client.name": "ClaudeDesktop",
    "mcp.client.version": "1.0.0",
    "mcp.capabilities.tools": true,
    "mcp.capabilities.resources": false,
  },
  "events": [
    {
      "name": "mcp.handshake.success",
      "attributes": {
        "mcp.server.version": "2.1.0"
      }
    }
  ]
}
```

```
{
  "name": "tools/call fetch_logs",
  "kind": "SPAN_KIND_SERVER",
  "attributes": {
    "rpc.system": "mcp",
    "mcp.method.name": "tools/call",
    "gen_ai.operation.name": "execute_tool",
    "gen_ai.tool.name": "fetch_logs",
    "gen_ai.tool.call.arguments": "{\"service\":\"payment-api\",\"duration\":\"1h\"}",
    "gen_ai.tool.call.result": "{\"status\":200,\"errors_found\":12}",
  },
  "events": [
    {
      "name": "mcp.tool.execution_start",
      "timestamp": "2026-07-31T12:00:00.000Z"
    },
    {
      "name": "mcp.tool.execution_end",
      "timestamp": "2026-07-31T12:00:00.142Z"
    }
  ]
}
```

internal/tracing/mcp.go
dergraf: Thanks! The empty "kind" is the viewer reading `gen_ai.operation.name`, which the conventions only define for actual AI operations (`execute_tool` for `tools/call`) — there's no registry value for `tools/list`, `initialize`, etc. This PR deliberately focused on not inventing any new `gen_ai.*` semantics, so I left it off non-AI methods (they're still identifiable via `mcp.method.name` and the span name). Would you prefer we invent gen_ai values for them anyway?

internal/tracing/mcp.go
dergraf: Done. Added `RecordListResult`, invoked once from the aggregation path, recording `mcp.tools.count` / `mcp.resources.count` / `mcp.prompts.count` / `mcp.resource_templates.count`. I kept it to counts and left out the `tools.name_hash` for now: it's a low-cardinality fingerprint, agreed, but it's only actionable with an out-of-band hash→tool-set mapping, and it's still a derived signal of the tool inventory we otherwise keep off spans by default. Happy to add it as a follow-up if you have a concrete correlation use for it.

internal/mcpproxy/handlers.go
dergraf: Added. The list aggregation is now bracketed with `"<method> aggregation begin"` / `"<method> aggregation end"` events on the span, so list operations get the begin/end timeline you sketched. I put the size on the span as the `*.count` attribute (queryable) rather than repeating it in the event.

internal/tracing/mcp.go
dergraf: Good call, and it turns out `gen_ai.tool.call.arguments` / `gen_ai.tool.call.result` are both real semconv v1.41.0 attributes, so I've used exactly those names. Because they're message content, they follow the GenAI opt-in (`OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT`) rather than being recorded unconditionally — arguments are captured at request time, the result from the proxied response, and both are omitted entirely when capture is off. Names are pinned against semconv in the tests.

internal/tracing/mcp.go
```diff
@@ -134,9 +204,19 @@ func getMCPParamsAsAttributes(p mcp.Params) []attribute.KeyValue {
 			)
 		}
 	case *mcp.CallToolParams:
-		attrs = append(attrs, attribute.String("mcp.tool.name", params.Name))
+		attrs = append(attrs,
+			attribute.String("gen_ai.operation.name", "execute_tool"),
```
aishwaryaraimule21: IIUC, these attributes would be populated even when user has set sem convention as `OpenInference`. And we are not following https://arize-ai.github.io/openinference/spec/semantic_conventions.html for MCP tool calls in that case. Is my understanding correct?

internal/tracing/mcp.go
```diff
@@ -62,25 +122,34 @@ type mcpTracer struct {
...
 		tracer:            tracer,
 		propagator:        propagator,
 		attributeMappings: attributeMappings,
+		captureContent:    captureContent,
 	}
 }
 
 // StartSpanAndInjectMeta implements [tracingapi.MCPTracer.StartSpanAndInjectMeta].
 func (m mcpTracer) StartSpanAndInjectMeta(ctx context.Context, req *jsonrpc.Request, param mcp.Params, headers http.Header) tracingapi.MCPSpan {
 	attrs := []attribute.KeyValue{
```
aishwaryaraimule21: can we check if [client.address](https://github.com/open-telemetry/semantic-conventions/blob/v1.43.0/docs/registry/attributes/client.md) and port can be added?

internal/tracing/mcp.go
```diff
@@ -30,19 +32,77 @@ var _ tracingapi.MCPTracer = (*mcpTracer)(nil)
...
 // RecordRouteToBackend implements [tracingapi.MCPSpan.RecordRouteToBackend].
-func (s mcpSpan) RecordRouteToBackend(backend string, sessionID string, isNew bool) {
+func (s mcpSpan) RecordRouteToBackend(backend string, sessionID string, isNew bool, serverAddr string, serverPort int) {
+	// The resolved backend, session, and server peer are only known after
+	// routing. Record them as span attributes for the OTel conventions
+	// (mcp.session.id, server.address, server.port) in addition to the
+	// gateway-specific "route to backend" event.
+	attrs := []attribute.KeyValue{attribute.String("mcp.session.id", sessionID)}
+	if serverAddr != "" {
+		attrs = append(attrs, attribute.String("server.address", serverAddr))
```
aishwaryaraimule21: This belongs to the MCP client span. Should we rather use the [MCP Server span](https://github.com/open-telemetry/semantic-conventions-genai/blob/main/docs/gen-ai/mcp.md#server)?

internal/tracing/mcp.go
dergraf: Your understanding is correct on the mechanics: `AI_GATEWAY_TRACING_SEMCONV` only selects the `recorderSet` for the LLM endpoints (`internal/tracing/semconv.go`). The MCP tracer is constructed unconditionally with GenAI semantics:

```go
// internal/tracing/tracing.go
mcpTracer: newMCPTracer(tracer, propagator, headerAttrs, otelgenai.NewConfigFromEnv().CaptureMessageContent),
```

So yes — an OpenInference user gets `gen_ai.operation.name` / `gen_ai.tool.name` / `gen_ai.tool.call.*` on MCP spans. Worth noting this PR is what introduces the divergence: before it, the MCP span was convention-neutral (`mcp.tool.name`, `mcp.prompt.name`, no `gen_ai.*`).

On "we are not following the OpenInference spec for MCP tool calls" — I went looking for the spec to follow and there isn't one to diverge from. OpenInference *does* ship MCP support, but it is scoped to context propagation only. From `openinference-instrumentation-mcp`:

> Currently, it only enables context propagation so that the span active when making an MCP tool call can be connected to those generated when executing it. It does not generate any telemetry.

And the semantic conventions page has no MCP section at all. The closest analogue is the generic `TOOL` span kind (`openinference.span.kind=TOOL`, `tool.name`, `tool.parameters`, `input.value`/`output.value`), which is not MCP-specific — it is what an agent framework emits around a tool invocation.

So there are two separable questions here, and I'd like your read on the first:

1. **Should `tools/call` also carry the OpenInference `TOOL` mapping when that convention is selected?** That would make Phoenix render it as a tool span rather than an untyped one. It is a genuine mapping for `tools/call`, but there is nothing to map the other dozen-plus methods onto, so they would stay OTel-MCP-shaped either way and the span vocabulary becomes mixed. My inclination is to keep MCP spans single-convention (OTel MCP semconv) and document that explicitly, but if Phoenix rendering is the main consumer here I am happy to add the `TOOL` mapping for `tools/call`.

2. **The content gate is a real defect and I'll fix it regardless.** Tool call arguments/results are gated on `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT` even when OpenInference is selected, so an operator who sets `OPENINFERENCE_HIDE_INPUTS` / `OPENINFERENCE_HIDE_OUTPUTS` gets no effect on MCP spans. Nothing leaks by default (the GenAI default is off), but a redaction knob that silently does not apply is precisely the failure mode `semconv.go`'s own doc comment says we refuse. I'll make the gate read the active convention's privacy config.

internal/tracing/mcp.go
dergraf: Right attribute, but it is the same decision as your other comment on `RecordRouteToBackend`, so let me tie them together: the MCP conventions put `client.address` / `client.port` on the **server** span and `server.address` / `server.port` on the **client** span — never both on one span. This span is currently `SpanKindClient` (that predates this PR), which is why it has `server.*` and not `client.*`. Adding `client.*` only makes sense as part of resolving the span-role question, so I'll take it up in that thread.

One feasibility note for when we do: the MCP proxy sits behind the local Envoy instance, so `r.RemoteAddr` is the loopback hop from Envoy, not the client. The real client address has to come from `X-Forwarded-For`, and there is no XFF trust handling anywhere in the codebase today. The tracer already receives the request `http.Header`, so no signature change is needed — but we do need a deliberate answer on whether XFF is trusted here, rather than silently recording whatever the header says.

internal/tracing/mcp.go
dergraf: Agreed, and this is the sharpest of your three. The gateway is a server to the MCP client and a client to each backend, and today a single `SpanKindClient` span plays both roles. The span kind predates this PR; adding `server.address` / `server.port` is what makes the conflation visible.

Two things make me think the fix is more than a kind flip:

**`server.address` is not actually the MCP server.** It is the local Envoy MCP listener:

```go
// internal/mcpproxy/config.go — "Talk to the backend MCP listener on the local Envoy instance."
if u, err := url.Parse(mcpConfig.BackendListenerAddr); err == nil {
    newConfig.backendListenerHost = u.Hostname()
    ...
```

The conventions say `server.address` SHOULD represent the server behind any intermediaries, if available. It is not available to the proxy: `filterapi.MCPBackend` carries no address at all — routing is by backend name and Envoy resolves it. So as written these attributes point at a loopback listener and carry close to zero value.

**One span cannot honestly carry one peer under fan-out.** `initialize` and the aggregating methods (`tools/list`, `resources/list`, `prompts/list`, `resources/templates/list`, `logging/setLevel`) dispatch to every backend in parallel and call `RecordRouteToBackend` once per backend from inside those goroutines — `internal/mcpproxy/session.go:334` and `internal/mcpproxy/mcpproxy.go:208`.

That surfaced a concrete defect I should flag: this PR promotes `mcp.session.id` from an event attribute to a span attribute, and those per-backend calls pass the **gateway↔backend** session (`cse.sessionID` / `initResult.sessionID`). On the fan-out paths the span attribute is therefore last-writer-wins across N backends — an arbitrary backend's session ID. On the single-backend paths (`tools/call`, `resources/subscribe`, `completion/complete`, `notifications/progress`) nothing is overwritten, but the recorded value is still the upstream session rather than the client-facing one, and the conventions define `mcp.session.id` as the session the request is part of — on a server span, that is the client↔gateway session (`session.id`). On `main` none of this bit, because the value only ever appeared on the per-backend event, where per-backend is the correct scope. I'll fix it independently of how we resolve the span roles.

So, two options:

**(a)** Flip this span to `SpanKindServer`, add `client.address` / `client.port`, drop `server.*`. Cheap and conformant for a single span, but it drops the backend peer entirely and is a second behavioral change for anyone already consuming these spans.

**(b)** A `SERVER` span for the inbound client request, with `client.*` and the client-facing `mcp.session.id`, plus a child `CLIENT` span per backend dispatch carrying `server.*`, `mcp.backend.name` and the per-backend session ID — replacing the `route to backend` event. More work, but it is the only shape that survives fan-out and it makes the per-backend latency visible, which the current single span hides.

I lean to (b). Interestingly OpenInference's MCP instrumentation went the same direction: `@arizeai/openinference-instrumentation-mcp` v0.2.0 was reworked to instrument the Transport rather than Client/Server, on the rationale that A→B→A is fine because the second A "just happens to be the same server, but is not the same RPC method". Per-hop instrumentation rather than a fixed role per process is the same conclusion.

Which would you prefer? (b) is a meaningful chunk of work on top of an already large PR, so I am also fine landing the `mcp.session.id` fix plus dropping the low-value `server.*` here, and doing the SERVER/CLIENT split as a tracked follow-up.

---

*Edited to correct the call-site citations: the per-backend fan-out is `session.go:334` and `mcpproxy.go:208`; the `handlers.go` sites I originally listed are single-backend routed paths, which I've folded into the paragraph above. The defect itself is unchanged.*

dergraf: Following up on my own point about `server.address` — I traced the data path and it is worse than "low value": the attribute is a constant.

`BackendListenerAddr` is hardcoded by the controller:

```go
// internal/controller/gateway.go
BackendListenerAddr: fmt.Sprintf("http://127.0.0.1:%d", internalapi.MCPBackendListenerPort),
```

`mcpproxy` parses exactly that into `backendListenerHost` / `backendListenerPort`, so every MCP span gets `server.address="127.0.0.1"`, `server.port=10088` — identical on every span, regardless of which backend the request was routed to.

For context on why, the MCP request path is two loopback hops through the local Envoy:

```
client → Envoy public listener
       → STATIC cluster 127.0.0.1:9856          (MCP proxy — its own http.Server in the
                                                  extproc container, separate from the
                                                  ext_proc gRPC service on its UDS)
       → MCP proxy issues plain HTTP to 127.0.0.1:10088
       → "aigateway-mcp-backend-listener"        (internal Envoy listener synthesized by
                                                  the extension server)
       → routed by the x-ai-eg-mcp-backend header → the real upstream MCP server
```

The inbound cluster rewrite is in `internal/extensionserver/mcproute.go` (`modifyMCPGatewayGeneratedCluster`), the internal listener in `createBackendListener`. So the proxy never learns the upstream MCP server's address — Envoy resolves it one hop later from the backend name — which matches the conventions' "if it's available" escape hatch on `server.address`.

That makes dropping `server.*` the right call under either option, so I'll take it out regardless of whether we land (a) or (b). If we do go with (b), the per-backend `CLIENT` child spans would carry `mcp.backend.name` as the meaningful peer identifier instead, since that is the only backend identity available at this layer.

internal/tracing/mcp.go
dergraf: Update after review feedback: `RecordListResult` is now gated behind `AI_GATEWAY_TRACING_SEMCONV=gen_ai`. The default convention keeps the attribute set it had before this PR, so `mcp.tools.count` and friends need that opt-in. The maintainers asked for no breaking changes to observability, and adding attributes by default counts. Nothing about the implementation changed, only when it applies.

internal/mcpproxy/handlers.go
dergraf: Same note as the list-count thread: the `<method> aggregation begin`/`end` events are now emitted only under `AI_GATEWAY_TRACING_SEMCONV=gen_ai`. The default span keeps the events it had before (`route to backend`, `exception`).

internal/tracing/mcp.go
dergraf: And the same here: `gen_ai.tool.call.arguments`/`.result` are now under `AI_GATEWAY_TRACING_SEMCONV=gen_ai`, on top of the existing `OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT` opt-in — so two things to enable, not one. The default convention never records tool call content. Sorry for the churn; the names and behavior are unchanged.

internal/tracing/mcp.go
dergraf: Your understanding was correct, and this is now fixed. `recorderSet` gained an `mcp` field, so MCP spans go through the same selector as the LLM endpoints: OpenInference keeps the gateway-specific `mcp.*` attributes unchanged, and the OTel MCP conventions require `AI_GATEWAY_TRACING_SEMCONV=gen_ai`. Content capture follows the selected convention too, so a GenAI variable can no longer affect MCP spans while OpenInference is active. The pre-existing `tests/data-plane-mcp` suite passes byte-identical to `main`, which is the check that the default did not move.

gavrissh:
@dergraf Could you resolve the conflicts in the PR

johnugeorge:
cc @nacx

internal/tracing/mcp.go
nacx: I'd rather keep the current approach and do not invent any name. Users could start relying on those invented values and we probably don't want that.

nacx (approved):
LGTM. Thanks for the work here, it looks really good!

Thanks @Hritik003 and @aishwaryaraimule21 for your thorough reviews. Much appreciated.

@dergraf Can you file an issue to track the change to implement option (b) for the CLIENT/SERVER spans? I agree that should be the way to go, but I don't think we should block this PR on it. Let's keep it tracked in an issue.

dergraf:
Thanks @nacx! Filed #2562 to track the SERVER/CLIENT span split (option (b)).

The issue captures why one span can't carry both peers, the two fan-out call sites that make it concrete (`session.go:334` and `mcpproxy.go:235`), and the X-Forwarded-For trust question that needs settling before `client.address` can be recorded at all — the MCP proxy sits behind the local Envoy, so `r.RemoteAddr` is the loopback hop rather than the client.

It also notes what this PR already settled so the follow-up stays narrow: `mcp.session.id` is the client-facing session on the span with the per-backend one left on the `route to backend` event, and `server.address`/`server.port` were dropped rather than shipped, since `BackendListenerAddr` is hardcoded and they'd have been a constant `127.0.0.1:10088` on every span.

And agreed on not inventing `gen_ai.operation.name` values for the non-AI MCP methods — leaving that as-is.

PR #2407, opened 2026-07-22
docs: add Eden AI to supported providers (EU, OpenAI-compatible)

MVS-source:
**Description**

Adds Eden AI [1] to the supported providers table. Eden AI is an EU based, GDPR compliant gateway that exposes models from many vendors (OpenAI, Anthropic, Mistral, Google, Cohere and others) behind a single OpenAI compatible key.

It works with the existing OpenAI schema, the same way Groq, Together AI, DeepSeek, Mistral and TARS already do, so no code change is needed:

- API schema config: `{"name":"OpenAI","prefix":"/v3"}`, endpoint host `api.edenai.run`
- Upstream auth: API Key

On verification: when I opened this PR I called `POST https://api.edenai.run/v3/chat/completions` directly and got HTTP 200 for `openai/gpt-4o-mini` and `mistral/mistral-large-latest`. Both ids are still served today, which you can confirm yourself without credentials because the models endpoint is public and returns the OpenAI list shape [2]. I have not run this through a live gateway deployment, so please check it against your provider conformance suite if you have one.

I work at Eden AI and am happy to keep this entry current.

N/A

Sorry about the style check. The original description opened with a "Commit Message" line instead of "**Description**", which is why it stayed red. Fixed here.

[1] https://www.edenai.co
[2] https://api.edenai.run/v3/models


MVS-source:
The `Description` check is green as of `fd978e3`. Sorry it sat red for a month.

Two things were wrong in the description, not one. It opened with a `**Commit Message:**` line instead of `**Description**`, which is the failure the log showed. Behind that, the "no markdown format link" step would have failed too, because the body used `[text](url)` links. I fixed both together and re-ran your three checks locally against the new body before touching the PR, so it should not bounce again. Links are now bare and numbered at the end, following the format in #1128.

I also dropped the model count from the table cell while I was in there. It would have gone stale, and the catalogue moves week to week.

Two notes that may help whoever picks this up:

The push reset workflow approval on this fork, so three check suites now show `action_required` and need a maintainer to run them. They were passing on the previous commit. The diff is one table row, so nothing in the Go build is affected either way.

The Netlify deploy preview on this PR does not show the change. I fetched it a few minutes after `deploy/netlify` went green on `fd978e3` and its provider table is identical to the one on aigateway.envoyproxy.io, with no Eden AI row. The row is really there, at line 32 of `site/docs/capabilities/llm-integrations/supported-providers.md`, and it has the same five columns as the other entries with Prettier clean. I suspect fork previews build from the base branch, but either way the preview is not a reliable way to review this one.

PR #2412, opened 2026-07-22
fix(mcp-apps): preserve ui:// scheme when namespacing MCP resource URIs

### LINKED ISSUE #2228: `resources/read` fails for MCP tool `_meta.ui.resourceUri` in multiplexing configuration

walsm232:
**Description**

> [!NOTE]  
> This PR was developed with AI assistance (**Claude Code - Fable**).

In MCP multiplexing mode, the gateway namespaces resource URIs as `<backend>+<scheme>://<path>` so that `resources/read` can route back to the originating backend. This breaks MCP Apps (ext-apps) UI resources in two ways:

1. `tools/call` results carrying `_meta.ui.resourceUri` (or `ResourceLink` / `EmbeddedResource` entries in `Content`) were passed through unmodified. Agents then issued `resources/read` with the bare upstream URI, and the gateway rejected it with HTTP 400 (`invalid resource URI: ui://...`).
2. Even where the rewrite did apply (e.g. `resources/list`), `backend+ui://` changes the URI scheme, which the MCP Apps specification: https://github.com/modelcontextprotocol/ext-apps/blob/main/specification/2026-01-26/apps.mdx requires to remain exactly `ui://`, so spec-compliant hosts refuse to render the resource.

This PR namespaces `ui://` URIs by inserting the backend name as the leading path segment (`ui://<backend>/<rest>`), so the scheme survives the rewrite, and extends the rewrite to the surfaces that previously leaked bare upstream URIs: `tools/call` results (`_meta`, `ResourceLink`, `EmbeddedResource`) and tool descriptors in `tools/list`. Both MCP Apps
metadata conventions are handled: nested `_meta.ui.resourceUri` and flat `_meta["ui/resourceUri"]`. Non-UI URIs keep the existing `<backend>+<scheme>://` encoding.

The encoding is deterministic and stateless: the backend is recoverable from the URI alone on any gateway replica, with no per-session lookup state. This keeps `resources/read` responses safely cacheable by shared intermediaries and aligns with the upcoming MCP specification revision (2026-07-28), which removes protocol-level sessions and prescribes self-contained, server-minted handles for cross-call state.

Verified end-to-end with two kind clusters running identical Envoy Gateway v1.8.1 + the same MCP backends, differing only in AI Gateway version (released v1.0.0 vs this branch), using MCP Inspector against a genuine MCP Apps server (`@modelcontextprotocol/server-basic-vanillajs`) multiplexed behind the gateway:

| | v1.0.0 | this PR |
|---|---|---|
| `tools/call` → `_meta.ui.resourceUri` | `ui://get-time/mcp-app.html` (bare) | `ui://vanillajs/get-time/mcp-app.html` |
| `resources/read` with that URI | HTTP 400 | HTTP 200, app renders |

Fixes #2228
Related PR: #2234 (alternative approach; the `backend+ui://` prefix was found there to break the `ui://` scheme hosts require, and a per-session URI→backend map conflicts with the stateless session design and cacheable reads. This PR keeps routing derivable from the URI itself instead)

- **Known limitation (intentionally out of scope):** once an app renders, interactive apps that invoke tools *from within the UI* call them by their upstream names (e.g. `get-time`), which multiplexing renames to `vanillajs__get-time`, so in-app tool calls fail with `invalid tool name`. The tool name is baked into server-authored HTML/JS the gateway cannot rewrite; fixing this requires a routing-semantics design discussion (unprefixed-name resolution, scoping calls to the backend that served the UI resource, and the interaction with per-tool authorization rules). I can file a follow-up issue with reproduction steps if needed. This PR fixes rendering, which is a strict prerequisite; display-only apps work fully with it.
- `rewriteMetaResourceURIs` is deliberately applied without an "already rewritten" guard: each response passes through the rewrite exactly once (JSON and SSE paths are mutually exclusive), and a prefix-based guard would misfire when an upstream URI's first path segment coincides with the backend name. A test covers that round-trip.
- The `metaResourceURIPaths` table is the extension point for future `_meta` conventions that carry resource URIs.
- Old-format `<backend>+ui://` URIs still decode via the fallback split, so clients holding URIs from a previous version keep working across an upgrade.


walsm232:
**Before**

Screenshot 2026-07-22 at 23 05 09

```
Error from MCP server: StreamableHTTPError: Streamable HTTP error: Error POSTing to endpoint: invalid resource name ui://get-time/mcp-app.html: invalid resource URI: ui://get-time/mcp-app.html
```

**After**

Screenshot 2026-07-22 at 23 06 29


```
Received POST message for sessionId 67157d99-36cc-482e-91c4-e71c972e6658
```

PatrickMcAfee-Elastic:
**`resources/read` response returns the namespaced URI to the client** (re: `tests/data-plane-mcp/mcp_test.go`, `testToolCallUIResourceURI`)

The test asserts `readRes.Contents[0].URI == defaultMCPBackendUIRendererURI` (the gateway-namespaced form). This happens because the test handler echoes `req.Params.URI` back in the content, and the upstream URI the gateway sends is the bare upstream form (`ui://prefab/tool/renderer.html`). `maybeResponseModify` then re-namespaces it, so the client sees `ui://backend/prefab/tool/renderer.html` in the content URI field.

This is self-consistent and arguably correct — all URIs a client receives are in the namespaced form it can round-trip through `resources/read`. Worth an explicit call-out though: a reader might expect the content URI to be the canonical upstream/resource-server URI. If any client inspects content URIs for identity rather than routing, it will see the namespaced form. Is this the intended contract?

PatrickMcAfee-Elastic (approved):

walsm232:
For the known limitation I flagged in the PR description (interactive in-app tool calls failing because they use the bare upstream name), I have opened a proposal to address it: https://github.com/envoyproxy/ai-gateway/pull/2425.

The short version is a `prefixMode: Never` option that exposes tool names unprefixed and resolves the target backend at call time, so the names hardcoded in app HTML match what the gateway routes on.

I think this rendering fix can and probably should merge as-is (provided any reviewers are happy with all of the implementation and the approach!), independently of that proposal. Rendering is a prerequisite and stands on its own (display-only apps work fully with it), and the interactive piece is a separable design decision that might be better handled as its own follow-up I think.

internal/mcpproxy/handlers.go
```diff
@@ -1224,24 +1235,97 @@ func upstreamResourceName(fullName string) (backendName, name string, err error)
...
 // can still match the resource URIs.
+//
+// ui:// URIs are the exception: the scheme must survive the rewrite, so the backend name is
+// inserted as the leading path segment instead: ui://<backend>/<rest>.
 func downstreamResourceURI(uri string, backendName string) string {
+	if rest, ok := strings.CutPrefix(uri, uiSchemePrefix); ok {
+		return uiSchemePrefix + backendName + "/" + rest
+	}
 	return fmt.Sprintf("%s+%s", backendName, uri)
 }
```
walsm232, line 1260: For context, while working on this I looked at how Agentgateway solved the same problem for inspiration, since as far as I can tell they might be the only other gateway with MCP Apps awareness so far 🤔 . They also keep the `ui://` scheme intact, but encode the backend differently: `ui://<target>+<rest>` (e.g. `ui://monitor+system-monitor/mcp-app.html`), joining the target to the original authority with a `+`. This PR uses `ui://<backend>/<rest>` instead, making the backend the authority and shifting the original authority into the path.

I don't have a strong argument for one over the other. The MCP Apps spec only requires the URI to start with `ui://`, and hosts treat the value opaquely, so both are compliant as far as I know. The main trade-offs I can think of are:

- The `+` form leaves the original path untouched, and matches the existing `backend+scheme://rest` encoding we already use for non-ui URIs, so it is probably more internally consistent.
- The `/` form keeps the authority a plain backend name and avoids `+`, which I wonder if naive form decoding could turn into a space.

Happy to switch to the `+` form if it's more consistent or if any reviewers prefer it.

aishwaryaraimule21 (approved):
lgtm

Hritik003 (commented):
lgtm

mathetake (approved):

PR #2413, opened 2026-07-23
fix(backendauth): resolve AWS signing host dynamically from headers

pbonuboinasnow:
**Description**

This PR implements dynamic host resolution for AWS SigV4 signing in the AWS Backend Auth handler. 

Previously, `awsHandler.Do` hardcoded the request URL and signed host to `https://bedrock-runtime.<region>.amazonaws.com`, which ignored custom HTTP/2 pseudo-headers (like `:authority`) and `host` headers. This restriction blocked routing targeting custom domains, such as AWS PrivateLink VPC endpoints or non-standard Bedrock proxies.

To resolve this:
- Extracted host resolution into a dedicated, documented `resolveHost` helper method inside `internal/backendauth/aws.go`.
- Supported extracting host from `AWSSigningHostHeader` (`x-ai-eg-aws-signing-host`), `:authority`, or `host` headers.
- Introduced a security safeguard suffix check: if a custom host does not end with `.amazonaws.com` (ignoring ports), we discard it and fallback to the default regional bedrock endpoint. This ensures AWS credentials are never accidentally signed/transmitted to arbitrary external domains.
- Updated `Do` to use the resolved host and properly set `req.Host` before signing.
- Added comprehensive unit tests inside `internal/backendauth/aws_test.go` to cover all fallback tiers, including non-.amazonaws.com fallback, standard and non-standard ports.

- Refactoring the host extraction logic into `resolveHost` allows us to cleanly unit test the fallback tiers in isolation without needing to mock complex request signatures or credentials providers.
- Port-splitting is used solely to extract the hostname for the suffix validation check, and the original custom port remains fully preserved in the final request host header.

### DIFF
--- a/internal/backendauth/aws.go
+++ b/internal/backendauth/aws.go
@@ -11,6 +11,7 @@ import (
 	"crypto/sha256"
 	"encoding/hex"
 	"fmt"
+	"net"
 	"net/http"
 	"os"
 	"strings"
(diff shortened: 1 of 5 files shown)


nacx (commented):
Thanks! Can you please:

* Fix the DCO.
* Since the previous PR broke the e2e tests that use real credentials, please run those real provider tests and copy the results here.

mathetake:
This PR is extremely weird: why a client needs to know the AWS endpoint as well as the routing condition which is necessary to set the weird header?

mathetake:
See https://github.com/envoyproxy/ai-gateway/pull/2416

pbonuboinasnow:
> See #2416

Yes this is what we want if approved. This is an alternative (actually best) but we were trying (we both are from same team) to minimize upstream changes as this is a blocker for us.

mathetake:
then let's land #2416

PR #2414, opened 2026-07-23
fix(controller): Add support for AWS Role based access

pbonuboinasnow:
**Description**

This PR adds AWS role-based authentication and authorization support using AWS STS `AssumeRole` in the `BackendSecurityPolicy` controller.

When configured, the controller reads the user-provided AWS credentials file from a secret and checks if a `role_arn` is present. If found, the controller performs STS `AssumeRole` using the base credentials, stores the resulting temporary security credentials in a generated secret, and automatically rotates them before they expire.

Key Changes:
- **AWSAssumeRoleRotator**: Added a new rotator (`aws_assume_role_rotator.go`) that performs periodic STS `AssumeRole` calls and updates the temporary credentials secret.
- **Credentials Discovery**: Controller parses the referenced credentials secret for a `role_arn` entry, allowing users to configure role assumption seamlessly directly in their credentials file.
- **Unified Gateway Controller Mounting**: Updated the gateway controller to read from the generated secret containing temporary AWS credentials instead of static key pairs when role assumption is in use.

Related to AWS Role Based Access.

None.


kanurag94 (commented):
I believe the dataplane already does this today. In `internal/backendauth/aws.go`, we write `CredentialFileLiteral` to a temp file and call `config.LoadDefaultConfig(WithSharedCredentialsFiles(...))`
AWS SDK Go v2 reads role_arn out of the credentials file (shared_config.go, roleArnKey) and after resolving the static keys, wraps them in an assume-role provider `(resolve_credentials.go: if len(sharedConfig.RoleARN) > 0 { credsFromAssumeRole(...) })`

Since awsHandler.Do calls credentialsProvider.Retrieve on every request, the cache refreshes on its own. So a user who puts role_arn in their credentials secret today already gets AssumeRole with automatic refresh, and it honors external_id, source_profile, role_session_name and duration_seconds as well.

If there is an argument for doing this in the control plane, which is that the extproc pods would then never hold long-lived keys and STS calls get centralized. If that's the motivation, we will need to evaluate this from that angle.

pbonuboinasnow:
> I believe the dataplane already does this today. In `internal/backendauth/aws.go`, we write `CredentialFileLiteral` to a temp file and call `config.LoadDefaultConfig(WithSharedCredentialsFiles(...))` AWS SDK Go v2 reads role_arn out of the credentials file (shared_config.go, roleArnKey) and after resolving the static keys, wraps them in an assume-role provider `(resolve_credentials.go: if len(sharedConfig.RoleARN) > 0 { credsFromAssumeRole(...) })`
> 
> Since awsHandler.Do calls credentialsProvider.Retrieve on every request, the cache refreshes on its own. So a user who puts role_arn in their credentials secret today already gets AssumeRole with automatic refresh, and it honors external_id, source_profile, role_session_name and duration_seconds as well.
> 
> If there is an argument for doing this in the control plane, which is that the extproc pods would then never hold long-lived keys and STS calls get centralized. If that's the motivation, we will need to evaluate this from that angle.

pbonuboinasnow:
> I believe the dataplane already does this today. In `internal/backendauth/aws.go`, we write `CredentialFileLiteral` to a temp file and call `config.LoadDefaultConfig(WithSharedCredentialsFiles(...))` AWS SDK Go v2 reads role_arn out of the credentials file (shared_config.go, roleArnKey) and after resolving the static keys, wraps them in an assume-role provider `(resolve_credentials.go: if len(sharedConfig.RoleARN) > 0 { credsFromAssumeRole(...) })`
> 
> Since awsHandler.Do calls credentialsProvider.Retrieve on every request, the cache refreshes on its own. So a user who puts role_arn in their credentials secret today already gets AssumeRole with automatic refresh, and it honors external_id, source_profile, role_session_name and duration_seconds as well.
> 
> If there is an argument for doing this in the control plane, which is that the extproc pods would then never hold long-lived keys and STS calls get centralized. If that's the motivation, we will need to evaluate this from that angle.

Did not realise that its using v2. Will test and come back

PR #2415, opened 2026-07-23
feat(mcpproxy): implement backend-only authorization pre-check for MCP initialize phase

pbonuboinasnow:
**Description**

The MCP Proxy currently opens upstream connections to **every** backend in a route during `initialize`, deferring authorization until `tools/list` / `tools/call` since the tool isn't known yet at `initialize` time. This wastes connections and causes avoidable upstream 401s against backends the client was never going to be allowed to use.

This PR adds a backend-only authorization pre-check, `authorizeBackendOnly` (`internal/mcpproxy/authorization.go`), that runs during `initialize`, before any connection is opened, and skips backends the client has no possible access to under `MCPRoute.spec.securityPolicy.authorization`. It's evaluated via CEL partial evaluation against the same rules `authorizeRequest` already uses, so it can only ever be *more permissive* than the real per-tool check, never more restrictive.

**The pre-check is disabled by default**, behind `MCP_PROXY_ENABLE_BACKEND_FILTER=true` - see "Known limitations" below.

**Known limitations**

- **Resources/prompts/completions aren't authorization-scoped at all today** - this predates this PR, is not specific to it, and pruning a backend at `initialize` also hides its resources/prompts as a result. Bringing  resources/prompts/completions under the same authorization rules as tools is real, separate follow-up work - not addressed here, to be discussed in community sync.
- **JWT scope changes mid-session.** A backend pruned at `initialize` (using  that request's token) stays pruned for the life of the session (is really not a limitation?), even if  the client later presents a refreshed token that would now pass. This  currently surfaces as a `400` on the next `tools/call`, recoverable only by calling `initialize` again. This is inherent to deciding access once at `initialize` time - **the same limitation applies to #2351's
  `backendSelector` approach**, not just to reading authorization rules. 

Because of these two limitations, the pre-check defaults to **disabled**. Operators comfortable with the tradeoff can opt in via `MCP_PROXY_ENABLE_BACKEND_FILTER=true`; everyone else is unaffected - when disabled, `newSession` never calls `authorizeBackendOnly` and behavior is byte-for-byte identical to before this PR. No new deployment plumbing is needed: settable today via the Helm chart's `extProc.extraEnvVars`, or per-Gateway via `GatewayConfig.spec.extProc.kubernetes.env`.

Test coverage: `internal/mcpproxy/authorization_test.go` (`TestAuthorizeBackendOnly`), `internal/mcpproxy/mcpproxy_test.go`, `internal/mcpproxy/handlers_test.go`. 
All local checks (`make lint` and `make test`) pass cleanly.

<b>How the pre-check works, in detail </b>

`newSession` extracts the caller's JWT claims/scopes once per request (identical for every backend), then calls `authorizeBackendOnly` for each backend before `initializeSession`. If it returns `false`, the backend is skipped entirely -no connection, no `initialize` call, no log noise, no metric for a doomed request.

The same compiled CEL expression used for the full per-tool check is evaluated again here, but as a second program compiled with `cel.OptPartialEval` (`compiledAuthorizationRule.backendOnlyProgram`), with `request.mcp.method`,
`request.mcp.tool`, and `request.mcp.params` marked as unknown CEL attributes - these describe whichever future `tools/call`/`tools/list` request eventually happens, so they can't be known yet. `request.method`, `request.host`, and `request.path`, by contrast, belong to the current HTTP request and are the same for every request on this session, so they're populated with their real values. This lets cel-go decide any part of the expression that doesn't depend on the future call's specifics (backend checks, HTTP-level checks, headers,
JWT claims - including combined with a tool check via `&&`/`||`, since CEL's short-circuiting resolves those independently of the unknown side) while correctly reporting the call-dependent part as ambiguous instead of coercing it to a specific (and potentially wrong) `true`/`false`.

Ambiguity is resolved differently for `Allow` vs `Deny`, asymmetrically on purpose. An ambiguous `Deny` rule is skipped - it can't be proven to fire for whichever tool ends up being requested, so it must not block the backend. An
ambiguous `Allow` rule is not treated as a non-match either; it falls through to the rule's `Target`/`Source` checks instead, rather than being excluded outright. This preserves the invariant that the pre-check may be more
permissive than the real per-tool check but must never be more restrictive. The same reasoning applies to a `Deny` rule scoped to one tool on a backend (`backendMatches` only sees the backend name, so it can't distinguish "deny
this one tool" from "deny the whole backend") - it's skipped for the same reason, regardless of whether the rule also has CEL, since a decided CEL result doesn't resolve the Target's separate tool-scoping ambiguity.

Net effect: a client whose token only grants `mcp:snuser` scope against the `servicenow` backend never causes the proxy to dial `github`, `slack`, etc. during `initialize` - those backends are pruned up front instead of being
connected to and then hidden from `tools/list`.

**Bonus capability, no new code:** the same mechanism generalizes to CEL-driven, dynamic backend/tool subsetting when an upstream `ext_authz`/ ExtProc filter computes per-request access and passes it in a header, e.g.
`x-ai-eg-mcp-tool-subset: database-mcp__read_record,email-mcp__send_alert`. Because `buildCELActivation` already exposes `request.headers[...]` to CEL, and `authorizeBackendOnly` evaluates the same expression with only the tool/method/params unknown, a rule that references the backend and headers only (no tool dependency) is decided during `initialize` exactly as it would be at `tools/call` time - no struct, field, or parsing logic for this header
exists anywhere in the repo; it's ordinary CEL against a header name chosen by the operator. Example rule (verified directly against both `authorizeBackendOnly` and `authorizeRequest`):

```
spec:
  securityPolicy:
    authorization:
      defaultAction: Deny
      rules:
        - action: Allow
          cel: |
            ("x-ai-eg-mcp-tool-subset" in request.headers) &&
            ("," + request.headers["x-ai-eg-mcp-tool-subset"]).contains("," + request.mcp.backend + "__") &&
            ("," + request.headers["x-ai-eg-mcp-tool-subset"] + ",").contains("," + request.mcp.backend + "__" + request.mcp.tool + ",")
```

This must stay one `&&`-joined rule, not split into separate rules per clause - if the exact-FQN-match clause were its own ownerless `Allow` rule, it would come back ambiguous for every backend at `initialize` and unconditionally permit all of them, defeating the backend-prefix clause's pruning. The header must also be stripped from client-supplied request headers and set only by a trusted upstream filter (Envoy's `HeaderMutation`/`Sanitize` or ExtProc header removal) - that enforcement is the operator's responsibility, not this PR's. Since the same rule still gates
`tools/call` at runtime (not just `tools/list`), a client can't bypass subsetting by calling a tool it was never shown.


pbonuboinasnow:
@kanurag94 @mathetake @nacx please review and provide feedback

pbonuboinasnow:
> @kanurag94 @mathetake @nacx please review and provide feedback

I have updated proposal [2359](https://github.com/envoyproxy/ai-gateway/pull/2359) with more details and it can work for dynamic header as well using CELL

internal/mcpproxy/authorization.go
```diff
@@ -420,6 +447,191 @@ func claimHasAllowedString(value any, allowed []string) bool {
...
+		// (request.mcp.backend, auth.jwt.claims, request.headers, etc.) while tool-specific
+		// conditions come back ambiguous.
+		if rule.backendOnlyProgram != nil {
+			if celActivation == nil {
+				// Create authorizationRequest with backend and available headers.
+				// Other fields (HTTPMethod, Host, HTTPPath, MCPMethod, Params, Tool) are
+				// unavailable during initialize phase and left unset; evalBackendOnlyCEL
+				// marks Tool unknown via toolUnknownAttr, which intercepts attribute
+				// resolution before it would reach this map.
+				req := &authorizationRequest{
```
kanurag94, line 559: Only request.mcp.tool is marked unknown. request.method, request.host, request.path, request.mcp.method, request.mcp.params still has their zero values, so rules on those evaluate against "" instead of returning unknown.

For example:

defaultAction: Deny, one Allow rule

- request.method == "POST"            pre=false  request=true
- request.host == "example.com"       pre=false  request=true
- request.path == "/mcp"              pre=false  request=true
- request.mcp.method == "tools/call"  pre=false  request=true

request.method == "POST" is the first example in the CEL doc comment which will start giving 500s.

mergeToolsList has the same gap already so tools/list was wrong for this rule before; initialize and tools/call weren't.

The comment says these are unavailable at initialize. Method, host and path are available.  mcp.method, mcp.tool and mcp.params are the per-request ones that need to be unknown.

internal/mcpproxy/authorization.go
```diff
@@ -420,6 +447,191 @@ func claimHasAllowedString(value any, allowed []string) bool {
...
+			// CEL-bearing rules are NOT skipped by this check: a CEL condition that doesn't
+			// reference request.mcp.tool is fully decidable here regardless of the unknown
+			// Tool attribute, and a CEL condition that DOES reference request.mcp.tool comes
+			// back Unknown via partial evaluation - which the ambiguous-handling above this
+			// Target check already resolved: an ambiguous Deny rule was skipped before ever
+			// reaching here, and an ambiguous Allow rule already fell through knowing it
+			// can't be ruled out. Either way, this check has nothing left to do for
+			// CEL-bearing rules; it only matters for the no-CEL, tool-scoped-target case
+			// described above.
+			if !action && rule.backendOnlyProgram == nil {
```
kanurag94: backendOnlyProgram == nil means Deny rules with CEL skip this
Only the tool is marked unknown, so a CEL that doesn't reference it resolve it and falls through to backendMatches, which ignores Tool.

For example:

 cel: 'request.headers["x-tenant"] == "acme"', target [{backend1, delete_file}], defaultAction: Allow:

- authorizeBackendOnly(backend1)           false
- authorizeRequest(backend1, read_file)    true
- authorizeRequest(backend1, delete_file)  false

 Without the CEL it returns true. The guard is right, it needs to cover CEL rules too.

internal/mcpproxy/mcpproxy.go
```diff
@@ -183,7 +185,23 @@ func (m *mcpRequestContext) newSession(ctx context.Context, p *mcp.InitializePar
...
+	var claims jwt.MapClaims
+	var scopeSet sets.Set[string]
+	if backends.authorization != nil && len(backends.authorization.Rules) > 0 {
+		claims, scopeSet = m.extractClaimsAndScopes(m.requestHeaders, "authorizeBackendOnly")
+	}
+
 	for _, backend := range backends.backends {
+		// Pre-check backend authorization during initialize phase to avoid unnecessary connections.
+		if backends.authorization != nil {
+			if !m.authorizeBackendOnly(backends.authorization, backend.Name, m.requestHeaders, claims, scopeSet) {
```
kanurag94: authorizeBackendOnly returns a bool, so requiredScopesForChallenge is never computed. Pruning the last backend gives 500 instead of 403.

After a token refresh or different headers, authorizeRequest allows but getCompositeSessionEntry is nil, so gives 400. Session id is encrypted, so re-initialize is the only way out.

internal/mcpproxy/authorization.go
```diff
@@ -420,6 +447,191 @@ func claimHasAllowedString(value any, allowed []string) bool {
...
+// backendMatches checks if the backend is in the target tools list (ignoring tool constraints).
+// Used for backend-only authorization pre-checks during the initialize phase.
+func (m *mcpRequestContext) backendMatches(backend string, tools []filterapi.ToolCall) bool {
+	// Empty tools means all backends match.
+	if len(tools) == 0 {
+		return true
+	}
+
+	for _, t := range tools {
+		if t.Backend == backend || t.Backend == "*" {
```
kanurag94: backendMatches accepts "*", toolMatches doesn't. "*" isn't used anywhere else and isn't documented on ToolCall.Backend. Drop it?

kanurag94 (commented):
Thanks for this.
This is a nice alternative to #2351, but is a bit more complicated.

The problem with driving it off the authorization rules is that those rules are only about tools. They say which tools a caller may invoke. Resources and prompts aren't authorized at all right now. So skipping a backend here also takes away its resources and prompts, which nothing ever said the caller couldn't have.

The other issue is timing. We decide at initialize using that request's headers and token, then never look again. If the client refreshes its token later and would now be allowed, the backend is still gone for the rest of the session.

The case where this does work is a backend whose rules don't mention tools at all: no tool in Target, no tool in the CEL. Then "can this caller use this backend" has a real answer and we can prune on it. That version doesn't need the partial-eval program either.

Also, #2351 adds backend pruning to newSession as well, off an explicit spec.backendSelector instead of reading the authorization rules. We shouldn't merge both. cc @PraveenKumarInjam

internal/mcpproxy/mcpproxy.go
pbonuboinasnow: > authorizeBackendOnly returns a bool, so requiredScopesForChallenge is never computed. Pruning the last backend gives 500 instead of 403.
This is taken care of

> After a token refresh or different headers, authorizeRequest allows but getCompositeSessionEntry is nil, so gives 400. Session id is encrypted, so re-initialize is the only way out.

This is not new. I think this is how it has to behave unless we dynamically evaluate, init for delta backends, create new session and return which client has to start using.

internal/mcpproxy/authorization.go
pbonuboinasnow: Fixed. Please review

internal/mcpproxy/authorization.go
pbonuboinasnow: Fixed. Please review

PR #2416, opened 2026-07-23
feat: support custom Bedrock signing host for VPC endpoints

bannu8615:
**Description**

Before this change, the AWS SigV4 signer always built the signing host from the region only, like `bedrock-runtime.<region>.amazonaws.com`. It did not use the real host of the upstream endpoint. This worked for the default public Bedrock endpoint, but it did not work for AWS PrivateLink VPC endpoints (VPCE) or other custom Bedrock hosts, because those hosts do not match the region-based pattern.

This PR makes the signer use the set upstream host when available, with a branch to fall back to the default host:

- The extension server now stamps the resolved upstream host on endpoint metadata. This happens at config-translation time, when a route is set up, not on every request.
- The ext_proc filter reads this metadata and copies the host into an internal header before the request reaches the AWS auth handler.
- The AWS auth handler reads this header to build the signing URL, and to work out the correct signing region from the host. If the header is missing, it falls back to the old region-based default, so existing setups keep working the same way.
- The internal header is always removed from the request before it goes upstream, including on a retry. This means a client cannot set this header itself and change which host gets signed.

All of this is covered by unit tests in `internal/internalapi`, `internal/backendauth`, `internal/extensionserver`, and `internal/extproc`.

**Related Issues/PRs**

This builds on the earlier AWS signing work discussed here https://github.com/envoyproxy/ai-gateway/pull/61#discussion_r1900599982.

- Deployments that never use a custom host are not affected. The handler still falls back to the region-based default when no metadata is present.
- Upstream-host metadata is preferred over cluster-level metadata, because a single cluster can have many endpoints with different real hosts.
- The region-detection regex only self-corrects the signing region. It does not decide which endpoint is safe to sign for; the endpoint host itself is set only by the extension server at config time, never by the request.
- In the `cluster.LoadAssignment == nil` branch no LbEndpoint is present to stamp as the upstream host. Thread: https://github.com/envoyproxy/ai-gateway/pull/2416#discussion_r3712421047 .


mathetake:
Posting the comment by @nacx here:

Since the previous PR broke the e2e tests that use real credentials, please run those real provider tests and copy the results here.

pbonuboinasnow:
> Since the previous PR broke the e2e tests that use real credentials, please run those real provider tests and copy the results here.

Here are the results. We had to make few local changes to make test suite as we use ARN for model names, GCPVertexAI and AzureOpenAI. Let us know if we are missing anything

[run_real_provider_tests.sh](https://github.com/user-attachments/files/30371989/run_real_provider_tests.sh)
[log.txt](https://github.com/user-attachments/files/30372194/log.txt)

kanurag94:
Nice approach — Two small suggestions from a parallel attempt I'm closing (#2429):

1. **Derive the SigV4 region from the host too.** The host is now dynamic but the credential scope still uses the configured `region`, so a config/endpoint region mismatch (e.g. `region: us-west-2` but a `us-east-1` VPCE) can still caus 403s. A small regex over the bedrock host — `bedrock-runtime.<region>.amazonaws.com`, including the `…vpce.amazonaws.com` form — can let a wrong/typo'd region self-correct. Non-bedrock hosts can  just keep the configured region.

2. **Add a test that recomputes SigV4 over the resolved host.** Nothing currently asserts the signature is actually computed over the host from resolved string — same issue as #2296. Can we sign with a static key/time, then compare `Authorization` against an independent recompute (vpce host vs region default). It's credential-free, so even a fork CI can run it.

internal/extensionserver/post_translate_modify.go
```diff
@@ -381,6 +381,9 @@ func (s *Server) maybeModifyCluster(ctx context.Context, cluster *clusterv3.Clus
 				}
 				for _, endpoint := range endpoints.LbEndpoints {
 					setEndpointMetadataBackendName(endpoint, namespace, name, aigwRoute.Name, httpRouteRuleIndex, i)
+					if host := endpointAWSSigningHost(endpoint); host != "" {
+						setEndpointMetadataAWSSigningHost(endpoint, host)
+					}
```
kanurag94: This is the only place that stamps the signing host. But the case `cluster.LoadAssignment == nil` and  `clusterName.backendRefIndex != noBackendRefIndex` only sets the backend name and not signing host. This can make a AWS backend routed through sign over `:authority` and produce 403s -- same as #2296. WDYT about stamping the host wherever the backend name is accepted?
Also can we add `setClusterMetadataAWSSigningHost` similar to `setClusterMetadataBackendName` for the nil assignment.

internal/extproc/server.go
```diff
@@ -434,6 +435,24 @@ func resolveRouteName(attributes *structpb.Struct) string {
...
+func setAWSSigningHost(headers map[string]string, attributes *structpb.Struct) {
+	if attributes == nil {
+		return
+	}
+	host := ""
+	if v, ok := attributes.Fields[internalapi.XDSUpstreamHostMetadataAWSSigningHostPath]; ok {
+		host = v.GetStringValue()
+	}
+	if host == "" {
+		if v, ok := attributes.Fields[internalapi.XDSClusterMetadataAWSSigningHostPath]; ok {
```
kanurag94: I think this cluster metadata fallback is dead code. I couldn't find anything that populates cluster level `aws_signing_host`. The comment https://github.com/envoyproxy/ai-gateway/pull/2416/changes#r3654733473 here should make it live

internal/backendauth/aws.go
```diff
@@ -127,3 +130,28 @@ func (a *awsHandler) Do(ctx context.Context, requestHeaders map[string]string, m
 	}
 	return headers, nil
 }
+
+func (a *awsHandler) signingHost(requestHeaders map[string]string) string {
+	host := requestHeaders[internalapi.AWSSigningHostHeader]
+	if host == "" {
+		host = requestHeaders[":authority"]
```
kanurag94: At signing time `:authority` is the downstream gateway authority, not the upstream endpoint so this sets the wrong host.  They're only reachable when metadata is absent, so they turn a config gap into a silent 403.
Can we drop both tiers and fall straight to region default when metadata header is absent.

internal/backendauth/aws.go
```diff
@@ -127,3 +130,28 @@ func (a *awsHandler) Do(ctx context.Context, requestHeaders map[string]string, m
...
+	}
+	if host == "" {
+		host = requestHeaders["host"]
+	}
+	if host == "" {
+		host = fmt.Sprintf("bedrock-runtime.%s.amazonaws.com", a.region)
```
kanurag94: Host is now dynamic but the SigV4 credential scope in `newAWSHandler` still uses the configured region  so a region/endpoint mismatch (e.g. region us-west-2 but a us-east-1 VPCE) can give 403s. Can you see if we can derive the host from from bedrock host  (bedrock-runtime.<region>.amazonaws.com, incl. the vpce form) so it can self correct?

internal/extensionserver/post_translate_modify.go
```diff
@@ -929,6 +934,20 @@ func routeNameFromEnvoyGatewayMetadata(route *routev3.Route) string {
...
+	}
+	if hostname := endpoint.GetHostname(); hostname != "" {
+		return hostname
+	}
+	if socketAddress := endpoint.GetAddress().GetSocketAddress(); socketAddress != nil {
+		return socketAddress.GetAddress()
+	}
+	return ""
+}
+
```
kanurag94: The socketAddress fallback stamps a bare IP as the signing host. Signing over an IP won't match any AWS endpoint. Do you think we can just use GetHostname() and skip IP/ address fallbacks.

internal/backendauth/aws.go
kanurag94: This seems to pass the region from the signer and not the actual request, can we infer from request correctly:
```
var bedrockHostRE = regexp.MustCompile(`(?:^|\.)bedrock-runtime\.([a-z0-9-]+)\.(?:vpce\.)?amazonaws\.com$`)

func regionFromBedrockHost(hostname string) string {
      if m := bedrockHostRE.FindStringSubmatch(hostname); m != nil {
              return m[1]
      }
      return ""
}

host := a.signingHost(requestHeaders)
region := a.region
if r := regionFromBedrockHost(host); r != "" {
    region = r
}
```

internal/backendauth/aws.go
kanurag94: https://github.com/envoyproxy/ai-gateway/pull/2416/changes#r3711340846

internal/extensionserver/post_translate_modify.go
bannu8615: From what I see:

```go
type BackendEndpoint struct {
    Hostname *string       // optional hostname override, independent of the three below
    FQDN  *FQDNEndpoint    // fqdn:  { hostname (required), port }
    IP    *IPEndpoint      // ip:    { address, port }
    Unix  *UnixSocket      // unix:  { path }
    Zone  *string          // optional
}
```

1. An endpoint is a one-of `{fqdn | ip | unix}`; only `fqdn` surfaces a hostname into the xDS
   endpoint (STRICT_DNS). `ip`/`unix` don't.
2. There are therefore **two** no-hostname shapes, and they hit different branches:
   - `LoadAssignment == nil` — EDS cluster, no inline endpoints (e.g. a future K8s Service
     backendRef, #902). There's no `LbEndpoint` to iterate, so `stampAWSSigningMetadata` is
     never reached — this branch.
   - a non-`fqdn` inline endpoint (`ip`/STATIC, #950) — `LoadAssignment != nil`, so
     `stampAWSSigningMetadata` *is* called, but `endpointAWSSigningHost` returns `""` and it skips.
3. Both fall back to the region-default host at signing time; the real fix reads
   `spec.endpoints[].fqdn.hostname` from the Backend CR via the backendRef.

internal/backendauth/aws.go
bannu8615: I think you meant extracting the region from the `host` info. 
Implemented a regex for the same : [Reference](https://github.com/envoyproxy/ai-gateway/pull/2416/changes#diff-fb5ef3630fee19cfda34448c1dd562052b3dc9b96508cf21252d4b3b4fd574adR58)

internal/extproc/server.go
```diff
@@ -434,6 +436,25 @@ func resolveRouteName(attributes *structpb.Struct) string {
 	return ""
 }
 
+func setAWSSigningAttributes(headers map[string]string, attributes *structpb.Struct) {
+	if attributes == nil {
+		return
+	}
+
```
kanurag94, line 449: can we delete before setting so a client supplied header doesn't overwrite?

internal/extensionserver/post_translate_modify.go
```diff
@@ -51,6 +52,11 @@ type aiGatewayClusterName struct {
 	backendRefIndex int
 }
 
+// awsBedrockRegionRegexp extracts the region from a Bedrock runtime host, including the PrivateLink
+// (VPCE) form. Both bedrock-runtime.us-east-1.amazonaws.com and
+// vpce-<id>.bedrock-runtime.us-east-1.vpce.amazonaws.com yield "us-east-1".
+var awsBedrockRegionRegexp = regexp.MustCompile(`bedrock-runtime\.([a-z0-9-]+)\.(?:vpce\.)?amazonaws\.com`)
```
kanurag94: nit: What about `(?:^|\.)…amazonaws\.com$.`

internal/extensionserver/post_translate_modify.go
```diff
@@ -341,9 +347,23 @@ func (s *Server) maybeModifyCluster(ctx context.Context, cluster *clusterv3.Clus
...
+			// `fqdn` surfaces a hostname into the xDS endpoint (STRICT_DNS) — that case is stamped in the
+			// inline branches below. This EDS branch is the future K8s Service backendRef (currently
+			// CEL-blocked, ai-gateway#902): its endpoints are pod IPs, which SigV4 cannot sign. (The `ip`
+			// Backend variant, ai-gateway#950, is the sibling no-hostname case, but it yields a STATIC
+			// cluster with an inline load_assignment, so it is handled in the inline branches where an
+			// empty host is skipped — it does not reach here.) Both fall back to the region-based default
+			// host at signing time. The real fix resolves the hostname from the Backend CR
+			// (spec.endpoints[].fqdn.hostname) via the backendRef rather than the translated endpoint;
+			// feeding EDS endpoints to the extension server would need envoyproxy/gateway#8150
+			// (PostEndpointsModify).
```
kanurag94: nit: can you file an issue instead?
```suggestion
			// TODO(aws-signing): no LbEndpoint to stamp here, so an AWS backend on an EDS cluster falls
			// back to the region-based default signing host and a VPCE host is lost. See ai-gateway#XXXX.
```

internal/backendauth/aws.go
kanurag94: yes, and doing it in the handler instead.
It should delete InternalMetadataAWSSigningRegionKey, XDSUpstreamHostMetadataAWSSigningRegionPath, AWSSigningRegionHeader, setEndpointMetadataAWSSigningRegion, the region branch in setAWSSigningAttributes, signingRegion, and envoy.yaml change

kanurag94: Currently both host and region are implemented parallely. But we don't need the region half as it is already inferred from host

internal/extensionserver/post_translate_modify.go
bannu8615: I will shorten the comment. Regarding filing an issue, I see if the dependent fields are updated in the issues that already open, we should be able to update code here as required.

internal/extensionserver/post_translate_modify.go
bannu8615: I have updated the regex.

internal/extproc/server.go
bannu8615: Yes that is better. Added it.

internal/backendauth/aws.go
bannu8615: I have moved the regex logic to `aws.go` and removed other pieces of code is cleaned up.

internal/extensionserver/post_translate_modify.go
bannu8615: Added the check to ignore IP before returning `socketAddress.GetAddress()` https://github.com/envoyproxy/ai-gateway/pull/2416/changes#diff-fb5ef3630fee19cfda34448c1dd562052b3dc9b96508cf21252d4b3b4fd574adR968

This helps to het the hostname from two levels,

```
| Order | Field                                  | Condition               | YAML path                               |
|-------|----------------------------------------|-------------------------|-----------------------------------------|
| 1     | Endpoint.Hostname                      | non-empty               | endpoint.hostname                       |
| 2     | Endpoint.Address.SocketAddress.Address | non-empty and not an IP | endpoint.address.socket_address.address |
```

internal/backendauth/aws.go
bannu8615: https://github.com/envoyproxy/ai-gateway/pull/2416#discussion_r3719435842

internal/backendauth/aws.go
bannu8615: Removed this piece of code that was considering `":authority"`.

.gitignore
```diff
@@ -43,6 +43,9 @@ site/temp
 # auto generated conformance report
 inference-extension-conformance-test-report.yaml
 
+# e2e debug log export (kind/pod logs dumped on failure)
+tests/e2e/logs-export/
```
kanurag94: nit: can you revert this please?

.licenserc.yaml
```diff
@@ -25,6 +25,7 @@ header:
     - "**/*.md"
     - "**/*.json"
     - "**/*.txt"
+    - "**/*.log"
```
kanurag94: nit: can you revert this please?

internal/extensionserver/post_translate_modify.go
```diff
@@ -939,6 +947,39 @@ func routeNameFromEnvoyGatewayMetadata(route *routev3.Route) string {
...
+	}
+	return ""
+}
+
+// stampAWSSigningMetadata stamps the SigV4 signing host on the endpoint's metadata at config-translation
+// time, so the data plane signs over the real upstream endpoint instead of re-deriving it at request
+// time. The signing region is derived from this host later, in the AWS backend auth handler.
+func stampAWSSigningMetadata(endpoint *endpointv3.LbEndpoint) {
+	if host := endpointAWSSigningHost(endpoint); host != "" {
+		setEndpointMetadataAWSSigningHost(endpoint, host)
```
kanurag94: I think this gets set on every request. Even if it is for openai.com.
Can we use a non aws specific header instead. Maybe `x-ai-eg-upstream-host`

internal/extproc/server.go
```diff
@@ -434,6 +436,24 @@ func resolveRouteName(attributes *structpb.Struct) string {
 	return ""
 }
 
+func setAWSSigningAttributes(headers map[string]string, attributes *structpb.Struct) {
+	// The signing host header is an internal, controller-derived value. A downstream client could also
+	// send it, so drop any inbound copy before overlaying trusted xDS metadata — otherwise, for an
+	// endpoint with no derivable metadata (non-Bedrock host, EDS), the client-supplied value would
+	// survive and drive the SigV4 host. The region is derived from this host in the AWS handler.
+	delete(headers, internalapi.AWSSigningHostHeader)
```
kanurag94: is it possible to add an e2e for this? So that a spoofed header isn't seen at upstream?

internal/extensionserver/post_translate_modify.go
```diff
@@ -361,6 +366,7 @@ func (s *Server) maybeModifyCluster(ctx context.Context, cluster *clusterv3.Clus
 				}
 				for _, endpoint := range endpoints.LbEndpoints {
 					setEndpointMetadataBackendName(endpoint, aigwRoute.Namespace, backendRef.Name, aigwRoute.Name, httpRouteRuleIndex, clusterName.backendRefIndex)
+					stampAWSSigningMetadata(endpoint)
```
kanurag94: similarly we can make this non aws agnostic since it is present for all endpoints

internal/backendauth/aws.go
```diff
@@ -127,3 +138,43 @@ func (a *awsHandler) Do(ctx context.Context, requestHeaders map[string]string, m
...
+// regionFromBedrockHost returns the AWS region encoded in a Bedrock signing host, or "" for a
+// non-Bedrock host (in which case the caller keeps the configured region).
+func regionFromBedrockHost(hostname string) string {
+	if m := bedrockHostRE.FindStringSubmatch(hostname); m != nil {
+		return m[1]
+	}
+	return ""
+}
+
+func normalizeAWSSigningHost(host string) string {
```
kanurag94: nit: this may never be hit with `host:port` and only `host`

internal/backendauth/aws.go
```diff
@@ -85,6 +87,14 @@ func newAWSHandler(ctx context.Context, awsAuth *filterapi.AWSAuth) (filterapi.B
 func (a *awsHandler) Do(ctx context.Context, requestHeaders map[string]string, mutatedBody []byte) ([]internalapi.Header, error) {
 	method := requestHeaders[":method"]
 	path := requestHeaders[":path"]
+	host := a.signingHost(requestHeaders)
+	// Derive the SigV4 region from the resolved Bedrock host so the credential scope self-corrects to
+	// the endpoint's region (e.g. a us-east-1 VPCE under a us-west-2 configured region). Non-Bedrock
+	// hosts keep the configured region.
```
kanurag94: nit: I think credential retrieval will not self heal. Fine for non bedrock hosts

bannu8615: Correct, updated the comment accordingly.

internal/backendauth/aws.go
bannu8615: Removed `normalizeAWSSigningHost`.

internal/extensionserver/post_translate_modify.go
bannu8615: Adding a generic header `x-ai-eg-upstream-host` feels broader in scope than what the changes in the PR are for.
We will now set aws signing host only if matches the AWS host regex.
This ensures , the new attribute is not added for every request.

internal/extensionserver/post_translate_modify.go
bannu8615: https://github.com/envoyproxy/ai-gateway/pull/2416#discussion_r3728208237

.gitignore
bannu8615: This change is now reverted.

.licenserc.yaml
bannu8615: This change is now reverted.

internal/extproc/server.go
bannu8615: Added a test case in this commit https://github.com/envoyproxy/ai-gateway/pull/2416/commits/ff6dc7b68407bfdf6786b2f1d6cf37efad49f9a4

internal/extensionserver/post_translate_modify.go
kanurag94: I see that you are now stamping on the regex which fails for a bunch of strings:

```
bedrock-runtime-fips.us-east-1.amazonaws.com
bedrock-runtime.us-east-1.api.aws
bedrock.corp.internal
```

This reverts the original intent of this PR itself

internal/extensionserver/post_translate_modify.go
```diff
@@ -939,6 +947,40 @@ func routeNameFromEnvoyGatewayMetadata(route *routev3.Route) string {
...
+	}
+	return ""
+}
+
+// stampAWSSigningMetadata stamps the SigV4 signing host on the endpoint's metadata at config-translation
+// time, so the data plane signs over the real upstream endpoint instead of re-deriving it at request
+// time. It only stamps AWS Bedrock endpoints (public or VPCE hosts); other backends (e.g. api.openai.com)
+// do not need a signing host and must not carry a misleading aws_signing_host on every endpoint.
+func stampAWSSigningMetadata(endpoint *endpointv3.LbEndpoint) {
+	if host := endpointAWSSigningHost(endpoint); internalapi.AWSBedrockRegionFromHost(host) != "" {
```
kanurag94: Here the regex doesn't match all possible endpoints. 
The gate only exists because the field is called aws_signing_host and we don't want it showing up on an OpenAI endpoint. Rename it to upstream_host and x-ai-eg-upstream-host  is a small correct change I feel.

internal/extproc/processor_impl.go
```diff
@@ -394,6 +394,11 @@ func (u *upstreamProcessor[ReqT, RespT, RespChunkT, EndpointSpecT]) ProcessReque
 		u.requestHeaders[h.Header.Key] = string(h.Header.RawValue)
 	}
 
+	// x-ai-eg-aws-signing-host is an internal, controller-derived value consumed by the AWS backend
+	// auth handler (via the upstream-host metadata attribute). Strip any copy — including one a
+	// downstream client spoofed — so it never egresses to the upstream provider.
+	headerMutation.RemoveHeaders = append(headerMutation.RemoveHeaders, internalapi.AWSSigningHostHeader)
```
kanurag94: Thanks for adding this.

Can you also add a unit test where a spoofed `x-ai-eg-aws-signing-host` gets removed with `onRetry() true`?

internal/extensionserver/post_translate_modify.go
bannu8615: Fixed in https://github.com/envoyproxy/ai-gateway/pull/2416/commits/94d649f579492c0cb0d9fb700c0a9426c2211f25

internal/extproc/processor_impl.go
bannu8615: Added the UT to verify removal of spoofed case in https://github.com/envoyproxy/ai-gateway/pull/2416/commits/fe5191d055ebd3a806e9b6b57be7a79745fb035f

internal/extensionserver/post_translate_modify.go
bannu8615: Updated the header name and code references for generic name in https://github.com/envoyproxy/ai-gateway/pull/2416/commits/fe5191d055ebd3a806e9b6b57be7a79745fb035f

kanurag94 (approved):
Thanks, LGTM!

johnugeorge (approved):

johnugeorge:
Thanks @bannu8615

Issue #2411, opened 2026-07-22
Gateway-side `anthropic-beta` compatibility filtering across upstream providers


hustxiayang:
## Motivation

Anthropic API clients, such as Claude Code, send their enabled betas in the `anthropic-beta` request header (comma-separated). Upstream providers (GCP Vertex, AWS Bedrock) each support a *different shifting* subset of those values, and reject unknown ones with an HTTP 400 that, absent any client-side per-turn retry, terminates the turn and surfaces a raw error to the user.

**How the header works.** A client sends all its enabled betas together as a single comma-separated `anthropic-beta` header. The gateway currently forwards this header to the upstream provider unchanged (passthrough), so any value the provider does not support reaches it and is rejected with a 400.

**The failure.** Take the concrete case of enabling tool search in Claude Code. Because there is no per-value control, turning it on causes Claude Code to send its full experimental beta set in the `anthropic-beta` header — the tool-search value (`advanced-tool-use-2025-11-20`) alongside others such as `thinking-token-count-2026-05-13`. Vertex accepts the tool-search value but does not recognize `thinking-token-count-2026-05-13`, and rejects the entire request:

```
400 invalid_request_error
"Unexpected value(s) `thinking-token-count-2026-05-13` for the `anthropic-beta` header."
```

Because all betas share one header, this single unsupported value fails the whole request — even though the feature actually wanted (tool search) is itself supported. Claude Code does not retry the turn, so the turn dies and the user sees the raw 400. The net effect: tool search cannot be enabled on the Vertex path, because turning it on necessarily sends a value that Vertex rejects.

## Proposed change

This class of failure can be prevented in ai-gateway. As in [#2040](https://github.com/envoyproxy/ai-gateway/pull/2040), we cannot control GCP's behavior, but we can explicitly drop unsupported beta values before forwarding upstream. This achieves three things:

- **No unsupported value can fail a turn** — on any Anthropic upstream, not just the one value that blocks tool search today.
- **Durability as both sides change** — when Anthropic ships new betas (which clients send automatically), or a provider starts or stops accepting a value, the filter adapts through a config change, not a code deploy. It is a general mechanism, not a one-off patch for today's bad value.
- **Tool search unblocked** on the Vertex path, as the first concrete beneficiary.

### Filtering strategy: denylist vs allowlist

Two mechanisms, opposite failure modes:

| | **Denylist** (drop known-bad) | **Allowlist** (keep sanctioned only) |
|---|---|---|
[...]

PR #2420, opened 2026-07-23
feat: add per-backend header value filter

hustxiayang:
**Description**

Anthropic clients (e.g. Claude Code) send all enabled betas in a single comma-separated `anthropic-beta` header. Some upstream providers reject specific values with an HTTP 400 (e.g. GCP Vertex rejecting `thinking-token-count-2026-05-13`), and because clients don't retry per-turn, one unsupported value fails the whole request — even when the feature actually wanted (e.g. tool search, `advanced-tool-use-2025-11-20`) is itself supported. Check https://github.com/envoyproxy/ai-gateway/issues/2411 for more details.

This adds an optional, per-backend `headerValueFilters` field to `AIServiceBackend`: a list of filters, at most one per header name, that drop (`Denylist`) or restrict to (`Allowlist`) individual values of a comma-separated request header before the request is forwarded upstream. The rest of the header is forwarded untouched, so the gateway reacts to whatever the client actually sent rather than replacing the header with a fixed value.

### Example

```yaml
kind: AIServiceBackend
spec:
  schema:
    name: GCPAnthropic
  headerValueFilters:
    - name: anthropic-beta
      mode: Denylist            # Denylist (default) | Allowlist
      values: ["thinking-token-count-2026-05-13"]
```

### Why not `HTTPHeaderMutation`?

`HTTPHeaderMutation.Set` can already force `anthropic-beta` to a fixed literal value, which covers a narrower case. It doesn't read the client's actual per-request header, so it can't selectively drop one known-bad value while passing through everything else the client sent (including future beta values not yet known at config time). This filter reconciles against the client's real request instead of replacing it wholesale.

### Scope

- Applies to the `/v1/messages` (Anthropic Messages) path only. Whether the chat-completions path needs the same is still open.
- Honored by `GCPAnthropic` and `AWSAnthropic` backends; a CEL rule rejects the field on any other schema.
- Today only the `anthropic-beta` header is actually filtered by the translators — a filter naming another header is accepted but has no effect (use `HeaderMutation` for those).


api/v1beta1/shared_types.go
```diff
@@ -165,6 +165,26 @@ const (
...
+// comma-separated header, but upstream providers accept different subsets and reject unknown values
+// with an HTTP 400. This lets an operator drop (or restrict to) specific values per backend.
+type AnthropicBetaFilter struct {
+	// Mode selects the filtering semantics:
+	//   - "denylist": drop the values listed in Values, keep everything else.
+	//   - "allowlist": keep only the values listed in Values, drop everything else.
+	//
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Enum=denylist;allowlist
+	Mode string `json:"mode"`
```
siddharth1036: Should we make separate type for Mode?

api/v1beta1/ai_service_backend.go
```diff
@@ -76,6 +76,11 @@ type AIServiceBackendSpec struct {
 	// +optional
 	BodyMutation *HTTPBodyMutation `json:"bodyMutation,omitempty"`
 
+	// AnthropicBetaFilter filters values from the `anthropic-beta` request header before sending the
+	// request to this backend. Only applies to Anthropic backends (GCP Vertex and AWS Bedrock).
+	// +optional
+	AnthropicBetaFilter *AnthropicBetaFilter `json:"anthropicBetaFilter,omitempty"`
```
siddharth1036: Should we add it to alpha CRD as well?

api/v1beta1/shared_types.go
aabchoo: what's the use case for mode? don't we already know the values allowed for each provider?

hustxiayang: `don't we already know the values allowed for each provider`. Anthropic keeps releasing new beta values continuously, and each provider picks up a different subset on its own schedule.

hustxiayang: sure, it's definitely better if we do keep it.

api/v1beta1/ai_service_backend.go
hustxiayang: i thought alpha is deprecated? `// +kubebuilder:deprecatedversion:warning="aigateway.envoyproxy.io/v1alpha1 is deprecated; use aigateway.envoyproxy.io/v1beta1 instead"`

api/v1beta1/shared_types.go
hustxiayang: `What's the use case for mode?` — currently, it supports two modes: `denylist` and `allowlist`:

` denylist`: strip the one known-broken value, let everything else through. Client betas we've never heard of keep working with  no config change. This is the motivating case.
`allowlist`: only a vetted set reaches upstream. For a strict backend or a conservative operator, an unexpected client beta  can't cause a 400 or unvetted behavior.

Users could choose different options based on their needs, but one option should be enough as well.

aabchoo:
Could you make this generic to remove headers based on a key and values?

hustxiayang:
@aabchoo Sure, it makes sense not to patch anthropic specific data into a generic structure. I would define a generic data structure `HTTPHeaderValueFilter` and add a `name` field into it. For  anthropic, it would be called `anthropic_beta`. In this way, it should be generic enough?

api/v1beta1/shared_types.go
hustxiayang: updated.

api/v1beta1/shared_types.go
```diff
@@ -165,26 +165,61 @@ const (
...
-// comma-separated header, but upstream providers accept different subsets and reject unknown values
-// with an HTTP 400. This lets an operator drop (or restrict to) specific values per backend.
-type AnthropicBetaFilter struct {
-	// Mode selects the filtering semantics:
-	//   - "denylist": drop the values listed in Values, keep everything else.
-	//   - "allowlist": keep only the values listed in Values, drop everything else.
+// HTTPHeaderValueFilter filters individual values out of a multi-valued request header before the
+// request is forwarded to this backend.
+//
+// Clients often send several values in a single comma-separated header, and upstream providers
```
aabchoo: 171 - 178 seems unnecessarily verbose can remove

internal/translator/anthropic_awsanthropic.go
```diff
@@ -45,11 +45,17 @@ type anthropicToAWSAnthropicTranslator struct {
...
+	a.anthropicBetas = parseCommaSeparatedHeader(headers, anthropicBetaHeaderName)
 }
 
-// SetAnthropicBetaFilter implements [AnthropicBetaFilterSetter].
-func (a *anthropicToAWSAnthropicTranslator) SetAnthropicBetaFilter(mode string, values []string) {
+// SetHeaderValueFilter implements [HeaderValueFilterSetter]. Only anthropic-beta is handled here:
+// Bedrock reads these values from the request body rather than the header, so the filtered set has
+// to be mirrored into the body's anthropic_beta field by RequestBody below. Filters on any other
+// header are applied by Envoy's header mutation instead.
+func (a *anthropicToAWSAnthropicTranslator) SetHeaderValueFilter(name, mode string, values []string) {
```
aabchoo, line 124: this implies that any header that is not anthropic beta will be ignored? this should not be the case?

aabchoo: this feels still too specific to anthropic beta -- isn't the point to filter out header values based on the list you defined in the cr?

hustxiayang: I am a bit hesitant on this. Currently there is only anthropic beta use case, so I did not want to allow any other headers filter because it would make no sense for any custom filters. If there is use case for other filters in the future, we could add it explicitly?

api/v1beta1/shared_types.go
hustxiayang: would update it

internal/translator/anthropic_awsanthropic.go
aabchoo: Wouldn't mean we require code change for each time config-admins need to filter a specific header? Feels like it defeats the purpose of making it generic

hustxiayang: `Feels like it defeats the purpose of making it generic` This is just to prevent a header filter which does not exist in reality. By generic, I think it means it can support other filters when the filters do exist?

hustxiayang: but I do feel I should remove the check:
```
 x-kubernetes-validations:
            - message: headerValueFilters is only honored by GCPAnthropic and AWSAnthropic
                backends
              rule: '!has(self.headerValueFilters) || self.schema.name == ''GCPAnthropic''
                || self.schema.name == ''AWSAnthropic'''
```
in file manifests/charts/ai-gateway-crds-helm/templates/aigateway.envoyproxy.io_aiservicebackends.yaml...

johnugeorge:
As we discussed, https://github.com/envoyproxy/ai-gateway/pull/2512#issuecomment-5268681077 is not the best solution. But, adding to AIServiceBackend is also not scalable in adding capabilities. anthropicBetaFilter is also odd to make it as a general field in AIServiceBackend

internal/translator/anthropic_awsanthropic.go
aabchoo: I'm advocating for a generic logic that allows/prevents specific header/value combinations from being passed in the request. It's not anthropic specific and I don't think we can assume what are valid header/values to remove. 

Ie I could specific key: `service_tier`  values: [`default`, `priority`] and mode: `allowList` -> this will make all requests sent to my gateway allow those service tiers but not others.

aabchoo: would this change be acceptable to you w.r.t your concerns @johnugeorge -- this change would be generic and not anthropic specific

hustxiayang:
> anthropicBetaFilter is also odd to make it as a general field in AIServiceBackend

it's no longer `anthropicBetaFilter`

aabchoo (approved):
Approving this PR to unblock Claude Code features. This change is technically two parts: translation and API but I will help Yang put a follow up PR to clean up the generic implementation for HeaderValuesFilter.

Issue #2422, opened 2026-07-24
MCP: api-key forwardClientIDHeader identity is not surfaced to MCP metrics/spans/access logs (only OAuth claimToHeaders is)


mohitgurnani:
## Description

When an `MCPRoute` is protected with **api-key auth** and `apiKeyAuth.forwardClientIDHeader` is set, the matched client id is **not** surfaced to the MCP proxy request, so it never appears on MCP metrics (`mcp_*` `mcpKeyName`/configured header attribute), spans, or access logs. Authentication itself works (bad/missing keys are rejected), but every authenticated request is recorded as an empty caller.

By contrast:

- The **LLM path** attributes the caller correctly: `SecurityPolicy.apiKeyAuth.forwardClientIDHeader: x-llm-client-id` + `--requestHeaderAttributes=x-llm-client-id:client.id` yields a populated `client.id` on `gen_ai_*` metrics. The LLM request is processed by the in-chain extproc filter on the same Envoy request, so it sees the header EG added.
- The **OAuth path** for MCP has an explicit bridge so claims reach the proxy: the JWT provider's `ClaimToHeaders` injects request headers (consumed by `mcpMetrics.WithRequestAttributes(r)`), and `mcpConfig` additionally forwards them to backends (`internal/controller/gateway.go`, `mcpRoute.ForwardHeaders = append(..., ctoh.Header)`).

There is **no equivalent path for api-key auth**: nothing makes the `forwardClientIDHeader` identity available to the MCP proxy request, and `#1342` (MCP metrics/spans honoring `requestHeaderAttributes`) can therefore never attribute an api-key caller.

## Environment

- `ai-gateway-controller`: `v1.0.0` (contains #1342, verified: `220b90c` is an ancestor of `v1.0.0`)
- Envoy Gateway `v1.7.0`, Envoy `distroless-v1.37.0`
- Controller flag deployed: `--requestHeaderAttributes=x-llm-client-id:client.id,x-mcp-client-id:mcpKeyName`
- `MCPRoute.securityPolicy.apiKeyAuth` with `forwardClientIDHeader: x-mcp-client-id`, `sanitize: true`

## Expected

An authenticated MCP request over api-key auth is attributed to the matched credential id (like the LLM path): the configured header attribute (e.g. `x-mcp-client-id:mcpKeyName`) is populated on `mcp_*` metrics/spans, and the id is available for access logs.

## Actual

[...]

PR #2423, opened 2026-07-24
mcp: forward api-key client-id header to backends (parity with OAuth)

mohitgurnani:
**Description**

When an `MCPRoute` is secured with `apiKeyAuth` and `apiKeyAuth.forwardClientIDHeader` is set, the matched caller id is never bridged into the MCP proxy config. As a result the id has no path to the backends, nor to the MCP request-attribute plumbing that populates metrics/spans/access-logs (`#1342`). Authenticated api-key MCP requests are therefore recorded with an **empty** caller, while the OAuth path attributes correctly.

The OAuth path already has an explicit bridge in `mcpConfig` (`internal/controller/gateway.go`):

```go
// Forward OAuth claim-to-header mappings to all backends in this route.
if route.Spec.SecurityPolicy != nil && route.Spec.SecurityPolicy.OAuth != nil {
    for _, ctoh := range route.Spec.SecurityPolicy.OAuth.ClaimToHeaders {
        mcpRoute.ForwardHeaders = append(mcpRoute.ForwardHeaders, ctoh.Header)
    }
}
```

There is no equivalent for `apiKeyAuth.forwardClientIDHeader`. This PR adds one, mirroring the OAuth bridge: when `SecurityPolicy.APIKeyAuth.ForwardClientIDHeader` is set, its header name is appended to the route's `ForwardHeaders`.

**Refs #2422**

## Why

- api-key auth is the only auth mode where per-caller attribution comes from the key itself (key == tenant/department). Without this bridge the caller identity injected by `apiKeyAuth.forwardClientIDHeader` never reaches the MCP backends and cannot be surfaced by the MCP request-attribute mapping, so `mcp_*` metrics/spans/logs record an empty caller.
- Brings api-key auth to parity with OAuth (`claimToHeaders`), which is already forwarded.

## Changes

- `internal/controller/gateway.go`: in `mcpConfig`, bridge `apiKeyAuth.forwardClientIDHeader` into `mcpRoute.ForwardHeaders` (guarded on non-nil / non-empty), directly mirroring the OAuth `claimToHeaders` handling immediately above it.
- `internal/controller/gateway_test.go`: `Test_mcpConfig_APIKeyForwardClientIDHeader` covering set / empty / unset / no-security-policy.

## Testing

```
$ go test ./internal/controller/ -run 'Test_mcpConfig_APIKeyForwardClientIDHeader|Test_mcpConfig_ForwardHeaders' -v
--- PASS: Test_mcpConfig_ForwardHeaders
--- PASS: Test_mcpConfig_APIKeyForwardClientIDHeader
    --- PASS: .../forwards_the_api-key_client-id_header_to_backends
    --- PASS: .../no_client-id_header_configured
    --- PASS: .../empty_client-id_header_is_ignored
    --- PASS: .../no_security_policy
PASS
```

`gofmt` clean, `go vet ./internal/controller/` clean.

## Notes / scope

- This is the controller-side bridge (parity with OAuth). The broader item in #2422 — ensuring the `forwardClientIDHeader` value is present on the request the standalone MCP proxy processes so `WithRequestAttributes` promotes it to a metric/span label (the analog of how JWT `ClaimToHeaders` surfaces claims to the proxy request) — may involve the extension-server route handling and/or Envoy Gateway's `api_key_auth` behaviour; happy to extend this PR based on maintainer guidance.
- Existing e2e (`tests/e2e/mcp_route_api_key_auth_test.go` / `testdata/mcp_route_api_key_auth.yaml`) only asserts auth enforcement and does not set `forwardClientIDHeader`. A full e2e assertion of the forwarded caller id would need the test MCP server to surface received request headers (the current echo tool only echoes the text argument); called out as a follow-up.

By submitting this PR, I confirm the contribution is made under the terms of the DCO (commit is signed off).


nacx (approved):
Thanks!

nacx:
Can you fix the failing job about hte PR description?

nacx:
Can you take a look at the conflict?

mohitgurnani:
> Can you take a look at the conflict?

Thanks @nacx for approving it, I have resolved conflicts and PR description.

internal/controller/gateway.go
```diff
@@ -688,6 +688,16 @@ func mcpConfig(mcpRoutes []aigv1b1.MCPRoute) (_ *filterapi.MCPConfig, hasEffecti
...
+		// route, mirroring the OAuth claim-to-header bridge above. Without this the
+		// caller id injected by apiKeyAuth.forwardClientIDHeader has no path to the
+		// MCP proxy request, so it never reaches the backends nor the MCP
+		// request-attribute plumbing (metrics/spans/logs). See #2422.
+		if route.Spec.SecurityPolicy != nil && route.Spec.SecurityPolicy.APIKeyAuth != nil {
+			if h := route.Spec.SecurityPolicy.APIKeyAuth.ForwardClientIDHeader; h != nil && *h != "" {
+				mcpRoute.ForwardHeaders = append(mcpRoute.ForwardHeaders, *h)
+			}
+		}
 		mc.Routes = append(mc.Routes, mcpRoute)
```
Hritik003, line 733: qq: what if the same header is added as part of oauth claimToHeaders? will this not append duplicates to the forward headers? Runtime behavior is unchanged but we might have to dedup the config right?

mohitgurnani: Good catch, and it's more than theoretical — `ensureSecurityPolicy` will happily build a single SecurityPolicy with both JWT and APIKeyAuth populated if both are configured, and pointing `claimToHeaders`/`forwardClientIDHeader` at the same header name is actually the natural way to get one unified caller-attribution label regardless of which auth method fired. So yes, a duplicate is a real (if narrow) case.

Runtime stays harmless either way (both consumers are idempotent — map-keyed or repeated `Header.Set`), but agreed it's worth deduping rather than relying on that. Pushed a fix covering both this new branch and the pre-existing OAuth loop above it (same latent gap, not introduced by this PR), plus a test for the same-header case.

Hritik003 (approved):
Thanks @mohitgurnani! LGTM

johnugeorge:
Thanks @mohitgurnani

johnugeorge (approved):

Issue #2316, opened 2026-07-02
mcp: allow disabling/customizing the `<backend>__` tool name prefix


dmitrysDV:
### Description:

The MCP proxy unconditionally prefixes every downstream tool/resource/prompt name with `<backendName>__` and there's no config knob on `MCPRoute` (or elsewhere) to opt out or customize it — confirmed still the case as of v1.0.0.

```go
// Using "__" as the separator to avoid collision with any character in k8s resource names as well as base64 encoding.
// We can't use special characters as tool names must match the regex `[a-zA-Z0-9._-]+`.
const nameSeparator = "__"

// downstreamResourceName converts the upstream resource/prompt name to the downstream resource/prompt name by
// prefixing the backend name.
func downstreamResourceName(name string, backendName string) string {
	return fmt.Sprintf("%s%s%s", backendName, nameSeparator, name)
}
```

#### Problem

This causes real usability problems for API consumers of multi-backend MCP gateways:
- Tool names get needlessly long and redundant, e.g. a backend named `acme-billing-mcp` produces tool IDs like `acme-billing-mcp__get-invoices`, where both the company prefix and the "mcp" suffix are pure noise once the tool already appears behind the gateway.
- It leaks internal backend/service naming (team names, "mcp" suffixes, company prefixes) into surfaces meant to be consumed by LLM agents/clients, where names ideally read as natural, domain-scoped intents (e.g. `billing__get-invoices` instead of `acme-billing-mcp__get-invoices`).
- Because renaming the *backend* is the only lever, teams are forced to rename their whole MCP server/deployment just to shorten a tool ID prefix.

#### Desired behavior

We'd like a way to either:
1. Disable prefixing entirely when there's only one backend / no collision risk, or
2. Configure a custom alias per `MCPRoute`/backend that's used for the prefix instead of the raw backend name (so the "friendly" prefix can differ from the k8s resource name), or
3. Configure the separator/format (e.g. allow no separator, or a shorter delimiter).

We understand the prefix exists to avoid tool name collisions across multiplexed backends and to route `tools/call` back to the correct upstream (`upstreamResourceName` strips everything before the first `__`), so any solution needs to preserve that routing guarantee — e.g. an explicit per-backend `toolPrefix`/`alias` field would satisfy both.

#### How it could be used

```yaml
apiVersion: aigateway.envoyproxy.io/v1beta1
kind: MCPRoute
metadata:
  name: example-mcp-route
spec:
  backendRefs:
[...]

PR #2425, opened 2026-07-25
docs: proposal for configurable MCP tool name prefixing (prefixMode)

walsm232:
**Description**

In MCP multiplexing mode, tool and prompt names are unconditionally prefixed with `<backend>__`. This breaks interactive MCP Apps (`ext-apps`): a rendered app calls tools by names hardcoded in server-authored HTML/JS (e.g. `get-time`), which never carry the prefix, so the gateway rejects them with `invalid tool name`. https://github.com/envoyproxy/ai-gateway/pull/2412 fixes the *rendering* half of MCP Apps support (preserving the `ui://` scheme when namespacing resource URIs) but in-app tool calls remain broken, and the tool names cannot be rewritten in the served HTML.

This proposal adds a `prefixMode` field to `MCPRoute` with three values: `Always` (current behavior, default), `Conditional` (prefix only with multiple backends), and `Never` (bare names, routed by resolving the name against the route's backends at call time). `Never` mode is what makes interactive MCP Apps work. The design follows the shape Agentgateway shipped for the same problem in https://github.com/agentgateway/agentgateway/pull/2531, with the trade-offs (unique-name constraint, resolution fan-out) and one refinement (a distinct ambiguity error instead of a silent unknown-tool failure for colliding names) documented in the proposal.

The proposal also covers: why the default stays `Always` (no advertised name changes on upgrade), why the field is route-level, the interaction with per-tool authorization (resolve first, then authorize), why resource URIs are unaffected (they keep the https://github.com/envoyproxy/ai-gateway/pull/2412 namespacing in all modes), and the limitations that cannot be solved at the gateway with today's ext-apps spec as far as I know.

- Related issue: https://github.com/envoyproxy/ai-gateway/issues/2316 (`prefixMode` covers its "disable for single backend" and unique-names cases; the custom per-backend alias ask remains open there and is not what I am mainly trying to solve for here. My goal is to solve for the MCP Apps case although this does benefit what was asked for there)
- Related issue: https://github.com/envoyproxy/ai-gateway/issues/2228 (The original MCP Apps report; https://github.com/envoyproxy/ai-gateway/pull/2412 (below) fixes its rendering half, and this proposal covers the remaining interactive tool call half)
- Related PR: https://github.com/envoyproxy/ai-gateway/pull/2412 (MCP Apps rendering fix; prerequisite for this: display-only apps work with it, interactive ones additionally need this proposal)

Two decisions where I would especially like maintainer input, both discussed in the proposal with the alternatives kept open:

- **Default value**: `Always` is proposed to keep upgrades name-stable, whereas Agentgateway defaults to `Conditional` (which was already its pre-existing behavior). Happy to switch if renaming single-backend routes' tools on upgrade is considered acceptable.
- **Duplicate names in `Never` mode**: like Agentgateway, colliding names are omitted from `tools/list`; unlike Agentgateway, calls to them get a distinct "ambiguous tool name" error (generic to the client, backend names in logs only) instead of a plain unknown-tool failure. This is a diagnostics refinement, not a safety requirement. Exact parity with Agentgateway is equally safe if that is preferred.

### DIFF
--- a/docs/proposals/012-mcp-tool-name-prefix-mode/proposal.md
+++ b/docs/proposals/012-mcp-tool-name-prefix-mode/proposal.md
@@ -1,0 +1,318 @@
+# Configurable Tool Name Prefixing for MCP Multiplexing (`prefixMode`)
+
+## Table of Contents
+
+- [Configurable Tool Name Prefixing for MCP Multiplexing (`prefixMode`)](#configurable-tool-name-prefixing-for-mcp-multiplexing-prefixmode)
+  - [Table of Contents](#table-of-contents)
+  - [Background and Motivation](#background-and-motivation)
+  - [Current State](#current-state)
+    - [How Prefixing Works](#how-prefixing-works)
+    - [Where It Breaks: Interactive MCP Apps](#where-it-breaks-interactive-mcp-apps)
+    - [Relationship to the Rendering Fix (PR #2412)](#relationship-to-the-rendering-fix-pr-2412)
+  - [Goals and Non-Goals](#goals-and-non-goals)
+    - [Goals](#goals)
+    - [Non-Goals](#non-goals)
+  - [Design](#design)
+    - [The `prefixMode` Field](#the-prefixmode-field)
+    - [How `Never` Mode Routes a Call](#how-never-mode-routes-a-call)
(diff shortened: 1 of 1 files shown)


nacx (approved):
Thanks for the detailed proposal and the research.
The proposal LGTM!

PR #2427, opened 2026-07-27
controller: trigger extproc pod rollout on injected-config drift

### LINKED ISSUE #2395: Changing `extProcExtraEnvVars` (or `GatewayConfig` env) does not trigger rollout of managed Envoy pods
**Description**:

`extProcExtraEnvVars` is applied by the pod mutating webhook (`gatewayMutator.mutatePod`), so env vars are only injected when a gateway pod is created. The drift check that decides whether to roll existing pods, `checkPodHasSideCar` (`internal/controller/gateway.go`), only compares the extproc container's image, the `-logLevel` arg, and `-mcpAddr` presence — it never inspects container env.

As a result, restarting the controller with a new `--extProcExtraEnvVars` value (or editing `GatewayConfig.spec.extProc.kubernetes.env`) has no effect on running pods. New values apply only to pods recreated for unrelated reasons, which can leave the fleet in a mixed state.

The same blind spot applies to other webhook-injected settings not covered by the check: `extProcImagePullSecrets`, `-endpointPrefixes`, `-enableRedaction`, `-maxRecvMsgSize`, header-attribute flags, and `GatewayConfig` resources/securityContext/volumeMounts.

**Repro steps**:
1. Deploy a Gateway with AIGatewayRoutes; confirm extproc sidecar is running.
2. Restart the controller with a changed `--extProcExtraEnvVars`.
3. Observe no rollout occurs and existing extproc containers keep the old env.

**Proposed fix**:

Instead of enumerating individual fields, have the webhook stamp a hash of the full injected extproc config as a pod annotation, and have the reconciler compare it against the hash of the currently-desired config (built by the same shared function), triggering the existing rollout path on mismatch. This closes the whole class of drift bugs rather than adding another special case.

**Worth considering:**
  - **Filter config content changes** stay on the existing fast path. The UUID pod-annotation trick that accelerates secret propagation without restarts. The hash mechanism is complementary, not a replacement; config-routing changes shouldn't cost you a pod rollout.
  - **One-time rollout on upgrade**: existing pods have no hash annotation, so the first controller version shipping this will roll every managed proxy once. That's arguably correct (it converges the fleet onto known state) but should be called out in release notes. If it's unacceptable, "missing annotation" can fall back to the legacy image/logLevel check for one release.
  - **Rollout-loop risk is the failure mode to guard in review**: if webhook and reconciler ever compute different hashes for identical state, you get an infinite roll. Sharing literally one function and hashing its output (not two parallel implementations) is what prevents this. A unit test asserting "mutate a pod, then recompute desired hash → equal" pins it.

Note this is a follow-on to the same drift-check that #1857 hardened: that fix addressed sidecar *presence* getting out of sync, while this issue is about sidecar config *content* drifting without triggering a rollout.

Aias00:
**Description**

The mutating webhook injects the extproc container config only at pod creation time, but the previous drift check only compared the container image, `-logLevel`, and `-mcpAddr` presence. Restarting the controller with changed extproc settings, or editing `GatewayConfig.spec.extProc.kubernetes`, could leave existing pods running stale injected config until they were recreated for some unrelated reason.

This PR computes a stable desired hash from the extproc injection inputs and writes it to the Deployment/DaemonSet pod template annotation `aigateway.envoyproxy.io/extproc-config-hash`. Kubernetes then owns the rollout whenever injected extproc config changes.

Upgrade behavior: workloads created before this annotation existed will not have the desired hash on their pod template. On the next Gateway reconcile, Deployments/DaemonSets with effective routes get the template hash annotation and Kubernetes performs one rollout to converge pods onto the current injected extproc config. After that first convergence rollout, reconciles are no-op unless the desired hash changes again.

The existing direct pod UUID annotation path remains for filter-config Secret refresh because it speeds projected-volume updates without restarting pods. Workload pod template patching is limited to cases where pod template state must change: sidecar add/remove, inconsistent pod state self-heal, or extproc config hash drift.

To avoid rollout loops from Kubernetes API shape/defaulting changes, the hash serializes an explicit digest of the inputs that affect extproc injection instead of hashing a full `corev1.Container`. The hash covers controller flags, GatewayConfig extproc settings, image pull secrets, and MCP input. It intentionally excludes secret-presence-driven `-configPath` / `-configBundlePath` args and legacy/bundle volume mounts.

Fixes #2395

- The shared builder in `internal/controller/extproc_builder.go` is the single source of truth for the injected container and the desired template hash.
- `GatewayConfig.spec.extProc.kubernetes.image` / `imageRepository` are used when checking whether an existing extproc sidecar is current.
- The reconciler uses the cached route-list path; the no-cache fallback remains only on the admission webhook path.
- Tests cover deterministic hashing, drift-relevant inputs, GatewayConfig-resolved image checks, and Deployment/DaemonSet template hash rollout behavior.

### DIFF
--- a/internal/controller/extproc_builder.go
+++ b/internal/controller/extproc_builder.go
@@ -24,8 +24,8 @@ import (
 	"github.com/envoyproxy/ai-gateway/internal/internalapi"
 )
 
-// extProcConfigHashAnnotationKey stores the webhook's extproc config hash. The
-// gateway reconciler compares it with the desired hash to detect config drift.
+// extProcConfigHashAnnotationKey stores the desired extproc config hash on
+// workload pod templates so Kubernetes rolls pods when injected config changes.
 const extProcConfigHashAnnotationKey = "aigateway.envoyproxy.io/extproc-config-hash"
 
 // extProcAdminPort is the admin port of the extproc container.
@@ -34,8 +34,8 @@ const extProcAdminPort = 1064
 // newExtProcBuilder constructs the shared extproc builder from the controller
 // options plus the runtime-resolved extProcAsSideCar flag. It is called once in
 // StartControllers; the resulting builder is shared by the mutating webhook (to
-// inject the container and stamp the config hash) and the gateway reconciler (to
-// detect drift), guaranteeing the two sides compute identical hashes.
+// inject the container) and the gateway reconciler (to compute the workload
+// template hash), guaranteeing the two sides use identical extproc config.
 func newExtProcBuilder(options *Options, extProcAsSideCar bool, logger logr.Logger) *extProcBuilder {
 	var parsedEnvVars []corev1.EnvVar
 	if options.ExtProcExtraEnvVars != "" {
--- a/internal/controller/extproc_builder.go
+++ b/internal/controller/extproc_builder.go
@@ -8,7 +8,7 @@ package controller
 import (
 	"crypto/sha256"
 	"encoding/hex"
-	stdjson "encoding/json" //nolint: depguard // determinism is required: the hash must be byte-identical across the webhook and reconciler, and sonic's Marshal does not guarantee stable field order, which would risk an infinite rollout loop.
+	stdjson "encoding/json" //nolint: depguard // the webhook and reconciler need byte-stable hashing; sonic does not guarantee stable field order.
 	"fmt"
 	"path/filepath"
 	"strconv"
@@ -24,12 +24,8 @@ import (
 	"github.com/envoyproxy/ai-gateway/internal/internalapi"
 )
 
-// extProcConfigHashAnnotationKey is stamped on each gateway pod by the mutating
-// webhook with a hash of the injected extproc container config. The gateway
-// reconciler recomputes the desired hash with the same builder and compares it
-// against this annotation to detect config drift (e.g. a changed
-// --extProcExtraEnvVars flag) and trigger a rollout. The controller only reads
-// this annotation; it never writes it, so a stale hash can never be persisted.
+// extProcConfigHashAnnotationKey stores the webhook's extproc config hash. The
+// gateway reconciler compares it with the desired hash to detect config drift.
 const extProcConfigHashAnnotationKey = "aigateway.envoyproxy.io/extproc-config-hash"
 
 // extProcAdminPort is the admin port of the extproc container.
@@ -81,17 +77,8 @@ func newExtProcBuilder(options *Options, extProcAsSideCar bool, logger logr.Logg
 	}
 }
 
-// extProcBuilder holds the controller-global extproc configuration and is the
-// single source of truth for the injected extproc container. It is constructed
-// once in StartControllers from Options and shared by the mutating webhook
-// (to inject the container and stamp the config hash) and the gateway reconciler
-// (to detect drift and trigger a rollout).
-//
-// Sharing one builder plus one build function is what prevents rollout loops:
-// both sides hash the output of the *same* function rather than maintaining two
-// parallel representations of the drift signal. If the webhook and reconciler
-// ever computed different hashes for identical state, the deployment would roll
-// forever — this structure makes that impossible by construction.
+// extProcBuilder holds the controller-global extproc configuration used by both
+// the mutating webhook and the gateway reconciler.
 type extProcBuilder struct {
 	image           string
 	imagePullPolicy corev1.PullPolicy
@@ -118,9 +105,7 @@ type extProcBuilder struct {
 	mcpFallbackSessionEncryptionIterations int
 }
 
-// extProcContainerInput is the per-gateway input the shared builder needs to
-// produce the extproc container. Everything not in this struct is global and
-// lives on the builder itself.
+// extProcContainerInput is the per-gateway input for extproc injection.
 type extProcContainerInput struct {
 	// gatewayConfig is the GatewayConfig referenced by the Gateway (nil if none).
 	gatewayConfig *aigv1b1.GatewayConfig
@@ -129,20 +114,14 @@ type extProcContainerInput struct {
 	needMCP bool
 }
 
-// buildExtProcContainer builds the extproc container *minus* the secret-presence-
-// driven parts: the -configPath / -configBundlePath args and the legacy/bundle
-// volumeMounts. Those depend on which filter-config secrets happen to exist at
-// pod-creation time, not on controller config, so they are excluded from the
-// drift signal (the UUID-annotation fast path handles filter-config content
-// updates). The mutating webhook adds them after this function returns; the
-// reconciler never adds them — it only hashes this base container.
+// buildExtProcContainer builds the extproc container without secret-presence
+// config args or mounts; those are handled by the webhook at pod creation time.
 func (b *extProcBuilder) buildExtProcContainer(input extProcContainerInput) corev1.Container {
 	var (
 		extProcSpec       *aigv1b1.GatewayConfigExtProc
 		kubernetesExtProc *egv1a1.KubernetesContainerSpec
 	)
-	if input.gatewayConfig != nil && input.gatewayConfig.Spec.ExtProc != nil {
-		extProcSpec = input.gatewayConfig.Spec.ExtProc
+	if extProcSpec = extProcSpecFromInput(input); extProcSpec != nil {
 		if extProcSpec.Kubernetes != nil {
 			kubernetesExtProc = extProcSpec.Kubernetes
 		}
@@ -155,7 +134,7 @@ func (b *extProcBuilder) buildExtProcContainer(input extProcContainerInput) core
 	}
 
 	envVars := b.mergeEnvVars(input.gatewayConfig)
-	image := b.resolveExtProcImage(extProcSpec)
+	image := b.extProcImage(input)
 
 	udsMountPath := filepath.Dir(b.udsPath)
 	securityContext := &corev1.SecurityContext{
@@ -221,39 +200,71 @@ func (b *extProcBuilder) buildExtProcContainer(input extProcContainerInput) core
 	return container
 }
 
-// extProcConfigDigest is the canonical representation of everything the webhook
-// injects that is drift-relevant: the extproc container (minus the secret-
-// presence-driven config-routing args/mounts, which buildExtProcContainer already
-// omits) plus the pod-spec-level imagePullSecrets, which are not part of the
-// container but are injected by the webhook and must also drift-trigger a rollout.
-// Hashing this struct — built by one shared function — is the drift signal both
-// the webhook and reconciler use.
+// extProcConfigDigest contains the config inputs that affect extproc injection.
+// It intentionally avoids hashing the full Kubernetes Container API object.
 type extProcConfigDigest struct {
-	Container        corev1.Container
-	ImagePullSecrets []corev1.LocalObjectReference
+	Image                                  string
+	ImagePullPolicy                        corev1.PullPolicy
+	LogLevel                               string
+	EnableRedaction                        bool
+	UDSPath                                string
+	RequestHeaderAttributes                *string
+	SpanRequestHeaderAttributes            *string
+	MetricsRequestHeaderAttributes         *string
+	LogRequestHeaderAttributes             *string
+	RootPrefix                             string
+	EndpointPrefixes                       string
+	ExtraEnvVars                           []corev1.EnvVar
+	ImagePullSecrets                       []corev1.LocalObjectReference
+	MaxRecvMsgSize                         int
+	ExtProcAsSideCar                       bool
+	MCPSessionEncryptionSeed               string
+	MCPSessionEncryptionIterations         int
+	MCPFallbackSessionEncryptionSeed       string
+	MCPFallbackSessionEncryptionIterations int
+	NeedMCP                                bool
+	GatewayConfigExtProc                   *aigv1b1.GatewayConfigExtProc
 }
 
-// extProcContainerHash returns a stable hex hash of the extproc config the
-// webhook injects for the given input. Both the webhook and the reconciler call
-// this with the same builder + input, so identical state yields identical
-// hashes. It is over the SAME object the builder injects (the container from
-// buildExtProcContainer plus pod-spec-level imagePullSecrets), so the drift
-// signal can never diverge from the injected config.
+// extProcContainerHash returns a stable hash of the extproc injection inputs.
 func (b *extProcBuilder) extProcContainerHash(input extProcContainerInput) string {
 	digest := extProcConfigDigest{
-		Container:        b.buildExtProcContainer(input),
-		ImagePullSecrets: b.imagePullSecrets,
+		Image:                                  b.extProcImage(input),
+		ImagePullPolicy:                        b.imagePullPolicy,
+		LogLevel:                               b.logLevel,
+		EnableRedaction:                        b.enableRedaction,
+		UDSPath:                                b.udsPath,
+		RequestHeaderAttributes:                b.requestHeaderAttributes,
+		SpanRequestHeaderAttributes:            b.spanRequestHeaderAttributes,
+		MetricsRequestHeaderAttributes:         b.metricsRequestHeaderAttributes,
+		LogRequestHeaderAttributes:             b.logRequestHeaderAttributes,
+		RootPrefix:                             b.rootPrefix,
+		EndpointPrefixes:                       b.endpointPrefixes,
+		ExtraEnvVars:                           b.extraEnvVars,
+		ImagePullSecrets:                       b.imagePullSecrets,
+		MaxRecvMsgSize:                         b.maxRecvMsgSize,
+		ExtProcAsSideCar:                       b.extProcAsSideCar,
+		MCPSessionEncryptionSeed:               b.mcpSessionEncryptionSeed,
+		MCPSessionEncryptionIterations:         b.mcpSessionEncryptionIterations,
+		MCPFallbackSessionEncryptionSeed:       b.mcpFallbackSessionEncryptionSeed,
+		MCPFallbackSessionEncryptionIterations: b.mcpFallbackSessionEncryptionIterations,
+		NeedMCP:                                input.needMCP,
+		GatewayConfigExtProc:                   extProcSpecFromInput(input),
 	}
-	// stdjson.Marshal of a struct is deterministic (fields in declaration
-	// order; map keys sorted). Since the builder always produces the same
-	// object for the same input, the bytes — and therefore the hash — are
-	// stable. stdlib json is used (not internal/json/sonic) because sonic
-	// does not guarantee stable field order, which would risk rollout loops.
-	// corev1.Container is a plain serializable struct, so Marshal cannot
-	// fail here; the error is ignored accordingly.
-	marshaled, _ := stdjson.Marshal(digest) // corev1.Container is serializable; see comment above
+	marshaled, _ := stdjson.Marshal(digest)
 	sum := sha256.Sum256(marshaled)
 	return hex.EncodeToString(sum[:])
+}
+
+func extProcSpecFromInput(input extProcContainerInput) *aigv1b1.GatewayConfigExtProc {
+	if input.gatewayConfig == nil {
+		return nil
+	}
+	return input.gatewayConfig.Spec.ExtProc
+}
+
+func (b *extProcBuilder) extProcImage(input extProcContainerInput) string {
+	return b.resolveExtProcImage(extProcSpecFromInput(input))
 }
 
 // buildExtProcBaseArgs builds the command line arguments for the extproc
--- a/internal/controller/extproc_builder.go
+++ b/internal/controller/extproc_builder.go
@@ -1,0 +1,381 @@
+// Copyright Envoy AI Gateway Authors
+// SPDX-License-Identifier: Apache-2.0
+// The full text of the Apache license is available in the LICENSE file at
+// the root of the repo.
+
+package controller
+
+import (
+	"crypto/sha256"
+	"encoding/hex"
+	stdjson "encoding/json" //nolint: depguard // determinism is required: the hash must be byte-identical across the webhook and reconciler, and sonic's Marshal does not guarantee stable field order, which would risk an infinite rollout loop.
+	"fmt"
+	"path/filepath"
+	"strconv"
+	"strings"
+
+	egv1a1 "github.com/envoyproxy/gateway/api/v1alpha1"
+	"github.com/go-logr/logr"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/utils/ptr"
+
+	aigv1b1 "github.com/envoyproxy/ai-gateway/api/v1beta1"
+	"github.com/envoyproxy/ai-gateway/internal/internalapi"
+)
+
+// extProcConfigHashAnnotationKey is stamped on each gateway pod by the mutating
+// webhook with a hash of the injected extproc container config. The gateway
+// reconciler recomputes the desired hash with the same builder and compares it
+// against this annotation to detect config drift (e.g. a changed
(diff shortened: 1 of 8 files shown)


nacx:
@kanurag94 can you take a look?

internal/controller/gateway.go
```diff
@@ -1058,20 +1063,36 @@ func (c *GatewayController) getBSPForInferencePool(ctx context.Context, namespac
...
+// annotation, and any subsequent config drift is then detected.
+func (c *GatewayController) checkPodHasSideCar(pod *corev1.Pod, needMCP bool, desiredHash string) (hasSideCar, needsRollout bool) {
 	podSpec := pod.Spec
-	hasSideCar := false
 
 	if c.extProcAsSideCar {
 		for i := range podSpec.InitContainers {
 			// If there's an extproc sidecar container with the current target image, we don't need to roll out the deployment.
-			if podSpec.InitContainers[i].Name == extProcContainerName && podSpec.InitContainers[i].Image == c.extProcImage {
+			if podSpec.InitContainers[i].Name == extProcContainerName && podSpec.InitContainers[i].Image == c.image {
```
kanurag94: since this PR already does `resolveExtProcImage`, we can fix this and instead of using the global extproc image, use the one from gatewayConfig `GatewayConfig.spec.extProc.kubernetes.image`

internal/controller/extproc_builder_test.go
```diff
@@ -0,0 +1,284 @@
...
+		{name: "with gateway config", input: extProcContainerInput{gatewayConfig: testGatewayConfig}},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			h1 := b.extProcContainerHash(tc.input)
+			h2 := b.extProcContainerHash(tc.input)
+			require.NotEmpty(t, h1)
+			require.Equal(t, h1, h2, "hash must be stable across recomputes")
+		})
+	}
+}
```
kanurag94, line 65: I think this test will be more useful if it really mutates a pod against a fake client and validates it. Currently it just is a UT to check hash calculation

internal/controller/gateway.go
kanurag94: I think it differs from `gateway_mutator.go`. This only uses the cached index while webhook uses the fallback. There can be a small window where cache is not updated.

internal/controller/extproc_builder.go
```diff
@@ -0,0 +1,381 @@
...
+
+// extProcContainerHash returns a stable hex hash of the extproc config the
+// webhook injects for the given input. Both the webhook and the reconciler call
+// this with the same builder + input, so identical state yields identical
+// hashes. It is over the SAME object the builder injects (the container from
+// buildExtProcContainer plus pod-spec-level imagePullSecrets), so the drift
+// signal can never diverge from the injected config.
+func (b *extProcBuilder) extProcContainerHash(input extProcContainerInput) string {
+	digest := extProcConfigDigest{
+		Container:        b.buildExtProcContainer(input),
```
kanurag94: What happens when k8s api changes for some reason? This rolls out all the pods. Maybe we could hash the actual inputs. WDYT?

internal/controller/gateway.go
Aias00: Fixed in 17673eed. The Gateway reconciler now gets mgr.GetAPIReader() and uses the same cache-index + no-cache fallback route lookup as the mutating webhook for both AIGatewayRoute and MCPRoute. Added controller tests for the cache-empty/API-reader-has-route window.

Verified with:
- go test ./internal/controller -run 'TestGatewayController_listAIGatewayRoutesForGateway_NoCacheReaderFallback|TestGatewayController_listMCPRoutesForGateway_NoCacheReaderFallback|TestGatewayMutator_listAIGatewayRoutesForGateway_NoCacheReaderFallback|TestGatewayMutator_listMCPRoutesForGateway_NoCacheReaderFallback|TestGatewayController_annotateGatewayPods_ConfigHashDrift'
- go test ./internal/controller

kanurag94 (commented):
This is close, but we need to ensure this doesn't break anything. The comments are too verbose right now. If possible, can we make them just `what and why`.

Alternative to the approach in this PR, wdyt about just patching pod annotations through deployment so kubernetes takes care of rollout itself?

internal/controller/extproc_builder_test.go
Aias00: Addressed in f3e96d21. Added GatewayController coverage that creates pods and deployments in the fake Kubernetes client, runs the rollout path, and validates the deployment template annotation is patched when the extproc config hash drifts.

The smaller hash unit test remains to pin deterministic hash behavior.

internal/controller/gateway.go
Aias00: Fixed in 6284456e. The Gateway reconciler now resolves the desired extproc image from the same extProcContainerInput used for hashing, so GatewayConfig.spec.extProc.kubernetes.image / imageRepository are honored when checking whether the existing sidecar is current.

Added regression coverage for the GatewayConfig-resolved image path.

internal/controller/extproc_builder.go
Aias00: Fixed in 6284456e. The drift hash now serializes an explicit extProcConfigDigest of the inputs that affect extproc injection instead of serializing the full corev1.Container object. That keeps Kubernetes Container API shape/defaulting changes from changing the rollout signal, while controller flags, GatewayConfig extproc settings, image pull secrets, and MCP input still move the hash.

Verified with go test ./internal/controller.

Aias00:
Addressed the review summary in 6284456e: comments are trimmed to what/why, the hash no longer serializes the full Kubernetes Container API object, and GatewayConfig-resolved extproc images are used when checking existing sidecars.

On rollout behavior: I kept direct pod annotation patching for ordinary filter-config Secret updates because it speeds projected-volume refresh without restarting pods. Deployment/DaemonSet template patching is still limited to cases where the pod template must change: missing sidecar, sidecar removal, inconsistent pod state, or extproc config drift.

Verified with go test ./internal/controller.

internal/controller/gateway.go
```diff
@@ -1058,29 +1060,31 @@ func (c *GatewayController) getBSPForInferencePool(ctx context.Context, namespac
...
-func (c *GatewayController) checkPodHasSideCar(pod *corev1.Pod, needMCP bool) bool {
+// checkPodHasSideCar checks the extproc container against the controller-global
+// image. It is kept for tests and legacy callers without GatewayConfig input.
+func (c *GatewayController) checkPodHasSideCar(pod *corev1.Pod, needMCP bool, desiredHash string) (hasSideCar, needsRollout bool) {
+	return c.checkPodHasSideCarWithImage(pod, needMCP, c.image, desiredHash)
+}
+
+// checkPodHasSideCarWithImage returns whether the extproc container is current,
+// and whether a stamped config hash mismatch requires a rollout.
+func (c *GatewayController) checkPodHasSideCarWithImage(pod *corev1.Pod, needMCP bool, desiredImage, desiredHash string) (hasSideCar, needsRollout bool) {
```
kanurag94: This is unnecessary duplicated I think. Is there a reason for it?

internal/controller/gateway.go
```diff
@@ -1164,38 +1169,54 @@ func (c *GatewayController) annotateGatewayPods(ctx context.Context,
 	uuid string,
 	hasEffectiveRoute bool,
 	needMCP bool,
+	desiredHash string,
+) (ctrl.Result, error) {
+	return c.annotateGatewayPodsWithDesiredImage(ctx, pods, deployments, daemonSets, uuid, hasEffectiveRoute, needMCP, c.image, desiredHash)
+}
+
+func (c *GatewayController) annotateGatewayPodsWithDesiredImage(ctx context.Context,
+	pods []corev1.Pod,
```
kanurag94, line 1139: Again, this is duplicated

internal/controller/gateway.go
```diff
@@ -46,43 +46,52 @@ const (
...
-	extProcImage string, extProcLogLevel string, standAlone bool, uuidFn func() string, extProcAsSideCar bool,
+	standAlone bool, uuidFn func() string, options *Options, extProcAsSideCar bool,
+) *GatewayController {
+	return NewGatewayControllerWithAPIReader(client, nil, kube, logger, envoyGatewayNamespace, standAlone, uuidFn, options, extProcAsSideCar)
+}
+
+func NewGatewayControllerWithAPIReader(
+	client client.Client, noCacheReader client.Reader, kube kubernetes.Interface, logger logr.Logger, envoyGatewayNamespace string,
+	standAlone bool, uuidFn func() string, options *Options, extProcAsSideCar bool,
 ) *GatewayController {
```
kanurag94, line 55: this seems unnecessary, and just passes nil

internal/controller/gateway_mutator.go
```diff
@@ -265,41 +141,57 @@ func ParseImagePullSecrets(s string) ([]corev1.LocalObjectReference, error) {
...
-	if g.noCacheReader == nil {
+	if noCacheReader == nil {
 		return routes, cacheErr
 	}
 	// noCacheReader doesn't have access to cache indexes, so list then filter.
 	var all aigv1b1.AIGatewayRouteList
-	if err := g.noCacheReader.List(ctx, &all); err != nil {
+	if err := noCacheReader.List(ctx, &all); err != nil {
 		return routes, fmt.Errorf("failed to list routes: %w", err)
 	}
```
kanurag94, line 166: I think it's fine to not have noCache reader here. Reconcile can be hit multiple times and it may make listing without cache too expensive. But the current callers are fine.
Eventual consistency should be okay here so we can avoid it as well.

internal/controller/extproc_builder_test.go
kanurag94: I couldn't find the change. Can you please point me? 
We need something to mutate the pod and calculate hash

internal/controller/gateway.go
```diff
@@ -1116,7 +1117,15 @@ func (c *GatewayController) checkPodHasSideCar(pod *corev1.Pod, needMCP bool) bo
 		}
 	}
 
-	return hasSideCar
+	if hasSideCar {
+		if stamped := pod.Annotations[extProcConfigHashAnnotationKey]; stamped != "" && stamped != desiredHash {
+			needsRollout = true
+			c.logger.Info("extproc config hash mismatch, triggering rollout",
```
kanurag94: Can you please add stamped and desiredHash here?
If these two ever disagree permanently, this rollout repeats forever with a fresh UUID each time, and this log line is the only signal.
A  counter metric would be even better

internal/controller/gateway_test.go
```diff
@@ -1837,6 +1934,253 @@ func TestGatewayController_annotateGatewayPods(t *testing.T) {
...
+	})
+
+	t.Run("sidecar mode: logLevel mismatch triggers rollout", func(t *testing.T) {
+		c := newTestGatewayController(fakeClient, kube, ctrl.Log, egNamespace, image, logLevel, false, nil, true)
+		pod := &corev1.Pod{Spec: corev1.PodSpec{InitContainers: []corev1.Container{extProcContainer("-logLevel", "debug")}}}
+		hasSideCar, _ := c.checkPodHasSideCar(pod, false, "")
+		require.False(t, hasSideCar, "logLevel mismatch must clear hasSideCar")
+	})
+
+	t.Run("sidecar mode: uses GatewayConfig resolved image", func(t *testing.T) {
```
kanurag94, line 1988: This passes desiredImage in by hand, so it tests checkPodHasSideCarWithImage but not the wiring that produces the image. We should test `c.extProcImage(input)` here

internal/controller/gateway.go
```diff
@@ -1116,7 +1117,15 @@ func (c *GatewayController) checkPodHasSideCar(pod *corev1.Pod, needMCP bool) bo
 		}
 	}
 
-	return hasSideCar
+	if hasSideCar {
+		if stamped := pod.Annotations[extProcConfigHashAnnotationKey]; stamped != "" && stamped != desiredHash {
```
kanurag94: The stamped != "" check is the decision that a pod with no annotation is never rolled by the hash.
It means an operator who upgrades and then changes a flag sees nothing happen until something else recreates the pods.
Could we add the comment that was previously there?

internal/controller/gateway.go
```diff
@@ -1213,11 +1234,6 @@ func (c *GatewayController) annotateGatewayPods(ctx context.Context,
 		}
 	}
 
-	// We annotate the deployments and daemonsets under three scenarios:
-	// 1. If there's an effective route but no sidecar container, we need to add the sidecar container.
-	// 2. If there's no effective route but has sidecar container,
-	//    we need to roll out the deployment to trigger the mutation webhook to remove the sidecar container.
-	// 3. If pods are inconsistent even when rollout isn't in progress, force rollout to self-heal.
```
kanurag94, line 1220: Can we keep this explanation please?

kanurag94:
> Addressed the review summary in [6284456](https://github.com/envoyproxy/ai-gateway/commit/6284456ebe5c2ad379ffa4a2d69890fb4c0670c1): comments are trimmed to what/why, the hash no longer serializes the full Kubernetes Container API object, and GatewayConfig-resolved extproc images are used when checking existing sidecars.
> 
> On rollout behavior: I kept direct pod annotation patching for ordinary filter-config Secret updates because it speeds projected-volume refresh without restarting pods. Deployment/DaemonSet template patching is still limited to cases where the pod template must change: missing sidecar, sidecar removal, inconsistent pod state, or extproc config drift.
> 
> Verified with go test ./internal/controller.

Thanks. I agree on keeping the pod-annotation fast path for filter-config Secret refresh.
My question was about the hash specifically. The reconciler already computes desiredHash. Could it write that to `Deployment.spec.template.metadata.annotations` and let Kubernetes handle the rollout, instead of the webhook stamping it on pods and the reconciler comparing?

I am just curious to know if that was thought of and rejected as an approach. If so, why?

internal/controller/extproc_builder_test.go
Aias00: Added this directly to the mutator fake-client test. `TestGatewayMutator_mutatePod_UsesNoCacheReader` now calls `mutatePod` and verifies the pod's `extproc-config-hash` annotation equals
  `g.extProcContainerHash(extProcContainerInput{})`.

internal/controller/gateway.go
Aias00: Removed the duplicate wrapper. `checkPodHasSideCar` now takes the desired image directly

internal/controller/gateway.go
Aias00: Removed the duplicate annotate wrapper.

internal/controller/gateway.go
Aias00: Removed `NewGatewayControllerWithAPIReader` and the `noCacheReader` field from `GatewayController`. The reconciler now uses the cached client path only.

internal/controller/gateway_mutator.go
Aias00: Updated the reconciler route-list helpers to pass `nil` for the no-cache reader. The no-cache fallback remains only on the admission webhook/mutator path.

internal/controller/gateway.go
Aias00: Added both fields to the hash-mismatch log line: `stamped` and `desiredHash`.

internal/controller/gateway_test.go
Aias00: Updated the GatewayConfig image test to build an `extProcContainerInput` and use `c.extProcImage(input)` as the desired image, instead of passing a hand-written image value.

internal/controller/gateway.go
Aias00: Restored the comment explaining that a missing hash means a pre-upgrade pod and should not trigger a rollout only because the annotation is absent.

internal/controller/gateway.go
Aias00: sure

Aias00:
> > Addressed the review summary in [6284456](https://github.com/envoyproxy/ai-gateway/commit/6284456ebe5c2ad379ffa4a2d69890fb4c0670c1): comments are trimmed to what/why, the hash no longer serializes the full Kubernetes Container API object, and GatewayConfig-resolved extproc images are used when checking existing sidecars.
> > On rollout behavior: I kept direct pod annotation patching for ordinary filter-config Secret updates because it speeds projected-volume refresh without restarting pods. Deployment/DaemonSet template patching is still limited to cases where the pod template must change: missing sidecar, sidecar removal, inconsistent pod state, or extproc config drift.
> > Verified with go test ./internal/controller.
> 
> Thanks. I agree on keeping the pod-annotation fast path for filter-config Secret refresh. My question was about the hash specifically. The reconciler already computes desiredHash. Could it write that to `Deployment.spec.template.metadata.annotations` and let Kubernetes handle the rollout, instead of the webhook stamping it on pods and the reconciler comparing?
> 
> I am just curious to know if that was thought of and rejected as an approach. If so, why?

Updated this to follow the pod template annotation approach.

  The webhook no longer stamps the config hash on individual pods. The reconciler now writes the desired extproc config hash to the Deployment/DaemonSet pod template annotation, so Kubernetes handles the rollout
  when the injected config changes.

  The existing UUID annotation is still used only for sidecar add/remove and inconsistent pod self-heal cases.

internal/controller/extproc_builder_test.go
```diff
@@ -0,0 +1,283 @@
...
+		{name: "tag after last slash", image: "registry.io/foo:v1", expected: ":v1"},
+		{name: "no tag and no digest", image: "registry.io/foo", expected: ""},
+		{name: "colon before slash is not a tag", image: "localhost:5000/foo", expected: ""},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			require.Equal(t, tt.expected, imageTagOrDigest(tt.image))
+		})
+	}
+}
```
kanurag94, line 346: Can we add test for asserting number of fields?
```
require.Equal(t, 19, reflect.TypeOf(extProcBuilder{}).NumField(),...)
```

internal/controller/gateway_mutator.go
```diff
@@ -265,41 +140,57 @@ func ParseImagePullSecrets(s string) ([]corev1.LocalObjectReference, error) {
...
+	return listMCPRoutesForGateway(ctx, g.c, g.noCacheReader, gatewayName, gatewayNamespace)
+}
+
+func (c *GatewayController) listAIGatewayRoutesForGateway(ctx context.Context, gatewayName, gatewayNamespace string) (aigv1b1.AIGatewayRouteList, error) {
+	return listAIGatewayRoutesForGateway(ctx, c.client, nil, gatewayName, gatewayNamespace)
+}
+
+func (c *GatewayController) listMCPRoutesForGateway(ctx context.Context, gatewayName, gatewayNamespace string) (aigv1b1.MCPRouteList, error) {
+	return listMCPRoutesForGateway(ctx, c.client, nil, gatewayName, gatewayNamespace)
+}
```
kanurag94: nit: I think reconciler will never need a non cached reader as it waits for warmup. We can drop these methods and call them inline.

internal/controller/gateway.go
```diff
@@ -152,10 +143,9 @@ func (c *GatewayController) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		return ctrl.Result{}, err
 	}
 
-	// Finally, we need to annotate the pods of the gateway deployment with the new uuid to propagate the filter config Secret update faster.
-	// If the pod doesn't have the extproc container, it will roll out the deployment altogether which eventually ends up
-	// the mutation hook invoked.
-	result, err := c.annotateGatewayPods(ctx, pods, deployments, daemonSets, uid, hasEffectiveRoutes, len(mcpRoutes.Items) > 0)
+	input := extProcContainerInput{gatewayConfig: gwConfig, needMCP: len(mcpRoutes.Items) > 0}
+	result, err := c.annotateGatewayPods(ctx, pods, deployments, daemonSets, uid,
+		hasEffectiveRoutes, input.needMCP, c.extProcImage(input), c.extProcContainerHash(input))
```
kanurag94, line 148: Can you add one test which goes through Reconcile with a Gateway whose GatewayConfig overrides the image, with pods already running the overridden image, and assert no workload patch happens?

internal/controller/gateway.go
```diff
@@ -1213,39 +1190,57 @@ func (c *GatewayController) annotateGatewayPods(ctx context.Context,
...
-				fmt.Appendf(nil,
-					`{"spec":{"template":{"metadata":{"annotations":{"%s":"%s"}}}}}`, aigatewayUUIDAnnotationKey, uuid),
-				metav1.PatchOptions{})
-			if err != nil {
-				return ctrl.Result{}, fmt.Errorf("failed to patch daemonset %s: %w", daemonSet.Name, err)
-			}
+	for i := range daemonSets {
+		daemonSet := &daemonSets[i]
+		// Hash changes are written to the pod template so Kubernetes owns the rollout.
+		needsHashRollout := hasEffectiveRoute && desiredHash != "" && daemonSet.Spec.Template.Annotations[extProcConfigHashAnnotationKey] != desiredHash
```
kanurag94, line 1218: I couldn't find a test for desiredHash not being nil for daemonSet. WDYT about a test for this as well?

internal/controller/extproc_builder_test.go
```diff
@@ -0,0 +1,283 @@
...
+	// needMCP toggling the MCP args must move the hash.
+	require.NotEqual(t, baseHash, base.extProcContainerHash(extProcContainerInput{needMCP: true}),
+		"needMCP must alter the config hash")
+}
+
+// TestExtProcContainerHash_ExcludesConfigRoutingArgs ensures the secret-
+// presence-driven -configPath / -configBundlePath args are NOT part of the hash
+// (they are added by the webhook after buildExtProcContainer returns). The base
+// container's Args must contain neither flag, so secret-existence transitions
+// do not spuriously trigger rollouts.
```
kanurag94, line 233: I think this test only does the second part of this comment and not computes hash

internal/controller/extproc_builder_test.go
```diff
@@ -0,0 +1,283 @@
...
+		name  string
+		input extProcContainerInput
+	}{
+		{name: "no gateway config, no MCP", input: extProcContainerInput{}},
+		{name: "need MCP", input: extProcContainerInput{needMCP: true}},
+		{name: "with gateway config", input: extProcContainerInput{gatewayConfig: testGatewayConfig}},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			h1 := b.extProcContainerHash(tc.input)
+			h2 := b.extProcContainerHash(tc.input)
```
kanurag94: nit: You can do something like
```
for range 20 {
  require.Equal(t, h1, b.extProcContainerHash(tc.input))
}
```
to ensure determinism

internal/controller/extproc_builder_test.go
Aias00: Added the field-count check and expanded the drift coverage for the builder fields.

internal/controller/gateway_mutator.go
Aias00: Done

internal/controller/gateway.go
Aias00: done

internal/controller/gateway.go
Aias00: Added a DaemonSet test for non-empty desiredHash.

internal/controller/extproc_builder_test.go
Aias00: done

internal/controller/extproc_builder_test.go
Aias00: done

kanurag94 (commented):
Very good shape now. Can you please add the upgrade rollout behavior in PR description? We can also think about it a little more if the first rollout should do restarts or not.

Aias00:
Updated the PR description with the upgrade behavior: workloads without the template hash get one convergence rollout on the next Gateway reconcile, then future reconciles are no-op unless the desired hash changes again.

kanurag94 (approved):
Thanks for working on this, LGTM

nacx (approved):

PR #2429, opened 2026-07-27
fix: sign AWS Bedrock SigV4 over the backend endpoint host

kanurag94:
**Description**

AWS Bedrock backends whose endpoint is an interface VPC endpoint (`vpce-….bedrock-runtime.<region>.vpce.amazonaws.com`) or any custom host return `403 InvalidSignatureException` on every request. The extproc SigV4 signer builds the request it signs from the hardcoded `bedrock-runtime.<region>.amazonaws.com`, but Envoy's Backend host-rewrite puts the endpoint FQDN on the wire as `Host`. Bedrock recomputes the signature over the `Host` it receives, so it never matches.

This signs over the host Envoy actually sends upstream — the referenced Backend's FQDN:

- The controller resolves the Envoy Gateway `Backend` referenced by the AIServiceBackend and plumbs its FQDN via a new `filterapi.AWSAuth.Hostname`. The signer signs over that host, falling back to the region default when it's empty (so existing deployments are unchanged).
- The SigV4/STS region is derived from the Bedrock host when it is one, so a typo'd configured region (e.g. `us-east1`) can't break signing.
- The AIServiceBackend controller now watches `egv1a1.Backend`, so a Backend create or FQDN edit regenerates the extproc config instead of leaving the signing host stale.

**Related Issues/PRs**

This is an alternative to #2413, which addresses the same bug. Both change the same signing line; the difference is where the signing host comes from. #2413 reads it from request headers (`x-ai-eg-aws-signing-host`, then `:authority`, then `host`) gated by a `.amazonaws.com` suffix check. Those headers are client-reachable and not stripped at ingress, and at signing time `:authority` is the downstream gateway host rather than the upstream host — which is also why the earlier attempt (#2296, reverted in #2298) signed the wrong host.

| | #2413 | this PR |
|---|---|---|
| Signing host source | request headers (`x-ai-eg-aws-signing-host` / `:authority` / `host`) | controller-resolved Backend FQDN |
| Client can influence it | yes (headers not stripped at ingress) | no (control-plane only) |
| Guard | `.amazonaws.com` suffix check | none needed |
| AWS China / custom proxy | rejected by suffix check | works |
| Stale on FQDN edit | n/a | Backend watch requeues config |
| Region on typo'd config | unchanged | derived from the Bedrock host |
| Covered by fork-PR CI | no (asserts headers exist only) | yes (credential-free, recomputes SigV4) |

- Empty `Hostname` preserves the current behavior, so non-vpce deployments are unaffected.
- `bspToFilterAPIBackendAuth` allocates a fresh `BackendAuth` per backendRef, so the per-backend host is stamped in place; there's a comment noting this must copy-on-write if that ever becomes shared.
- Controller RBAC already grants `gateway.envoyproxy.io` access, so the Backend watch needs no RBAC change.
- Tests are credential-free so fork PRs actually exercise them (#2296 broke on main precisely because the credentialed e2e is skipped on forks): a backendauth test recomputes SigV4 independently and asserts the signature is over the vpce host and not the region default; a controller test asserts the reconciled filter config carries the Backend FQDN as the signing host; plus Backend-FQDN resolution and the Backend→AIServiceBackend requeue mapping.

### DIFF
--- a/internal/backendauth/aws.go
+++ b/internal/backendauth/aws.go
@@ -13,6 +13,7 @@ import (
 	"fmt"
 	"net/http"
 	"os"
+	"regexp"
 	"strings"
 	"time"
 
@@ -29,6 +30,9 @@ type awsHandler struct {
 	credentialsProvider aws.CredentialsProvider
 	signer              *v4.Signer
 	region              string
+	// hostname is the upstream endpoint host to sign over. Empty falls back to the
+	// region-default bedrock-runtime.<region>.amazonaws.com.
+	hostname string
 }
 
 func newAWSHandler(ctx context.Context, awsAuth *filterapi.AWSAuth) (filterapi.BackendAuthHandler, error) {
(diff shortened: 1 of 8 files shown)


kanurag94:
Closing in favor of #2416, which fixes the same bug with a stronger design: the signing host comes from per-endpoint cluster metadata rather than being resolved once at config-build time, so it also covers weighted/multi-endpoint clusters (which this PR can't). I've left two suggestions there drawn from this PR — deriving the SigV4 region from the host, and a recompute-SigV4 test to pin the signed host in CI.

PR #2431, opened 2026-07-27
docs: proposal for new MCP Release candidate ( 28-07-2026 )

### LINKED ISSUE #2323: feat: mcp spec Release Candidate (2025-07-28)
## Summary

The next MCP protocol revision reaches Release Candidate on July 28, 2026. This
tracking issue coordinates the work needed to bring the AI Gateway's MCP proxy into
conformance, while preserving backward compatibility with clients and backends still
on 2025-11-25 and earlier.

Spec changelog: https://modelcontextprotocol.io/specification/draft/changelog

## Requires gateway work
 
| Sub-issue | SEP / Change | Title | Severity | Gateway effort | Key evidence |
|-----------|--------------|-------|----------|----------------|--------------|
| #A1 | SEP-2567 | Stateless Streamable HTTP sessions (remove `Mcp-Session-Id`) | Major / breaking | Heavy | `session.go` composite session ID; `handlers.go:329` requires session for all non-`initialize` |
| #A2 | SEP-2575 | Stateless protocol: remove `initialize`/`ping`/`logging/setLevel`, add `server/discover`, `subscriptions/listen`, drop SSE resumability | Major / breaking | Heavy | `handlers.go:364/384/472/1128`, `sse.go` |
| #A3 | SEP-2322 | MRTR — `InputRequiredResult` + required `resultType` | Major / breaking | Heavy | `handlers.go:982/992/1002` server-initiated methods |
| #B1 | SEP-2549 | Cacheable results (`ttlMs`, `cacheScope`) as a proxy cache | Minor | Moderate | `handleToolsListRequest` aggregates across backends |
| #B2 | — (non-SEP) | Deterministic `tools/list` ordering across aggregated backends | Minor | Light | `handlers.go:1620` merge order |
| #C1 | SEP-2468 | Authorization response `iss` validation (via generated Envoy `jwt_authn` config) | Minor | Moderate | Envoy does JWT verify (`mcproute.go:32`); gateway generates filter config, parses claims unverified (`authorization.go:139`) |
| #C2 | SEP-2352 | Client credentials bound to issuing authz server | Minor | Moderate | `tokenprovider/`, `rotators/` persist credentials |
| #D1 | SEP-2243 | Required `Mcp-Method`/`Mcp-Name` headers + `x-mcp-header` + reject-on-mismatch | Minor | Moderate | `handlers.go` body-parse dispatch |
| #D2 | SEP-414 | OTel `_meta` trace-context conventions (`traceparent`/`tracestate`/`baggage`) | Minor | Light (verify) | `tracing/mcp.go:76` already injects `_meta` |
| #D3 | — (non-SEP) | `extensions` field on capabilities (merge-drop fix) | Minor | Light | `session.go:647` merges field-by-field, drops `extensions` |
| #D4 | — (non-SEP) | JSON-RPC error-code renumbering + allocation policy (`-32002`→`-32602`, `-32020/21/22`) | Minor | Moderate | `handlers.go` switches on `jsonrpcErr.Code`, mints own errors |
| #E1 | SEP-2577 | Deprecate Roots / Sampling / Logging (plan migration, don't remove) | Deprecation | Light | `handlers.go` handles all three today |
| #E2 | SEP-2596 | Feature lifecycle policy + deprecate HTTP+SSE transport & `includeContext` | Deprecation | Light | `sse.go` transport handling |
| #F1 | — (spec: Versioning) | Backward compatibility: dual-era intermediary + per-origin era detection/caching | Cross-cutting | Heavy (design) | `handlers.go` `initialize` anchor; spec `-32022` probe |
 
## Light / pass-through / verify
 
| Sub-issue | SEP / Change | Title | Severity | Why light |
|-----------|--------------|-------|----------|-----------|
| #E3 | SEP-2663 | Tasks moved to `io.modelcontextprotocol/tasks` extension | Minor | No task handling today; pass-through only |
| — | SEP-2106 | JSON Schema 2020-12 adoption | Minor | Gateway doesn't validate `inputSchema`/`outputSchema`; opaque pass-through |
 
## Not relevant
 
| SEP | Title | Why excluded |
|-----|-------|--------------|
| SEP-837 | `application_type` in Dynamic Client Registration | OAuth *client*-registration concern; gateway is a resource server / token validator, never originates DCR |
| PR #2858 | Deprecate DCR → Client ID Metadata Documents (CIMD) | Same — client-registration mechanism; gateway only passes `registration_endpoint` through in served metadata (`mcp_route_security_policy.go:642`) |
| SEP-1850 | Formalized PR-based SEP workflow | Governance process only; no gateway code |
 
## Cross-cutting (not SEP-specific)
 
| Sub-issue | Title | Purpose |
|-----------|-------|---------|
| #F2 | Docs, examples, versioned capability pages | Update `site/docs/capabilities/mcp`, `examples/mcp` |
| #F3 | Conformance & E2E testing for RC | Dual-era test matrix + benchmarks |
---

Hritik003:
**Description**

**What's changing in the spec:**
- Sessions (`Mcp-Session-Id`) and `initialize`/`ping`/`logging/setLevel` are removed.
- Every request now carries protocol version, client info, and capabilities in `_meta` fields.
- `server/discover` replaces `initialize` for capability discovery; `subscriptions/listen` replaces the GET SSE stream.
- Multi Round-Trip Requests (MRTR) with `InputRequiredResult` replace server-to-client requests (sampling/elicitation/roots).
- New required headers: `Mcp-Method`, `Mcp-Name`, `MCP-Protocol-Version`.

**Approach:**
- The gateway becomes **dual-era**: it detects whether each client/backend speaks legacy or modern protocol and handles both.
- For the modern↔modern path, routing no longer needs sessions — the route comes from the `x-ai-eg-mcp-route` header (already on every request), the target backend from `Mcp-Name`/name-prefix encoding (already exists), and per-backend capabilities from a lazy `server/discover` cache.
- A 2×2 compatibility matrix (legacy/modern × client/backend) is defined for all four cells.

**Immediate scope (Phase 0 + Phase 1):**
- Phase 0: Foundation types, constants, era detection, error codes , split handlers into a shared handlers.go and legacy.go so that changes can be removed easily when legacy spec gets deprecated— zero behavior change.
- Phase 1: Full modern↔modern stateless path (Cell 2) while legacy↔legacy (Cell 1) remains completely unchanged. Covers `server/discover`, stateless routing, `_meta`/header injection, MRTR passthrough, `subscriptions/listen`, and rejection of removed methods.

**Deferred (Phases 2–4):**
- Phase 2: Cross-era translation (modern client ↔ legacy backend and vice versa) — deferred until mixed-era deployments become common.
- Phase 3: Proxy-level result caching, deterministic ordering, minor spec items.
- Phase 4: Auth hardening, deprecation annotations, docs/examples.

**Notes to the reviewers:**
- With Phase 0 and 1, Envoy AI Gateway MCP Route will be **ONLY** able to support either of the two i.e, all the backends attached are `old spec` or all the backends attached are `new spec` and `NOT MIXED`.

Closes: #2323


Hritik003:
Open questions on the proposal:

1. at the moment, Initialize method sends a hardcoded protocol version though the upstream MCP backends might support higher. But as per the new spec, the new method `server/discovery` which replaces `initialize` sends out a list of supported versions in the json response. Refer - [here](https://modelcontextprotocol.io/specification/2026-07-28/server/discover#discoverresult). 
2. In the newer spec, there was caching that got introduced which gives us two fields `ttlms` and `cacheScope` in each of the json response from a backend. Now given that we are multiplexing multiple backends how do we plan out to show `what?` in the aggregated response from all? Refer - [here](https://modelcontextprotocol.io/specification/2026-07-28/server/utilities/caching)
3. Till now, authorize request was only called on initialize(recently added) , tools/list, tools/call, but now should we call on all methods for modern routing?
4. Planning to add independent PRs till the actual integration in phase 0 as mentioned in the pr desc, how do we plan to write e2e tests for uncallable code? 


Jotting down the discussions we had on the above open questions in the community meeting ( 17/08/2026 ):
 
2. one approach would be to return the most restrictive config across the caches. Return the caching config with the
smallest TTL that you have seen from all the responses and whether if there are many that are public or private, if there's one that's private, just return private
    - we shouldn't complicate and  probably having the most returning, the most restrictive thing, it's a reasonable way to get started 
3.  Call it everywhere, add a generalize logic for the cel expression to call any MCP method.
4. Unit tests should be a good testing till phase 1, and as soon as phase 1 lands we can work on writing e2e tests.


cc: @nacx @aishwaryaraimule21

missBerg (approved):
Good for initial iteration of this proposal. We will learn as we go and if needed can be updated with learnings, as this covers a large piece of work I suspect we will become wiser as we work through the plan.

docs/proposals/012-mcp-new-spec-rc/proposal.md
```diff
@@ -0,0 +1,638 @@
...
+| ----------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| **Route name**                      | Already arrives on **every** request in the `x-ai-eg-mcp-route` header (`internalapi.MCPRouteHeader`), set by the Envoy frontend `HTTPRouteFilter`. Today read only on `initialize` (`handlers.go:376`); must be read on every request.                                  |
+| **Backend for single-target calls** | Already encoded in the **namespaced name/URI**: `downstreamResourceName` → `backend__tool`, `downstreamResourceURI` → `backend+scheme://…` (`handlers.go:1204-1243`). `upstreamResourceName`/`upstreamResourceURI` decode it. Now also appears in the `Mcp-Name` header. |
```
aishwaryaraimule21: this would be the name of the MCPRoute

docs/proposals/012-mcp-new-spec-rc/proposal.md
```diff
@@ -0,0 +1,638 @@
...
+	return eraLegacy // conservative default
+}
+```
+
+### D2 — Upstream (Backend) Era Detection
+
+Three strategies, combined:
+
+1. **Auto-detect + cache (spec fallback):** Send `server/discover` to the backend; a valid `DiscoverResult` or a modern JSON-RPC error (`-32022`/`-32020`/`-32021`) ⇒ modern. A `400/404/405` without a recognized modern error body ⇒ fall back to `initialize` ⇒ legacy. Cache era per `route+backend` origin; re-probe on failure.
+2. **Explicit config field** on `MCPBackend` (`internal/filterapi/mcpconfig.go`), e.g., `ProtocolVersion string` or `Stateless *bool`. Useful for deterministic behavior in air-gapped setups.
```
aishwaryaraimule21: I don’t think we should add ProtocolVersion as a MCP Backend field as enforcement is an issue. `Stateless: true` sounds good to me.

aishwaryaraimule21 (approved):
added nits which can be fixed, rest LGTM! 
Thanks @Hritik003 !

swiftdiaries:
The overall direction, phases makes sense to me. I was taking a look at the downstream PRs and the phases and staging seems confusing to me. If we could possible define the boundaries for each phase and make the ownership and dependencies of each explicit, that'd be very helpful.

A suggestion: Can we make the phases inline with the request lifecycle? This would really help staging the downstream PRs into logically independent, reviewable pieces. 

An example of the current boundary problem is validation: #2518 introduces `detectClientEra` with validation, while #2545 independently infers the client era.

Trying to see if we can establish a boundary for PR #2518 and split PR #2545 into two PRs each with it's own boundary. Move to caching, subscription, activation after that. It looks like the dual-era logic introduces a lot of changes that might need more careful review.

### Staging recommendation (AI-generated)

#### Showing my work

I reviewed the PRs separately and used that as the "grounding" for the staging. Used `/grill-with-docs` skill to try and understand the review comments, suggestions and wrote the above review comment.


#### Implementation staging

1. **Downstream ingress and classification — #2518 / Phase 0.1**
   - Parse the request once.
   - Detect the client era.
   - Validate the modern headers and `_meta`, including `Mcp-Method` and `Mcp-Name`.
   - Define shared protocol errors and header codecs.
   - Separate the legacy handler without changing its HTTP or observability behavior.
     - ref: https://github.com/envoyproxy/ai-gateway/pull/2518/changes#r3885804318

   Acceptance: later phases receive either a validated legacy request, a validated modern request, or a protocol error.

2. **Backend selection and discovery — Phase 0.2**
   - Apply `backendSelector`.
   - Discover every selected backend.
   - Preserve positive, unknown, and contradictory backend evidence.
   - Calculate protocol versions common to the gateway and all usable selected backends.
   - Reject mixed or unresolved backend sets according to the homogeneous-route requirement.
     - ref: https://github.com/envoyproxy/ai-gateway/pull/2545#discussion_r3886022710

3. **Modern forwarding and response handling — Phase 0.2/0.3**
   - Consume the validated request from Phase 0.1 rather than performing another client-era check.
   - Construct the required upstream headers.
   - Implement fanout and single-target routing.
   - Validate upstream JSON-RPC envelopes before merging results.
   - Define partial-success and zero-success behavior explicitly.
     - ref:
       - https://github.com/envoyproxy/ai-gateway/pull/2545#discussion_r3886843734
       - https://github.com/envoyproxy/ai-gateway/pull/2545#discussion_r3886794200

4. **Subscriptions**
   - Keep cache scope, invalidation, and `subscriptions/listen` ownership explicit. If #2545 implements only part of Phase 0.2, identify the remaining PR rather than treating the full Phase 0.2 checklist as complete.
   - note: It looks like we're deferring caching from the latest changes
  
5. **Activation — Phase 1**
   - Wire `detectClientEra` into `servePOST`.
   - Dispatch only validated requests.
   - Require positive homogeneous backend-era evidence before enabling the modern route.
   - Verify the complete legacy and modern paths through E2E tests and observability checks.

PR #2433, opened 2026-07-27
fix: allow `include_thoughts` + `thinking_level` together for Gemini 3.x

hustxiayang:
**Description**
For GCP Vertex Gemini models, the gateway could not send `thinking_level` and `include_thoughts` in the same request. As a result, callers using `reasoning_effort` (which maps to `thinking_level`) could **never** receive thought summaries (`reasoning_content`), even though Gemini 3.x returns them when both fields are set. 

This pr is to allow a separate bool value: `include_thoughts` to determine whether it returns the thought summary.

```python
client.chat.completions.create(
    model="gemini-3.5-flash",
    messages=[...],
    reasoning_effort="high",                 # depth  -> ThinkingLevel  (unchanged for all models)
    extra_body={"include_thoughts": True},   # opt-in summary -> IncludeThoughts  (new, Gemini-only)
)
```

This would not break existing interface. There is a bit duplication with the budget setup, but it's fine as the budget setup is deprecating.


gavrissh:
We should document this somewhere? because the gemini spec expects to be sent within thinking_config?

```
 extra_body={
        "google": {
            "thinking_config": {
                "include_thoughts": True,
            }
        }
    }
```

PR #2434, opened 2026-07-27
fix: fix tool-search in response api

hustxiayang:
**Description**

Currently, ai-gateway did not recognize `tool_search_call` and `tool_search_output` in responses input/output item unions. This caused two errors:

1. When parsing a response output item for a non-streaming request, ai-gateway can fail a valid upstream response with:  `unknown_type_field_value_'tool_search_call'_for_response_output_item_union`
This is because the response output union does not know that type, parsing hits the output-union default error: https://github.com/envoyproxy/ai-gateway/blob/6722cca8d33896c4464c12f2de5aaf1238a569b6/internal/apischema/openai/openai.go#L6452.  For streaming requests, this error is not visible because ai-gateway skips SSE event decode errors.

2. When parsing a later request that resends prior output items as `input`, ai-gateway would reject the request with:
`cannot unmarshal unknown input type: tool_search_call` This is because response input union does not know that type: https://github.com/envoyproxy/ai-gateway/blob/6722cca8d33896c4464c12f2de5aaf1238a569b6/internal/apischema/openai/openai.go#L3984

**Fix** Add `tool_search_call` and `tool_search_output` to both unions.
**Other Schema Items Added**:
- `additional_tools`
- `compaction_trigger`
- `computer_call_output`
- `custom_tool_call_output`
- `mcp_approval_response`


aabchoo:
@hustxiayang could you please fix the tests

hustxiayang:
@aabchoo this is due to an independent pr https://github.com/envoyproxy/ai-gateway/pull/2459 added some field. Would update it.

aabchoo (approved):

PR #2441, opened 2026-07-28
fix: emit dynamic metadata on error responses for access-log attribution

### LINKED ISSUE #2440: extproc: error responses lose all LLM access-log dimensions (backend/model/tokens incl. reasoning_tokens)
## Summary

When an upstream LLM returns a non-2xx response (4xx/5xx), the Envoy access log loses every LLM-specific dimension: `gen_ai.provider.name`, `gen_ai.response.model`, and all `gen_ai.usage.*` token fields (input/output/total/reasoning) are emitted as `null`, even though the request was successfully routed to a backend (`upstream_host` is populated).

This defeats the main purpose of the LLM access log for exactly the requests operators most need to debug — failed calls — and specifically nullifies the `gen_ai.usage.reasoning_tokens` field that was just surfaced for reasoning models.

## Reproduction

Run `aigw run` against any OpenAI-compatible backend and send a request the upstream rejects (e.g. an unsupported model):

```
curl -s -X POST http://localhost:1975/v1/chat/completions \
  -H "Authorization: Bearer $OPENAI_API_KEY" \
  -d '{"model":"gpt-4o-mini","messages":[{"role":"user","content":"hi"}]}'
```

The upstream returns 400. The resulting access log line:

```
{"gen_ai.provider.name":null,"gen_ai.request.model":"gpt-4o-mini","gen_ai.response.model":null,
 "gen_ai.usage.input_tokens":null,"gen_ai.usage.output_tokens":null,"gen_ai.usage.reasoning_tokens":null,
 "response_code":400,"upstream_host":"104.21.69.189:443", ...}
```

`upstream_host` proves the request reached a backend, but `provider.name` / `response.model` / every usage token is null. Compare with a 200 response where the same fields are fully populated.

## Root cause

`internal/extproc/processor_impl.go`, `ProcessResponseBody`:

When the response status is not 2xx (`!isGoodStatusCode(code)`), the method calls `u.translator.ResponseError(...)` and returns early, before the `buildDynamicMetadata(...)` call that runs only on the success path. As a result `resp.DynamicMetadata` stays nil for error responses, so `backend_name`, `ai_service_backend_name`, `model_name_override`, `route_name`, and all token-cost keys are never written to dynamic metadata — which the access log `%DYNAMIC_METADATA(io.envoy.ai_gateway:...)%` expressions then read as null.

The contract is even documented as intentional in the `buildDynamicMetadata` doc comment: "It is not called for failed requests or partial responses." But `backend_name` and `model_name_override` are already known at request time (`SetBackend`) and do not depend on a successful response body, so there is no reason to withhold them from the access log on errors.

## Impact

- Operators cannot attribute failed LLM calls to a backend or route from the access log alone.
- For reasoning models (gpt-5.x, o-series), failed calls produce no `reasoning_tokens` (nor input/output) in the access log, the field recently surfaced for visibility.
- Quota/rate-limit consumers are unaffected today because the stream-done `hits_addend` reads the token cost, which is 0 on error — so emitting the metadata on errors would not charge quota (the per-token addend is 0).

## Suggested fix

On the error branch of `ProcessResponseBody`, when `body.EndOfStream` is true and request costs are configured, call `buildDynamicMetadata(...)` with `responseModel=""` (error responses carry no model) and attach it to `resp.DynamicMetadata`, mirroring the success path. `response_model` stays absent (it is only emitted when non-empty), while `backend_name` / `model_name_override` / `route_name` / token costs (0) become available to the access log.

The token-cost values are 0 on errors (no tokens were consumed), so the rate-limit `hits_addend` remains 0 — no quota behavioral change, only access-log observability.

Aias00:
**Description**

When an upstream LLM returns a non-2xx response, ProcessResponseBody returns
early after translator.ResponseError and never calls buildDynamicMetadata.
resp.DynamicMetadata stays nil, so the access log reads every LLM dimension as
null for failed requests: gen_ai.provider.name, gen_ai.response.model, and all
gen_ai.usage.* token fields (input/output/total/reasoning), even though the
request was routed to a backend (upstream_host is populated).

This defeats the access log for exactly the requests operators most need to
debug, and nullifies gen_ai.usage.reasoning_tokens for failed reasoning-model
calls.

backend_name and model_name_override are known at request time (SetBackend) and
do not depend on a successful response body, so withholding them on errors is
unjustified.

## Fix

On the error branch of ProcessResponseBody, when body.EndOfStream and request
costs are configured, call buildDynamicMetadata with responseModel empty and
attach it to resp.DynamicMetadata, mirroring the success path. Updated the
buildDynamicMetadata doc comment accordingly.

- Token costs are 0 on errors (no tokens consumed), so the rate-limit
  stream-done hits_addend (which reads the cost via quotaHitsAddend) stays 0:
  observability only, no quota behavioral change.
- response_model stays absent on errors because error responses carry no model
  (buildDynamicMetadata only emits it when non-empty).

## Test

Added a subtest "non-2xx status emits dynamic metadata at end of stream"
asserting backend_name, ai_service_backend_name, model_name_override, route_name
are present and token costs are 0, response_model absent, on a 500 response.

go test ./internal/extproc/... passes.

## Verified live

aigw run against an OpenAI-compatible backend, sending a request the upstream
rejects (gpt-4o-mini, 400):

Before: gen_ai.provider.name=null, gen_ai.usage.input_tokens=null,
gen_ai.usage.output_tokens=null, gen_ai.usage.reasoning_tokens=null
(upstream_host present).

After: gen_ai.provider.name="default/openai/route/aigw-run/rule/0/ref/0",
gen_ai.usage.input_tokens=0, gen_ai.usage.output_tokens=0.

reasoning_tokens is null on this branch only because the reasoning cost rule
itself is added by a separate change (#2438); once configured, it follows the
same path and will read 0 on errors instead of null.

Fixes #2440

### DIFF
--- a/internal/extproc/processor_impl.go
+++ b/internal/extproc/processor_impl.go
@@ -521,7 +521,7 @@ func (u *upstreamProcessor[ReqT, RespT, RespChunkT, EndpointSpecT]) ProcessRespo
 		}
 		// Mark so the deferred handler records failure.
 		recordRequestCompletionErr = true
-		return &extprocv3.ProcessingResponse{
+		resp := &extprocv3.ProcessingResponse{
 			Response: &extprocv3.ProcessingResponse_ResponseBody{
 				ResponseBody: &extprocv3.BodyResponse{
 					Response: &extprocv3.CommonResponse{
@@ -530,7 +530,34 @@ func (u *upstreamProcessor[ReqT, RespT, RespChunkT, EndpointSpecT]) ProcessRespo
 					},
 				},
 			},
-		}, nil
+		}
+		// Emit dynamic metadata at end of stream so failed requests still carry
+		// backend_name / model_name_override / route_name / token costs (0) in the
+		// access log. backend_name and model_name_override are known from the request
+		// phase (SetBackend), so unlike token usage they do not depend on a successful
+		// response body. Token costs are 0 on errors (no tokens consumed), so the
+		// rate-limit stream-done hits_addend stays 0 — observability only, no token
+		// quota charge (verified by an e2e in tests/e2e/backend_quota_ratelimit_test.go).
+		// responseModel is empty because error responses carry no model.
+		//
+		// If a CEL cost expression fails to evaluate here, skip emitting metadata
+		// and log rather than failing the response: the access log simply stays
+		// null for this request, same as for any failed request without costs.
+		// (Note: llmcostcel.NewProgram sanity-checks each expression at all-zero
+		// token counts — the exact state of a failed request — so division-by-zero
+		// on zero tokens is already rejected at program creation; this guard is
+		// defensive against any other CEL runtime error.)
+		if body.EndOfStream && u.parent.config != nil &&
+			(len(u.parent.config.GlobalRequestCosts) > 0 || len(u.parent.config.RequestCosts) > 0) {
+			metadata, mdErr := buildDynamicMetadata(u.parent.config.GlobalRequestCosts, u.parent.config.RequestCosts, &u.costs, u.requestHeaders, u.backendName, u.routeName, internalapi.ResponseModel(""))
+			if mdErr != nil {
+				u.logger.Warn("failed to build dynamic metadata for error response, skipping access-log metadata",
+					slog.Any("error", mdErr))
+			} else {
+				resp.DynamicMetadata = metadata
+			}
+		}
+		return resp, nil
 	}
 
 	newHeaders, newBody, tokenUsage, responseModel, err := u.translator.ResponseBody(u.responseHeaders, decodingResult.reader, body.EndOfStream, u.parent.span)
@@ -816,9 +843,12 @@ func evalRuntimeRequestCost(rc *filterapi.RuntimeRequestCost, costs *metrics.Tok
 }
 
 // buildDynamicMetadata creates metadata for rate limiting and cost tracking.
-// This function is called by the upstream filter only at the end of the stream (body.EndOfStream=true)
-// when the response is successfully completed. It is not called for failed requests or partial responses.
-// The metadata includes token usage costs and model information for downstream processing.
+// This function is called by the upstream filter only at the end of the stream
+// (body.EndOfStream=true), on both successful and failed responses. For failed
+// responses (non-2xx), token costs are 0 (no tokens were consumed) and
+// responseModel is empty, but backend_name / model_name_override / route_name
+// are still emitted so the access log can attribute failed requests. The metadata
+// includes token usage costs and model information for downstream processing.
 // Two-tier precedence: for each metadataKey, check route-scoped requestCosts first (matching RouteName == routeName).
 // If found, use it. Otherwise, fall back to globalRequestCosts. If neither exists, the key is not emitted.
 func buildDynamicMetadata(globalRequestCosts []filterapi.RuntimeGlobalRequestCost, requestCosts []filterapi.RuntimeRequestCost, costs *metrics.TokenUsage, requestHeaders map[string]string, backendName, routeName, responseModel string) (*structpb.Struct, error) {
--- a/internal/extproc/processor_impl_test.go
+++ b/internal/extproc/processor_impl_test.go
@@ -399,6 +399,168 @@ func Test_chatCompletionProcessorUpstreamFilter_ProcessResponseBody(t *testing.T
 		mm.RequireRequestFailure(t)
 	})
 
+	// Verify failed (non-2xx) responses still emit dynamic metadata at end of
+	// stream so the access log can attribute them to backend/model/route. Token
+	// costs are 0 since no tokens were consumed. See issue #2440.
+	t.Run("non-2xx status emits dynamic metadata at end of stream", func(t *testing.T) {
+		inBody := &extprocv3.HttpBody{Body: []byte("error-body"), EndOfStream: true}
+		mm := &mockMetrics{}
+		mt := &mockTranslator{
+			t: t, expResponseBody: inBody,
+			retHeaderMutation: []internalapi.Header{{"foo", "bar"}},
+			retBodyMutation:   []byte("error-body"),
+		}
+		p := &chatCompletionProcessorUpstreamFilter{
+			translator: mt,
+			metrics:    mm,
+			parent: &chatCompletionProcessorRouterFilter{
+				config: &filterapi.RuntimeConfig{
+					RequestCosts: []filterapi.RuntimeRequestCost{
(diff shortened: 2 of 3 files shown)


nacx:
@kanurag94 can you take a look?

internal/extproc/processor_impl.go
```diff
@@ -530,7 +530,23 @@ func (u *upstreamProcessor[ReqT, RespT, RespChunkT, EndpointSpecT]) ProcessRespo
...
+		// Emit dynamic metadata at end of stream so failed requests still carry
+		// backend_name / model_name_override / route_name / token costs (0) in the
+		// access log. backend_name and model_name_override are known from the request
+		// phase (SetBackend), so unlike token usage they do not depend on a successful
+		// response body. Token costs are 0 on errors (no tokens consumed), so the
+		// rate-limit stream-done hits_addend stays 0 — observability only, no quota
+		// behavioral change. responseModel is empty because error responses carry no model.
+		if body.EndOfStream && u.parent.config != nil &&
+			(len(u.parent.config.GlobalRequestCosts) > 0 || len(u.parent.config.RequestCosts) > 0) {
+			metadata, mdErr := buildDynamicMetadata(u.parent.config.GlobalRequestCosts, u.parent.config.RequestCosts, &u.costs, u.requestHeaders, u.backendName, u.routeName, internalapi.ResponseModel(""))
```
kanurag94, line 552: I am worried about division by zero for CEL expressions since input_tokens/ total_tokens can become 0 now and this will error. Can you see if we can skip and log it confidently instead of returning an error?

internal/extproc/processor_impl.go
```diff
@@ -530,7 +530,23 @@ func (u *upstreamProcessor[ReqT, RespT, RespChunkT, EndpointSpecT]) ProcessRespo
...
 				},
 			},
-		}, nil
+		}
+		// Emit dynamic metadata at end of stream so failed requests still carry
+		// backend_name / model_name_override / route_name / token costs (0) in the
+		// access log. backend_name and model_name_override are known from the request
+		// phase (SetBackend), so unlike token usage they do not depend on a successful
+		// response body. Token costs are 0 on errors (no tokens consumed), so the
+		// rate-limit stream-done hits_addend stays 0 — observability only, no quota
```
kanurag94: The `hits_addend` stays 0, but I don't think the descriptor stays out of it.

`baseDescriptorActions()` builds the stream-done entry from the `ai_service_backend_name` and `model_name_override` metadata keys and neither action sets `default_value` or `skip_if_absent`, so today those keys are absent on a non-2xx and Envoy never populates that descriptor - no stream-done ratelimit call goes out at all.
After this change all three keys are present (`quota_cost` renders as "0" since the controller injects the quota cost expression as an LLMRequestCost), so the `apply_on_stream_done` entry will fire on every failed request.

Can we get an e2e confirming a 500 doesn't decrement quota?

internal/extproc/processor_impl_test.go
```diff
@@ -399,6 +399,55 @@ func Test_chatCompletionProcessorUpstreamFilter_ProcessResponseBody(t *testing.T
 		mm.RequireRequestFailure(t)
 	})
 
+	// Verify failed (non-2xx) responses still emit dynamic metadata at end of
+	// stream so the access log can attribute them to backend/model/route. Token
+	// costs are 0 since no tokens were consumed. See issue #2440.
+	t.Run("non-2xx status emits dynamic metadata at end of stream", func(t *testing.T) {
+		inBody := &extprocv3.HttpBody{Body: []byte("error-body"), EndOfStream: true}
+		mm := &mockMetrics{}
+		mt := &mockTranslator{t: t, expResponseBody: inBody,
```
kanurag94:
```suggestion
		mt := &mockTranslator{
			t: t, expResponseBody: inBody,
```

internal/extproc/processor_impl_test.go
```diff
@@ -399,6 +399,55 @@ func Test_chatCompletionProcessorUpstreamFilter_ProcessResponseBody(t *testing.T
...
+		// the access log reads 0 instead of null.
+		require.Equal(t, float64(0), fields["input_token_usage"].GetNumberValue())
+		require.Equal(t, float64(0), fields["output_token_usage"].GetNumberValue())
+		require.Equal(t, float64(0), fields["reasoning_token_usage"].GetNumberValue())
+		// Error responses carry no model; response_model must be absent (not "").
+		_, ok := fields["response_model"]
+		require.False(t, ok)
+		mm.RequireRequestFailure(t)
+	})
+
```
kanurag94, line 451: Can you please add:

1. Case with `EndOfStream: false` asserting we don't emit metadata
2. Case with `GlobalRequestCosts`

internal/extproc/processor_impl.go
Aias00: Good catch. The error path now logs a warning and skips emitting metadata instead of returning the error: `if mdErr != nil { u.logger.Warn(...); } else { resp.DynamicMetadata = metadata }`.

One nuance worth noting: `llmcostcel.NewProgram` sanity-checks every CEL expression at all-zero token counts (cel.go), which is the exact state of a failed request (the error path never parses tokens, so all costs stay 0). So a division-by-zero on zero tokens is already rejected at program creation time, before it can ever reach the error path. The log-and-skip guard is defensive against any other CEL runtime error (e.g. unsigned integer overflow), and there is now a unit test that pre-sets costs to trigger such an overflow and asserts the response still succeeds with nil metadata.

internal/extproc/processor_impl.go
Aias00: You are right — the stream-done descriptor does fire on failed requests now, whereas before it never did (backend_name/model_name_override were absent). The token-cost hits_addend is still 0, so no token quota is charged, but the descriptor is sent.

To confirm the user-visible effect is unchanged, I added an e2e in tests/e2e/backend_quota_ratelimit_test.go: "non-2xx response does not charge token quota". It makes the upstream return 500 with a body claiming 100 total_tokens, then asserts the next request still succeeds (200). If those 100 tokens were charged, the 10-token-per-hour limit would be exceeded and the next request would 429. CI runs this as part of Test_Examples_BackendQuotaRateLimit.

internal/extproc/processor_impl_test.go
Aias00: Applied, thanks.

internal/extproc/processor_impl_test.go
Aias00: Added both:
1. "non-2xx status does not emit metadata before end of stream" — EndOfStream:false, asserts res.DynamicMetadata is nil.
2. "non-2xx status emits global request costs" — uses GlobalRequestCosts (OutputToken + TotalToken) and asserts the keys are present with 0 and backend_name is set.

Also added "non-2xx status skips metadata on CEL error" covering the skip-and-log path above.

PR #2444, opened 2026-07-28
fix(tracing): don't record an empty streaming model as llm.model_name

### LINKED ISSUE #2443: Streaming chat-completion spans record empty llm.model_name when the first SSE chunk carries no model
_Description_:

For streaming chat completions, the OpenInference span's `llm.model_name` attribute is silently overwritten with an empty string whenever the upstream's first SSE chunk doesn't carry a model name. This happens deterministically with Azure OpenAI / Azure AI Foundry backends, which send a leading content-filter "prologue" chunk with `model:""` before the chunk that actually names the model.

| | Non-streaming | Streaming (Azure OpenAI backend) |
|---|---|---|
| `llm.model_name` on the span | correct, e.g. `gpt-5.4-nano-2026-03-17` | `""` (empty) |
| Cost attribution in an OpenInference consumer (e.g. Arize Phoenix) keyed on `llm.model_name` | correct | null — no price row matches an empty model |
| Token counts (prompt/completion) | correct | correct — unaffected |

**Root cause**

Confirmed at both `v1.0.0` and current `main` (identical line numbers on both):

1. `internal/tracing/openinference/openai/request_attrs.go:36` — `buildRequestAttributes` sets `llm.model_name` from the *request-time* model (`chatRequest.Model`). This is applied via `span.SetAttributes(...)` in `RecordRequest`, `internal/tracing/openinference/openai/chat_completion.go:59`. At this point the attribute is correct.
2. For streaming, `RecordResponseChunks` (`chat_completion.go:63-69`) calls `convertSSEToJSON(chunks)`, which at `internal/tracing/openinference/openai/sse_converter.go:112` builds the aggregated response with `Model: firstChunk.Model` — it unconditionally takes the model from the *first* SSE chunk, with no check for it being non-empty.
3. That aggregated response is passed to `RecordResponse` (`chat_completion.go:77-91`), which calls `buildResponseAttributes` (`response_attrs.go:20-22`) — this **unconditionally** appends `attribute.String(openinference.LLMModelName, resp.Model)` — and then does a second `span.SetAttributes(attrs...)` at `chat_completion.go:90`.

Because OTel span attributes are last-write-wins, step 3's `SetAttributes` call overwrites the correct value set in step 1 with whatever `resp.Model` is — which, per step 2, is whatever the *first* SSE chunk carried, empty or not.

Notably, the Responses API path in the same file already guards against this: `buildResponsesResponseAttributes` (`response_attrs.go:177-181`) only appends the attribute `if resp.Model != ""`. The Chat Completions streaming path (`buildResponseAttributes`, used by both the non-streaming and streaming-aggregated-to-JSON code paths) does not have the same guard.

**Trigger**

Azure OpenAI / Azure AI Foundry backends emit a leading SSE chunk carrying content-filter results with `"model":""` and no `choices`, before the chunk(s) that carry the actual generation and model name. `convertSSEToJSON`'s `firstChunk`/model-capture logic (`sse_converter.go:29-40`) takes the very first chunk seen, so this prologue chunk's empty model wins.

**Precedent**

The same class of bug — first/request model blindly overriding the response's authoritative model — was already identified and fixed for **metrics** (not tracing) in issue #1218 ("metrics: response.model label should use the model name in the response body if available"), fixed by PR #1219 ("feat(metrics): use authoritative model from response and test with otel", merged 2025-09-25) and PR #1230 ("feat(metrics): track original model separately from overridden request model", merged 2025-10-01). Both first shipped in `v0.4.0`. Neither PR touched the OpenInference tracing package, so the SSE-aggregation path (`sse_converter.go`) and the Chat Completions response-attribute path (`response_attrs.go`) were never brought in line with that fix.

**OpenInference reference behavior**

The official Python instrumentation (`Arize-ai/openinference`, `python/instrumentation/openinference-instrumentation-openai/src/openinference/instrumentation/openai/_response_attributes_extractor.py`) only emits the model attribute when it's truthy:

```python
if model := getattr(completion, "model", None):
    yield SpanAttributes.LLM_MODEL_NAME, model
```

An empty/falsy model is never yielded, so it can never clobber an already-correct attribute. This Go implementation's unconditional write in `buildResponseAttributes` diverges from that reference behavior.

**Impact**

Any OpenInference-consuming observability backend that keys cost calculation off `llm.model_name` (e.g. Arize Phoenix) gets **null cost for all streaming chat-completion traffic proxied through Azure OpenAI / Azure AI Foundry backends**. This is deterministic (100% of streaming requests through those backends), not intermittent, since the content-filter prologue chunk is always first in the stream.

**Environment**

- envoyproxy/ai-gateway: reproduced against `v1.0.0` and current `main` (`sse_converter.go` / `response_attrs.go` unchanged between the two for the relevant lines).
- Backend: Azure OpenAI / Azure AI Foundry, streaming chat completions (`stream: true`).
- OpenInference consumer: Arize Phoenix (any OpenInference consumer keying cost on `llm.model_name` is equally affected).

A fix is proposed in a companion PR: it captures the first *non-empty* model across all SSE chunks (rather than only the first chunk), and adds the same `resp.Model != ""` guard to `buildResponseAttributes` that `buildResponsesResponseAttributes` already has.

jeremypng:
**Description**

On streaming chat completions, `convertSSEToJSON` took the model from the first SSE chunk only. Azure OpenAI's leading content-filter chunk carries `model:""`, so the span's `llm.model_name` was overwritten with an empty string (the request-time value is last-write-lost), breaking cost attribution in OpenInference consumers.

This fixes it by capturing the first *non-empty* model across all SSE chunks in the stream, so a leading empty-model chunk (or any later chunk reporting a different value) can't clobber it. It also adds the same `resp.Model != ""` guard to `buildResponseAttributes` that `buildResponsesResponseAttributes` already has for the Responses API path, so an empty response model can never overwrite the request-time `llm.model_name` attribute.

This intentionally differs from the translator's `streamingResponseModel` (used for metrics), which takes the **last** non-empty value across the stream. For tracing, taking the **first** non-empty value is a defensive choice: it means a later, possibly-anomalous chunk (e.g. a routed fallback reporting a different model) can't silently override the model that was already correctly captured earlier in the stream.

Two behavioral guarantees after this change:
- Streamed spans carry the upstream-resolved model as soon as any chunk in the stream reports one, regardless of chunk order.
- An empty model value never overwrites the request-time `llm.model_name` value, on either the streaming or non-streaming Chat Completions path.

Fixes #2443

- Test coverage added (`internal/tracing/openinference/openai`):
  1. `TestBuildResponseAttributesEmptyModelOmitsModelName` (`response_attrs_test.go`) — an empty response model must not emit `llm.model_name` at all.
  2. `sse_converter_test.go`, case "model from first non-empty chunk (Azure content-filter prologue)" — model comes from the first chunk that actually carries one, skipping the empty-model prologue chunk.
  3. `sse_converter_test.go`, case "model retains first non-empty value even if a later chunk differs" — a later chunk reporting a different model does not override the first non-empty value.
- Commit is signed off (DCO) per CONTRIBUTING.md.
- Generative AI disclosure: this fix and its tests were prepared with AI assistance (Claude); I've reviewed and understand the change and take full ownership of it, per CONTRIBUTING.md's generative AI policy.


nacx (approved):
Thanks!

johnugeorge:
LGTM

PR #2446, opened 2026-07-28
feat: allow excluding models from /v1/models

### LINKED ISSUE #2445: Support hidden model aliases
We use model aliases to expose a stable name to clients and to keep older model names working after moving them to a newer upstream model.

Currently every model matched with `x-ai-eg-model` is listed in `/v1/models`. For aliases this is not always wanted, for example when the alias is only for backwards compatibility or for selected clients.

It would be useful to allow routing and model rewriting as today, but not advertise the matched model in `/v1/models`.

Maybe a rule option like `excludeFromModelsEndpoint: true` could work.

Killusions:
**Description**

Adds `excludeFromModelsEndpoint` to `AIGatewayRoute` rules.

Models matched by a rule with this option remain routeable and support `modelNameOverride`, but are excluded from the OpenAI-compatible `/v1/models` response. This supports backwards-compatible aliases and aliases only intended for selected clients.

Fixes #2445

Tested manually.

⚒️ with ❤️ by @siemens

### DIFF
--- a/api/v1alpha1/ai_gateway_route.go
+++ b/api/v1alpha1/ai_gateway_route.go
@@ -306,6 +306,16 @@ type AIGatewayRouteRule struct {
 	// +optional
 	// +kubebuilder:validation:Format=date-time
 	ModelsCreatedAt *metav1.Time `json:"modelsCreatedAt,omitempty"`
+
+	// ExcludeFromModelsEndpoint prevents models matched by this rule from being returned by the
+	// OpenAI-compatible "/models" endpoint. Requests still match this rule and apply its
+	// configured transformations, including model name overrides.
+	//
+	// This is useful for internal or backwards-compatible aliases, as well as duplicate model
+	// definitions with different schemas.
+	//
+	// +optional
+	ExcludeFromModelsEndpoint bool `json:"excludeFromModelsEndpoint,omitempty"`
 }
 
 // AIGatewayRouteRuleBackendRef is a reference to a backend with a weight.
--- a/internal/controller/gateway.go
+++ b/internal/controller/gateway.go
@@ -391,6 +391,10 @@ func (c *GatewayController) reconcileFilterConfigSecret(
 		injectedQuotaCosts := make(map[string]struct{})
 		for ruleIndex := range spec.Rules {
 			rule := &spec.Rules[ruleIndex]
+			if rule.ExcludeFromModelsEndpoint {
+				continue
+			}
+
 			for _, m := range rule.Matches {
 				for _, h := range m.Headers {
 					// If explicitly set to something that is not an exact match, skip.
@@ -421,6 +425,11 @@ func (c *GatewayController) reconcileFilterConfigSecret(
 					}
 				}
 			}
+		}
+		// Second pass: backends are collected for every rule, including rules excluded from
+		// /v1/models — those rules still route traffic and need their backends in the config.
+		for ruleIndex := range spec.Rules {
+			rule := &spec.Rules[ruleIndex]
 			for backendRefIndex := range rule.BackendRefs {
 				backendRef := &rule.BackendRefs[backendRefIndex]
 				b := filterapi.Backend{}
--- a/internal/controller/gateway.go
+++ b/internal/controller/gateway.go
@@ -418,6 +418,10 @@ func (c *GatewayController) reconcileFilterConfigSecret(
 		injectedQuotaCosts := make(map[string]struct{})
 		for ruleIndex := range spec.Rules {
 			rule := &spec.Rules[ruleIndex]
+			if rule.ExcludeFromModelsEndpoint {
+				continue
+			}
+
 			for _, m := range rule.Matches {
 				for _, h := range m.Headers {
 					// If explicitly set to something that is not an exact match, skip.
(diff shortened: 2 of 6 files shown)


siddharth1036:
@Killusions can you explain the usecase for exclusion of models?

Killusions:
> @Killusions can you explain the usecase for exclusion of models?

@siddharth1036 There are many, the most common one would be backwards-compatible aliases which people and new clients should no longer use. But also generic aliases, multiple definitions (openai + anthropic schema without translation since our backend vLLM supports both), private test models, etc.

At Siemens we have all of these and currently need to use a workaround by using regex patterns instead of exact model name matches.

internal/controller/gateway.go
```diff
@@ -393,6 +393,10 @@ func (c *GatewayController) reconcileFilterConfigSecret(
 			rule := &spec.Rules[ruleIndex]
 			for _, m := range rule.Matches {
 				for _, h := range m.Headers {
+					if rule.ExcludeFromModelsEndpoint {
+						continue
+					}
+
```
siddharth1036: nit, 
```suggestion
if !rule.ExcludeFromModelsEndpoint {
    for _, m := range rule.Matches {
        for _, h := range m.Headers { /* ... collect models ... */ }
    }
}

```

Killusions: Applied, thanks.

internal/controller/gateway.go
```diff
@@ -391,33 +391,35 @@ func (c *GatewayController) reconcileFilterConfigSecret(
...
-					model := filterapi.Model{
-						Name:      h.Value,
-						CreatedAt: ptr.Deref[metav1.Time](rule.ModelsCreatedAt, aiGatewayRoute.CreationTimestamp).UTC(),
-						OwnedBy:   ptr.Deref(rule.ModelsOwnedBy, defaultOwnedBy),
-					}
-					ec.Models = append(ec.Models, model)
-					if len(hostnames) > 0 {
-						if ec.ModelsByHost == nil {
-							ec.ModelsByHost = make(map[string][]filterapi.Model)
+			if !rule.ExcludeFromModelsEndpoint {
```
nacx: style nit: prefer following the [line of sight](https://medium.com/@matryer/line-of-sight-in-code-186dd7cdea88) pattern:
```go
if rule.ExcludeFromModelsEndpoint {
    continue
}
```

api/v1alpha1/ai_gateway_route.go
```diff
@@ -306,6 +306,13 @@ type AIGatewayRouteRule struct {
 	// +optional
 	// +kubebuilder:validation:Format=date-time
 	ModelsCreatedAt *metav1.Time `json:"modelsCreatedAt,omitempty"`
+
+	// ExcludeFromModelsEndpoint prevents models matched by this rule from being returned by the
+	// OpenAI-compatible "/models" endpoint. Requests still match this rule and apply its
+	// configured transformations, including model name overrides.
```
nacx, line 312: Can you also include the example use cases in the API docs? This would help readers better understand when to sue this.

nacx (commented):
Thank you! Just a couple minor comments

internal/controller/gateway.go
nacx: Once the scope of the conditional is fixed, in general, prefer the `continue` pattern for a better [line of sight](https://medium.com/@matryer/line-of-sight-in-code-186dd7cdea88). It makes code easier to read.

internal/controller/gateway.go
Killusions: Done.

api/v1alpha1/ai_gateway_route.go
Killusions: Done, added in all places but very minimal, can expand if you want.

Killusions:
@nacx Pretty sure it's an unrelated failure (after I fixed the precommit), re-triggered the tests by re-pushing.

internal/controller/gateway_test.go
```diff
@@ -429,6 +429,23 @@ func TestGatewayController_reconcileFilterConfigSecret_HostnameScopedModels(t *t
...
+						ExcludeFromModelsEndpoint: true,
+						Matches: []aigv1b1.AIGatewayRouteRuleMatch{
+							{Headers: []gwapiv1.HTTPHeaderMatch{
+								{Name: internalapi.ModelNameHeaderKeyDefault, Value: "hidden-alias"},
+							}},
+						},
+					},
+				},
+			},
+		},
```
kanurag94, line 566: The reason why we need two loops is not tested here. What do you think about adding a test so that an excludedEndpoint still has it's backendRefs

internal/controller/gateway.go
```diff
@@ -421,6 +425,9 @@ func (c *GatewayController) reconcileFilterConfigSecret(
 					}
 				}
 			}
+		}
+		for ruleIndex := range spec.Rules {
```
kanurag94, line 458:
```suggestion
        Second pass: backends are collected for every rule, including rules excluded from
        /v1/models — those rules still route traffic and need their backends in the config.
		for ruleIndex := range spec.Rules {
```

kanurag94 (approved):

kanurag94:
Thanks, LGTM! Deferring to @envoyproxy/ai-gateway-maintainers

siddharth1036 (approved):
Please take a look at the upgrade failure. Rest LGTM!

Killusions:
@mathetake does this still make sense? Let me know if you want me to change anything.

Killusions:
> Please take a look at the upgrade failure. Rest LGTM!

Rebased, it was just a transitive CI failure.

PR #2450, opened 2026-07-28
feat(tracing): opt-in to record the caller's remote parent as a span link

### LINKED ISSUE #2449: extproc OpenInference spans join the caller's trace — double-counted token/cost rollups in Phoenix for self-instrumented callers
### Description

The extproc always continues the caller's propagated `traceparent`: `StartSpanAndInjectHeaders` extracts the incoming headers and starts the LLM span with that remote context as parent (`internal/tracing/tracer.go`). For callers that also instrument their own LLM client (e.g. `openinference-instrumentation-openai` — the standard app-side setup for the Phoenix integration this project documents), every request produces **two** LLM spans in the same trace: the app's client span and the gateway's.

Phoenix — the backend featured in the tracing docs — resolves a span's **project** from the trace it joins, not from the span's own resource attributes: trace IDs are globally unique in its schema (`UniqueConstraint("trace_id")` on the traces table) and `insert_span` attaches a span to the existing trace row's project, explicitly ignoring the incoming project for existing traces ([span.py](https://github.com/Arize-ai/phoenix/blob/arize-phoenix-v19.8.0/src/phoenix/db/insertion/span.py#L32-L60)). Its per-project token/cost rollups then sum every `span_kind=LLM` span in the project ([token_counts.py](https://github.com/Arize-ai/phoenix/blob/arize-phoenix-v19.8.0/src/phoenix/server/api/dataloaders/token_counts.py#L97-L129)).

Consequences for any deployment with a self-instrumented caller:

1. **The gateway span lands in the caller's Phoenix project** (the extproc's `OTEL_RESOURCE_ATTRIBUTES=openinference.project.name=...` has no effect for existing traces), and that project's **token/cost totals double-count** every request — we measured exactly 2x on LLM token rollups in production-like traffic (paired ChatCompletion spans, identical timing and token totals, one serialized by Python `json.dumps`, one by Go).
2. **The gateway's own project has no root spans**, so its default root-span trace views appear empty.

### Repro

1. Gateway with tracing to Phoenix (per the docs), any consumer that propagates W3C `traceparent` from an OpenInference-instrumented OpenAI client.
2. Make one chat completion through the gateway.
3. The consumer's Phoenix project shows two `ChatCompletion` spans with the same token usage in one trace; project token totals count both.

### Why existing knobs don't help

`OTEL_PROPAGATORS=none` stops the parent continuation but is a sledgehammer: it also disables `Inject`, so the upstream provider stops receiving any `traceparent`, MCP `_meta` propagation dies, and all correlation with the caller is lost.

### Proposal

Opt-in env (default = current behavior): when set, gateway spans start their own trace as **root spans**, recording the caller's context as an OTel **span link** plus `caller.trace_id`/`caller.span_id` attributes (Phoenix currently drops OTLP links at ingestion — [`links: None  # TODO`](https://github.com/Arize-ai/phoenix/blob/arize-phoenix-v19.8.0/src/phoenix/server/api/routers/v1/spans.py#L419) — so the attributes keep correlation queryable). Inject is unchanged, so the upstream provider joins the gateway-rooted trace.

We've been running this in production as a fork patch; PR to follow. Naming/config-surface (env vs. tracing config API) flexible.

jeremypng:
**Description**

Fixes #2449.

By default the extproc continues the caller's propagated traceparent, so gateway spans join the calling application's trace. When the caller also instruments its own LLM client (e.g. OpenInference), one request produces two LLM spans in the same trace, and trace backends that group traces into projects — such as Phoenix, whose span ingestion assigns a span to the project of the existing trace row — double-count the request's token/cost rollups in the caller's project. The gateway's own project is also left with no root spans.

This adds `AI_GATEWAY_TRACING_REMOTE_PARENT_AS_LINK`: when `"true"`, a wrapper on the shared tracer starts every span as the root of a new trace whenever the start context carries a remote span context, recording that context as a span link plus `caller.trace_id`/`caller.span_id` attributes (some backends, including Phoenix, do not ingest OTLP links). One seam covers the generic request tracer behind every LLM endpoint and the MCP tracer. Unset, behavior is byte-for-byte unchanged.

`OTEL_PROPAGATORS=none` is not a substitute: it also disables Inject, so the upstream provider stops receiving any traceparent, MCP `_meta` propagation dies, and all correlation with the caller is lost.

Notes:

- Default is unchanged behavior; the feature is pure opt-in and the env var is read once at `NewTracingFromEnv`.
- Tests cover: link/root/attribute/carrier semantics with a propagated traceparent, default-off parent continuation, inertness with no incoming context, and the MCP `_meta` path.
- Docs: added a section to `site/docs/capabilities/observability/tracing.md`.
- We run this in production (Azure OpenAI backend, Phoenix 19.8.0) — verified live that gateway spans land only in the gateway's own Phoenix project as root spans with `caller.*` correlation attributes, and consumer projects stop double-counting.
- Happy to move the knob to the tracing config API or rename the env/attributes if maintainers prefer — naming follows the `AI_GATEWAY_TRACING_*` precedent from #2405.


nacx (approved):

johnugeorge:
@jeremypng Can you fix the CI errors?

jeremypng:
@johnugeorge rebased onto current main (the branch had conflicted with the v1.1 tracing refactor — recorders struct etc.). Squashed to a single clean commit; behavior is unchanged from the revision @nacx approved: opt-in env, unset = byte-identical upstream behavior. `go test ./internal/tracing/` green locally; CI should be able to run now.

nacx:
Thanks! Mind fixing the DCO?

PR #2455, opened 2026-07-30
fix(openai): support Codex agent input items

### LINKED ISSUE #2442: Responses API rejects Codex-emitted `agent_message` and `agent_reasoning` input item types
## Summary

The Envoy AI Gateway Responses API handler strictly rejects request input item types it does not recognize, returning:

```
malformed request: failed to parse JSON for /v1/responses: cannot unmarshal unknown input type: agent_message
```

The OpenAI **Codex** client emits two input item types on the `/v1/responses` endpoint that the gateway currently does not model:

- `agent_message`
- `agent_reasoning`

Both fall through to the `default` branch of `ResponseInputItemUnionParam.UnmarshalJSON` and are rejected before the request reaches the upstream provider, which breaks Codex traffic routed through the gateway.

## Reproduction

Send a minimal `/v1/responses` request whose `input` array contains one of these types, e.g.:

```json
{
  "model": "<codex-targeted-model>",
  "input": [
    {"type": "agent_message", "role": "agent", "content": [{"type": "text", "text": "prior context"}]},
    {"type": "message", "role": "user", "content": "hi"}
  ]
}
```

Observed: HTTP 400 with `cannot unmarshal unknown input type: agent_message`.
Expected: the request passes through to the upstream provider.

## Location

`internal/apischema/openai/openai.go` — the input item union switch hits its default error path for any unhandled `type`:

https://github.com/envoyproxy/ai-gateway/blob/main/internal/apischema/openai/openai.go#L4181

## Proposal

These types appear to be Codex client-emitted (rather than part of the publicly documented Responses API input item union), so a full schema model may be out of scope. The lowest-risk, backward-compatible fix would be to **gracefully ignore unrecognized input item types** (leave the union field zero-valued and continue parsing the remaining items) rather than failing the whole request — consistent with how other pass-through cases behave, and in the spirit of the open work adding more Codex-emitted types (#2164, #2303, and the in-flight #2434 which already proposes adding `additional_tools` and several others).

## Environment

Reproduced against latest `main` (`6722cca8`).

0YHR0:
**Description**

Allow Responses API requests containing Codex-emitted agent_message and agent_reasoning input items to pass through OpenAI-compatible backends.

**Root cause**

The Responses input-item union rejected these types before the request reached the upstream provider.

**Fix**

Preserve the raw JSON for both types, so their evolving Codex-specific fields are forwarded unchanged without requiring an incomplete schema model.

**Related Issue**

Fixes #2442

**Testing**

- go test ./internal/apischema/openai -run '^TestResponseInputItemUnionParam(MarshalJSON|UnmarshalJSON)$' -count=1

### DIFF
--- a/internal/apischema/openai/openai.go
+++ b/internal/apischema/openai/openai.go
@@ -3920,6 +3920,10 @@ type ResponseInputItemUnionParam struct {
 	OfCustomToolCallOutput *ResponseCustomToolCallOutputParam
 	OfCustomToolCall       *ResponseCustomToolCall
 	OfItemReference        *ResponseInputItemItemReferenceParam
+	// Codex-emitted item types whose schema is not yet part of the public Responses API.
+	// Preserve their raw JSON so OpenAI-compatible backends can handle them unchanged.
+	OfAgentMessage   *json.RawMessage
+	OfAgentReasoning *json.RawMessage
 }
 
 func (r ResponseInputItemUnionParam) MarshalJSON() ([]byte, error) { // nolint:gocritic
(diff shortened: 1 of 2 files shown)


nacx (approved):
Thanks!

Issue #2460, opened 2026-07-30
`QuotaPolicy` does not enforce as expected


AyushSawant18588:
_Description_:

**What is being seen:**
1. A **single-model** `QuotaPolicy` (only `defaultBucket`) works incorrectly when its used on AIGatewayRoute containing more than one backendRefs, each with a different modelNameOverride: Once the model in the quota is exhausted, requests to the route return a hard 429 immediately, even though other models are also configured on the same AIGatewayRoute. The request is never routed to other models.

2. A **multi-model** `QuotaPolicy` can do incorrect routing:
Setup:
```
AIGatewayRoutewith one rule containing twobackendRefs(weight 90/10), sameAIServiceBackend, differentmodelNameOverride:gpt-4o-mini and gpt-5-mini.
One QuotaPolicy targeting that backend with two simple per-model quotas:
gpt-4o-mini: default Bucket limit 100 / 1h
gpt-5-mini: default Bucket limit 200 / 1h
```
Observed:
```
When gpt-4o-mini is exhausted (≥100) but gpt-5-mini is still under (e.g. 30/200), requests can still route to and served by gpt-4o-mini. Its counter keeps climbing past 100.
A 429 only starts once both gpt-4o-mini ≥ 100 and gpt-5-mini ≥ 200.
```

3. Quota Policy `bucketRules` + `clientSelectors` (per-tenant limits) are **counted in Redis but never enforced**  the per-tenant counter climbs far past the limit, `ratelimit.ok` keeps incrementing, `ratelimit.over_limit` never does, and no `429` is returned.

**What should happen instead:**
- When one model/backend is exhausted, the request should be routed to another model/backend that still has quota; `429` only when all are exhausted.
- `bucketRules` + `clientSelectors` should enforce per-tenant limits (tenant over its own limit → `429`), independent of the default bucket.

**Why adding `bucketRules` + `clientSelectors` in Quota Policy breaks quota enforcement** 
As per the logic used for Quota policy in rate limit service:
- Simple policy: one request → one quota descriptor (backend/model). RLS trips OVER_LIMIT as soon as it's over. Enforcement works.
- With clientSelectors: one request → two quota descriptors, the matched tenant rule leaf and an always-firing default bucket.
- RLS quota rule is AND: OVER_LIMIT only when all quota descriptors are over. So the tenant leaf being over its own limit isn't enough the default bucket must be over too.
[...]

PR #2462, opened 2026-07-31
feat: carry backend weight/priority into ext_proc config

Aias00:
**Description**

`AIGatewayRouteRuleBackendRef` already exposes `Weight` and `Priority`, but the controller dropped both before building `filterapi.Backend`, so the ext_proc had no visibility into a backend's traffic weight or failover priority.

This change carries `Weight` and `Priority` through to `filterapi.Backend` as new optional fields (mirroring the CRD defaults of 1 and 0). It is pure data plumbing with no runtime behavior change: Envoy Gateway's own load balancing continues to use the HTTPRoute weighted clusters and the controller's priority annotation; these fields are consumed by a later change that implements quota-aware backend selection in the router-level ext_proc.

Why separate this out: the upcoming quota-aware routing (proposal 009) needs the ext_proc to know each backend's weight and priority to select among backends with available quota. Landing the plumbing independently keeps that later change focused and reviewable, with no behavior delta today.

Refs #2460

- No behavior change; existing tests in `internal/filterapi` and `internal/controller` pass unchanged.
- New `Weight`/`Priority` fields are optional with `omitempty`, so the serialized filter-config Secret stays backward compatible.
- The full proposal-009 design (filter reorder, descriptor-index to backend mapping, ORIGINAL_DST quota pool, dep bump) is larger and will follow in separate PRs pending maintainer sign-off on a few architectural points; this PR intentionally ships only the non-controversial plumbing.


internal/filterapi/filterconfig.go
```diff
@@ -196,7 +196,17 @@ type Backend struct {
 	// Sensitive Headers to be removed from the request before sending to the backend. Optional.
 	HeaderMutation *HTTPHeaderMutation `json:"httpHeaderMutation,omitempty"`
 	// Body mutations to be applied to the request before sending to the backend. Optional.
-	BodyMutation *HTTPBodyMutation `json:"httpBodyMutation,omitempty"`
+	BodyMutation *HTTPBodyMutation `json:"bodyMutation,omitempty"`
```
kanurag94: Is this change intended?

internal/controller/gateway.go
```diff
@@ -426,6 +426,13 @@ func (c *GatewayController) reconcileFilterConfigSecret(
 				b := filterapi.Backend{}
 				b.Name = internalapi.PerRouteRuleRefBackendName(aiGatewayRoute.Namespace, backendRef.Name, aiGatewayRoute.Name, ruleIndex, backendRefIndex)
 				b.ModelNameOverride = backendRef.ModelNameOverride
+				// Carry Weight/Priority through to the ext_proc config for quota-aware
+				// routing. Envoy Gateway's own load balancing does not read these (it uses
+				// the HTTPRoute weighted clusters and the priority annotation generated
+				// elsewhere in the controller); they are consumed by the router-level
+				// ext_proc when selecting a backend with available quota.
+				b.Weight = backendRef.Weight
+				b.Priority = backendRef.Priority
```
kanurag94: Can you please add tests for this?

internal/controller/gateway.go
```diff
@@ -426,6 +426,13 @@ func (c *GatewayController) reconcileFilterConfigSecret(
 				b := filterapi.Backend{}
 				b.Name = internalapi.PerRouteRuleRefBackendName(aiGatewayRoute.Namespace, backendRef.Name, aiGatewayRoute.Name, ruleIndex, backendRefIndex)
 				b.ModelNameOverride = backendRef.ModelNameOverride
+				// Carry Weight/Priority through to the ext_proc config for quota-aware
+				// routing. Envoy Gateway's own load balancing does not read these (it uses
+				// the HTTPRoute weighted clusters and the priority annotation generated
+				// elsewhere in the controller); they are consumed by the router-level
+				// ext_proc when selecting a backend with available quota.
```
kanurag94: the explanation already lives inside the `filterconfig.go`

internal/filterapi/filterconfig.go
```diff
@@ -196,7 +196,17 @@ type Backend struct {
...
+	// Weight is the traffic weight of this backend within its route rule (mirrors
+	// AIGatewayRouteRuleBackendRef.Weight; default 1). Populated by the controller for
+	// quota-aware routing; not used by Envoy Gateway's own load balancing (that path uses
+	// the HTTPRoute weighted clusters generated by the controller).
+	Weight *int32 `json:"weight,omitempty"`
```
kanurag94: Having omitEmpty on these pointer fields may lead to CRD defaulting the weights to 0. Can you please check if it is the case?

kanurag94 (changes requested):

internal/filterapi/filterconfig.go
Aias00: Fixed. That tag change was unintentional; I restored `httpBodyMutation` and added a serialization test to prevent it from regressing.

internal/controller/gateway.go
Aias00: Added coverage in `TestGatewayController_reconcileFilterConfigSecret`

internal/filterapi/filterconfig.go
Aias00: Checked and added `TestBackendMarshalOptionalRoutingFields`. Nil `Weight`/`Priority` pointers are omitted from the marshaled config, so unset values do not get serialized as zero; explicit zero pointers are still
  serialized as `0` when intentionally set.

internal/controller/gateway.go
```diff
@@ -426,6 +426,8 @@ func (c *GatewayController) reconcileFilterConfigSecret(
 				b := filterapi.Backend{}
 				b.Name = internalapi.PerRouteRuleRefBackendName(aiGatewayRoute.Namespace, backendRef.Name, aiGatewayRoute.Name, ruleIndex, backendRefIndex)
 				b.ModelNameOverride = backendRef.ModelNameOverride
+				b.Weight = backendRef.Weight
+				b.Priority = backendRef.Priority
```
kanurag94: I think the fakeClient doesn't do any defaulting. So if we do this here anyway, we can write tests for the default behavior even when this field is omitted.
```suggestion
				b.Weight = ptr.To(ptr.Deref(backendRef.Weight, 1))
				b.Priority = ptr.To(ptr.Deref(backendRef.Priority, 0))
```

internal/controller/gateway.go
kanurag94, line 445: The priority seems to be copied for InferencePools too, but the api doc says they aren't used https://github.com/envoyproxy/ai-gateway/blob/main/api/v1alpha1/ai_gateway_route.go#L390. We should follow the doc or have a counter why this make sense and then update api doc as well.

internal/controller/gateway.go
Aias00: Done. The controller now defaults omitted AIServiceBackend backend refs to weight 1 and priority 0, and the reconcile test covers that path with a backend ref that does not set either field.

internal/controller/gateway.go
Aias00: Done. Priority is no longer copied for InferencePool refs, matching the API doc. I also added a test to cover that case.

kanurag94: Same for weight. It also needs to be moved

internal/controller/gateway.go
```diff
@@ -426,6 +426,7 @@ func (c *GatewayController) reconcileFilterConfigSecret(
 				b := filterapi.Backend{}
 				b.Name = internalapi.PerRouteRuleRefBackendName(aiGatewayRoute.Namespace, backendRef.Name, aiGatewayRoute.Name, ruleIndex, backendRefIndex)
 				b.ModelNameOverride = backendRef.ModelNameOverride
+				b.Weight = ptr.To(ptr.Deref(backendRef.Weight, int32(1)))
```
kanurag94: Same comment as https://github.com/envoyproxy/ai-gateway/pull/2462/changes#r3697734631

Aias00: Done. Weight is no longer copied for InferencePool refs either, and the test now checks that both weight and priority are omitted for InferencePool.

internal/controller/gateway.go
```diff
@@ -446,6 +446,9 @@ func (c *GatewayController) reconcileFilterConfigSecret(
 						continue
 					}
 				} else {
+					b.Weight = ptr.To(ptr.Deref(backendRef.Weight, int32(1)))
+					b.Priority = ptr.To(ptr.Deref(backendRef.Priority, uint32(0)))
```
siddharth1036: Why not a direct assignment? The CRD ensures default values
```suggestion
					b.Weight =backendRef.Weight
					b.Priority = backendRef.Priority
```

siddharth1036:
@Aias00 Could you double-check if weights and priority are actually required for the quota policy change?

I could be mistaken, but since backend selection is handled directly by Envoy, I suspect ai-gateway-extproc might not need the weight and priority information. Let me know what you think!

Aias00:
> @Aias00 Could you double-check if weights and priority are actually required for the quota policy change?
> 
> I could be mistaken, but since backend selection is handled directly by Envoy, I suspect ai-gateway-extproc might not need the weight and priority information. Let me know what you think!

I double-checked the quota path, and I think you're right.

  For the current quota flow, extproc does not need weight or priority. Envoy still selects the backend, and extproc only uses the selected backend name, route, and model to match the quota cost entry.

  I'll remove the weight/priority plumbing from the filter config. If we later move backend selection into extproc, we can add these fields together with that actual consumer.

Aias00:
I double-checked this and removed the weight/priority plumbing from the filter config. extproc does not need these fields for the current quota flow because Envoy still selects the backend, and extproc only uses
  the selected backend name, route, and model to match the quota cost entry.

  After rebasing/merging latest main, this PR no longer has any file changes. I think we can close this PR and add those fields later only if backend selection actually moves into extproc.

PR #2464, opened 2026-07-31
fix(translator): parse CRLF SSE event framing

### LINKED ISSUE #2463: Responses API streaming parser ignores standard SSE framing
## Description

The OpenAI Responses passthrough translator inspects streamed SSE events to extract token usage, the resolved response model, and tracing chunks. Its parser recognizes only LF-delimited event boundaries (`\n\n`) and `data: ` lines with a space after the colon.

SSE permits CRLF line endings (`\r\n\r\n`) and makes that space optional. For a compliant stream using either form, the parser keeps the event buffered or skips its data line, so Gateway does not observe the event.

## Impact

The response bytes are still passed through to the client unchanged. However, Gateway cannot extract response usage or the resolved model from the stream, so usage-based metrics, access logging, and tracing lack that information.

## Reproduction

Send a streamed `/v1/responses` request through the passthrough translator. Have the upstream send `response.created` and `response.completed` events using CRLF separators, or use `data:{...}` without the optional space. The final event's usage is not recorded.

## Expected behavior

Parse both LF and CRLF SSE event boundaries, and accept both `data:` and `data: ` field forms.

0YHR0:
**Description**

Make the OpenAI Responses and Speech translators recognize complete SSE events framed with either LF or CRLF line endings while inspecting streamed responses for observability data.

**Root Cause**

Both parsers buffered events by searching only for `\n\n`. Standards-compliant backends using `\r\n\r\n` therefore left complete events buffered and prevented token usage, response model, or tracing data from being extracted. Parsing of the optional space after `data:` is already handled by the shared field helper on current main and is not changed here.

**Impact**

Client response bytes remain unchanged. The bug only affects Gateway-side inspection of CRLF-framed streams.

**Fix**

Use a shared event-boundary helper that preserves partial input and accepts both LF and CRLF framing in the Responses and Speech parsers.

Fixes #2463

**Testing**

- `go test ./internal/translator -count=1`

### DIFF
--- a/internal/translator/openai_responses.go
+++ b/internal/translator/openai_responses.go
@@ -69,9 +69,7 @@ func (o *openAIToOpenAITranslatorV1Responses) RequestBody(original []byte, req *
 	// Always set the path header to the responses endpoint so that the request is routed correctly.
 	newHeaders = []internalapi.Header{{pathHeaderName, o.path}}
 
-	if forceBodyMutation && len(newBody) == 0 {
-		newBody = original
-	}
+	newBody = forceOriginalBodyIfEmpty(forceBodyMutation, newBody, original)
 
 	if len(newBody) > 0 {
 		newHeaders = append(newHeaders, internalapi.Header{contentLengthHeaderName, strconv.Itoa(len(newBody))})
--- a/internal/translator/openai_responses.go
+++ b/internal/translator/openai_responses.go
@@ -169,12 +169,10 @@ func (o *openAIToOpenAITranslatorV1Responses) extractUsageFromBufferEvent(span t
 		}
 		o.buffered = remaining
 		for line := range bytes.SplitSeq(event, []byte("\n")) {
-			// Per the SSE specification, a space after the colon is optional.
-			data, ok := bytes.CutPrefix(line, []byte("data:"))
+			data, ok := sseData(line)
 			if !ok {
 				continue
 			}
-			data = bytes.TrimSpace(data)
 			if len(data) == 0 || bytes.Equal(data, sseDoneMessage) {
 				continue
 			}
@@ -208,22 +206,6 @@ func (o *openAIToOpenAITranslatorV1Responses) extractUsageFromBufferEvent(span t
 				span.RecordResponseChunk(&eventUnion)
 			}
 		}
-	}
-}
-
-// nextSSEEvent returns the next complete SSE event and preserves partial events in buffer.
-// SSE permits either LF or CRLF line endings.
-func nextSSEEvent(buffer []byte) (event, remaining []byte, ok bool) {
-	lfEnd := bytes.Index(buffer, []byte("\n\n"))
-	crlfEnd := bytes.Index(buffer, []byte("\r\n\r\n"))
-
-	switch {
-	case crlfEnd >= 0 && (lfEnd < 0 || crlfEnd < lfEnd):
-		return buffer[:crlfEnd], buffer[crlfEnd+len("\r\n\r\n"):], true
-	case lfEnd >= 0:
-		return buffer[:lfEnd], buffer[lfEnd+len("\n\n"):], true
-	default:
-		return nil, buffer, false
 	}
 }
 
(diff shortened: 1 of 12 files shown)


kanurag94 (approved):
LGTM
One thing I noticed is that this bug exists in other translators too.
```
- openai_completions.go:157
- openai_openai.go:195
- openai_speech.go:154
- openai_transcription.go:137
- anthropic_anthropic.go:146
- openai_gcpvertexai.go:318
```

They split on a single `\n` and not `\n\n`. A backend sending `data:{}` will break them the same way.
Is it possible to create a followup/ or fix them here (create a shared helper).

0YHR0:
@kanurag94 Thanks, I took the shared-helper approach in 4c81fc2. It now accepts both data field forms across the existing stream parsers, and shares LF/CRLF event-boundary handling for the parsers that buffer full events. I also added helper coverage and ran go test ./internal/translator -count=1. Could you please take another look?

kanurag94 (approved):
Thanks for the great PR

0YHR0:
@kanurag94 I rebased this onto current main and resolved the conflict with #2470. The PR now keeps main's shared data: parsing and focuses on the remaining LF/CRLF event-boundary handling for Responses and Speech. go test ./internal/translator -count=1 passes. Could you please take a final look at the refreshed diff?

0YHR0:
@kanurag94 The rebased code-specific checks are green. The remaining failed jobs are infrastructure failures (upgrade rollout timeout, inference-extension resource timeouts, and a Docker CloudFront connection reset), and GitHub will not let me rerun upstream jobs without repository admin permission. Could you please rerun the failed jobs and take a final look when they settle?

kanurag94 (approved):
LGTM, deferring to @envoyproxy/ai-gateway-maintainers

internal/translator/openai_responses.go
```diff
@@ -69,7 +69,9 @@ func (o *openAIToOpenAITranslatorV1Responses) RequestBody(original []byte, req *
 	// Always set the path header to the responses endpoint so that the request is routed correctly.
 	newHeaders = []internalapi.Header{{pathHeaderName, o.path}}
 
-	newBody = forceOriginalBodyIfEmpty(forceBodyMutation, newBody, original)
+	if forceBodyMutation && len(newBody) == 0 {
+		newBody = original
+	}
```
mathetake: why this change?

internal/translator/openai_speech.go
```diff
@@ -65,7 +65,9 @@ func (o *openAIToOpenAITranslatorV1Speech) RequestBody(original []byte, req *ope
 	// Always set the path header to the speech endpoint so that the request is routed correctly.
 	newHeaders = []internalapi.Header{{pathHeaderName, o.path}}
 
-	newBody = forceOriginalBodyIfEmpty(forceBodyMutation, newBody, original)
+	if forceBodyMutation && len(newBody) == 0 {
+		newBody = original
```
mathetake: same comment

mathetake:
can you not conflate unrelated changes with being untsted?

internal/translator/openai_responses.go
0YHR0: Good catch. This was unrelated to the SSE parsing fix and duplicated forceOriginalBodyIfEmpty. I restored the existing helper in 3f7825db; go test ./internal/translator -count=1 passes.

internal/translator/openai_speech.go
0YHR0: Same issue here. I restored the existing helper in 3f7825db so this PR no longer changes RequestBody behavior; go test ./internal/translator -count=1 passes.

0YHR0:
@mathetake I removed both unrelated RequestBody changes in 3f7825db, replied to and resolved the two review threads, and updated the title/description to reflect the now-focused CRLF SSE framing change. The refreshed CI run is fully green. Could you please take a final maintainer look?

0YHR0:
@mathetake The two unrelated RequestBody changes you flagged remain removed, both threads are resolved, and the focused CRLF event-framing diff has a maintainer approval from @kanurag94 with full CI green. Is there anything else you would like changed before merge, or could you take the final maintainer look?

mathetake (approved):

Issue #2454, opened 2026-07-29
Support OpenAI prompt_cache_breakpoint for explicit prompt caching


xiaolin593:
OpenAI now supports explicit prompt cache breakpoints for GPT-5.6 models and later model families. AI Gateway should support forwarding OpenAI Chat Completions requests that include
  `prompt_cache_breakpoint` on supported content blocks, along with the related request-level fields `prompt_cache_key` and `prompt_cache_options`.

  This enables users to mark stable prompt prefixes for reuse, reducing latency and cost for workloads with large repeated system prompts, files, tools, examples, or other static context.

  Example Chat Completions request shape:

```json
  {
    "model": "gpt-5.6",
    "prompt_cache_key": "tenant:acme:support-assistant-v1",
    "prompt_cache_options": {
      "mode": "explicit"
    },
    "messages": [
      {
        "role": "system",
        "content": [
          {
            "type": "text",
            "text": "You are a support assistant. ... long stable instructions ...",
            "prompt_cache_breakpoint": {
              "mode": "explicit"
            }
          }
        ]
      },
      {
        "role": "user",
        "content": "What should I do next?"
      }
    ]
  }
```

  Expected behavior:

  - AI Gateway accepts and forwards prompt_cache_breakpoint in OpenAI-compatible Chat Completions content blocks.
  - AI Gateway accepts and forwards request-level prompt_cache_options, including mode: "explicit".
  - AI Gateway accepts and forwards prompt_cache_key.
  - The gateway does not drop, reject, or rewrite these fields for OpenAI upstreams.
  - Usage fields returned by OpenAI, including prompt_tokens_details.cached_tokens and prompt_tokens_details.cache_write_tokens, are preserved in the response.
  - Tests cover at least one Chat Completions request with an explicit breakpoint on a text content block.

  This is useful for gateway users serving repeated long-context prompts through OpenAI models where explicit cache placement is needed for predictable prompt caching behavior.

  [optional Relevant Links:]

  OpenAI prompt caching guide:
  https://developers.openai.com/api/docs/guides/prompt-caching?prompt-cache-api=chat-completions

aggieturing-code:
Supporting explicit prompt-cache breakpoints/key is table stakes now that multiple providers accept them. GateLLM unifies prompt caching across providers — it auto-injects cache breakpoints where supported and maps the explicit cache-key/breakpoint fields across protocols, so explicit caching keeps working on a mixed-provider route.

PR #2468, opened 2026-08-01
feat: model OpenAI prompt cache fields for transparent forwarding

Aias00:
**Description**

OpenAI GPT-5.6 and later support explicit prompt cache breakpoints and related request-level fields for marking stable prompt prefixes for reuse. The OpenAI to OpenAI translator already forwards request and response bodies opaquely (raw bytes, only sjson-patching the model), so these fields already reach the upstream and client unchanged. This change promotes them to first-class fields in the OpenAI type model so the gateway recognizes them when parsing request/response bodies (and preserves them through the redaction/debug-log re-marshal path) instead of relying on opaque passthrough alone.

Modeled fields:
- request-level `prompt_cache_key` and `prompt_cache_options` (with `mode`) on `ChatCompletionRequest`
- `prompt_cache_breakpoint` (with `mode`) on `ChatCompletionContentPartTextParam`; the content union marshaler delegates to the concrete struct marshal, so the field round-trips
- `prompt_tokens_details.cache_write_tokens` on `PromptTokensDetails` (`cached_tokens` was already modeled)

The response body remains opaque passthrough, so `cache_write_tokens` reaches the client unchanged regardless; modeling it makes it available when the response is parsed. New metrics for `cache_write_tokens` are intentionally out of scope here (cache-write vs cache-creation semantics warrant a separate maintainer decision); the field is modeled but not yet emitted as a metric.

Tests cover a Chat Completions request with an explicit breakpoint on a text content block (field capture plus re-marshal preservation), byte-for-byte opaque request forwarding, and `cache_write_tokens` preservation/parsing in the response.

Refs #2454

- No behavior change on the data path; opaque passthrough already forwarded these fields. The change is additive type modeling plus regression tests.
- Local golangci-lint reports two pre-existing `SA5008 json:"type,"` warnings on `ThinkingDisabled`/`ThinkingAdaptive` (present on `main`, not introduced here).
- Metrics wiring for `cache_write_tokens` is deliberately deferred (see Description).


PatilHrushikesh:
redact these keys in RedactSensitiveInfoFromRequest  logs

Aias00:
> redact these keys in RedactSensitiveInfoFromRequest logs

Done.

gavrissh:
As this is supported in standard OpenAI Chat Completions, I believe passing of the Prompt Cache will working without this change in place. We dont drop any field passed from the client unless there is a translation to different schema. Hence this change seems to be redundant

Aias00:
> As this is supported in standard OpenAI Chat Completions, I believe passing of the Prompt Cache will working without this change in place. We dont drop any field passed from the client unless there is a translation to different schema. Hence this change seems to be redundant

Understood. Looking at this again, the practical value of this PR is pretty limited since the normal OpenAI-to-OpenAI path already passes these fields through. I'll close this PR for now.

PR #2471, opened 2026-08-02
feat: add gen_ai.backend attribute to GenAI metrics

CodePrometheus:
**Description**

`gen_ai.provider.name` is derived from the backend API schema, so backends that share a schema are indistinguishable in metrics: a route fronting several OpenAI-compatible backends reports every request under the same provider value. 
Since backendRefs exist to route across multiple backends, this means the metrics cannot attribute error rate, latency, or token usage to the backend that served the request.

This change adds a `gen_ai.backend` attribute to the GenAI metrics, set to the `namespace/name` of the AIServiceBackend that served the request. The attribute is not part of the GenAI semantic conventions. unknown is reported when no backend has been selected, and cardinality is bounded by the number of configured AIServiceBackends.


CodePrometheus:
Gentle ping @kanurag94, would you mind taking a look at this when you have time?

kanurag94:
Thanks!
This is on similar lines as https://github.com/envoyproxy/ai-gateway/pull/2485, we should be good to add the metric

internal/metrics/metrics_impl_test.go
```diff
@@ -697,3 +712,38 @@ func TestSetBackendProviderName(t *testing.T) {
...
+
+func TestAIServiceBackendAttribute(t *testing.T) {
+	// Two backends sharing an API schema must stay distinguishable, which is what
+	// gen_ai.provider.name alone cannot express.
+	for _, backendName := range []string{
+		internalapi.PerRouteRuleRefBackendName("default", "primary", "some-route", 0, 0),
+		internalapi.PerRouteRuleRefBackendName("default", "fallback", "some-route", 0, 1),
+	} {
+		mr := metric.NewManualReader()
+		meter := metric.NewMeterProvider(metric.WithReader(mr)).Meter("test")
```
kanurag94: shouldn't the reader be out of for loop to really test this?

internal/metrics/metrics_impl.go
```diff
@@ -16,6 +16,11 @@ import (
 	"github.com/envoyproxy/ai-gateway/internal/internalapi"
 )
 
+// aiAttributeBackend is the AIServiceBackend that served the request, as "namespace/name".
+// It is not part of the GenAI semantic conventions: gen_ai.provider.name only carries the
+// API schema, so backends sharing one are otherwise indistinguishable.
+const aiAttributeBackend = "ai.backend"
```
kanurag94: What about moving it to the same place as other constants. There are already other non convention semantic ones there

kanurag94 (commented):
Overall looks good.
Few things:
1. The `ai.backend` usage is against the `gen_ai` namespace usage this project has.
2. If this is a metric, why only keep `namespace/name` -- I don't think we are leaking anything so we can be granular.
3. Can you please review #2485 ?

CodePrometheus:
> Overall looks good. Few things:
> 
> 1. The `ai.backend` usage is against the `gen_ai` namespace usage this project has.
> 2. If this is a metric, why only keep `namespace/name` -- I don't think we are leaking anything so we can be granular.
> 3. Can you please review #2485 ?

Thanks! On `#2`: my concern with using the full internal backend name is metric-label stability. The `/rule/<ruleIndex>/ref/<refIndex>` components are positional: reordering rules or backendRefs, or inserting a new backendRef before an existing one, changes the label value even when the reference still targets the same `AIServiceBackend`. Prometheus treats a changed label value as a different time series, so subsequent samples go to a new series, and exact-match dashboards or alerts using the previous value no longer match the new samples. Using `namespace/name` keeps the label stable across these ordering-only edits and avoids creating a separate backend label value for every `route/rule/ref` occurrence.

kanurag94 (commented):
LGTM, thanks for adding this!
Can you please rebase?

CodePrometheus:
Thanks, Conflicts are resolved.

CodePrometheus:
Gentle ping~ @nacx @mathetake mind taking a look when you have time?

PR #2472, opened 2026-08-02
fix: carry Anthropic tool input schemas as-is

CodePrometheus:
**Description**

An Anthropic tool declares its input as a JSON schema, but the gateway modeled that schema as a struct with only "type", "properties" and "required". 
Everything else a client sends is silently dropped on the way to the backend: "additionalProperties", which backends need to enable strict schema adherence, the per-property "description" the model relies on to call the tool correctly, "$defs", and so on.
We hit this with a tool that takes no arguments: its empty "properties" was dropped too, and the OpenAI-compatible backend rejected the whole request with 400 invalid_function_parameters, "object schema missing properties".

The schema is now carried as raw JSON, so tools reach the backend as the client declared them. This covers both the Anthropic-to-OpenAI and the Anthropic-to-Bedrock conversions, which shared the struct.


johnugeorge:
Can you rebase?

johnugeorge (approved):

PR #2478, opened 2026-08-03
fix(translator): translate anthropic cache_control on the bedrock path

### LINKED ISSUE #2477: anthropic to bedrock translator drops cache_control on every request path
_Description_:

The Anthropic to Bedrock translator (`internal/translator/anthropic_awsbedrock.go`) silently drops
`cache_control` on every request path, so a native Anthropic Messages API client routed to a
Bedrock backend gets no prompt caching at all. There is no `CachePoint` anywhere in the file: the
request is translated into a `Converse` payload with no `cachePoint` block in `system[]`,
`messages[].content[]` or `toolConfig.tools[]`, whatever the client asked for.

Concretely, `cache_control` is dropped by:

- `convertSystemPrompt` — `SystemPrompt.Texts[i].CacheControl`
- `convertUserMessage` — `TextBlockParam.CacheControl`, `ImageBlockParam.CacheControl`
- `convertAssistantMessage` — `TextBlockParam.CacheControl`, `ToolUseBlockParam.CacheControl`
- `convertToolResultMessage` and `convertToolResultBlock` — `ToolResultBlockParam.CacheControl`
- `convertTools` — `Tool.CacheControl`

The failure is silent rather than an error: requests succeed, they are just billed and served
uncached. It is easy to miss because the response path is already wired up for caching —
`ResponseBody` maps `CacheReadInputTokens` and `CacheWriteInputTokens` back onto
`usage.cache_read_input_tokens` and `usage.cache_creation_input_tokens` — so a client sees a
well-formed response whose cache counters are simply always zero.

The equivalent OpenAI to Bedrock path does support this: `getCachePoint` plus a standalone cache
point block appended to the relevant array. The Anthropic path just never grew the same handling.

For anyone comparing the two paths: #2475 is a separate defect, in the OpenAI path, where the tool
cache point is emitted on the same `tools[]` element as `toolSpec`. This issue is about the
Anthropic path not emitting cache points at all.

_Repro steps_:

Send an Anthropic Messages API request to a Bedrock backend with a cached system prompt:

```json
{
  "model": "us.anthropic.claude-sonnet-4-5-20250929-v1:0",
  "max_tokens": 1024,
  "system": [
    {"type": "text", "text": "<long shared prefix>", "cache_control": {"type": "ephemeral"}}
  ],
  "messages": [{"role": "user", "content": "hello"}]
}
```

Expected: the translated `Converse` payload carries a `cachePoint` block after the system text
block, and a repeat request reports a non-zero `usage.cache_read_input_tokens`.

Actual: no `cachePoint` is emitted, and `cache_read_input_tokens` and
`cache_creation_input_tokens` are always zero. The same holds for `cache_control` on message
content blocks, tool results and tool definitions.

_Environment_:

Present on `main` and in v1.0.0.

_Fix_:

Translate `cache_control` into a standalone cache point element on each of those paths — a
`SystemContentBlock{CachePoint: ...}` after the system text block, a `ContentBlock{CachePoint: ...}`
after the message content block, and a `Tool{CachePoint: ...}` after the tool spec — mirroring how
the OpenAI path does it. Note that `Tool.ToolSpec` needs `omitempty` for the tools case, which
#2475 also covers.

I have a fix ready and will open a PR against this issue.

pmerrison:
**Description**

The Anthropic to Bedrock translator dropped `cache_control` on every request path. There was no
`CachePoint` anywhere in `anthropic_awsbedrock.go`, so a native Anthropic Messages API client
routed to a Bedrock backend got no prompt caching at all, whatever it asked for. The failure is
silent rather than an error: requests succeed, they are just billed and served uncached. It is easy
to miss because the response path is already wired up for caching and maps
`CacheReadInputTokens` and `CacheWriteInputTokens` back onto `usage`, so the client sees a
well-formed response whose cache counters happen to always be zero.

This emits a standalone Bedrock cache point wherever `cache_control` asks for one:

- `convertSystemPrompt` — after the system text block
- `convertUserMessage` — after a text, image or tool_result block
- `convertAssistantMessage` — after a text or tool_use block
- `convertToolResultMessage` — after a tool_result block
- `convertTools` — after the tool spec
- `promoteAnthropicSystemMessagesToParam` — system messages promoted from the messages array
  (the Claude Code mid-conversation-system beta pattern) now keep their text blocks, and with
  them `cache_control`, instead of being flattened into one joined string

Bedrock's `ContentBlock`, `SystemContentBlock` and `Tool` are all unions, so the cache point has to
be its own element rather than riding on the block it follows. Anthropic's ephemeral TTL has no
Bedrock equivalent and is dropped.

One behavioural side effect of the promotion change: multiple promoted system messages now become
separate Bedrock system blocks rather than a single newline-joined block, which is semantically
equivalent and more faithful to the original request.

Tests: `TestAnthropicToAWSBedrockTranslator_RequestBody_CacheControl` covers the first five paths
and asserts the cache-point-only `tools[]` element marshals without a `toolSpec` key, with a
companion case asserting no `cachePoint` is emitted when the request does not ask for one, and
`TestAnthropicToAWSBedrockTranslator_RequestBody_CacheControlOnPromotedSystemMessage` covers the
promoted-system-message path end to end. All fail on `main` and pass with this change.

Fixes #2477

Related PR: #2476, which fixes a separate defect in the OpenAI to Bedrock path where the tool cache
point is emitted on the same `tools[]` element as `toolSpec`. The one-line `omitempty` on
`Tool.ToolSpec` is needed by both and is identical in each, so whichever lands second should merge
cleanly; happy to rebase if you would rather it lived in only one of them.

Per the generative AI policy: an AI assistant was used while preparing this change. I have
reviewed it, understand it, and take full ownership of it.


nacx (approved):

kanurag94 (approved):

joseluisjimenez1:
Any news on this ? this is hitting hard the happiness of the engineers 😄 (budget)

johnugeorge:
@pmerrison 
Any reason for skipping cache TTL translation?

johnugeorge:
Can you also rebase?

codemug:
This is skyrocketing our token costs, can we please prioritize this?

johnugeorge:
@codemug @pmerrison @joseluisjimenez1 Can you resolve the conflicts and rebase this PR?  
Also, TTL config is not configured. Otherwise, this looks good.

johnpanos:
Does this affect AWSAnthropic? Or just AWSBedrock

codemug:
@johnpanos I think cache_control will passthrough: [ref](https://github.com/envoyproxy/ai-gateway/blob/main/internal/translator/anthropic_awsanthropic.go#L148)

Waiting for @pmerrison to rebase

PR #2480, opened 2026-08-04
fix: keep AIGatewayRoute backendRef indices aligned with the generated HTTPRoute

kanurag94:
**Description**

Fixes an index-out-of-range panic in the extension server, and the mislabelled endpoints the panic was masking.

An unresolvable reference now keeps its slot. `newHTTPRoute` emits a placeholder backendRef at the same index, with weight 0, naming a Backend in the route's own namespace that cannot exist. The rest of the route is programmed normally and the reference is reported through a new `ResolvedRefs` condition.

**Example**

```yaml
rules:
  - backendRefs:
      - name: openai
        priority: 0
      - name: typo-backend   # does not exist
        priority: 1
      - name: bedrock
        priority: 2
```

Before, the whole route is withheld. No HTTPRoute exists, so `openai` and `bedrock` do not serve either:

```
===RECONCILE_ERR=== failed to construct a new HTTPRoute: failed to get AIServiceBackend typo-backend.default: AIServiceBackend typo-backend.default not found
===HTTPROUTE=== httproutes.gateway.networking.k8s.io "myroute" not found
===COND=== type=NotAccepted status=False reason=ReconciliationFailed message=failed to construct a new HTTPRoute: failed to get AIServiceBackend typo-backend.default: AIServiceBackend typo-backend.default not found
```

After, the route is programmed and the unresolvable reference carries no traffic:

```
===RECONCILE=== err=<nil> requeueAfter=0s
===BACKENDREFS===
- name: openai-backend
  kind: Backend
  name: ai-eg-unresolved-5773aa97-r0-b1
  namespace: default
  weight: 0
- name: bedrock-backend

===CONDITION=== type=Accepted status=True reason=ReconciliationSucceeded
  message=AI Gateway Route reconciled successfully
===CONDITION=== type=ResolvedRefs status=False reason=BackendNotFound
  message=1 backend reference(s) could not be resolved and will not receive traffic; the remaining backends in the rule are serving: rules[0].backendRefs[1] (default/typo-backend): BackendNotFound: AIServiceBackend typo-backend.default not found
```

`bedrock` stays at index 2, and priority failover between `openai` and `bedrock` still works.

None.

None.

**Verification**

UTs and e2e


internal/extensionserver/post_translate_modify.go
```diff
@@ -372,6 +372,25 @@ func (s *Server) maybeModifyCluster(ctx context.Context, cluster *clusterv3.Clus
...
+						"endpoint_index", lbEndpointIndex,
+						"endpoints_len", len(cluster.LoadAssignment.Endpoints),
+						"backend_refs_len", len(httpRouteRule.BackendRefs),
+						"backend_index", i,
+						"backend_name", backendRef.Name,
+						"route_name", aigwRoute.Name,
+						"route_namespace", aigwRoute.Namespace,
+						"route_rule_index", httpRouteRuleIndex,
+						"guidance", "check AIGatewayRoute status for backend validation errors")
+					continue
```
johnugeorge: This avoid panic. But, positional correspondence between backend refs and endpoints will be misaligned right ? (depending on the skipped backends)

kanurag94: Thank you for reviewing closely. You are right. I have thus changed this PR to add a placeholder backend that will be skipped at all places.

PR #2482, opened 2026-08-04
feat: implement `/cohere/v2/embed` endpoint

cjackal:
**Description**

Current support of Cohere provider is limited to rerank endpoint `/cohere/v2/rerank`. This PR complements it by adding the Cohere native Embed V2 endpoint `/cohere/v2/embed` [1].

No relevant PRs submitted so far.

This PR closely follows the OpenAI `/v1/embeddings` endpoints, but as the API spec and feature range of the two providers are slightly different, some features (mostly the nonstandard output dtypes including ubinary) are left as todo.

1: https://docs.cohere.com/reference/embed

### DIFF
--- a/cmd/extproc/mainlib/main.go
+++ b/cmd/extproc/mainlib/main.go
@@ -323,6 +323,8 @@ func Main(ctx context.Context, args []string, stderr io.Writer) (err error) {
 		imageGenerationMetricsFactory, tracing.ImageGenerationTracer(), endpointspec.ImageGenerationEndpointSpec{}))
 	server.Register(path.Join(flags.rootPrefix, endpointPrefixes.Cohere, "/v2/rerank"), extproc.NewFactory(
 		rerankMetricsFactory, tracing.RerankTracer(), endpointspec.RerankEndpointSpec{}))
+	server.Register(path.Join(flags.rootPrefix, endpointPrefixes.Cohere, "/v2/embed"), extproc.NewFactory(
+		embeddingsMetricsFactory, tracing.EmbedTracer(), endpointspec.EmbedEndpointSpec{}))
 	server.Register(path.Join(flags.rootPrefix, endpointPrefixes.OpenAI, "/v1/models"), extproc.NewModelsProcessor)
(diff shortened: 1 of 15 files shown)


gavrissh:
I'm thinking does this need to be an ingress API support or translation from the existing OpenAI Embed API as that would be the API most clients would support

cjackal:
> I'm thinking does this need to be an ingress API support or translation from the existing OpenAI Embed API as that would be the API most clients would support

I think an ingress is more urgent, some of our engineers are using vLLM's cohere /v2/embed API in production and they have a good reason for their choice compared to openai /v1/embeddings API, one is that cohere API supports wider serialization format (openai supports only base64 and float). I also reckon cohere API as a modern extension of openai's, openai hasn't updated their embedding model since 2024.

daily-kim:
> > I'm thinking does this need to be an ingress API support or translation from the existing OpenAI Embed API as that would be the API most clients would support
> 
> I think an ingress is more urgent, some of our engineers are using vLLM's cohere /v2/embed API in production and they have a good reason for their choice compared to openai /v1/embeddings API, one is that cohere API supports wider serialization format (openai supports only base64 and float). I also reckon cohere API as a modern extension of openai's, openai hasn't updated their embedding model since 2024.

I agree with prioritizing support for Cohere’s /v2/embed API.

From the perspective of building RAG services and operating embedding APIs on-premises, OpenAI’s /v1/embeddings is a useful common denominator, but it does not represent many capabilities of modern embedding models.

Models such as [Qwen3-VL-Embedding](https://huggingface.co/Qwen/Qwen3-VL-Embedding-8B), Jina, Cohere, and Gemini-embedding increasingly support [Matryoshka Representation Learning](https://arxiv.org/abs/2205.13147), quantized outputs, multimodal inputs, and explicit input types. These cannot be faithfully recovered by translating an OpenAI request because its schema cannot express them in the first place.

[Cohere ](https://docs.cohere.com/reference/embed)[/v2/embed](https://docs.cohere.com/reference/embed) is already an [officially supported vLLM format](https://docs.vllm.ai/en/v0.27.1/api/vllm/entrypoints/pooling/embed/protocol/) and provides a practical, model-agnostic interface for these capabilities. Native ingress support would preserve them while allowing existing Cohere-compatible clients to work without custom extensions.

PR #2485, opened 2026-08-05
feat: emit backend name in dynamic metadata from the request phase

kanurag94:
**Description**

The backend that serves a request is written to the `io.envoy.ai_gateway:backend_name` dynamic metadata key, but only at end of stream and only when request costs are configured, so nothing can rely on it as a general source of backend identity.

This emits it from the request headers phase instead, where `SetBackend` has already resolved it.

Writing it during decode rather than on the response is what makes it usable. Envoy applies route-level response header mutations in the router filter, which runs before ext_proc when encoding, so a value written on the response path arrives too late for them. Written on the request path it is there in time, and filters placed ahead of ext_proc in the chain can read it too.

**Example**

A route that wants to name the serving backend to the client can do it in config, so nothing is exposed by default:

```yaml
response_headers_to_add:
  - header:
      key: x-ai-eg-upstream-backend-name
      value: "%DYNAMIC_METADATA(io.envoy.ai_gateway:backend_name)%"
```

Two edges worth knowing:

- The namespace in the name is the AIGatewayRoute's, not the AIServiceBackend's, so for a cross-namespace backendRef it is not where the backend itself lives.
- The short `namespace/name` projection (`ai_service_backend_name`) is left where it is, since its only consumer is the rate limit descriptor path.

**Testing**

UTs.

None.

None.


nacx (commented):
Thanks!

I'm not sure about hte security implications of this, and about the convenience of disclosing the serving backend to users by default.
Can we make this an opt-in feature, so that we keep the current behaviour? I can see many cases where admins don't want infor about the backends leaking to users.

kanurag94:
> Thanks!
> 
> I'm not sure about hte security implications of this, and about the convenience of disclosing the serving backend to users by default. Can we make this an opt-in feature, so that we keep the current behaviour? I can see many cases where admins don't want infor about the backends leaking to users.

Thanks for the review!
I agree, I have moved it to be emitted to metadata. #2471 will add it to request metric. That should be good

CodePrometheus (approved):
Solid work!

johnugeorge:
LGTM

siddharth1036 (approved):
LGTM!

johnugeorge (approved):

PR #2467, opened 2026-08-01
fix: error on unsupported Anthropic tool types instead of silently dropping them

hustxiayang:
**Description**
 When translating OpenAI tools to Anthropic, non-function tools (and function tools missing their definition) were silently skipped, so the model never received them and requests failed in confusing, hard-to-debug ways. This now returns a 400 (ErrInvalidRequestBody) with a clear message, matching the existing fail-closed behavior of the Gemini translator

PR #2486, opened 2026-08-05
feat: preserve unknown Responses tool types instead of rejecting the request

kanurag94:
**Description**

Makes `/v1/responses` tolerate tool types this schema does not enumerate, instead of rejecting the whole request.

`ResponseToolUnion.UnmarshalJSON` ends in:

```go
default:
	return fmt.Errorf("unknown tool type %s", typ.String())
```

so a request carrying any tool type not in the switch fails to parse and the client gets an error, even though nothing about that tool is the gateway's business.

This adds `OfUnknown json.RawMessage` to the union: unrecognized tools are captured verbatim on unmarshal and emitted verbatim on marshal.

**Example**

```shell
curl $GATEWAY_URL/v1/responses -d '{
  "model": "gpt-5",
  "input": "hi",
  "tools": [{"type": "some_future_tool", "execution": "client"}]
}'
```

Before, the request never leaves the gateway:

```
unknown tool type some_future_tool
```

After, it parses and is forwarded unchanged, and upstream decides whether the tool is valid.

#2434 ("fix tool-search in response api", open) is adjacent but not a substitute: it adds
`tool_search_call`/`tool_search_output` to the response *output* union, whereas this tolerates
unrecognised types in the request *tool* union. They do not overlap, and both are needed.

- **This is a deliberate behavior change and there was a test pinning the old behavior.** `TestResponseToolUnion_UnmarshalJSON_UnknownType` now asserts the tool round-trips.


nacx (approved):
LGTM. Tagging @johnugeorge and @aabchoo for a final review

johnugeorge:
Why do we need this change?   We have different behaviour from chat/completions. 
Ref: https://github.com/envoyproxy/ai-gateway/pull/2467

kanurag94:
> Why do we need this change? We have different behaviour from chat/completions. Ref: #2467

2467 is a translation concern: `translateOpenAItoAnthropicTools` has to map every tool into Anthropic's schema, and what it doesn't understand can't be mapped — so dropping it silently means the model never sees it. Failing with a 400 is right there.

chat/completions is already lenient at the parse layer. It only fails when a translator can't represent it, which is what #2467 fixes. Responses is the outlier — the only endpoint that rejects at parse time, for tools it was going to forward unchanged anyway.

After this change:
- chat/completions -> OpenAI: unknown tool forwarded (unchanged)
- chat/completions -> Anthropic: 400 at translation (#2467, unchanged)
- responses -> OpenAI/Azure: unknown tool forwarded (this PR)

The motivation is that every tool type OpenAI ships gives 400 until this schema catches up.  If a Responses -> non-OpenAI translator is added, it should fail closed exactly like #2467 does.

PR #2488, opened 2026-08-05
fix: preserve system blocks and cache_control when promoting system messages

kanurag94:
**Description**

`promoteAnthropicSystemMessagesToParam` in the Anthropic→Bedrock translator flattens `role: "system"` messages into one joined string and assigns it over `body.System`:

```go
if len(systemTexts) > 0 {
	systemText := strings.Join(systemTexts, "\n")
	body.System = &anthropicschema.SystemPrompt{Text: systemText}
}
```

Two problems.

**A top-level `system` is silently discarded.** The assignment is unconditional, so a request carrying both a top-level `system` and a `role: "system"` message loses the former entirely:

```
input : System.Texts = ["TOP LEVEL SYSTEM"], messages[0] = {role: system, "MID CONVERSATION SYSTEM"}
output: {"system":[{"text":"MID CONVERSATION SYSTEM"}]}     <- TOP LEVEL SYSTEM gone
```

The system prompt the client sent never reaches Bedrock. That changes model behaviour, not just billing. Per its own comment, this path exists for clients that send system prompts as messages rather than as the top-level parameter.

**`cache_control` is dropped.** Only `.Text.Text` is collected, so any `cache_control` on those blocks is lost, and a joined string could not carry per-block cache control anyway.

**Fix**

Promoted messages become `SystemPrompt.Texts` entries carrying their `CacheControl`, appended after any existing top-level `system` blocks instead of replacing them.

The function no longer mutates the request body (to avoid duplication on retries). It returns the effective system prompt alongside the filtered messages, and the caller hands that to `convertSystemPrompt`:

```go
messages, systemPrompt := promoteAnthropicSystemMessagesToParam(body)
```

- N promoted system messages used to become a single Bedrock block joined with `\n`; they are now N blocks. A single promoted message is byte-identical to before.
- One of those subtests was already named `system message is added to existing system param` while asserting the opposite, with a comment saying `// The existing system param is overwritten by the promoted messages.` It now does what its name says.

The `cache_control` becomes observable once #2478 lands, which adds the Bedrock cache-point emission in `convertSystemPrompt`.


nacx (approved):
LGTM. @aabchoo @johnugeorge can you take a look?

johnugeorge:
LGTM

PR #2494, opened 2026-08-06
feat: resolve backend credentials from the informer cache

kanurag94:
**Description**

`getSecretData` used to make a call to the api server directly. This PR changes it to use informer cache

None.

None.


internal/controller/gateway.go
```diff
@@ -751,7 +754,38 @@ func resolveCredentialOverride(bspType aigv1b1.BackendSecurityPolicyType, overri
 	return result, nil
 }
 
-func (c *GatewayController) bspToFilterAPIBackendAuth(ctx context.Context, backendSecurityPolicy *aigv1b1.BackendSecurityPolicy) (*filterapi.BackendAuth, error) {
+// secretCache memoizes Secret reads within a single reconcile pass: getSecretData reads through the
+// clientset rather than the informer cache, so every call is a round trip, and backends on a Gateway
```
nacx: Why can't we use the informer cache here instead of using this memoization?

kanurag94: I will move to informer cache. I thought from exisitng code that we were not using informer cache for some security reason but now I see that we do that at other places.

kanurag94: Updated

johnugeorge:
LGTM

aabchoo (approved):

PR #2504, opened 2026-08-10
feat(controller): add version subcommand

### LINKED ISSUE #2322: Add a `version` command to probe the ai controller workloads version
_Description_:

The envoy controller itself provides a command to fetch the version of the deployed controller workload - `envoy-gateway version`

But there is no equivalent for the `ai-gateway` workload. We are trying to automate the deployment and management of the the ai-gateway as a [juju](https://canonical.com/juju) charm and having a similar `version` command or a reliable way to probe the version from the ai gateway controller application would help us sync the versions of the ai-gateway and the the extproc images more reliably.

Relevant downstream issue: https://github.com/canonical/service-mesh/issues/126

akdok:
**Description**

Adds a `version` subcommand to the AI Gateway controller binary, mirroring `envoy-gateway version`:

```console
$ ai-gateway-controller version
Envoy AI Gateway Controller: v1.0.0
```

This gives operators (e.g. the Juju charm automation in the linked issue) a reliable way to probe the deployed controller workload version. The controller also now logs its version at startup, matching what the external processor already does.

The issue discussion also floated a `/version` HTTP endpoint as a possibly more idiomatic alternative. This PR implements the command for parity with `envoy-gateway` since it is small and unblocks version probing now; happy to follow up with an endpoint (or rework this PR) if maintainers prefer that direction.

Fixes #2322

**Notes**
- `make precommit test` passes locally.
- Per the contributing guidelines' transparency policy: this change was written with AI assistance; I understand and take full ownership of the code.

### DIFF
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -10,6 +10,7 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	"io"
 	"log/slog"
 	"net"
 	"os"
@@ -35,6 +36,7 @@ import (
 	"github.com/envoyproxy/ai-gateway/internal/internalapi"
 	"github.com/envoyproxy/ai-gateway/internal/pprof"
 	"github.com/envoyproxy/ai-gateway/internal/ratelimit/runner"
+	"github.com/envoyproxy/ai-gateway/internal/version"
 )
 
 type flags struct {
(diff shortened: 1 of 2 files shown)


cmd/controller/main.go
```diff
@@ -408,7 +410,21 @@ func newZapOpts(logFormat string, level zapcore.LevelEnabler) []zap.Opts {
...
+func handleVersion(args []string, stdout io.Writer) bool {
+	if len(args) == 0 || args[0] != "version" {
+		return false
+	}
+	_, _ = fmt.Fprintf(stdout, "Envoy AI Gateway Controller: %s\n", version.Parse())
+	return true
+}
+
 func main() {
+	if handleVersion(os.Args[1:], os.Stdout) {
```
mathetake, line 424: this panics when people execute without args

mathetake (changes requested):

cmd/controller/main.go
akdok: Thanks for the review! I tried to reproduce this but couldn't get a panic — could you share the invocation that panicked for you?

What I checked:

- `os.Args[1:]` is a slice expression, not an index, so it's safe when the binary is run with no arguments: it yields an empty slice, and `handleVersion` returns `false` on the `len(args) == 0` check. This path is covered by the `no args` subtest in `Test_handleVersion`.
- Running the built binary with no arguments behaves the same as `main` does today: it falls through to `ctrl.GetConfigOrDie()`, which (outside a cluster) logs the usual stack traces and exits 1. The stack traces look panic-like but it's the pre-existing `GetConfigOrDie` behavior, unchanged by this PR — I built `origin/main` and got byte-identical output.

Happy to fix whatever you hit if it's a different invocation.

mathetake: oh right

PR #2505, opened 2026-08-10
fix: tolerate non-string OpenAI error fields when translating backend errors

### LINKED ISSUE #1911: Tool calling behavior discrepancy between vLLM and Envoy AI Gateway for Anthropic to OpenAI translator
_Description_:

Follow-up to https://github.com/envoyproxy/ai-gateway/pull/1878#issuecomment-3963811887.
CC @changminbark @mathetake @groundsada

The issue: **For Claude Code, one thing that is happening is that on vLLM, web search fails and it moves to different tools (fetch). Via Envoy, it remains stuck in a loop of failing.**

_Repro steps_:

Use Claude Code on vLLM and Envoy AI Gateway, and compare.

E.g., MiniMax-M2.5, only reproduced in the PR and falls back to `fetch` when directly through vLLM.

```
⏺ Web Search("Super Bowl LX winner")
  ⎿  Did 0 searches in 2s                                                         
                                                          
⏺ The search isn't returning any results. This could be because:                  
  1. The web search tool isn't functioning properly                                           
  2. The information isn't available online               
                                                                                              
  Given that the system date is February 25, 2026, and Super Bowl 60 (LX) is typically played 
  in early February, it would have just happened recently. However, I don't have reliable     
  search results to provide an answer.                                                        
                  
  Let me be honest with the user about this limitation.


  I'm having trouble retrieving the search results. Based on today's date (February 25, 2026),
   the most recent Super Bowl would be Super Bowl LX (60), which is typically played in early
  February. However, I'm not getting search results to confirm the winner.

  Based on my knowledge, Super Bowl 60 (played February 2026) would have just concluded
  recently. Without working search results, I cannot provide definitive information on who
  won.

  Sources: None available due to search failure

✻ Cooked for 5m 1s
```

_Environment_:

Version: Nightly.

_Logs_:

Conversation in https://github.com/envoyproxy/ai-gateway/pull/1878:

@changminbark : It seems that web search tool use in Claude Code has a different ContentBlock being passed [here](https://platform.claude.com/docs/en/build-with-claude/streaming#streaming-request-with-web-search-tool-use). Have you tried using any other tool use?

Note the type being server_tool_use instead of tool_use and the last content block start having a type web_search_tool_result and compare it to the output with non-web search tool usage [here](https://platform.claude.com/docs/en/build-with-claude/streaming#streaming-request-with-tool-use).

```
event: content_block_start
data: {"type":"content_block_start","index":1,"content_block":{"type":"server_tool_use","id":"srvtoolu_014hJH82Qum7Td6UV8gDXThB","name":"web_search","input":{}}}

event: content_block_delta
data: {"type":"content_block_delta","index":1,"delta":{"type":"input_json_delta","partial_json":""}}

event: content_block_delta
data: {"type":"content_block_delta","index":1,"delta":{"type":"input_json_delta","partial_json":"{\"query"}}

event: content_block_delta
data: {"type":"content_block_delta","index":1,"delta":{"type":"input_json_delta","partial_json":"\":"}}

event: content_block_delta
data: {"type":"content_block_delta","index":1,"delta":{"type":"input_json_delta","partial_json":" \"weather"}}

event: content_block_delta
data: {"type":"content_block_delta","index":1,"delta":{"type":"input_json_delta","partial_json":" NY"}}

event: content_block_delta
data: {"type":"content_block_delta","index":1,"delta":{"type":"input_json_delta","partial_json":"C to"}}

event: content_block_delta
data: {"type":"content_block_delta","index":1,"delta":{"type":"input_json_delta","partial_json":"day\"}"}}

event: content_block_stop
data: {"type":"content_block_stop","index":1 }

event: content_block_start
data: {"type":"content_block_start","index":2,"content_block":{"type":"web_search_tool_result","tool_use_id":"srvtoolu_014hJH82Qum7Td6UV8gDXThB","content":[{"type":"web_search_result","title":"Weather in New York City in May 2025 (New York) - detailed Weather Forecast for a month","url":"https://world-weather.info/forecast/usa/new_york/may-2025/","encrypted_content":"Ev0DCioIAxgCIiQ3NmU4ZmI4OC1k...","page_age":null},...]}}
May I ask what backend you used if you did not use vLLM? If you used vLLM, the reason why it might be failing is because vLLM does not support internet connection (and therefore web searching). That is why I cannot find a direct mapping of the web search tool use in vLLM's repo [here](https://github.com/vllm-project/vllm/blob/ee59a7c61574485cf4ddbc6037ba557941be5c56/vllm/entrypoints/anthropic/protocol.py).
```

I could add support for web search tool capability, but this would mean that we would only be able to use this for backend/upstreams that support this like the official OpenAI API (not vLLM). What are your thoughts?

May I see how it looks like when you use vLLM directly to conduct a web search tool call? I just want to see how the output looks like if you directly use vLLM and then compare it to when you use it through Envoy.

Do you mind running this curl command on vLLM (adjust model name and port accordingly)?
```
cat > /tmp/request.json << 'EOF'
{
  "model": "Qwen/Qwen3-0.6B",
  "messages": [{"role": "user", "content": “what is the weather like in New York?“}],
  "tools": [{"type": "function", "function": {"name": "search", "description": "search the web", "parameters": {"type": "object", "properties": {"query": {"type": "string"}}, "required": ["query"]}}}],
  "tool_choice": "auto"
}
EOF

curl http://localhost:8000/v1/chat/completions \
  -d @/tmp/request.json | jq
```
Then can you run this curl command through Envoy (adjust model name and port accordingly)?
```
cat > /tmp/request.json << 'EOF'
{
  "model": "Qwen/Qwen3-0.6B",
  "messages": [{"role": "user", "content": "what is the weather like in New York?"}],
  "tools": [
    {
      "name": "search",
      "description": "search the web",
      "input_schema": {
        "type": "object",
        "properties": {
          "query": {"type": "string"}
        },
        "required": ["query"]
      }
    }
  ]
}
EOF

curl http://localhost:8000/v1/messages \
  -d @/tmp/request.json | jq
```
This will help me determine whether tool call structure translation is broken or if it's only present in the streaming response.

I tried doing this on my local machine.

vLLM
Directly calling /v1/messages using vLLM
```
$ curl http://localhost:8000/v1/messages   -H "Content-Type: application/json"   -d '{"model":"Qwen/Qwen3-0.6B","max_tokens":1024,"messages":[{"role":"user","content":"search the web for the year america was founded"}],"tools":[{"name":"search","description":"search the web","input_schema":{"type":"object","properties":{"query":{"type":"string"}},"required":["query"]}}]}' | jq  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  1519  100  1231  100   288   1510    353 --:--:-- --:--:-- --:--:--  1863
{
  "id": "chatcmpl-b0753bfdda7893af",
  "type": "message",
  "role": "assistant",
  "content": [
    {
      "type": "text",
      "text": "<think>\nOkay, the user is asking to search the web for the year America was founded. Let me think about how to approach this.\n\nFirst, I need to figure out the correct historical date. America was founded on July 4, 1776. But the user wants to know the year, not the date. So, the answer should be 1776. However, the user is asking for a web search, so I should use the search function provided.\n\nI should construct a query to search for \"year america was founded\" to get relevant information. The function requires a \"query\" parameter, so I'll input that. Let me make sure there are no typos. The function name is \"search\", and the parameters are a JSON object with \"query\". \n\nI think that's all. Just need to format the tool call correctly within the XML tags. Let me double-check the syntax to ensure it's valid JSON. Once confirmed, the tool call should return the result.\n</think>\n\n"
    },
    {
      "type": "tool_use",
      "id": "chatcmpl-tool-a02c6f995701b000",
      "name": "search",
      "input": {
        "query": "year america was founded"
      }
    }
  ],
  "model": "Qwen/Qwen3-0.6B",
  "stop_reason": "tool_use",
  "usage": {
    "input_tokens": 152,
    "output_tokens": 229
  }
}
Streaming
```
```
$ curl http://localhost:8000/v1/messages \
  -d '{"model":"Qwen/Qwen3-0.6B","max_tokens":1024,"stream":true,"messages":[{"role":"user","content":"search the web for the year america was founded"}],"tools":[{"name":"search","description":"search the web","input_schema":{"type":"object","properties":{"query":{"type":"string"}},"required":["query"]}}]}'
event: message_start
data: {"type":"message_start","message":{"id":"chatcmpl-854bb730eed1434f","content":[],"model":"Qwen/Qwen3-0.6B","usage":{"input_tokens":152,"output_tokens":0}}}

event: content_block_start
data: {"type":"content_block_start","content_block":{"type":"text","text":""},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":"<think>"},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":"\n"},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":"Okay"},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":","},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":" the"},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":" user"},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":" wants"},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":" to"},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":" know"},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":" the"},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":" year"},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":" America"},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":" was"},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":" founded"},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":"."},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":" Let"},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":" me"},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":" think"},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":"."},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":" I"},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":" need"},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":" to"},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":" search"},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":" the"},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":" web"},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":" for"},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":" that"},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":" information"},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":"."},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":" The"},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":" available"},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":" tool"},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":" is"},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":" the"},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":" '"},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":"search"},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":"'"},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":" function"},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":","},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":" which"},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":" takes"},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":" a"},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":" query"},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":" parameter"},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":"."},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":" The"},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":" query"},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":" here"},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":" would"},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":" be"},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":" \""},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":"year"},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":" america"},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":" was"},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":" founded"},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":"\"."},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":" I"},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":" should"},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":" make"},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":" sure"},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":" to"},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":" use"},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":" the"},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":" correct"},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":" function"},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":" name"},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":" and"},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":" parameters"},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":"."},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":" Let"},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":" me"},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":" construct"},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":" the"},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":" tool"},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":" call"},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":" properly"},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":".\n"},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":"</think>"},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"text_delta","text":"\n\n"},"index":0}

event: content_block_stop
data: {"type":"content_block_stop","index":0}

event: content_block_start
data: {"type":"content_block_start","content_block":{"type":"tool_use","id":"chatcmpl-tool-b5b2602282b39cb5","name":"search","input":{}},"index":1}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"input_json_delta","partial_json":"{\"query\": \""},"index":1}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"input_json_delta","partial_json":"year"},"index":1}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"input_json_delta","partial_json":" america"},"index":1}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"input_json_delta","partial_json":" was"},"index":1}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"input_json_delta","partial_json":" founded"},"index":1}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"input_json_delta","partial_json":"\"}"},"index":1}

event: content_block_stop
data: {"type":"content_block_stop","index":1}

event: message_delta
data: {"type":"message_delta","delta":{"stop_reason":"tool_use"},"usage":{"input_tokens":152,"output_tokens":101}}

event: message_stop
data: {"type":"message_stop"}

data: [DONE]
AI Gateway
$ curl http://localhost:8000/v1/messages   -H "Content-Type: application/json"   -d '{"model":"Qwen/Qwen2.5-0.5B-Instruct","max_tokens":1024,"messages":[{"role":"user","content":"search the web for the year america was founded"}],"tools":[{"name":"search","description":"search the web","input_schema":{"type":"object","properties":{"query":{"type":"string"}},"required":["query"]}}]}' | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   636  100   337  100   299   2378   2109 --:--:-- --:--:-- --:--:--  4510
{
  "id": "chatcmpl-be92bf95eb4f5d6d",
  "type": "message",
  "role": "assistant",
  "content": [
    {
      "type": "text",
      "text": ""
    },
    {
      "type": "tool_use",
      "id": "chatcmpl-tool-8f41db8c125e33eb",
      "name": "search",
      "input": {
        "query": "year america was founded"
      }
    }
  ],
  "model": "Qwen/Qwen2.5-0.5B-Instruct",
  "stop_reason": "tool_use",
  "usage": {
    "input_tokens": 168,
    "output_tokens": 22
  }
}
$ curl http://localhost:8000/v1/messages \ 
  -d '{"model":"Qwen/Qwen2.5-0.5B-Instruct","max_tokens":1024,"stream":true,"messages":[{"role":"user","content":"search the web for the year america was founded"}],"tools":[{"name":"search","description":"search the web","input_schema":{"type":"object","properties":{"query":{"type":"string"}},"required":["query"]}}]}'
event: message_start
data: {"type":"message_start","message":{"id":"chatcmpl-a2a79c35b0a88cfe","content":[],"model":"Qwen/Qwen2.5-0.5B-Instruct","usage":{"input_tokens":168,"output_tokens":0}}}

event: content_block_start
data: {"type":"content_block_start","content_block":{"type":"tool_use","id":"chatcmpl-tool-b846494e0a70f02e","name":"search","input":{}},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"input_json_delta","partial_json":"{\"query\": \""},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"input_json_delta","partial_json":"America"},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"input_json_delta","partial_json":" was"},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"input_json_delta","partial_json":" founded"},"index":0}

event: content_block_delta
data: {"type":"content_block_delta","delta":{"type":"input_json_delta","partial_json":"\"}"},"index":0}

event: content_block_stop
data: {"type":"content_block_stop","index":0}

event: message_delta
data: {"type":"message_delta","delta":{"stop_reason":"tool_use"},"usage":{"input_tokens":168,"output_tokens":21}}

event: message_stop
data: {"type":"message_stop"}

data: [DONE]
```

@changminbark : It seems that the format that Envoy is outputting is identical to the format the vLLM is outputting. I am not too sure why your tool call is not working. Isn't Claude Code also executing the tool call in the screenshot that you showed? Doesn't that mean that it is properly processing the tool call request from the model? Maybe it's a problem with Claude Code?

My thoughts: I think that web search tool in Claude Code requires a specific streaming SSE format as seen below. However, other tool calls (like ones that you custom define should work as intended). You can see that the type of the content block in the SSE for web search tool is server_tool_use instead of tool_use [here](https://platform.claude.com/docs/en/agents-and-tools/tool-use/web-search-tool#streaming).

@changminbark : It looks like web search tool is a Anthropic-specific tool that only Claude models can generate as part of their output SSE, so even if the request comes from Claude Code that says that this tool is available, the OpenAI backend cannot generate the web search tool specific SSE response needed for Claude Code to do the web search. Claude Code can only conduct web searches with this specific tool defined by Anthropic. The way LiteLLM goes around this is that they intercept the user-defined (not Anthropic-defined) web search tool and then executes it server-side as described [here](https://docs.litellm.ai/release_notes/v1-81-0#:~:text=Claude%20Code%20%2D%20Web%20Search%20Across%20All%20Providers%E2%80%8B,or%20any%20other%20supported%20provider.). Please correct me if my understanding is wrong.

@groundsada : okay, I ran some tests. You're 100% right. This is a Claude Code specific issue. Web search is failing on both vLLM and Envoy. IMO, PR looks great to merge. Can always ask CC to use a DDG MCP server and disable web search tool and it works flawlessly. One thing that is happening is that on vLLM, web search fails and it moves to different tools (fetch), via Envoy it remains stuck on a loop of failing.

pujitha24:
**Description**

While investigating issue 1911, a maintainer posted a gateway log line hit while reproducing the report on vLLM:

```
error processing request message: cannot process response body: failed to
transform response error: failed to unmarshal OpenAI error body: Mismatch
type string with value  number "at index 97"
```

This comes from `ResponseError` in `internal/translator/anthropic_openai.go`, which decodes an upstream error body into `openai.Error`. `openai.ErrorType` (the nested "error" object) declares `Type`, `Message`, `Code`, `Param`, and `EventID` as string-typed fields, but the repo's JSON layer (`internal/json`, a bytedance/sonic wrapper configured with `CaseSensitive: true`) type-checks strictly during decode. Not every OpenAI-compatible backend (e.g. vLLM) sends these fields as strings -- a prior PR already special-cased "code" to tolerate a JSON number, but "type", "message", "param", and "event_id" were left strict. Any one of them arriving as a non-string value makes the whole error body fail to decode with exactly the "Mismatch type string with value number" error above, and `ResponseError` returns a hard Go error instead of a translated Anthropic-format error response.

This PR generalizes the existing code-only leniency to all five fields of `ErrorType`, using gjson (already a dependency, already used the same way for flexible-type parsing in `internal/apischema/anthropic/anthropic.go`) to read each field and coerce it to its string representation regardless of whether the upstream sent a string, number, or boolean. A field that is absent or JSON null still unmarshals to nil for the pointer fields, matching prior behavior.

This only fixes the JSON-unmarshal defect that produces the log line above; it does not by itself reproduce or resolve the broader Claude Code web-search retry-loop behavior discussed in issue 1911, which needs a live vLLM/Claude Code environment to confirm end-to-end.

Related: #1911

Validation performed:
- Added a table-driven case to `TestAnthropicToOpenAITranslator_ResponseError` in `internal/translator/anthropic_openai_test.go` with a body where "type", "message", "param", and "event_id" are all JSON numbers. Confirmed by temporarily reverting `internal/apischema/openai/openai.go` that this case fails with the exact "Mismatch type string with value number" error before the fix, and passes after restoring it.
- Added `TestErrorTypeUnmarshalJSON` in `internal/apischema/openai/openai_test.go` covering all-string fields, numeric fields, a boolean field, and null/missing optional fields.
- `go build ./...` and `go test ./internal/...` pass.
- `make precommit` (tidy, spellcheck, apigen, codegen, apidoc, format/license headers, golangci-lint, actionlint, editorconfig, helm-lint/package) passes with only the intended files modified.

---

Fixes #1911


johnugeorge:
Why can't we target to fix it in vllm instead?

pujitha24:
Fair question. A few reasons I think it's worth fixing here regardless of what vLLM does:

- The gateway sits in front of many OpenAI-compatible backends, not just vLLM. Any of them sending a non-string value for one of these error fields currently makes the gateway hard-fail the translation with a Go decode error, instead of relaying *some* usable error to the client. Even a fully-fixed vLLM wouldn't help users on older vLLM versions or other backends with the same non-compliance.
- This repo already accepted the same reasoning once: #2161 special-cased the `code` field for exactly this class of problem. This PR just extends that existing precedent to the other four fields rather than introducing a new approach.

I haven't looked into whether there's a vLLM-side fix for the specific error body that triggered this (the `ValueError: Failed to transform json schema into a grammar` case from the issue), and I'm not against pursuing that too — they're not mutually exclusive, this one just stops the gateway from throwing away the upstream error entirely in the meantime. Happy to narrow this PR's scope or hold off if you'd rather only pursue the vLLM-side route.

johnugeorge:
@pujitha24 Have you seen any other backends that produce similar non OpenAI error types

johnugeorge:
LGTM. But, this is a slippery slope

pujitha24:
I haven't independently confirmed other specific backends doing this — my evidence is limited to the vLLM case from this issue and the fact that #2161 already had to special-case `code` for the same reason. So no, I can't point to a second backend beyond vLLM right now.

On the slippery-slope point: that's fair, and it's why I changed the approach here instead of just adding `type`/`message`/`param`/`event_id` as more one-off special cases next to the existing `code` handling. `ErrorType.UnmarshalJSON` now reads all five fields the same way via gjson and coerces whatever value shows up to a string, so a sixth field showing up as a number later wouldn't need a new special case — it'd already be covered by the same code path. Tests for the numeric/boolean/null cases are in `TestErrorTypeUnmarshalJSON`.

johnugeorge:
Are responses ai generated?

pujitha24:
Yes, I'm using AI assistance to research and draft the responses in this thread, including the ones above about vLLM scope and the `ErrorType.UnmarshalJSON` approach. Wanted to be upfront about that rather than let it go unanswered. Happy to keep addressing feedback here if that works for you — let me know if you'd rather handle this differently.

PR #2508, opened 2026-08-10
fix(aigw): cap Envoy http/ext_proc debug logging to stop credential leak

### LINKED ISSUE #2436: extproc/Envoy debug logs leak injected backend credential (redaction gap)
**Description**

When extproc debug logging is enabled (e.g. `AIGW_DEBUG=true`, or any deployment running the extproc at debug level with a backend that has an API key), the backend API credential — the `Authorization` header the extproc upstream filter injects from the backend's APIKey — can appear in cleartext in debug logs, even though the extproc already redacts the client-supplied `authorization` header via `filterSensitiveHeadersForLogging`.

The redaction coverage is inconsistent across three leak points observed in a live `aigw run`:

1. extproc Go server's own `request headers processed` debug log dumps the header mutation it returns to Envoy. The injected `Authorization` was only redacted when `enableRedaction` was on; with debug on but redaction off (the `aigw run` default), the real `Bearer ...` key was logged, while the incoming client `authorization` header on the same request showed `[REDACTED]`.
2. Envoy's `[debug][http]` `conn_manager_impl` "request headers complete" log emits all request headers verbatim, including the client's `authorization` and the upstream's injected `authorization`.
3. Envoy's `[debug][ext_proc]` `ext_proc.cc` "Received request headers response" log dumps the full gRPC `ProcessingResponse` proto returned by the extproc, including the `Authorization` header mutation `raw_value`.

Net effect: an operator who turns on debug logging to troubleshoot a request ends up with real backend API keys in their log files, with no indication that redaction is incomplete. The intent to keep credentials out of debug logs is already there (the incoming header is redacted); the gap is coverage, not policy.

**Repro steps**:

1. `OPENAI_API_KEY=sk-... OPENAI_BASE_URL=https://... AIGW_DEBUG=true aigw run` (any OpenAI-compatible backend with a real key works; the backend need not accept the model).
2. Send a chat completions request, e.g. `curl -s http://localhost:1975/v1/chat/completions -H 'Authorization: Bearer sk-...' -d '{"model":"...","messages":[{"role":"user","content":"hi"}]}'`.
3. Grep the extproc+envoy debug log for the key substring. It appears in (1) the extproc `request headers processed` line's `Authorization` mutation `raw_value`, (2) Envoy's `[http]` request-headers-complete lines (downstream and upstream), and (3) Envoy's `[ext_proc]` "Received request headers response" proto dump.

**Proposed fix**:

- extproc Go side (covered by #2435): always redact the header mutation in the `request headers processed` debug log when debug logging is on, independent of `enableRedaction` — mirroring `redactRequestBodyResponse`, which already redacts header mutations unconditionally and only gates body-content redaction on the flag. This closes leak point 1.
- Envoy side (remaining): leak points 2 and 3 cannot be redacted from extproc code, since Envoy logs the headers/proto it shuttles around itself. The fix here is log-level policy: the debug mode should not enable Envoy's `http` and `ext_proc` loggers at debug (or expose finer-grained Envoy log-level control so credential-bearing traffic isn't dumped), and/or `AIGW_DEBUG` should be documented as credential-leaky at the Envoy layer and recommended only for trusted/local environments.

**Worth considering**:

- **No functional risk on the extproc side.** The redaction only touches the `logContent` copy used for the debug log; the actual `ProcessingResponse` sent to Envoy (and upstream) is unchanged. A unit test in #2435 pins "log shows `[REDACTED]`, response carries the real key."
- **Decouple credential redaction from `enableRedaction`.** `enableRedaction` governs the more expensive body-content redaction (user prompts) and is opt-in. Credential/header redaction in debug logs should be always-on — a backend key must never reach a debug log regardless of that flag. #2435 makes it so for the extproc log; the same principle should guide the Envoy-side decision.
- **The Envoy-side gap is the residual exposure.** Even after #2435, an operator running full Envoy debug still gets plaintext keys from Envoy's own logs. Whatever the chosen Envoy-side mitigation (log-level default, finer knobs, or a documented warning), it should be the thing that actually closes points 2 and 3, since extproc redaction cannot reach them.
- **Consider whether a "redact known-secret headers in Envoy debug" path is even reachable.** Envoy has no built-in per-header redaction in its debug logging, so the realistic lever is log levels, not redaction. If a reviewer wants true redaction at the Envoy layer, that is an Envoy/EG-level change, not an ai-gateway one.

This is a follow-on to the same redaction surface that #2428 widened: that PR extended body-content redaction to the 7 passthrough endpoints, while this issue is about credential-header redaction in the extproc and Envoy debug logs — a different leak surface that #2435 partially closes.

pujitha24:
**Description**

When `aigw run` is started with `AIGW_DEBUG=true`, the generated EnvoyProxy resource sets `logging.level.default: debug`, which enables debug logging for every Envoy logger component. Two of those components dump the injected backend Authorization header in cleartext at debug level: Envoy's own `[debug][http]` "request headers complete" log (full request headers) and `[debug][ext_proc]` "Received request headers response" log (the full gRPC ProcessingResponse proto, including the header mutation's raw_value). This is separate from, and not fixed by, the extproc-side redaction already merged for the same credential in a prior PR: that fix only covers the extproc Go server's own debug log, not Envoy's own logging of the header mutation it receives back from the extproc.

This commit caps the `http` and `ext_proc` components at `info` in the generated EnvoyProxy logging config whenever debug mode is on, while leaving `default: debug` (and every other component) untouched. Envoy Gateway's `ProxyLogging.GetEnvoyProxyComponentLevel()` passes non-default map entries straight through to Envoy's own `--component-log-level` flag; `ext_proc` is a real Envoy logger id even though it falls outside Envoy Gateway's Go-level `ProxyLogComponent` enum documentation, and that enum isn't represented in the generated CRD's OpenAPI schema for map keys, so `ext_proc: info` validates and is accepted the same way whether running locally via `aigw run` or against a real Kubernetes-deployed Envoy Gateway. No functional behavior changes: the actual headers sent between Envoy, the extproc, and the backend are unchanged, only the volume of local debug logging for these two components is reduced.

Fixes #2436

Validation performed:
- `go build ./...`
- `go vet ./internal/autoconfig/... ./cmd/aigw/...`
- `go test ./internal/autoconfig/... ./cmd/aigw/...`
- Reproduced the leak and confirmed the fix locally: built `aigw` from both the pre-change and post-change code, ran each with `OPENAI_API_KEY=sk-<sentinel> OPENAI_BASE_URL=<local fake backend> AIGW_DEBUG=true aigw run`, and sent a real chat-completions request through the gateway. Before the fix, the sentinel key appeared in Envoy's own `[debug][ext_proc]` "Received request headers response" dump, and `[debug][http]` lines showed full header dumps. After the fix, `ps` confirmed the real Envoy process was launched with `--component-log-level ext_proc:info,http:info` and started successfully; the same request produced zero occurrences of the sentinel key anywhere in the log, `[debug][http]` and `[debug][ext_proc]` line counts both dropped to zero, while `[debug][upstream]`, `[debug][router]`, and `[debug][main]` lines remained present, and the HTTP response returned to the client was identical before and after.

---


johnugeorge:
Either way, we should not log credentials even in debug mode
cc @nacx

pujitha24:
Agreed, and that's exactly the goal here. The extproc-side leak (the Go server's own debug log) was already closed by #2435. This PR addresses the remaining gap: Envoy's own `[debug][http]` and `[debug][ext_proc]` loggers dump the injected `Authorization` header regardless of what extproc logs, so this caps just those two components at `info` when debug mode is on, leaving everything else at `debug` as before. Happy to adjust if @nacx sees a better approach here.

mathetake (approved):

PR #2509, opened 2026-08-10
feat: per-request AWS credentials for SigV4 signing

### LINKED ISSUE #2076: Feature Request: Support per-request AWS credentials via headers for per-user cost attribution

kanurag94:
**Description**

#2253 added `credentialOverride` so a trusted upstream filter can supply the backend credential per-request, but left `AWSCredentials` out: SigV4 needs three values and the plumbing carried one string. This adds it.

`fromRequestHeaders` treats the header field as a prefix rather than a full name, defaulting to `x-aigw-aws-`:

```yaml
spec:
  type: AWSCredentials
  awsCredentials:
    region: us-east-1
  credentialOverride:
    fromRequestHeaders: {}   # x-aigw-aws-access-key-id, -secret-access-key, -session-token
```

All three are stripped before the request reaches Bedrock, using the `HeaderMutation.Remove` mechanism from #2253.

If one already inject `x-aws-*` from an ext-auth service, `header: "x-aws-"` reproduces those exact names, so the injector needs no change.

`fromDynamicMetadata` reads one struct with fixed `accessKeyId` / `secretAccessKey` / `sessionToken` fields:

```yaml
  credentialOverride:
    fromDynamicMetadata:
      namespace: envoy.filters.http.ext_authz
      key: x-aigw-aws-credentials
```

A partial credential fails the request rather than falling back, even with `fallbackToConfigured: true`. Half a credential means the filter upstream is misconfigured.

`fallbackToConfigured` stays usable for AWS, since the credentials file and the default credential chain both count as a fallback.

Closes #2076
Related: #2253, #2416

**Testing**

UT and E2e

Some follow up PRs needed to fix functionality for using dynamic metadata in this field.


pbonuboinasnow:
When there are more than one backends and they use different credentials, fixed headers are not going to be enough. I think naming them with backend name helps.  For aws-ptu1, headers will be x-aigw-aws-access-key-id-aws-ptu1, -secret-access-key-aws-ptu1, -session-token-aws-ptu1. Same for other providers

kanurag94:
> When there are more than one backends and they use different credentials, fixed headers are not going to be enough. I think naming them with backend name helps. For aws-ptu1, headers will be x-aigw-aws-access-key-id-aws-ptu1, -secret-access-key-aws-ptu1, -session-token-aws-ptu1. Same for other providers

It is currently possible to specify a prefix for each backend so you can possibly use `x-aigw-aws-ptu1-` and it will create `x-aigw-aws-ptu1-access-key-id` etc.
IOW, header is already a prefix per policy rather than a fixed name. I'd rather keep that explicit than auto-suffix: auto-naming breaks reusing an existing x-aws-* injector.

Though one part I can agree is that if you supply surplus header, then it would not strip all of them but leak it to upstream whichever is not actually used. That part however is something the ext_authz should own. But I can take care of stripping in a followup for all types.

pbonuboinasnow:
> > When there are more than one backends and they use different credentials, fixed headers are not going to be enough. I think naming them with backend name helps. For aws-ptu1, headers will be x-aigw-aws-access-key-id-aws-ptu1, -secret-access-key-aws-ptu1, -session-token-aws-ptu1. Same for other providers
> 
> It is currently possible to specify a prefix for each backend so you can possibly use `x-aigw-aws-ptu1-` and it will create `x-aigw-aws-ptu1-access-key-id` etc. IOW, header is already a prefix per policy rather than a fixed name. I'd rather keep that explicit than auto-suffix: auto-naming breaks reusing an existing x-aws-* injector.
> 
> Though one part I can agree is that if you supply surplus header, then it would not strip all of them but leak it to upstream whichever is not actually used. That part however is something the ext_authz should own. But I can take care of stripping in a followup for all types.

Let's take an example where there are 3 backends and fallback is setup to attempt of all them. In this case extproc has to set x-aigw-aws-ptu-*, x-aigw-gcp-ptu-* and x-aigw-azure-ptu-*, correct? and will it make sure it does not leak ptu2 and ptu3 crdes when attempting  ptu1 etc?

kanurag94:
> > > When there are more than one backends and they use different credentials, fixed headers are not going to be enough. I think naming them with backend name helps. For aws-ptu1, headers will be x-aigw-aws-access-key-id-aws-ptu1, -secret-access-key-aws-ptu1, -session-token-aws-ptu1. Same for other providers
> > 
> > 
> > It is currently possible to specify a prefix for each backend so you can possibly use `x-aigw-aws-ptu1-` and it will create `x-aigw-aws-ptu1-access-key-id` etc. IOW, header is already a prefix per policy rather than a fixed name. I'd rather keep that explicit than auto-suffix: auto-naming breaks reusing an existing x-aws-* injector.
> > Though one part I can agree is that if you supply surplus header, then it would not strip all of them but leak it to upstream whichever is not actually used. That part however is something the ext_authz should own. But I can take care of stripping in a followup for all types.
> 
> Let's take an example where there are 3 backends and fallback is setup to attempt of all them. In this case extproc has to set x-aigw-aws-ptu-_, x-aigw-gcp-ptu-_ and x-aigw-azure-ptu-*, correct? and will it make sure it does not leak ptu2 and ptu3 crdes when attempting ptu1 etc?

I want to tackle this in followup as mentioned in last comment. EAIGW needs to strip all known headers not relevant for this request during fallback. I will also add docs why Dynamic Metadata should be preferred to avoid such cases.

kanurag94:
@johnugeorge @nacx
Gentle ping if we can add this to next release

johnugeorge:
Thanks @kanurag94

johnugeorge (approved):

PR #2512, opened 2026-08-11
fix(translator): allowlist anthropic-beta flags forwarded to AWS Anthropic

johnpanos:
**Description**

This commit filters the `anthropic-beta` request header values forwarded to AWS Anthropic (Bedrock) backends against an allowlist of flags Bedrock actually accepts, instead of forwarding every value verbatim into the `anthropic_beta` request body field.

AWS Bedrock rejects the entire request with a 400 `"invalid beta flag"` error when `anthropic_beta` contains a flag it does not recognize. Clients such as Claude Code send beta flags targeting the Anthropic API (e.g. `prompt-caching-2024-07-31`, `extended-cache-ttl-2025-04-11`, `files-api-2025-04-14`), so a single unsupported flag previously broke the whole request through the gateway. Unsupported flags are now stripped, and the `anthropic_beta` field is omitted entirely when nothing supported remains.

The allowlist contains 17 flags, each confirmed accepted (HTTP 200) by live probes against Bedrock:

`context-1m-2025-08-07`, `context-management-2025-06-27`, `interleaved-thinking-2025-05-14`, `dev-full-thinking-2025-05-14`, `fine-grained-tool-streaming-2025-05-14`, `token-efficient-tools-2025-02-19`, `token-counting-2024-11-01`, `tool-search-tool-2025-10-19`, `computer-use-2024-10-22`, `computer-use-2025-01-24`, `mcp-client-2025-04-04`, `mcp-client-2025-11-20`, `web-search-2025-03-05`, `search-results-2025-06-09`, `output-128k-2025-02-19`, `pdfs-2024-09-25`, `model-context-window-exceeded-2025-08-26`

Probed and confirmed rejected (documented in the code comment so they are not re-added from Anthropic docs): `prompt-caching-2024-07-31`, `extended-cache-ttl-2025-04-11`, `thinking-token-count-2026-05-13`, `files-api-2025-04-14`, `code-execution-2025-05-22`, `memory-2025-08-18`.

Unit tests cover the full allowlisted set, mixed supported/unsupported headers, and the all-unsupported case.

A second commit rewrites Anthropic-API-only flag names to their Bedrock equivalents before the allowlist check. The same feature can be gated under different flag names depending on the backend: the Anthropic API gates Tool Search (plus Programmatic Tool Calling and Tool Use Examples) under the umbrella flag `advanced-tool-use-2025-11-20`, while Bedrock exposes Tool Search under the older-dated `tool-search-tool-2025-10-19`. The `tool_search_tool_regex_20251119` / `tool_search_tool_bm25_20251119` tool type strings are identical on both. Clients speaking the Anthropic API dialect (e.g. Claude Code) send the umbrella flag, which the allowlist alone would strip, silently disabling tool search; the gateway now maps it to the Bedrock flag (deduplicating when both spellings are sent), matching the translation litellm ships in `anthropic_beta_headers_config.json`. An e2e matrix (`TestAnthropicBetaHeaderTranslationMatrix`) asserts the exact upstream `InvokeModel` body for the umbrella flag, the native Bedrock spelling, dedup, mixed supported/unsupported, and all-unsupported cases.

N/A

I raised this PR because requests from Claude Code to Bedrock-backed routes failed with 400s whenever the client sent a beta flag Bedrock does not support, even though the remaining flags were valid. Stripping unsupported flags at the gateway matches how the gateway already normalizes other Bedrock-specific request shapes.

The allowlist is intentionally limited to probe-verified flags rather than Anthropic's documented beta list, since an incorrect entry would reintroduce the exact 400 this change prevents. Extending it is a one-line change per newly verified flag.

I used an AI coding assistant to help implement this change, and I reviewed and verified the code and tests, including the live Bedrock probes behind the allowlist.

### DIFF
--- a/internal/translator/anthropic_awsanthropic.go
+++ b/internal/translator/anthropic_awsanthropic.go
@@ -70,13 +70,43 @@ var awsBedrockSupportedAnthropicBetas = map[string]struct{}{
 	"web-search-2025-03-05":                    {},
 }
 
+// anthropicAPIToBedrockBetaAliases maps anthropic-beta flag names that only the Anthropic API
+// accepts onto the equivalent Bedrock flag name. The same feature is often exposed under
+// different flag names depending on the backend, and clients speaking the Anthropic API
+// dialect (e.g. Claude Code) send the Anthropic API name, which would otherwise be stripped
+// by the allowlist above, silently disabling the gated feature.
+//
+// advanced-tool-use-2025-11-20 is the Anthropic API umbrella flag for Tool Search,
+// Programmatic Tool Calling and Tool Use Examples. Bedrock supports only Tool Search,
+// exposed under the older-dated tool-search-tool-2025-10-19; the tool type strings
+// (tool_search_tool_regex_20251119 / tool_search_tool_bm25_20251119) are identical on both.
+// The umbrella flag's non-Bedrock-supported siblings are dropped by the allowlist as usual.
+//
+// Aliases are applied before the allowlist check in SetRequestHeaders.
+var anthropicAPIToBedrockBetaAliases = map[string]string{
+	"advanced-tool-use-2025-11-20": "tool-search-tool-2025-10-19",
+}
+
 // SetRequestHeaders implements [RequestHeadersSetter].
 func (a *anthropicToAWSAnthropicTranslator) SetRequestHeaders(headers map[string]string) {
 	var anthropicBetas []string
+	seen := map[string]struct{}{}
 	if betaHeader := headers["anthropic-beta"]; betaHeader != "" {
 		for _, beta := range strings.Split(betaHeader, ",") {
 			beta = strings.TrimSpace(beta)
+			// Translate Anthropic-API-only flag names to their Bedrock equivalents
+			// before the allowlist check.
+			if alias, ok := anthropicAPIToBedrockBetaAliases[beta]; ok {
+				beta = alias
+			}
 			if _, ok := awsBedrockSupportedAnthropicBetas[beta]; ok {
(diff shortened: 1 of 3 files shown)


johnugeorge:
How do we plan to keep this updated? The concern is with the maintainence of the list of flags

johnpanos:
> How do we plan to keep this updated? The concern is with the maintainence of the list of flags

This is a question I also had, and don't yet have an answer to. I can do some more research but I wasn't able to find an automated/authoritative source for this list yet.

johnugeorge:
Is this list complete? How long are they valid? Need a strategy to keep this updated. Otherwise, it doesn't serve the purpose if it gets outdated soon.

johnpanos:
This list is complete as of today. I brute forced every Anthropic flag mentioned in their docs.

I'll search the AWS docs today for a similar list. If there's no API for this what would a good solution in your eyes?

I don't mind owning a reoccurring task here to update.

rpbarnes (approved):

johnpanos:
https://docs.litellm.ai/blog/claude-code-beta-headers-incident

Looks like LiteLLM is handling this in a similar way, as well.

mathetake (approved):

hustxiayang:
@johnpanos @mathetake @johnpanos  I think the issue is the same as https://github.com/envoyproxy/ai-gateway/issues/2411, and the solution is https://github.com/envoyproxy/ai-gateway/pull/2420.  I had the same concern on the maintaince of the list, thus I put the deny list in the config, so that whenever there are some updates, we do not need to get blocked by the code changes of this repo. Instead, we could just config them, cc @aabchoo

PR #2513, opened 2026-08-11
fix: do not reprocess gateway local replies as upstream responses

CodePrometheus:
**Description**

When the gateway rejects a request itself — a 422 for an invalid request body, a 401 for a missing upstream credential — Envoy returns that as a local reply, which travels back through the response path and is handled as if it came from the upstream. 
The error body gets re-translated, which on a cross-schema route drops the message entirely: a client sending temperature: 2.0 to an AWSAnthropic backend gets `{"type":"error","error":{"type":"","code":"422"}}` instead of being told the value is outof range. 
The request completion is also recorded a second time, so two failed requests appear as four in `gen_ai_server_request_duration_seconds_count`.

Such replies are now passed through untouched and counted once, so callers get the message that tells them what to fix and failure metrics reflect the real request count.


siddharth1036:
Overall LGTM! Good catch!

@johnugeorge @anuraaga @nacx can you take a look?

johnugeorge:
Thanks @CodePrometheus

johnugeorge (approved):

PR #2515, opened 2026-08-11
fix: drop content-encoding from streamed responses before headers are sent

CodePrometheus:
**Description**

Streaming responses with content-encoding: gzip are broken when the translator rebuilds the body, e.g. an Anthropic client against an OpenAI-compatible backend.

`extproc` decompresses the upstream body and the translator emits new plaintext, but the client already received `content-encoding:gzip`. It tries to inflate plaintext, fails with incorrect header check, and the stream dies mid-response.

`removeContentEncodingIfNeeded` handles this for buffered responses. It cannot for streamed ones: it runs in the body phase, and the headers are already on the wire by then.

The fix removes the header in the header phase, where it is still mutable, and sends the decompressed bytes even when the translator leaves the body alone.


CodePrometheus:
Gentle ping @johnugeorge mind taking a look when you have time?

PR #2516, opened 2026-08-11
fix: update structured output of guided_json, guided_regex, guided_choice of gcp to make it compatible with gcp

hustxiayang:
**Description**
vLLM v0.12.0 removed the legacy top-level guided-decoding request fields (`guided_json`, `guided_regex`, `guided_choice`, `guided_grammar`, …) and replaced them with a single nested `structured_outputs` object: https://github.com/vllm-project/vllm/blob/v0.12.0/vllm/sampling_params.py#L33 . Clients migrating to the new format got no guided decoding on GCP/Gemini models, because the Gemini translator only understood the deprecated fields. This adds `structured_outputs` support so the new format is honored on the GCP path, while keeping the legacy `guided_*` fields working for backward compatity.


hustxiayang:
@nacx Do you mind to review it?

PR #2517, opened 2026-08-11
fix: honour InferencePool failureMode when injecting the EPP ext_proc filter

### LINKED ISSUE #2502: InferencePool failureMode: FailOpen is not honored: injected ext_proc filters omit failure_mode_allow
**Description**:

An InferencePool whose `endpointPickerRef.failureMode` is `FailOpen` should degrade to normal load
balancing across the pool when the Endpoint Picker (EPP) is unavailable. Instead, all traffic to that
pool is dropped: the behaviour is FailClose regardless of the configured `failureMode`.

The cause is that the ext_proc filter AI Gateway injects for the EPP callout does not set
`failure_mode_allow`. Envoy's ext_proc default when that field is unset is to fail the request when
the processor is unreachable, so the InferencePool's `failureMode` never reaches the filter that
would enforce it.

This is not a user-configuration problem. The `extensionManager` injects the ext_proc filter chain
*after* Envoy Gateway has translated the Gateway API resources into xDS, and an `EnvoyPatchPolicy`
patches the pre-injection xDS, so it cannot reach the injected filters to add the field. At this
version there is no configuration surface that reaches them, so the fix has to happen where the
filter is injected.

**Repro steps**:

1. Install Envoy Gateway v1.8.3, the Gateway API Inference Extension v1.5.0 InferencePool CRDs, and
   Envoy AI Gateway v1.0.0.
2. Create an InferencePool with an EPP and set FailOpen:

```yaml
apiVersion: inference.networking.k8s.io/v1
kind: InferencePool
metadata:
  name: pool-a
  namespace: demo
spec:
  selector:
    matchLabels: { app: model-server }
  targetPortNumber: 8000
  endpointPickerRef:
    name: pool-a-epp
    failureMode: FailOpen
Route to the pool via an HTTPRoute backendRef of kind InferencePool. Confirm 2xx while the EPP is up.
Scale the EPP deployment to 0 (the outage).
Send 8 probes. Expected under FailOpen: requests continue, load balanced across the pool. Observed: 0 of 8 return 2xx.
Restore the EPP and confirm recovery.
Confirm the cause on the Envoy proxy:

curl -s localhost:19000/config_dump | grep -c failure_mode_allow    # 0
curl -s localhost:19000/config_dump | grep -c ExternalProcessor     # 2 (the injected filters)
Environment:

Envoy AI Gateway v1.0.0
Envoy Gateway v1.8.3
Gateway API Inference Extension v1.5.0
Kubernetes 1.34 (Amazon EKS)
Two InferencePools, one EPP each, behind a single Gateway
Logs:

Probe results, 2xx / non-2xx, one tier configured FailClose and one FailOpen:


baseline, both EPPs up:      tier-a 6/0    tier-b 6/0
tier-a EPP down (FailClose): tier-a 0/8    -> clean 5xx, as designed
tier-a restored:             tier-a 6/0    -> recovered
tier-b EPP down (FailOpen):  tier-b 0/8    -> DROPPED, expected degraded routing
tier-b restored:             tier-b 6/0    -> recovered
The FailOpen pool behaves identically to the FailClose pool, which is the bug.

The decisive artifact is the running configuration rather than the request logs. Both injected
ExternalProcessor filters carry only these fields, with failure_mode_allow absent from both
(0 occurrences across the entire config_dump):


{
  "@type": "type.googleapis.com/envoy.extensions.filters.http.ext_proc.v3.ExternalProcessor",
  "grpc_service": { "envoy_grpc": { "cluster_name": "<endpoint-picker-cluster>" }, "timeout": "..." },
  "message_timeout": "300s",
  "processing_mode": { "...": "..." }
}
Raw Envoy access logs from that run were not retained; the probe tallies above and the config_dump
grep are what the run recorded. I can capture fresh access logs against the same setup if useful.

Workaround, for others hitting this:

Set failureMode: FailClose on the pool so the manifest matches the delivered behaviour, rather than
asserting a FailOpen the stack does not deliver.

Offer:

Happy to test a patch that sets failure_mode_allow on the injected ext_proc filter, gated by the
InferencePool failureMode, against this setup and report back with a fresh config_dump and probe
results.

### PR DESCRIPTION
**Pull request by pujitha24**
**Description**

This commit fixes InferencePool `failureMode: FailOpen` being treated the same as `FailClose`.

`buildHTTPFilterForInferencePool` (internal/extensionserver/inferencepool.go) previously hardcoded `FailureModeAllow: false` on the injected `envoy.extensions.filters.http.ext_proc.v3.ExternalProcessor` HTTP filter. It now sets `FailureModeAllow` to `pool.Spec.EndpointPickerRef.FailureMode == gwaiev1.EndpointPickerFailOpen`.

The InferencePool consumed at that point is reconstructed by `getInferencePoolByMetadata` from a `/`-delimited metadata string, which previously encoded only 6 fields (`namespace/name/serviceName/port/bodyMode/allowModeOverride`) and dropped `failureMode`. The encoding is extended to a 7th field:

- `clusterRefInferencePool` (internal/extensionserver/post_cluster_modify.go) takes a new `failureMode` parameter and appends it to the formatted string.
- `buildEPPMetadata` (internal/extensionserver/inferencepool.go) appends `string(inferencePool.Spec.EndpointPickerRef.FailureMode)` when building the metadata value, and its doc comment is updated to reflect the new format.
- `getInferencePoolByMetadata` now requires 7 `/`-delimited fields (was 6), reads the 7th as `failureMode`, and sets it on the reconstructed pool's `EndpointPickerRef.FailureMode`.

Tests updated accordingly:

- `extensionserver_test.go`: existing `per_route_rule_inference_pool` metadata fixtures in `TestMaybeModifyClusterExtended`, `TestMaybeModifyListenerAndRoutes`, `TestPatchVirtualHostWithInferencePool`, and `TestBuildClustersForInferencePoolEndpointPickers` are updated to include the trailing `/FailClose` field.
- `inferencepool_test.go`:
  - `TestBuildAndParseMetadata_RoundTrip` sets `FailureMode: gwaiev1.EndpointPickerFailOpen` on the input pool, asserts the encoded string ends in `/FailOpen`, and asserts the parsed pool's `FailureMode` matches.
  - `TestGetInferencePoolByMetadata_Malformed` fixture string is updated to the 7-field format.
  - `TestBuildHTTPFilterForInferencePool_Defaults` gains an assertion that `FailureModeAllow` is `false` when `failureMode` is unset.
  - New `TestBuildHTTPFilterForInferencePool_FailureMode` table test covers `FailOpen`, `FailClose`, and unset `failureMode` against `buildHTTPFilterForInferencePool`.

Fixes InferencePool `failureMode: FailOpen` being treated the same as `FailClose`.

None.

---

Fixes #2502

### DIFF
--- a/internal/extensionserver/inferencepool.go
+++ b/internal/extensionserver/inferencepool.go
@@ -99,7 +99,7 @@ func getInferencePoolByMetadata(meta *corev3.Metadata) *gwaiev1.InferencePool {
 	}
 
 	result := strings.Split(metadata, "/")
-	if len(result) != 6 {
+	if len(result) != 7 {
 		return nil
 	}
 	ns := result[0]
@@ -111,6 +111,7 @@ func getInferencePoolByMetadata(meta *corev3.Metadata) *gwaiev1.InferencePool {
 	}
 	processingBodyMode := result[4]
 	allowModeOverride := result[5]
+	failureMode := result[6]
 	return &gwaiev1.InferencePool{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      name,
@@ -122,15 +123,16 @@ func getInferencePoolByMetadata(meta *corev3.Metadata) *gwaiev1.InferencePool {
 		},
 		Spec: gwaiev1.InferencePoolSpec{
 			EndpointPickerRef: gwaiev1.EndpointPickerRef{
-				Name: gwaiev1.ObjectName(serviceName),
-				Port: ptr.To(gwaiev1.Port{Number: gwaiev1.PortNumber(port)}),
+				Name:        gwaiev1.ObjectName(serviceName),
+				Port:        ptr.To(gwaiev1.Port{Number: gwaiev1.PortNumber(port)}),
+				FailureMode: gwaiev1.EndpointPickerFailureMode(failureMode),
 			},
 		},
 	}
 }
 
 // buildMetadataForInferencePool adds InferencePool metadata to the cluster for reference by other components.
-// encoded as a string in the format: "namespace/name/serviceName/port/bodyMode/allowModeOverride".
+// encoded as a string in the format: "namespace/name/serviceName/port/bodyMode/allowModeOverride/failureMode".
 func buildEPPMetadataForCluster(cluster *clusterv3.Cluster, inferencePool *gwaiev1.InferencePool) {
 	// Initialize cluster metadata structure if not present.
 	if cluster.Metadata == nil {
@@ -179,6 +181,7 @@ func buildEPPMetadata(metadata *corev3.Metadata, inferencePool *gwaiev1.Inferenc
 			portForInferencePool(inferencePool),
 			processingBodyMode,
 			allowModeOverride,
+			string(inferencePool.Spec.EndpointPickerRef.FailureMode),
 		),
 	)
 }
@@ -311,7 +314,11 @@ func buildHTTPFilterForInferencePool(pool *gwaiev1.InferencePool) *extprocv3.Ext
 		},
 		AllowModeOverride: allowModeOverride,
 		MessageTimeout:    durationpb.New(300 * time.Second),
-		FailureModeAllow:  false,
+		// FailureModeAllow controls what Envoy does when the EPP is unreachable: true lets the
+		// request continue (fail open), false drops it (fail close). This must mirror the
+		// InferencePool's own failureMode since nothing downstream of this injected filter can
+		// influence it.
+		FailureModeAllow: pool.Spec.EndpointPickerRef.FailureMode == gwaiev1.EndpointPickerFailOpen,
 	}
 }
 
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: internal/extensionserver/inferencepool.go [hunk 1]
```diff
@@ -99,7 +99,7 @@ func getInferencePoolByMetadata(meta *corev3.Metadata) *gwaiev1.InferencePool {
 	}
 
 	result := strings.Split(metadata, "/")
-	if len(result) != 6 {
+	if len(result) != 7 {
```
siddharth1036, line 102: This strict len(result) == 7 check introduces a mixed-version upgrade risk.
During rolling upgrades, some extension callbacks can still produce the old 6-field metadata (namespace/name/serviceName/port/bodyMode/allowModeOverride), and this parser will return nil for those entries. That can make PostTranslateModify miss InferencePool routes/clusters and skip EPP patching.

Could we make this backward-compatible by accepting both 6 and 7 fields, defaulting failureMode to FailClose when absent?
Suggested shape:

6 fields => parse old format + failureMode = FailClose
7 fields => parse new format with explicit failureMode

kanurag94 (commented):
Thanks for this!
When `FailOpen` is used, once ext_proc is bypassed the client's own `x-gateway-destination-endpoint` reaches the ORIGINAL_DST cluster,  which is configured with use_http_header: true on exactly that header. That means when EPP is down, it lets a client choose it's upstream. We probably should strip this client header. 

**This may not fix #2502** completely as #2502 also proposes an acceptable degraded behavior when EPP fails. I think this'd just 503. Please CMIIW

Jay-Pd:
Hi @pujitha24, are you still working on this?

PR #2524, opened 2026-08-13
translator: serialize non-streaming Bedrock reasoning_content as plain string

### LINKED ISSUE #2519: AWS Bedrock: reasoning_content serialized as an object instead of a plain string
## What happens

Routing an OpenAI `/chat/completions` request to an AWS Bedrock model that returns reasoning (e.g. the gpt-oss models on Bedrock) produces a response where `reasoning_content` is a nested **object** rather than a plain string.

Non-streaming:
```json
"message": {
  "role": "assistant",
  "content": "hello world",
  "reasoning_content": { "reasoningContent": { "reasoningText": { "text": "…reasoning…" } } }
}
```

Streaming delta:
```json
"delta": { "role": "assistant", "reasoning_content": { "text": "…reasoning…" } }
```

OpenAI-compatible clients expect `reasoning_content` to be a string (the litellm convention). A client that concatenates reasoning deltas as strings breaks — Open WebUI raises `sequence item 0: expected str instance, dict found`.

## Why this is a bug, not a preference

The intended contract is already documented in the code: `reasoning_content` is a plain string, and structured signature/redacted metadata goes in `thinking_blocks`.

`internal/apischema/openai/openai.go`, on `ChatCompletionResponseChoiceMessage`:
```go
// ThinkingBlocks holds structured thinking metadata (signatures, redacted content) …
// This preserves full fidelity without breaking clients that expect
// reasoning_content to be a plain string.
```

The Gemini path follows it (`internal/translator/gemini_helper.go:795`):
```go
message.ReasoningContent = &openai.ReasoningContentUnion{Value: thoughtSummary} // string
message.ThinkingBlocks = append(message.ThinkingBlocks, openai.ThinkingBlock{ … })
```

The Bedrock path does not (`internal/translator/openai_awsbedrock.go:468`):
```go
case output.ReasoningContent != nil:
    choice.Message.ReasoningContent = &openai.ReasoningContentUnion{
        Value: &openai.ReasoningContent{ReasoningContent: output.ReasoningContent},
    }
```

Streaming has the same issue — `StreamReasoningContent` has no string marshaler, so it always serializes as `{"text": …}` (`openai_awsbedrock.go:608` populates it; `openai.go` defines the struct with no `MarshalJSON`).

`ReasoningContentUnion.MarshalJSON` already emits a plain string when `Value` is a string, so the non-streaming fix is just to populate it with the text.

## Suggested fix

Non-streaming — `internal/translator/openai_awsbedrock.go`:
```go
case output.ReasoningContent != nil:
    if rt := output.ReasoningContent.ReasoningText; rt != nil {
        choice.Message.ReasoningContent = &openai.ReasoningContentUnion{Value: rt.Text}
        choice.Message.ThinkingBlocks = append(choice.Message.ThinkingBlocks, openai.ThinkingBlock{
            Type: "thinking", Thinking: rt.Text, Signature: rt.Signature,
        })
    }
```

Streaming — give `StreamReasoningContent` a string marshaler (or emit `thinking_blocks` deltas to retain the signature):
```go
func (s StreamReasoningContent) MarshalJSON() ([]byte, error) {
    return json.Marshal(s.Text)
}
```

The existing tests encode the current object shape and would need updating: `TestOpenAIToAWSBedrockTranslatorV1ChatCompletion_ResponseBody_WithReasoning` (`openai_awsbedrock_test.go:2377`) asserts `reasoning_content` is a map with a nested `reasoningContent` key, and `…_Streaming_WithReasoning` (2294) asserts the delta is a map.

## Version

v1.0.0 (also reproduces on `main` @ `a05ff66`).

dragon3:
**Description**

The non-streaming AWS Bedrock path still serialized `reasoning_content` as the legacy nested object (`reasoning_content.reasoningContent.reasoningText.text`) instead of a plain string. OpenAI-compatible clients that follow the de-facto `reasoning_content: string` convention break on this — Open WebUI raises `sequence item 0: expected str instance, dict found` when concatenating reasoning.

This closes the remaining gap in the reasoning-content convention:

| Path | Non-streaming | Streaming |
|---|---|---|
| Gemini/Vertex | ✅ #1995 | ✅ #2319 |
| AWS Bedrock | ❌ **this PR** | ✅ #2319 |

`openai_awsbedrock.go` now mirrors the streaming path and the Gemini helper: the reasoning text goes to the plain-string `reasoning_content`, and the signature / redacted content move to `thinking_blocks` (they cannot be represented in a plain string). Existing translator goldens are updated to the new wire shape, plus a new regression test covering signature and redacted content.

Per the generative AI policy: this change was developed with an AI assistant (Claude); I take ownership of the code and will address review feedback.

Fixes #2519
Related PR: #2319 (streaming counterpart), #1995 (Gemini non-streaming)

This is the non-streaming follow-up @johnugeorge asked for in #2519. The signature/redacted-content routing to `thinking_blocks` matches what #2319 already does for streaming.

### DIFF
--- a/internal/translator/openai_awsbedrock.go
+++ b/internal/translator/openai_awsbedrock.go
@@ -471,10 +471,18 @@ func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) ResponseBody(_ map[string
 			// redacted content go to thinking_blocks (they cannot be represented in a
 			// plain string). This mirrors the streaming path and the Gemini helper.
 			if rt := output.ReasoningContent.ReasoningText; rt != nil {
-				choice.Message.ReasoningContent = &openai.ReasoningContentUnion{Value: rt.Text}
-				choice.Message.ThinkingBlocks = append(choice.Message.ThinkingBlocks, openai.ThinkingBlock{
-					Type: "thinking", Thinking: rt.Text, Signature: rt.Signature,
-				})
+				// Only set reasoning_content when there is actual text; a
+				// signature-only block must not emit an empty string.
+				if rt.Text != "" {
+					choice.Message.ReasoningContent = &openai.ReasoningContentUnion{Value: rt.Text}
+				}
+				// Surface the block (text and/or signature) so a signature still
+				// round-trips even without accompanying text.
+				if rt.Text != "" || rt.Signature != "" {
+					choice.Message.ThinkingBlocks = append(choice.Message.ThinkingBlocks, openai.ThinkingBlock{
+						Type: "thinking", Thinking: rt.Text, Signature: rt.Signature,
+					})
+				}
 			}
 			if rc := output.ReasoningContent.RedactedContent; rc != nil {
 				choice.Message.ThinkingBlocks = append(choice.Message.ThinkingBlocks, openai.ThinkingBlock{
--- a/internal/translator/openai_awsbedrock.go
+++ b/internal/translator/openai_awsbedrock.go
@@ -467,10 +467,19 @@ func (o *openAIToAWSBedrockTranslatorV1ChatCompletion) ResponseBody(_ map[string
 				choice.Message.Content = output.Text
 			}
 		case output.ReasoningContent != nil:
-			choice.Message.ReasoningContent = &openai.ReasoningContentUnion{
-				Value: &openai.ReasoningContent{
-					ReasoningContent: output.ReasoningContent,
-				},
+			// Reasoning text goes to the plain-string reasoning_content; signature and
+			// redacted content go to thinking_blocks (they cannot be represented in a
+			// plain string). This mirrors the streaming path and the Gemini helper.
+			if rt := output.ReasoningContent.ReasoningText; rt != nil {
+				choice.Message.ReasoningContent = &openai.ReasoningContentUnion{Value: rt.Text}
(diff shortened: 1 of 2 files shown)


dragon3:
@johnugeorge

The only failing check is `E2E Test for Inference Extensions`, which timed out waiting for the conformance Gateway/InferencePool to become Programmed (`Reason: NoResources` → `context deadline exceeded`) — the traffic assertions never ran. That's unrelated to this change, which only touches the OpenAI→Bedrock reasoning serialization; Unit, Controller, and Data Plane tests all pass. Could you re-run the `E2E Test for Inference Extensions` job? 🙏

johnugeorge:
what is the expected behaviour  for a scenario with no reasoning text but only signature content ?

dragon3:
Good question. For a signature-only reasoning block (signature present, no text), the intent is to match the streaming path (#2319): we should not emit an empty `reasoning_content` string. Instead the signature surfaces via `thinking_blocks` so it still round-trips to the provider.

I've updated the non-streaming path accordingly (aa905077):
- `reasoning_content` is set only when there is actual reasoning text.
- A `thinking_block` is emitted whenever text or a signature is present, so a signature-only block yields `{"type":"thinking","signature":"…"}` with no `reasoning_content` on the wire.

Added `..._WithReasoningSignatureOnly` to cover exactly this case.

johnugeorge (approved):

PR #2530, opened 2026-08-14
translator: configure Anthropic Messages translation for OpenAI

johnpanos:
**Description**

Adds explicit API control for translating Anthropic Messages requests sent to OpenAI-schema backends:

```yaml
schema:
  name: OpenAI
  capabilities:
    messagesTranslation: Responses
```

`messagesTranslation` accepts `ChatCompletions` or `Responses`. Omission defaults to `ChatCompletions`, preserving compatibility with vLLM and other OpenAI-compatible backends. Endpoint selection is entirely configuration-driven; this PR does not add model-name heuristics.

When `Responses` is selected, the translator covers system and multimodal messages, tools and tool results, tool choice, parallel tool use, reasoning configuration, streaming and non-streaming responses, errors, and token usage.

For stateless multi-turn reasoning, Responses requests use `store: false` and request `reasoning.encrypted_content`. Complete reasoning items are encoded as Anthropic `redacted_thinking` blocks and restored before the next upstream request, preserving the item ID, type, summary, content, status, and exact encrypted content. Streaming emits the same replayable blocks and suppresses duplicates between `response.output_item.done` and `response.completed`.

The Responses usage parser also supports `cache_write_tokens` and maps OpenAI cache usage to Anthropic fields without double-counting input tokens.

The CRDs, generated API documentation, examples, and provider/endpoint guides are updated. CEL validation restricts `capabilities` to the OpenAI schema and validates/defaults the endpoint selection.

None.

This remains a draft while reviewer feedback is collected.

Validated with:

- `make precommit`
- `env -u AWS_PROFILE -u AWS_DEFAULT_REGION make test`
- `make test-crdcel GO_TEST_ARGS='-count=1'`
- `go test -race ./internal/translator`
- production Docusaurus build under Node 22 (passes with pre-existing generated API-anchor warnings)
- live GPT-5.6 stateless reasoning replay through proxy
- live prompt-cache probe: first turn wrote 4,213 tokens; replayed second turn read all 4,213 cached tokens and wrote 16 new tokens


tests/crdcel/testdata/aiservicebackends/openai-capabilities-default.yaml
```diff
@@ -0,0 +1,18 @@
...
+
+apiVersion: aigateway.envoyproxy.io/v1beta1
+kind: AIServiceBackend
+metadata:
+  name: openai-capabilities-default
+  namespace: default
+spec:
+  schema:
+    name: OpenAI
+    capabilities: {}
```
nacx, line 14: Let's just remove it for the default test case, to also ensure this is backwards-compatible and we do not require this field?
```suggestion
```

internal/translator/anthropic_openai_responses.go
```diff
@@ -0,0 +1,685 @@
...
+		"input":             input,
+		"include":           []string{"reasoning.encrypted_content"},
+		"max_output_tokens": body.MaxTokens,
+		"store":             false,
+	}
+	if body.Stream {
+		request["stream"] = true
+	}
+	if body.Temperature != nil {
+		request["temperature"] = *body.Temperature
```
nacx, line 101: GPT-5 models may not support temperature, which led to 400 errors.
Is this OK, or do we want to handle this more gracefully?

internal/translator/anthropic_openai_responses.go
```diff
@@ -0,0 +1,685 @@
...
+		}
+		return a.streamState.handleChunk(responsesToolCallChunk(done.OutputIndex, nil, done.Name, done.Arguments), out)
+	case event.OfResponseOutputItemDone != nil && event.OfResponseOutputItemDone.Item.OfReasoning != nil:
+		return a.emitResponsesReasoningItem(event.OfResponseOutputItemDone.Item.OfReasoning, out)
+	case event.OfResponseCompleted != nil:
+		return a.completeResponsesStream(&event.OfResponseCompleted.Response, out)
+	case event.OfResponseIncomplete != nil:
+		return a.completeResponsesStream(&event.OfResponseIncomplete.Response, out)
+	case event.OfResponseFailed != nil:
+		return a.completeResponsesStream(&event.OfResponseFailed.Response, out)
```
nacx, line 575: Should failed ones also carry the error, similar to the `OfError` case?

internal/translator/anthropic_openai_responses.go
```diff
@@ -0,0 +1,685 @@
...
+	}
+	result, err := json.Marshal(request)
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal OpenAI Responses request: %w", err)
+	}
+	return result, nil
+}
+
+func chatMessagesToResponsesInput(value any) ([]any, error) {
+	messages, ok := value.([]any)
```
nacx, line 136: If Anthropic yields zero messages (`chat["messages"]` is `nil`) this assertion would fail. We should probably treat `nil` as empty.

internal/translator/anthropic_openai_responses.go
```diff
@@ -0,0 +1,685 @@
...
+	}
+	if parallel := anthropicParallelToolCalls(body.ToolChoice); parallel != nil {
+		request["parallel_tool_calls"] = *parallel
+	}
+	if body.Thinking != nil {
+		switch {
+		case body.Thinking.Disabled != nil:
+			request["reasoning"] = map[string]any{"effort": "none"}
+		case body.Thinking.Enabled != nil:
+			request["reasoning"] = map[string]any{"effort": "high"}
```
nacx, line 123: Here we just default to high, ignoring the budget, which would lead to users always using the OpenAI expensive tier. Would it make sense to map budgets to low/medium/high?

internal/translator/anthropic_openai_responses.go
```diff
@@ -0,0 +1,685 @@
...
+	if choice, ok := chat["tool_choice"]; ok {
+		request["tool_choice"] = chatToolChoiceToResponsesToolChoice(choice)
+	}
+	if parallel := anthropicParallelToolCalls(body.ToolChoice); parallel != nil {
+		request["parallel_tool_calls"] = *parallel
+	}
+	if body.Thinking != nil {
+		switch {
+		case body.Thinking.Disabled != nil:
+			request["reasoning"] = map[string]any{"effort": "none"}
```
nacx, line 121: For models that don't support it is probably better to just omit it, to prevent 400 errors.

PR #2532, opened 2026-08-14
mcp: include backend extensions in merged capabilities

### LINKED ISSUE #2531: mcpproxy: extensions capability is dropped when aggregating backends
The MCP proxy does not propagate the `extensions` capability from backends, so an extension a backend advertises is invisible to clients connecting through the gateway. This makes MCP Apps (`io.modelcontextprotocol/ui`) impossible to negotiate through an MCPRoute.

A backend (FastMCP 3.4.4) queried directly returns:

```json
"capabilities": {
  "resources": {"subscribe": false, "listChanged": true},
  "tools": {"listChanged": true},
  "extensions": {"io.modelcontextprotocol/ui": {}}
}
```

`mergedCapabilities()` in internal/mcpproxy/session.go unions Tools, Prompts, Logging, Resources and Completions, but never sets Extensions, although `ServerCapabilities` in go-sdk v1.6.1 has the field.

Effect: the host sees a server with no extensions, so it ignores `_meta.ui.resourceUri` on the tools and renders results as plain text. Tools and resources themselves route fine, and #2412 already fixed the `ui://` URI namespacing, so this capability gap is the remaining blocker.

Possibly related: `handleInitializeRequest` answers with a fixed `protocolVersion` (2025-06-18) instead of negotiating against the version the client asked for, while backends may speak a newer one.

umerfarok:
**Description**

`mergedCapabilities()` unions Tools, Prompts, Logging, Resources and Completions but drops Extensions, so a backend advertising `io.modelcontextprotocol/ui` looks like it supports no extensions once it is behind the proxy, and clients never negotiate MCP Apps.

Found this with an MCP Apps server behind an MCPRoute: the backend advertises the extension in initialize and its tool carries `_meta.ui.resourceUri`, but the client sees no extensions downstream and renders the tool result as text instead of the view.

The proxy does receive it. Before this change, with a backend whose initialize advertises the extension, the session log shows `Extensions:map[io.modelcontextprotocol/ui:map[]]` while the initialize response written downstream has no extensions at all.

Extensions are unioned like the other fields. The first backend to declare an id wins, so the result does not depend on map iteration order. Two tests: a case in TestMergedCapabilities, and TestServePOST_InitializeRequest_ForwardsExtensions which drives servePOST against a backend advertising the extension and asserts it survives into the downstream response.

This may not be sufficient on its own: handleInitializeRequest answers with a fixed protocolVersion of 2025-06-18, which predates extension negotiation, so a client keyed on the negotiated version could still ignore the map. Happy to fold that in here or leave it separate, whichever you prefer.

`TestServePOST_ToolsCallRequest` fails on main with or without this change, so it looks unrelated.

Fixes #2531


umerfarok:
@nacx Bro its urgent can you please review

Hritik003 (commented):
LGTM

sqaisar:
Looks solid

aishwaryaraimule21 (approved):
LGTM

johnugeorge:
LGTM

johnugeorge (approved):

PR #2535, opened 2026-08-15
docs: fix broken Gateway API reference links

walsm232:
**Description**

Every link from this repository into the Gateway API reference documentation is currently broken, because the upstream site restructured its reference section.

Two separate moves are involved:

- `/api-types/httproute/` is now `/reference/api-types/httproute/`.
- `/reference/spec/` is now `/reference/api-spec/<version>/spec/`, and its anchors changed from version-qualified names such as `#gateway.networking.k8s.io/v1.HTTPRoute` to plain lowercase `#httproute`.

The first form returns a hard 404 but the second is subtler: a handful of links used the `/references/spec/` (plural) alias, which still redirects and therefore looks like it works, but its old anchor matches no element, so readers silently land at the top of a long reference page instead of the type they clicked.

This changes the source config and doc comments, regenerates the API reference, and applies the same rewrite to the published documentation snapshots.  It is a pure URL rewrite: 18 distinct URLs, 297 link occurrences across 284 changed lines, with no content changes. The `?h=httproutetimeouts` search-highlight parameters left on six links are dropped at the same time.

N/A (at least I am not aware of an issue for this)

The new spec links point at `/reference/api-spec/main/spec/` rather than a pinned version such as `1.5`, which would match this repository's `sigs.k8s.io/gateway-api v1.5.1` dependency. Pinned paths do exist and resolve today, but `/reference/api-spec/1.3/spec/` is already a 404 for example, so the site seems to prune old versions and a pinned link could reintroduce the rot at a later point.

As mentioned, `main` always resolves AFAIK. Happy to switch to `1.5` if folks would rather have version accuracy and treat the links as something to bump alongside the dependency.


johnugeorge:
cc @missBerg

PR #2539, opened 2026-08-16
fix(autoconfig): reject unsupported URL schemes with ports

### LINKED ISSUE #2538: aigw run accepts unsupported URL schemes with explicit ports
_Description_:

The `aigw run` autoconfiguration validates backend URL schemes only while selecting a default port. An unsupported scheme therefore fails when no port is present, but is accepted when an explicit port bypasses that branch.

_Repro steps_:

1. Configure a backend URL such as `ftp://example.com:21` (for example via `OPENAI_BASE_URL`).
2. Start `aigw run`.
3. Observe that autoconfiguration accepts the URL and renders a non-TLS Backend on port 21.

Expected: autoconfiguration rejects every scheme except `http` and `https`, regardless of whether the URL contains an explicit port.

The validation should run before default or explicit port handling, with a regression test for the explicit-port case.

0YHR0:
**Description**

Autoconfiguration currently validates the backend URL scheme only when it needs to select a default port. As a result, URLs such as `ftp://example.com:21` bypass the scheme check and produce a non-TLS Kubernetes Backend instead of failing early.

This change validates the scheme before port handling so only `http` and `https` are accepted whether or not the URL has an explicit port. It also adds a regression test for the explicit-port case.

Fixes #2538

Validation:
- `go test ./internal/autoconfig -count=1`
- `go vet ./internal/autoconfig`
- `git diff --check`


0YHR0:
@mathetake This is a focused autoconfig validation fix: unsupported schemes such as `ftp://example.com:21` previously bypassed validation whenever an explicit port was present. The branch is now rebased onto current main, with package tests, vet, changed-line coverage, and a fresh full CI run in progress. Could you please review it when the run settles?

nacx (approved):

PR #2540, opened 2026-08-17
fix: forward credentialOverride metadata namespaces to the upstream ext_proc

kanurag94:
**Description**

`credentialOverride.fromDynamicMetadata` needs the namespace to be forwarded to the extproc configuration.

This adds `GatewayConfig.spec.extProc.metadataForwardingNamespaces`. The extension server resolves it once per xDS snapshot and sets `forwarding_namespaces` on the upstream ext_proc filter of every AI cluster in it.

Envoy Gateway watches Gateways but not GatewayConfigs, so the controller stamps a hash of the config spec into a Gateway annotation to trigger a re-translation. That also fixes `forwardProxy` edits, which previously applied only on the next unrelated translation.

#2253, #2509

Only untyped dynamic metadata is read.

The controller logs an error when a `BackendSecurityPolicy` reads a namespace the GatewayConfig does not declare, since the two halves of the configuration live on different objects.


internal/controller/gateway.go
```diff
@@ -1401,6 +1417,61 @@ func (c *GatewayController) getObjectsForGateway(ctx context.Context, gw *gwapiv
...
+	if desired == "" {
+		delete(gw.Annotations, gatewayConfigHashAnnotationKey)
+	} else {
+		if gw.Annotations == nil {
+			gw.Annotations = make(map[string]string)
+		}
+		gw.Annotations[gatewayConfigHashAnnotationKey] = desired
+	}
+	if err := c.client.Patch(ctx, gw, patch); err != nil {
+		return fmt.Errorf("failed to patch Gateway with GatewayConfig hash: %w", err)
```
PatilHrushikesh, line 1456: why not to log error everytime on repeated ns as well? user will know all backends referring non declared ns in one shot

kanurag94: thanks!, Makes sense. I have added a consolidated log for it

siddharth1036:
Nice work! One extra thought on the Gateway hash annotation: it writes to a user-owned resource, which feels a bit unclean.
Could something like [extension server policies](https://gateway.envoyproxy.io/docs/tasks/extensibility/extension-server/#targeting-resources-with-extension-server-policies) work here? If GatewayConfig were registered under extensionManager.policyResources and attached with targetRefs (kind Gateway), EG would watch it, re-translate on spec change, and pass the object into PostTranslateModify, so we might not need the hash at all.

Not sure it is a fit with the current annotation-based attachment; just wanted to float it.

PatilHrushikesh:
I think collecting namespaces across all gatewayconfig and applying to all clusters, doesnt sound good,  In clusters we have aigwroutes as well we can find its corresponding gateway and then gatewayconfig and fetch ns and apply only those as forewarding ns

kanurag94:
> I think collecting namespaces across all gatewayconfig and applying to all clusters, doesnt sound good, In clusters we have aigwroutes as well we can find its corresponding gateway and then gatewayconfig and fetch ns and apply only those as forewarding ns

I agree applying the list to all AI clusters is broader than it needs to be. But extProc is Gateway-level config, and the API reads as "these are the namespaces the Gateway hands to the processor," so I'd rather keep the source of the list Gateway-level. WDYT about narrowing it to the clusters that actually use the configuration. Would that answer your concern?

On resolving it from the route instead, I don't think it works. Without mergeGateways it returns the same set, since a snapshot is one Gateway — it just costs a Gateway and GatewayConfig lookup per cluster instead of one per snapshot.

Under `mergeGateways` it changes who decides. `ParentRefs` is a list, and clusters are deduplicated by name with no Gateway component — `httproute/<ns>/<name>/rule/N/backendref/M` — so an AIGatewayRoute attached to two Gateways has exactly one cluster and no single owning Gateway to resolve. One will need to union anyway, but from the route, which means adding a parentRef would start forwarding one Gateway's namespaces on another Gateway's traffic. Resolving it from the Gateways that own the snapshot prevents that.

kanurag94:
> Nice work! One extra thought on the Gateway hash annotation: it writes to a user-owned resource, which feels a bit unclean. Could something like [extension server policies](https://gateway.envoyproxy.io/docs/tasks/extensibility/extension-server/#targeting-resources-with-extension-server-policies) work here? If GatewayConfig were registered under extensionManager.policyResources and attached with targetRefs (kind Gateway), EG would watch it, re-translate on spec change, and pass the object into PostTranslateModify, so we might not need the hash at all.
> 
> Not sure it is a fit with the current annotation-based attachment; just wanted to float it.

It's a good idea but like you said the annotation-based attachment model is not fit for it. I would like to work on this in a followup, is that okay?

PatilHrushikesh:
> so I'd rather keep the source of the list Gateway-level.

Yeah i agree with keeping it gateway level (in gatewayconfig)

> it just costs a Gateway and GatewayConfig lookup per cluster instead of one per snapshot.

this we are still doing for foreward proxy

kanurag94:
> this we are still doing for foreward proxy

Yeah it's actually fine for proxy address but for something like forwarding namespaces there's no "pick one" so one would need to union it. In a followup I will add a follow up PR for forward proxy to use the same mechanism

kanurag94:
gentle ping @envoyproxy/ai-gateway-maintainers for review

internal/controller/gateway.go
```diff
@@ -1381,6 +1397,67 @@ func (c *GatewayController) getObjectsForGateway(ctx context.Context, gw *gwapiv
...
+		if ns == "" || slices.Contains(declared, ns) {
+			continue
+		}
+		if _, ok := backends[ns]; !ok {
+			undeclared = append(undeclared, ns)
+		}
+		backends[ns] = append(backends[ns], b.Name)
+	}
+	for _, ns := range undeclared {
+		c.logger.Error(nil, "credentialOverride reads a dynamic metadata namespace the GatewayConfig does not declare in extProc.metadataForwardingNamespaces; Envoy will not forward it, so these backends fall back to the configured credential",
```
siddharth1036: Hi @kanurag94 is it possible to reflect this misconfiguration as warning via BackendSecurityPolicy status?

nacx (approved):
Thanks!

PR #2518, opened 2026-08-12
feat(mcp): stateless spec era detection and legacy split (28-07-2026)

Hritik003:
**Description**

This PR adds the initial phase0 changes as per
-  #2431 that are required for the new MCP stateless spec. Introduces protocol era(can be modern/legacy) detection and splits the legacy handling out of the handler file to keep things separate as per legacy and modern routing. Reason it was done this way so that, in future when MCP deprecates legacy mode we could just delete the legacy code as it is not worrying about the behavior.

- internal/mcpproxy/era.go - protocol version constants (2025-11-25, 2026-07-28), modern MCP header/_meta keys, SDK error-code mappings, and the era type used to classify a client/backend as legacy vs. modern.
- Moves the existing legacy MCP request handlers from handlers.go into a new internal/mcpproxy/legacy.go, leaving handlers.go as the thin dispatch layer. No behavioral change to legacy handling.
- Adds era_test.go covering era detection.

This lays the groundwork for the modern-spec handlers without touching current legacy behavior.

**Related Issues/PRs**
- #2431 

**Special notes for reviewers (if applicable)**

- handlers.go → legacy.go changes are almost entirely code moves; and since e2e data-plane tests have passed we should be good wrt existingbehavior not getting disturbed.

**References taken for validations**
- https://modelcontextprotocol.io/seps/2243-http-standardization
- https://modelcontextprotocol.io/seps/2575-stateless-mcp

**Testing** 
Built and deployed the changes to my local cluster, tested the functionality using MCP inspector.

PR #2543, opened 2026-08-17
fix(mcpproxy): negotiate protocol version from backend responses

### LINKED ISSUE #2542: mcpproxy: handleInitializeRequest hardcodes protocolVersion instead of negotiating from backends
Found this while working on MCP Apps negotiation through an MCPRoute. The gateway always responds with `protocolVersion: "2025-06-18"` in the initialize result, even when backends report a newer version.

The constant lives in `handlers.go` around line 580. Each backend's actual `InitializeResult.ProtocolVersion` comes back fine during `initializeSession`, but it gets discarded. Only capabilities are stored.

This creates a weird situation after #2532 lands. Extensions will appear in merged capabilities, but the client sees a protocol version that predates extension negotiation. A spec-strict client could reasonably ignore the extensions map because the version says they shouldn't exist yet.

I think the gateway should store the backend's reported version in the session entry and return the minimum across all backends when responding to the client. It also shouldn't exceed what the client asked for, since that's what the spec requires of servers.

There's also a secondary instance at `sendRequestPerBackend` (line 367 in session.go) where the gateway sends the hardcoded version header to backends on subsequent requests.

elang2:
**Description**

handleInitializeRequest responded with a hardcoded protocolVersion of
2025-06-18 regardless of what backends actually support. A client keyed
on the negotiated version could ignore capabilities (like extensions)
that require a newer protocol.

This stores each backend's reported protocolVersion during initialization
and returns the minimum across all backends, capped by the client's
requested version. The gateway never advertises a version newer than
what all backends support or what the client asked for.

Changes:

session.go: added protocolVersion field to compositeSessionEntry and a
mergedProtocolVersion() method that computes the minimum across backends,
falling back to the 2025-06-18 constant when no backend reports a version.

mcpproxy.go: stores initResult.result.ProtocolVersion during backend init.

handlers.go: uses the negotiated version instead of the constant, capped
at the client's requested version.

Note: sendRequestPerBackend (session.go line 367) still sends the
hardcoded version header to backends on subsequent requests. That could
be addressed as a follow-up since it requires encoding the per-backend
version in the session ID.

Fixes #2542
Related: #2531, #2532


internal/mcpproxy/session.go
```diff
@@ -687,6 +688,28 @@ func (s *session) mergedCapabilities() *mcpsdk.ServerCapabilities {
...
 }
 
+// mergedProtocolVersion returns the minimum protocol version across all backends.
+// MCP protocol versions are ISO date strings (YYYY-MM-DD), so lexicographic
+// comparison gives chronological ordering. The minimum is returned because the
+// gateway can only guarantee features supported by all backends.
+// Returns the default version constant if no backend reported a version.
+func (s *session) mergedProtocolVersion() string {
+	var minVersion string
+	for _, entry := range s.perBackendSessions {
```
Hritik003: All the merge helpers currently reside in handlers.go, can we also place it there for consistency

Hritik003 (commented):
lgtm, 2 nits

internal/mcpproxy/handlers.go
```diff
@@ -577,7 +577,11 @@ func (m *mcpRequestContext) handleInitializeRequest(ctx context.Context, w http.
 		return err
 	}
 
-	result := mcp.InitializeResult{ProtocolVersion: protocolVersion20250618, ServerInfo: &mcp.Implementation{}}
+	negotiatedVersion := s.mergedProtocolVersion()
+	if p.ProtocolVersion != "" && p.ProtocolVersion < negotiatedVersion {
+		negotiatedVersion = p.ProtocolVersion
+	}
```
Hritik003: can also move this inside `mergedProtocolVersion` by passing `negotiatedVersion` as an argument


```suggestion
	protocolVersion := s.mergedProtocolVersion(negotiatedVersion)
```

elang2:
Makes sense, I'll fold the client cap into the method signature and move it to handlers.go. One question though: `mergedCapabilities()` also lives on the session type in session.go, so I assumed that was the convention for merge logic. Happy to move it either way, just want to confirm you'd prefer handlers.go even with that precedent.

Hritik003:
> Makes sense, I'll fold the client cap into the method signature and move it to handlers.go. One question though: `mergedCapabilities()` also lives on the session type in session.go, so I assumed that was the convention for merge logic. Happy to move it either way, just want to confirm you'd prefer handlers.go even with that precedent.

Ideally yes, we can let it be for now. We are planning to anyways handle the placement of all these helpers in #2518.

internal/mcpproxy/session.go
```diff
@@ -687,6 +688,7 @@ func (s *session) mergedCapabilities() *mcpsdk.ServerCapabilities {
 	return merged
 }
 
+
 // String implements fmt.Stringer.
```
Hritik003: nit: remove new line?

Hritik003 (commented):
one nit, else LGTM. Can you also fix the CI ?

Hritik003 (commented):
LGTM

kanurag94 (approved):

elang2:
Thanks for the review @kanurag94. @johnugeorge would you be able to merge when ready?

aishwaryaraimule21 (commented):
This change looks good to me overall as this strategy is better than hardcoding a protocol version.
However, I have the following concerns:

1. This negotiation strategy implicitly assumes that MCP protocol versions are backward compatible, which is not the case. We need to call it out in a comment.

2. This might break existing MCP Clients connecting to AI Gateway. 
e.g. when the client version is 2025-06-18 and one of the server backends is at a lower version 2024-11-05, the negotiated version returned would be 2024-11-05 which is different from what it returned earlier (2025-06-18). 


3. One good thing about the earlier hardcoded approach was, Client <> Gateway and Gateway <> Server shared the same protocol version. After this change, e.g. even if the negotiated version is 2025-11-25, the request will be forwarded with the 2025-06-18 payload, resulting in newer fields getting dropped. This should get solved after we pick up the change mentioned in the PR description. 

I am skeptical of merging just this PR without the above change for this month's release. Should we also take up the change in point 3 and have both as part of the next monthly release?

elang2:
Both points are valid. MCP versions aren't backward compatible across the transport boundary (2024-11-05 used HTTP+SSE, 2025-03-26+ uses Streamable HTTP).

The problem this fixes is the other side of the same coin though. Hardcoding 2025-06-18 means the gateway advertises capabilities that a backend on 2024-11-05 can't deliver. A client that trusts the negotiated version tries to use features that silently fail. Downgrading is at least honest about what's available.

That said, the right behavior when backends are on incompatible versions is probably to reject the configuration rather than silently pick the lowest. I can add a log warning when the minimum backend version differs from the gateway's default, flagging it as a misconfiguration. Would that address your concern, or would you prefer the gateway to error outright when backends span a breaking version boundary?

aishwaryaraimule21:
> The problem this fixes is the other side of the same coin though. Hardcoding 2025-06-18 means the gateway advertises capabilities that a backend on 2024-11-05 can't deliver. A client that trusts the negotiated version tries to use features that silently fail. Downgrading is at least honest about what's available.

I agree with this.

> That said, the right behavior when backends are on incompatible versions is probably to reject the configuration rather than silently pick the lowest. I can add a log warning when the minimum backend version differs from the gateway's default, flagging it as a misconfiguration. Would that address your concern, or would you prefer the gateway to error outright when backends span a breaking version boundary?

I am in favour of downgrading over hardcoding a protocol version. Having said that, a MCPRoute can have MCP Server backends of different protocol versions. The merged protocol version can be lower than one or more MCP Server backends.
Let's not reject the configuration in this case. If the merged protocol version is lower than a MCP Server backend, worst case, the API will fail at the backend. We can however add a log warning if this happens.

internal/mcpproxy/handlers.go
```diff
@@ -609,6 +609,29 @@ func (m *mcpRequestContext) handleInitializeRequest(ctx context.Context, w http.
...
+			continue
+		}
+		if minVersion == "" || v < minVersion {
+			minVersion = v
+		}
+	}
+	if minVersion == "" {
+		minVersion = protocolVersion20250618
+	}
+	if clientVersion != "" && clientVersion < minVersion {
```
aishwaryaraimule21, line 648: Returning the min of all the backends makes sense. Can you help me understand why Gateway's support version be `min(backendVersions, clientVersion)`?

aishwaryaraimule21:
> This change looks good to me overall as this strategy is better than hardcoding a protocol version. However, I have the following concerns:
> 
> 1. This negotiation strategy implicitly assumes that MCP protocol versions are backward compatible, which is not the case. We need to call it out in a comment.
> 2. This might break existing MCP Clients connecting to AI Gateway.
>    e.g. when the client version is 2025-06-18 and one of the server backends is at a lower version 2024-11-05, the negotiated version returned would be 2024-11-05 which is different from what it returned earlier (2025-06-18).
> 3. One good thing about the earlier hardcoded approach was, Client <> Gateway and Gateway <> Server shared the same protocol version. After this change, e.g. even if the negotiated version is 2025-11-25, the request will be forwarded with the 2025-06-18 payload, resulting in newer fields getting dropped. This should get solved after we pick up the change mentioned in the PR description.
> 
> I am skeptical of merging just this PR without the above change for this month's release. Should we also take up the change in point 3 and have both as part of the next monthly release?


@nacx what do you think?

elang2:
Got it, log warning and no config rejection. I'll add it in `mergedProtocolVersion` — single warning after computing the minimum when any backend reports a higher version, something like "backend X advertises 2025-06-18 but merged version negotiated to 2024-11-05; client may encounter unsupported features." Plus the backward-compat comment noting the transport break between 2024-11-05 and 2025-03-26.

My inclination is to ship the warning first since it's a strict improvement over the current behavior, but happy to bundle with payload translation if you'd prefer a single release. @nacx let me know.

internal/mcpproxy/handlers.go
aishwaryaraimule21: @elang2 can you take this?

elang2:
Done. Added the warning and backward-compat comment in fc3e652.

The warning fires once per session initialization when any backend gets downgraded. It lists the affected backends with their versions (e.g., "backend1(2025-03-26), backend2(2025-06-18)") and calls out the 2024-11-05 vs 2025-03-26+ transport boundary.

I went with structured logging fields (merged_version, downgraded_backends, note) so operators can filter or alert on specific patterns in centralized logging systems. The backend list is comma-joined rather than an array to keep it readable in basic log viewers while still being machine-parseable.

The comment above the method explains why this matters (transport incompatibility, feature degradation risk) without requiring doc lookup.

Let me know if you'd prefer different field names or a simpler message format.

nacx (commented):
Left a comment for a final discussion, but not a blocker. LGTM.

Hritik003 (commented):
@elang2 can you fix the conflicts?

internal/mcpproxy/handlers.go
```diff
@@ -622,6 +622,106 @@ func (m *mcpRequestContext) handleInitializeRequest(ctx context.Context, w http.
...
+		}
+	}
+
+	// No backend has reported a version: return the floor unconditionally.
+	// Preserves the pre-floor default behavior and matches the outbound
+	// header pin in session.go.
+	if backendMin == "" {
+		return protocolVersion20250618
+	}
+
```
Hritik003, line 665: will this even occur? given that every server is expected to throw a protocol version in the initialize response? 

ref - https://modelcontextprotocol.io/specification/2025-11-25/basic/lifecycle#initialization

internal/mcpproxy/handlers.go
```diff
@@ -622,6 +622,106 @@ func (m *mcpRequestContext) handleInitializeRequest(ctx context.Context, w http.
...
+// forces pre-floor clients to upgrade or disconnect.
+func (s *session) mergedProtocolVersion(clientVersion string) string {
+	// Minimum non-empty backend version. Empty means "no backend has
+	// reported a version yet."
+	var backendMin string
+	for _, entry := range s.perBackendSessions {
+		v := entry.protocolVersion
+		if v == "" {
+			continue
+		}
```
Hritik003, line 653: empty is not possible ig

internal/mcpproxy/handlers.go
```diff
@@ -622,6 +622,106 @@ func (m *mcpRequestContext) handleInitializeRequest(ctx context.Context, w http.
...
+			slog.String("floor", protocolVersion20250618),
+		}
+		if clientFloored {
+			attrs = append(attrs, slog.String("client_requested", clientVersion))
+		}
+		if len(flooredBackends) > 0 {
+			attrs = append(attrs, slog.String("floored_backends", strings.Join(flooredBackends, ", ")))
+		}
+		s.reqCtx.l.Warn("MCP version below floor: lifting to gateway-tested minimum", attrs...)
+	}
```
Hritik003, line 720: just trying to understand, why do we need this? we already logged warnings in the snippets above right?

Hritik003 (commented):
@elang2 Happy to take these changes as part of #2545 PR! let me know whatever works for you.

PR #2545, opened 2026-08-17
feat(mcp): stateless spec modern fanout routing and server discovery ( 28-07-2026 )

Hritik003:
**Description**

This PR adds phase0 support for modern fanout routing  as per
-  #2431 that are required for the new MCP stateless spec. It introduces a new modern request handling path for all the fan out APIs;
    -  server/discover,
    - tools/list,
    -  resources/list, 
    - resources/templates/list, 
    - prompts/list, 
validates modern protocol headers, fans out requests across backends, and aggregates responses into a single JSON-RPC result. It also extracts capability aggregation into a shared helper (unionServerCapabilities) so both stateful and stateless paths use the same capability-union behavior, and bumps github.com/modelcontextprotocol/go-sdk to v1.7.0 for modern MCP support.

**Related Issues/PRs**
- #2431 

**Dependent PR**
- #2518 

- Please focus review on fanout/merge behavior and protocol compliance in internal/mcpproxy/modern.go, especially:

   - modern header and version validation
   - partial-backend-failure handling during aggregation 
   - merged discover/list response 

- this PR also has some common code with #2518 which will be adjusted and rebased once the first PR goes in.

**Testing** 
Built and deployed the changes to local cluster, tested the functionality using MCP inspector.


internal/mcpproxy/modern.go
```diff
@@ -0,0 +1,699 @@
...
+
+	headerVersion := r.Header.Get(mcpProtocolVersionHeader)
+	if headerVersion == "" {
+		headerVersion = protocolVersion20260728
+	}
+	if !isSupportedVersion(headerVersion) {
+		errType = metrics.MCPErrorUnsupportedProtocolVersion
+		err = fmt.Errorf("unsupported protocol version: %s", headerVersion)
+		onErrorResponse(w, http.StatusBadRequest, fmt.Sprintf("unsupported protocol version: %s", headerVersion))
+		return
```
swiftdiaries, line 135: Since this PR depends on #2518 , should this handler consume the validated result from `detectClientEra` instead of independently validating or inferring the version?
Inline with that, what's the intended request flow after rebase? Is it ``servePOST` → `detectClientEra` → reject or dispatch a validated request → `serveModernPOST``?

Broader question: can we try and centralize validation to ensure it's consistent and there's one place to make changes/evolve it? Also, have cleaner separation of concerns.

internal/mcpproxy/modern.go
```diff
@@ -0,0 +1,699 @@
...
+	routeConfig, ok := m.routes[route]
+	if !ok {
+		onErrorResponse(w, http.StatusNotFound, "route not found")
+		return handlerResult{}, fmt.Errorf("%w: %s", errBackendNotFound, route)
+	}
+
+	var results []*mcp.DiscoverResult
+	for _, backend := range routeConfig.backends {
+		backendStartAt := time.Now()
+		result, err := m.discoverBackend(ctx, route, backend)
```
swiftdiaries, line 215: is it possible for a request to receive a backend (and it's tools/prompts/resources) that isn't authorized?
Can the backend selection be a separate step that produces the backend set consumed downstream (discovery, fanout)? Trying to see if we can clearly define the input interface here. I was thinking a validated request + authorized backend set for the request.

internal/mcpproxy/modern.go
```diff
@@ -0,0 +1,699 @@
...
+		backendMetrics.RecordMethodCount(ctx, req.Method, nil)
+		backendMetrics.RecordRequestDuration(ctx, backendStartAt, nil)
+		results = append(results, result)
+	}
+	if len(results) == 0 {
+		m.l.Error("server/discover failed for all backends", slog.String("route", route))
+		onErrorResponse(w, http.StatusInternalServerError, "failed to discover any backend")
+		return handlerResult{}, errors.New("failed to discover any backend")
+	}
+	merged := mergeDiscoverResults(results)
```
swiftdiaries, line 237: what happens when there are partial discovery results? when there are temporary failures on a backend?
could it create a different backend set for a similar request? trying to see what the best possible response should be when we have that ambiguity

internal/mcpproxy/modern.go
```diff
@@ -0,0 +1,699 @@
...
+// mergeSupportedProtocolVersionsFromBackends merges the supported protocol versions from all servers.
+func mergeSupportedProtocolVersionsFromBackends(results []*mcp.DiscoverResult) []string {
+	versions := make([]string, 0, len(results))
+	for _, r := range results {
+		if r == nil {
+			continue
+		}
+		for _, v := range r.SupportedVersions {
+			if !slices.Contains(versions, v) {
+				versions = append(versions, v)
```
swiftdiaries, line 314: should we intersect against the gateway's supported versions here? to prevent accidentally advertising versions the gateway doesn't support

internal/mcpproxy/modern.go
```diff
@@ -0,0 +1,699 @@
...
+		return nil, fmt.Errorf("parse response: %w (body: %.200s)", err, string(jsonPayload))
+	}
+	if errField, ok := rpcResp["error"]; ok && string(errField) != "null" {
+		return nil, fmt.Errorf("backend error: %s", string(errField))
+	}
+	result, ok := rpcResp["result"]
+	if !ok || string(result) == "null" {
+		return nil, fmt.Errorf("backend returned no result")
+	}
+	return result, nil
```
swiftdiaries: should we add some validation logic here?
some to things to validate:
- The payload is a valid JSON-RPC 2.0 response.
- The response ID matches the upstream request ID.
- Exactly one of `result` or `error` is present.
- Required modern fields are present and valid, including `resultType`.
- A malformed response is recorded as a backend failure and is not included in the aggregated result.

swiftdiaries (commented):
Thanks for the context and the work here! Areas I focused on for the review:
- request validation at ingress
- backend selection logic
- discovery logic
- response handling

P.S Tried my best with the small context I have of the project and the code. So feel free to disregard things if they are not relevant

internal/mcpproxy/modern.go
Hritik003: For initialize in the previous spec, this was the standard we followed during fanout;
-  if one of the backends failed, we don't return any error and pass the request if atleast one of the underlying backend returned a 200.
- and only throw the error, if all underlying backends fail to establish session. And I think it makes sense in a multiplexing scenario
- Ref: https://github.com/envoyproxy/ai-gateway/blob/79770ba7777caaa2870fa0c8cd0eeb1a2aa08ba5/internal/mcpproxy/mcpproxy.go#L224

internal/mcpproxy/modern.go
Hritik003: Good point.
planning to not have any hardcoded logic for showing supported versions. Was thinking to directly return the merged list of supported versions from all the backends beneath. This PR #2543 handles that logic, and would want to use the same for showing supported versions.

internal/mcpproxy/modern.go
Hritik003: Out of all the validations suggested, i think we can up few, 

but validations like `valid JSON-RPC 2.0` does not make sense for a proxy. Because once we are able to pull up `result` from the response, these extra checks do not change aggregation behavior

>Exactly one of result or error is present.

Also for this, we anyways inject the `resultType` in the ensureResultType function on the outbound response. `resultType` is only mattered during MRTR cases like `input_required`, but given that we are deferring those changes in this phase, lets keep it simple?

internal/mcpproxy/modern.go
Hritik003: >Since this PR depends on https://github.com/envoyproxy/ai-gateway/pull/2518 , should this handler consume the validated result from detectClientEra instead of independently validating or inferring the version?

Right, I think we should avoid these version checks here, and `detectClientEra` function in the previous PR. Nevertheless, all the necessary checks are already present in #2518 itself.

>Inline with that, what's the intended request flow after rebase? Is it servePOST` → `detectClientEra` → reject or dispatch a validated request → `serveModernPOST?

yes

>Broader question: can we try and centralize validation to ensure it's consistent and there's one place to make changes/evolve it? Also, have cleaner separation of concerns.

the `detectClientEra` function in the previous PR takes care of that

internal/mcpproxy/mcpproxy.go
```diff
@@ -322,6 +323,139 @@ func (m *mcpRequestContext) sessionFromID(id secureClientToGatewaySessionID, las
...
+func mergedProtocolVersion(l *slog.Logger, clientVersion string, backends []backendReportedVersions) string {
+	const floor = protocolVersion20250618
+
+	// Compute the minimum, across backends, of each backend's best (highest)
+	// supported version. Backends that reported nothing are ignored: every
+	// spec-compliant server MUST return a protocolVersion in its initialize
+	// response, so an empty entry means we failed to learn it, not that the
+	// backend supports "no version".
+	var backendMin string
+	var downgradedBackends []string // backends advertising newer than the merged min.
```
Hritik003, line 385: these changes are have been taken from #2543 and are required for modern flow as well. Depending on whether @elang2 wishes to close that PR or have it merged independently,  will rebase accordingly.

internal/mcpproxy/modern.go
```diff
@@ -0,0 +1,725 @@
...
+			if errors.As(err, &errToolCall) {
+				applicationError = true
+			}
+			if applicationError {
+				metricsInstance.RecordMethodErrorCount(ctx, req.Method, nil, metrics.MCPStatusFailed)
+			} else {
+				metricsInstance.RecordMethodErrorCount(ctx, req.Method, nil, metrics.MCPStatusError)
+			}
+			metricsInstance.RecordRequestErrorDuration(ctx, startAt, errType, nil)
+			return
```
Hritik003: we can make use of the recordPOSTCompletion we have

internal/mcpproxy/modern.go
```diff
@@ -0,0 +1,725 @@
...
+		return
+	}
+
+	headerVersion := r.Header.Get(mcpProtocolVersionHeader)
+	if headerVersion == "" {
+		headerVersion = protocolVersion20260728
+	}
+	if !isSupportedVersion(headerVersion) {
+		errType = metrics.MCPErrorUnsupportedProtocolVersion
+		err = fmt.Errorf("unsupported protocol version: %s", headerVersion)
```
Hritik003: can this even occur given era.go file takes care of it?

internal/mcpproxy/modern.go
```diff
@@ -0,0 +1,725 @@
...
+	if params := modernParamsForHeaderMetadata(req); params != nil {
+		span = m.tracer.StartSpanAndInjectMeta(ctx, req, params, r.Header)
+	}
+	defer func() {
+		if span == nil {
+			return
+		}
+		if err != nil {
+			span.EndSpanOnError(string(errType), err)
+		} else {
```
Hritik003: can remove? since it is taken care in recordPOSTCompletion function?

internal/mcpproxy/modern.go
```diff
@@ -0,0 +1,725 @@
...
+	}
+	if len(results) == 0 {
+		m.l.Error("server/discover failed for all backends", slog.String("route", route))
+		onErrorResponse(w, http.StatusInternalServerError, "failed to discover any backend")
+		return handlerResult{}, errors.New("failed to discover any backend")
+	}
+	merged := mergeDiscoverResults(m.l, results)
+	merged.Instructions = fmt.Sprintf("Envoy AI Gateway — MCP proxy aggregating %d backends", len(routeConfig.backends))
+	writeJSONRPCResult(w, req.ID, merged)
+	return handlerResult{}, nil
```
Hritik003, line 251: do we need this instructions?

internal/mcpproxy/modern.go
Hritik003: > is it possible for a request to receive a backend (and it's tools/prompts/resources) that isn't authorized? Can the backend selection be a separate step that produces the backend set consumed downstream (discovery, fanout)? Trying to see if we can clearly define the input interface here. I was thinking a validated request + authorized backend set for the request.

Agree, earlier on initialize we used to call newSession() which used to set the selected authorized backends, and tools/list prompts/list.... used to iterate through the selected backends returned from newSession(). Since in modern, we dont have any initialize and every request is stateless, reevaluating every request is actually the right stateless behavior, a new token can change which backends are visible. Legacy cannot do that mid session because the set is fixed at initialize.

internal/mcpproxy/modern.go
```diff
@@ -0,0 +1,725 @@
...
+	}
+
+	return &result, nil
+}
+
+func discoverParams() []byte {
+	return []byte(`{"_meta":{` +
+		`"` + metaProtocolVersion + `":"` + protocolVersion20260728 + `",` +
+		`"` + metaClientInfo + `":{"name":"envoy-ai-gateway","version":"1.0.0"},` +
+		`"` + metaClientCapabilities + `":{}` +
```
Hritik003: change the client name here to `agentrouter`?

internal/mcpproxy/modern.go
```diff
@@ -0,0 +1,695 @@
...
+// This is a fan-out handler: it records per-backend metrics itself and sets
+// perBackendMetricsRecorded so the generic recording in serveModernPOST is
+// skipped.
+func (m *mcpRequestContext) handleServerDiscover(ctx context.Context, w http.ResponseWriter, req *jsonrpc.Request, route filterapi.MCPRouteName, span tracingapi.MCPSpan) (handlerResult, error) {
+	m.perBackendMetricsRecorded = true
+	_, selectedBackends, err := m.resolveModernRouteBackends(w, route)
+	if err != nil {
+		return handlerResult{}, err
+	}
+
```
Hritik003, line 222: since we are calling it at every handler, can we call it at some top level?

internal/mcpproxy/handlers.go
```diff
@@ -391,55 +391,53 @@ func (m *mcpRequestContext) mergeToolsList(s *session, responses []broadCastResp
...
-				allowed, _ := m.authorizeRequest(route.authorization, &authorizationRequest{
-					Headers:   m.requestHeaders,
-					MCPMethod: "tools/call",
-					Backend:   r.backendName,
-					Tool:      tool.Name,
-				})
-				if !allowed {
+	if route != nil {
+		// Aggregate tools from all backends. Per-backend PrefixMode controls whether a backend's
+		// tools are prefixed with "<backendName>__" (Always) or exposed as bare names (Never).
```
Hritik003: fyi:
The reason this is being change from `if route == nil { return }` into `if route != nil { ... }` is for caching hints to be still applied on the missing route path

internal/mcpproxy/session.go
```diff
@@ -648,9 +655,45 @@ func decodeCapabilityFlags(hex string) *mcpsdk.ServerCapabilities {
 // If ANY backend supports a capability, the merged result includes it.
 // Sub-fields like ListChanged and Subscribe are OR'd across all backends.
 func (s *session) mergedCapabilities() *mcpsdk.ServerCapabilities {
```
aish1331, line 652: do you need the separate helper method `unionServerCapabilities` for this?

internal/mcpproxy/modern.go
```diff
@@ -0,0 +1,694 @@
...
+			errType: errType,
+			err:     err,
+			startAt: startAt,
+			span:    span,
+			result:  result,
+		})
+	}()
+
+	route := r.Header.Get(internalapi.MCPRouteHeader)
+	if route == "" {
```
aish1331, line 74: why do we need this?

internal/mcpproxy/modern.go
```diff
@@ -0,0 +1,694 @@
...
+		errType = metrics.MCPErrorUnsupportedMethod
+		err = errors.New("logging/setLevel removed in 2026-07-28")
+		onErrorResponse(w, http.StatusNotFound, "logging/setLevel removed in 2026-07-28; use _meta logLevel")
+		return
+	}
+
+	// Start a tracing span for the request, mirroring the legacy path. The span
+	// is closed in recordPOSTCompletion. Fan-out handlers additionally record
+	// per-backend routing on the span via RecordRouteToBackend.
+	if params := modernParamsForHeaderMetadata(req); params != nil {
```
aish1331: start span only for supportedMethod. Similar to `legacy.go`?

internal/mcpproxy/modern.go
```diff
@@ -0,0 +1,694 @@
...
+	if errType == "" {
+		errType = errorType(err)
+	}
+}
+
+// resolveModernRouteBackends looks up the route and evaluates backendSelector
+// against this request. The returned map is the only backend set discovery and
+// list fan-out may talk to. Writes 404 (unknown route) or 403 (no matching
+// backends) on failure.
+func (m *mcpRequestContext) resolveModernRouteBackends(w http.ResponseWriter, route filterapi.MCPRouteName) (*mcpProxyConfigRoute, map[filterapi.MCPBackendName]filterapi.MCPBackend, error) {
```
aish1331, line 188: `forwardHeader` extraction before authorizing against backend selector?

aish1331: I think you also need to extract per backend `forwardHeader`. Can you please compare this with legacy flow to find what's missing and fix it?

internal/mcpproxy/modern.go
```diff
@@ -0,0 +1,694 @@
...
+	var results []*mcp.DiscoverResult
+	for _, backend := range selectedBackends {
+		backendStartAt := time.Now()
+		result, err := m.discoverBackend(ctx, route, backend)
+		backendMetrics := m.metrics.WithBackend(backend.Name)
+		if err != nil {
+			m.l.Warn("server/discover failed for backend",
+				slog.String("backend", backend.Name),
+				slog.String("error", err.Error()))
+			backendMetrics.RecordMethodErrorCount(ctx, req.Method, nil, metrics.MCPStatusError)
```
aish1331, line 232: is this already being recorded in `recordPOSTCompletion`?

Hritik003: Nope, we set `perBackendMetrics=true` which ends the span when it goes to defer block. So its not a duplicate

internal/mcpproxy/modern.go
Hritik003: so in legacy, every session remembers which route it points to inside its session id. But its not the case with modern cause it being stateless, and if we skip this check, a missing header would look `route <name> not found` and return 404 masking the gateway bug

internal/mcpproxy/modern.go
```diff
@@ -0,0 +1,773 @@
...
+		if json.Unmarshal(params, &p) == nil {
+			return p.URI
+		}
+	}
+	return ""
+}
+
+// modernParamsForHeaderMetadata best-effort parses params for methods where
+// addMCPHeaders can enrich upstream metadata (tool name/resource URI).
+func modernParamsForHeaderMetadata(req *jsonrpc.Request) mcp.Params {
```
aish1331, line 630: can this be reused from legacy?

internal/mcpproxy/modern.go
```diff
@@ -0,0 +1,773 @@
...
+	case "resources/unsubscribe":
+		var p mcp.UnsubscribeParams
+		if json.Unmarshal(req.Params, &p) == nil {
+			return &p
+		}
+	}
+	return nil
+}
+
+func writeJSONRPCResult(w http.ResponseWriter, id jsonrpc.ID, result any) {
```
aish1331, line 659: same, can this be reused from legacy?

internal/mcpproxy/modern.go
Hritik003: those helpers aren’t in legacy

internal/mcpproxy/modern.go
Hritik003: those helpers aren’t in legacy

aish1331 (approved):
LGTM, thanks!!

johnugeorge (approved):

Issue #682, opened 2025-06-05
Remove content-encoding: gzip after decompressing the response body


sukumargaonkar:
*Description*:
>What issue is being seen? Describe what should be happening instead of
the bug, for example: Envoy should not crash, the expected value isn't
returned, etc.

Extrproc check response encoding header in the response from upstream and decompresses the response body if response-encoding it set to gzip. see [here](https://github.com/envoyproxy/ai-gateway/blob/21810cf8e29b429c88cdd3d3610866505002c088/internal/extproc/chatcompletion_processor.go#L260)

Since the extproc doesn't explicitly compress the response body the client get an uncompressed response but with response-encoding header set to 'gzip'.

we should remove the response-encoding header when we decompress the response-body in extproc.

*Repro steps*:
> Include sample requests, environment, etc. All data and inputs
required to reproduce the bug.

set up testupstream to return a gzip response and monitor response header on client (like curl)

>**Note**: If there are privacy concerns, sanitize the data prior to
sharing.

*Environment*:
>Include the environment like gateway version, envoy version and so on.

*Logs*:
>Include the access logs and the Envoy logs.

mathetake:
closed by #683 (for now)

PR #2554, opened 2026-08-19
fix: remove stale content-encoding header on translated error responses

### LINKED ISSUE #2553: ext_proc leaves stale content-encoding: gzip on translated error responses (error path misses removeContentEncodingIfNeeded)
_Description_:

When the upstream returns a compressed (`content-encoding: gzip`) **error** response, the ext_proc decompresses the body, translates it to the OpenAI error format, and replaces the response body — but does not remove the `content-encoding` header. The client then receives an uncompressed JSON body labeled `content-encoding: gzip`.

Consequences:
- Browsers fail the whole response with `ERR_CONTENT_DECODING_FAILED`, so the error body is unreadable exactly when users need it for debugging.
- SDKs mask the real error, e.g. `@ai-sdk/openai-compatible` surfaces only "Failed to process error response".

This is the error-path sibling of #682. That fix is applied on the success path via `removeContentEncodingIfNeeded` (https://github.com/envoyproxy/ai-gateway/blob/85746d5b41513b014231aa1d88739c520e51a171/internal/extproc/processor_impl.go#L603), but the non-2xx branch in `ProcessResponseBody` (https://github.com/envoyproxy/ai-gateway/blob/85746d5b41513b014231aa1d88739c520e51a171/internal/extproc/processor_impl.go#L567-L592) returns early after `translator.ResponseError` without calling it. The translators' `ResponseError` implementations set only `content-type` and `content-length` (e.g. `convertGCPVertexAIErrorToOpenAI`), so the stale `content-encoding` from the upstream response survives.

Expected: since `ResponseError` always produces an uncompressed body, the error path should remove `content-encoding` whenever the original body was encoded, same as the success path.

Happy to send a PR — the fix is adding the same `removeContentEncodingIfNeeded` call to the error branch.

_Repro steps_:

1. Set up an `AIGatewayRoute` with a `GCPVertexAI` backend (any backend whose upstream compresses error responses works; observed with Vertex AI Gemini).
2. Send a `/v1/chat/completions` request with `Accept-Encoding: gzip` that triggers an upstream 4xx. An easy trigger for Vertex: a `response_format` JSON schema complex enough to exceed Google's constrained-decoding limit (returns 400 INVALID_ARGUMENT "The specified schema produces a constraint that has too many states for serving").
3. Observe the response: status 400, `content-encoding: gzip`, but the body is plain (uncompressed) translated JSON:

```
$ curl --compressed https://<gateway>/v1/chat/completions -d @too-complex-schema.json ...
curl: (61) Error while processing content unencoding: incorrect header check
```

4. The same request with `Accept-Encoding: identity` returns readable JSON (upstream doesn't compress, so no stale header):

```
{"type":"error","error":{"type":"INVALID_ARGUMENT","code":"400","message":"Request contains an invalid argument."}}
```

Successful (2xx) responses are unaffected — the success path removes the header correctly.

_Environment_:

Observed on Envoy AI Gateway v1.0.0 (Envoy Gateway v1.7.4, GKE). Code inspection shows the bug is still present on current `main` (85746d5).

_Logs_:

n/a — reproducible with the curl output above; no crash, nothing relevant in Envoy logs (the mismatch is purely between the mutated body and the untouched header).

dragon3:
**Description**

When the upstream returns a compressed (`content-encoding: gzip`) error response, the non-2xx branch of `ProcessResponseBody` decompresses the body and replaces it with the uncompressed translated OpenAI-format error, but leaves the `content-encoding` header intact. Clients then fail to decode the response: browsers reject it with `ERR_CONTENT_DECODING_FAILED`, and SDKs such as `@ai-sdk/openai-compatible` mask the real error as "Failed to process error response" — exactly when the body is needed for debugging.

The success path already handles this via `removeContentEncodingIfNeeded` (added for #682); this change applies the same call to the error path, and adds a regression test that fails without the fix.

Fixes #2553

Per the generative AI policy: this change was developed with AI assistance (Claude); I have reviewed it and take full ownership.

### DIFF
--- a/internal/extproc/processor_impl.go
+++ b/internal/extproc/processor_impl.go
@@ -572,6 +572,8 @@ func (u *upstreamProcessor[ReqT, RespT, RespChunkT, EndpointSpecT]) ProcessRespo
 			return nil, fmt.Errorf("failed to transform response error: %w", err)
 		}
 		headerMutation, bodyMutation := mutationsFromTranslationResult(newHeaders, newBody)
+		// Remove content-encoding header if original body encoded but was mutated in the processor.
+		headerMutation = removeContentEncodingIfNeeded(headerMutation, bodyMutation, decodingResult.isEncoded)
 		if u.parent.span != nil {
 			b := bodyMutation.GetBody()
 			if b == nil {
--- a/internal/extproc/processor_impl_test.go
+++ b/internal/extproc/processor_impl_test.go
@@ -7,6 +7,7 @@ package extproc
 
 import (
 	"bytes"
+	"compress/gzip"
 	"context"
 	"errors"
 	"fmt"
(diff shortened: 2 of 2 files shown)


PatilHrushikesh:
LGTM
CC @siddharth1036

siddharth1036 (approved):
LGTM

mathetake (approved):

PR #2556, opened 2026-08-19
fix(controller): scope MCP OAuth routes by headers

### LINKED ISSUE #2544: bug(mcp): MCPRoute headers are not applied to the generated OAuth well-known HTTPRoute rules
_Description_:

`MCPRoute.spec.headers` is documented as "HTTP headers that must match for this route to be selected", but the controller only attaches it to the MCP proxy rule. The three OAuth discovery rules generated on the same HTTPRoute are built with path-only matches, so the header condition is silently dropped for every `/.well-known/*` endpoint.

The result is an asymmetric route: `/mcp` is header-scoped as configured, while `/.well-known/oauth-protected-resource/mcp`, `/.well-known/oauth-authorization-server/mcp` and `/.well-known/openid-configuration/mcp` match unconditionally. In cases when I want to route based on the Host header, Oauth is not functional, as it would return the wrong metadata. 

A workaround is to manually create an additional `HTTPRoute` object with correct rules for `/.well-known/*`, which would then take precedence over the Envoy GW-generated `HTTPRoute`.

_Repro steps_:

1. Create an MCPRoute with `.spec.headers` and `.spec.securityPolicy` defined.
```yaml
kind: MCPRoute
metadata:
  name: my-route
spec:
  path: "/mcp"
  headers:
    - name: "Host"
      value: "my-host.org"
  parentRefs:
      kind: Gateway
      name: ai-gateway
      namespace: ai-gateway
      sectionName: https
  backendRefs:
      kind: Backend
      name: my-backend
      path: "/mcp"
  securityPolicy:
    oauth:
     # config here
```
2. Check the HTTPRoute object, and specifically pay attention to the `.well-known` paths
```yaml
kind: HTTPRoute
metadata:
  name: ai-eg-mcp-main-my-route
spec:
  parentRefs:
  # trimmed, focus is on the rules
  rules:
  - backendRefs:
      kind: Backend
      name: my-backend
      namespace: ai-gateway
      port: 9856
      weight: 1
    filters:
    - requestHeaderModifier:
        set:
        - name: x-ai-eg-mcp-route
          value: ai-gateway/my-route
      type: RequestHeaderModifier
    matches:
    - headers: 
      - name: Host # << header only present on the /mcp path
        type: Exact
        value: my-host.org
      path:
        type: Exact
        value: /mcp
    timeouts:
      backendRequest: 30m
      request: 30m
  - filters:
    - extensionRef:
        group: gateway.envoyproxy.io
        kind: HTTPRouteFilter
        name: ai-eg-mcp-my-route-oauth-protected-resource-metadata
      type: ExtensionRef
    matches:
    - path:
        type: Exact
        value: /.well-known/oauth-protected-resource/mcp
    name: oauth-protected-resource-metadata
  - filters:
    - extensionRef:
        group: gateway.envoyproxy.io
        kind: HTTPRouteFilter
        name: ai-eg-mcp-my-route-oauth-authorization-server-metadata
      type: ExtensionRef
    matches:
    - path:
        type: Exact
        value: /.well-known/oauth-authorization-server/mcp
    name: oauth-authorization-server-metadata
  - filters:
    - extensionRef:
        group: gateway.envoyproxy.io
        kind: HTTPRouteFilter
        name: ai-eg-mcp-my-route-oauth-authorization-server-metadata
      type: ExtensionRef
    matches:
    - path:
        type: Exact
        value: /.well-known/openid-configuration/mcp
    name: oauth-authorization-server-metadata-oidc
```
3. `Host` header is only present on the `/mcp` route, it is not applied to the ` /.well-known/*` routes

_Environment_:

Envoy AI Gateway version 1.0

_Logs_:

N/A

0YHR0:
**Description**

`MCPRoute.spec.headers` currently scopes the generated MCP proxy rule, but the three OAuth metadata rules generated on the same `HTTPRoute` only match by path. This allows requests that do not satisfy the MCPRoute's header constraints to reach:

- `/.well-known/oauth-protected-resource{path}`
- `/.well-known/oauth-authorization-server{path}`
- `/.well-known/openid-configuration{path}`

This change applies the MCPRoute header matches to all three OAuth rules so the MCP endpoint and its discovery metadata share the same routing scope. The existing OAuth route test now verifies every generated metadata rule carries those headers.

Fixes #2544

Validation:
- `go test ./internal/controller -run 'Test_newHTTPRoute_MCP_(PathAndBackendsAndMetadata|Oauth)' -count=1`
- `go test ./internal/controller -count=1`
- `go vet ./internal/controller`
- `git diff --check`


0YHR0:
@mathetake This is a focused controller fix for #2544: the three generated OAuth metadata rules now inherit the same `MCPRoute.spec.headers` matches as the MCP endpoint. The controller package tests and vet pass, and the PR adds a regression assertion for every generated well-known rule. Could you please review it when CI is green?

nacx (approved):

PR #2561, opened 2026-08-19
fix: propagate QuotaPolicy changes to derived HTTPRoute

### LINKED ISSUE #2550: QuotaPolicy changes silently stop taking effect until the derived HTTPRoute is manually recreated
## Description

Editing a QuotaPolicy (bucketRules, defaultBucket, or serviceQuota) updates the rate-limit service's own config correctly (visible via /rlconfig), but the Envoy route's rate_limits actions, generated via the extension hook in internal/extensionserver/quota_ratelimit.go, silently stop updating. No error anywhere.

**Expected**: a QuotaPolicy change should reliably force Envoy Gateway to re-translate the affected route, without manual intervention.

**Root cause**: `QuotaPolicyController.notifyAIGatewayRoutes` (internal/controller/quota_policy.go) re-syncs the derived HTTPRoute on a QuotaPolicy change, but `newHTTPRoute()` builds that HTTPRoute purely from the AIGatewayRoute object, QuotaPolicy content never touches it. The resulting `client.Update()` is byte-identical, bumping `resourceVersion` but not `metadata.generation`, and labels/annotations don't change either.

Envoy Gateway's own HTTPRoute watch predicate (internal/provider/kubernetes/predicates.go, `metadataPredicate`) is `GenerationChanged OR LabelChanged OR AnnotationChanged`. None of these fire, so Envoy Gateway skips re-translating the route, and the extension hook that builds rate_limits never re-runs. Restarting ai-gateway-controller doesn't help, a fresh reconcile produces the same no-op update.

**Suggested fix**: have the QuotaPolicy-triggered HTTPRoute update bump something the predicate actually observes, e.g. a controller-managed annotation hashing the relevant QuotaPolicy state.

## Repro steps

1. Apply a QuotaPolicy with bucketRules targeting a route, confirm /rlconfig reflects it.
2. Check the route's `typed_per_filter_config` for `envoy.filters.http.ratelimit/ai-gateway-quota` (Envoy admin `/config_dump`), it still reflects the previous shape.
3. Restart ai-gateway-controller. No change.
4. Delete the AIGatewayRoute itself (this cascades to delete the owned HTTPRoute via its owner reference). Once the AIGatewayRoute is recreated, the derived HTTPRoute goes through `Create()` again instead of `Update()`, and the route immediately picks up the correct config.

## Environment

ai-gateway v1.0.0, Envoy Gateway v1.8.3, Kubernetes.

## Logs

Controller logs during step 3 show the reconcile firing with no effective change:

```
INFO  controller.quota-policy    Reconciling QuotaPolicy
INFO  controller.quota-policy    notifying AIGatewayRoute of QuotaPolicy change
INFO  controller.ai-gateway-route  Reconciling AIGatewayRoute
INFO  controller.ai-gateway-route  syncing AIGatewayRoute
INFO  controller.ai-gateway-route  updating HTTPRoute
```

In the same pass, the ratelimit service's own debug log confirms it reloaded the current bucketRules descriptor tree correctly, only the Envoy route side stayed stale.

BetterAndBetterII:
**Description**

QuotaPolicy edits requeue the AIGatewayRoute, but the derived HTTPRoute is rebuilt only from that object. Envoy Gateway never sees a generation, label, or annotation change, so quota rate_limits stay stale until the HTTPRoute is recreated by hand.

newHTTPRoute now lists QuotaPolicies that target the route's AIServiceBackends and stamps a deterministic spec hash on aigateway.envoyproxy.io/quota-policy, same idea as gateway.envoyproxy.io/backend-ref-priority. A QuotaPolicy spec change updates that annotation so Envoy Gateway reconciles.

Controller tests cover a QuotaPolicy update requeuing the AIGatewayRoute and changing the derived HTTPRoute annotation.

Fixes #2550


internal/controller/ai_gateway_route.go
```diff
@@ -475,3 +492,56 @@ func buildPriorityAnnotation(rules []aigv1b1.AIGatewayRouteRule) string {
...
+		p := &policies[i]
+		raw, err := stdjson.Marshal(p.Spec)
+		if err != nil {
+			parts = append(parts, fmt.Sprintf("%s/%s:%d", p.Namespace, p.Name, p.Generation))
+			continue
+		}
+		sum := sha256.Sum256(raw)
+		parts = append(parts, fmt.Sprintf("%s/%s:%s", p.Namespace, p.Name, hex.EncodeToString(sum[:8])))
+	}
+	return strings.Join(parts, ",")
```
siddharth1036, line 546: Just thinking should we hash this as well?

vignesh-chaturvedi (commented):
I was in the quota code recently for #2664, so I took a look at this. The
approach matches what the issue suggests and the mechanics check out: the
annotation changes whenever a targeting policy's spec changes, which trips
Envoy Gateway's `AnnotationChanged` predicate. Two things.

**Deletion misses routes in other namespaces.** On update,
`notifyAIGatewayRoutes` finds routes through the
`BackendToReferencingAIGatewayRoute` index, which is keyed on the backend's
namespace, so it reaches routes in any namespace that reference the backend
through a ReferenceGrant. On deletion, `notifyAllAIGatewayRoutesInNamespace`
lists routes with `client.InNamespace(policy.Namespace)` instead. A route in
namespace B that references a backend in namespace A is never re-reconciled
when a QuotaPolicy in A is deleted, so it keeps the deleted policy's entry in
its annotation, and Envoy Gateway keeps serving the old `rate_limits` until
something else re-translates the route.

The gap predates this PR, but this PR is what makes that notification
load-bearing. One fix is to have the deletion path list the AIServiceBackends
in the policy's namespace and notify through the same index the update path
uses. That change lives in `quota_policy.go` rather than anything touched
here, so I am happy to send it as a separate small PR if that is easier than
widening this one.

**On the hash question at line 546:** I think hashing the joined string makes
sense. The annotation currently grows with every targeting policy, and since
the value only needs to change rather than be readable, a single fixed-length
digest over the sorted parts keeps it constant size. Keeping each policy's
`namespace/name` in the hashed input, as now, means a rename or a retarget
still changes the digest.

PR #2569, opened 2026-08-21
fix: support parameterized JSON content types

### LINKED ISSUE #2568: MCP proxy silently drops backend JSON responses when Content-Type carries a charset parameter (exact string match in session.go)
### Summary

The MCP proxy's per-backend response collector decides how to parse a backend response with an **exact string comparison** on the `Content-Type` header:

https://github.com/envoyproxy/ai-gateway/blob/v1.0.0/internal/mcpproxy/session.go#L450

```go
if httpResp.Header.Get("Content-Type") == "application/json" {
```

A backend that responds `Content-Type: application/json;charset=UTF-8` — the default output of Java servlet containers when the servlet sets an explicit character encoding, which the official **MCP Java SDK** (`HttpServletStatelessServerTransport`, mcp-core 2.0.0) does (`setContentType("application/json")` followed by `setCharacterEncoding("UTF-8")`) — fails this check. The body then falls through to the SSE parser, which finds no `data:` lines in a plain JSON body, yields **zero events and zero errors**. Net effect: the backend's `tools/list` response is silently discarded, its tools never appear in the aggregated tool list, and nothing is logged.

Notably, the **session-initialize path uses the opposite (lenient) dispatch** — `mcpproxy.go` treats anything `!= "text/event-stream"` as JSON — so `initialize` against the same backend succeeds and the backend looks healthy. Only tool collection is lost, which makes this very hard to diagnose: no error at the client, no error in the extproc log, session established.

Still present on `main` as of 2026-08-20.

### Version

- AI Gateway v1.0.0 (helm), Envoy Gateway v1.8.1, also reproduced reading `main`'s source.

### Repro

1. MCPRoute with two backends: one Spring-AI-based MCP server (responds over SSE) and one built on the MCP Java SDK's `HttpServletStatelessServerTransport` (responds `application/json;charset=UTF-8` — servlet-container default).
2. `initialize` through the gateway: succeeds, no errors logged for either backend.
3. `tools/list` through the gateway: only the SSE backend's tools are returned. The servlet backend's tools are absent; the extproc log contains no entries for it.
4. Direct `tools/list` against the servlet backend (same token/headers): returns its tools, with `Content-Type: application/json;charset=UTF-8`.

### Expected

Content-type dispatch should parse the media type rather than compare strings — e.g. `mime.ParseMediaType` and match on the `application/json` media type, ignoring parameters like `charset` (per RFC 9110 §8.3; note `charset` is even meaningless for `application/json` per RFC 8259 §8.1, and servlet containers append it routinely). At minimum, a non-empty 200 body that produced zero parsed messages should log a warning instead of being dropped silently.

Related: #2218 is the mirror image of the same dispatch fragility (body labeled `text/event-stream` but containing unframed JSON, silently dropped by the same code path). A shared fix could be a robustness cascade: parse per declared media type first, then fall back to the other parser when the first yields zero messages from a non-empty body.

### Impact

Any MCP backend implemented on the official MCP Java SDK servlet transports (or any stack that appends `;charset=` to JSON responses) aggregates as connected-but-toolless, with no diagnostic anywhere. We hit this fronting four internal MCP servers: the three Spring-AI ones aggregated fine, the one raw-Java-SDK server vanished silently.

### Workaround

A servlet response filter on the backend that suppresses the charset parameter so the header is exactly `application/json` (swallow `setCharacterEncoding`, hand the servlet a UTF-8 writer over the raw output stream) restores aggregation.

Karthik-Chowdary:
**Description**

Parse MCP response media types instead of comparing the raw `Content-Type` header. This recognizes valid parameterized and case-varied `application/json` values across all MCP proxy response paths, rejects malformed media types, and preserves non-JSON/SSE handling.

Tests cover plain, parameterized, case-varied, malformed, missing, and SSE media types. Local verification passed with `go test ./internal/mcpproxy -count=1` and `make precommit`.

Fixes #2568

Hermes Agent was used to help investigate the issue, draft the implementation and tests, and review the change. I reviewed and understand the submitted code and take responsibility for it.

### DIFF
--- a/internal/mcpproxy/content_type.go
+++ b/internal/mcpproxy/content_type.go
@@ -1,0 +1,17 @@
+// Copyright Envoy AI Gateway Authors
+// SPDX-License-Identifier: Apache-2.0
+// The full text of the Apache license is available in the LICENSE file at
+// the root of the repo.
+
+package mcpproxy
+
+import (
+	"mime"
+	"strings"
+)
+
+// isJSONMediaType reports whether contentType is a valid application/json media type.
+func isJSONMediaType(contentType string) bool {
+	mediaType, _, err := mime.ParseMediaType(contentType)
+	return err == nil && strings.EqualFold(mediaType, "application/json")
+}
--- a/internal/mcpproxy/content_type_test.go
+++ b/internal/mcpproxy/content_type_test.go
@@ -1,0 +1,33 @@
+// Copyright Envoy AI Gateway Authors
+// SPDX-License-Identifier: Apache-2.0
+// The full text of the Apache license is available in the LICENSE file at
+// the root of the repo.
+
+package mcpproxy
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestIsJSONMediaType(t *testing.T) {
+	tests := []struct {
+		name        string
+		contentType string
+		want        bool
+	}{
+		{name: "plain", contentType: "application/json", want: true},
+		{name: "parameterized", contentType: "application/json; charset=UTF-8", want: true},
+		{name: "case variation", contentType: "Application/JSON; Charset=utf-8", want: true},
+		{name: "malformed", contentType: `application/json; charset="unterminated`, want: false},
+		{name: "missing", contentType: "", want: false},
+		{name: "non JSON", contentType: "text/event-stream", want: false},
+	}
+
+	for _, tt := range tests {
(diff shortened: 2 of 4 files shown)


Karthik-Chowdary:
Closing in favor of the earlier #2306, which already implements parameter-aware JSON media-type handling across these MCP proxy paths. I missed that existing PR during the duplication check—apologies for the noise.

PR #2576, opened 2026-08-24
feat(gcp): support pass-through credentials for GCP backends

DanielChrn:
**Description**
As part of our internal AI-gateway integration, we want to be able to take Vertex auth from the client and bypass it to the backend.
The current controller setup does not allow it because:

It always overwrites the path into something that does not exist (part of the full Vertex URL.)
Client does not have a way to introduce region/project in the request.
Because of that we want to introduce 2 changes.

Move path header translated from the GCP request to gcp-path .
Allow bypassing the region/project form client via gcp-project and gcp-region headers


internal/backendauth/gcp.go
```diff
@@ -91,8 +123,9 @@ func (g *gcpHandler) Do(_ context.Context, requestHeaders map[string]string, _ [
 		accessToken = g.gcpAccessToken
 	}
 
+	resultHeaders = append(resultHeaders, internalapi.Header{"Authorization", fmt.Sprintf("Bearer %s", accessToken)})
 	// Add the Authorization header with the GCP access token.
-	requestHeaders[":path"] = newPath
-	requestHeaders["Authorization"] = fmt.Sprintf("Bearer %s", accessToken)
-	return []internalapi.Header{{":path", newPath}, {"Authorization", fmt.Sprintf("Bearer %s", accessToken)}}, nil
+	// requestHeaders[":path"] = newPath
+	// requestHeaders["Authorization"] = fmt.Sprintf("Bearer %s", accessToken)
```
nacx:
```suggestion
```

internal/backendauth/gcp.go
```diff
@@ -63,21 +69,47 @@ func newGCPHandler(ctx context.Context, gcpAuth *filterapi.GCPAuth) (filterapi.B
...
+			if region, ok = requestHeaders["gcp-region"]; !ok {
+				return nil, fmt.Errorf("gcp-region header must be specified")
+			}
+		}
+	}
+
+	prefixPath := fmt.Sprintf("/v1/projects/%s/locations/%s", projectName, region)
 	// Find and update the ":path" header by prepending the prefix.
-	path := requestHeaders[":path"]
+	path := requestHeaders["gcp-path"]
```
nacx, line 101: This looks like a breaking change. Can you explain it? Is this guaranteed to always be present?

DanielChrn: Sure, 
This part is the follow-up to the changes in translators (you can see this in Anthropic and OpenAI to Vertex AI transformers) 
Basically, right now, the translator change path for GCP points to a URL that does not exist (a partial suffix of the GCP Vertex AI API). 
So when the path arrives at the backend, no matter what, it will be broken. 

Now, why exactly it bothered me is that me as a user/client - I want to be able to specify the full path in the URL to the Vertex API the same way we did for Anthropic and OpenAI (for examlple Claude code when you use it with vertex  use raw vertex API path).  Because of that, I decided to move path header transformation from `path` var to `gcp-path`. 

I would emphasize that this part of the MR is not the most important, so if you see it as a no-go, I can revert this part and keep only support for region/project headers.

PR #2578, opened 2026-08-24
fix: read thinking_delta content from the thinking field

khalidDaoud:
**Description**

`text_delta` and `thinking_delta` are separate variants of the SDK's
`RawContentBlockDeltaUnion` and carry their payload in different fields. The
Anthropic stream parser read `.Text` for both, so every extended thinking token
reached OpenAI clients as an empty string.

The union documents the split itself:

```go
// This field is from variant [TextDelta].
Text string `json:"text"`
...
// This field is from variant [ThinkingDelta].
Thinking string `json:"thinking"`
```

`thinking_delta` now reads `.Thinking`. Everything else about the handling is
unchanged — thinking is still folded into `delta.content` alongside text.

**Impact:** the Anthropic-to-OpenAI streaming path, for any backend using this
parser (`GCPAnthropic`, `AWSAnthropic`), whenever the model streams extended
thinking. The response is well-formed and returns 200; the reasoning content is
just missing, which makes it easy to miss in testing.

None that I could find — happy to open an issue first if you would prefer that.

`TestAnthropicStreamParser_ThinkingDelta` drives a stream with one thinking
block (two deltas plus a `signature_delta`) followed by one text block, and
collects the non-empty `delta.content` values in order. Before the change it
collects only the text block's content:

```
expected: []string{"Let me work through it. ", "Two plus two is four.", "4"}
actual  : []string{"4"}
```

`make precommit` is clean and `go test ./internal/translator/... -count=1`
passes. `make test` has four pre-existing failures in `internal/controller`
(`TestBackendSecurityPolicyController_{ReconcileOIDC_Fail,RotateCredential,
RotateExpiredCredential,ExecutionRotation}`) which reproduce unmodified on
`main` and are unrelated to this change.

Per the generative AI policy: this change was developed with AI assistance
(Claude). I have reviewed it, understand it, and can revise any part of it on
request.


gavrissh (approved):
Thanks for helping out with this change!

PR #2579, opened 2026-08-24
fix: route streaming tool state by content block index

khalidDaoud:
**Description**

Anthropic documents an accumulation contract for streamed tool input, and it is
keyed by the content block index:

> **The accumulation contract:**
> 1. On `content_block_start` with `type: "tool_use"`, initialize an empty string: `input_json = ""`
> 2. For each `content_block_delta` with `type: "input_json_delta"`, append: `input_json += event.delta.partial_json`
> 3. On `content_block_stop`, parse the accumulated string

— "Accumulating tool input deltas" [1], which also states that "the accumulation
contract is the same as for standard tool-use streaming, so this section applies
with and without `eager_input_streaming`."

`event.index` is the key in every SDK example on that page. The Go one is
explicit about what the map is:

```go
toolInputs := map[int64]string{} // content block index -> accumulated JSON
```

This matches the streaming reference [2]: "Each content block has an `index` that
corresponds to its index in the final Message `content` array", and "each
`content_block_delta` event contains a `delta` of a type that updates the
`content` block at a given `index`."

This parser is a client of that contract, and implemented steps 2 and 3 against
a different key. Tool state was stored under `p.toolIndex` — a counter reset at
`message_start` and incremented on each tool block start, i.e. the *OpenAI*
`tool_calls` index — and the delta and stop handlers both looked state up under
that counter:

```go
p.activeToolCalls[p.toolIndex] = &streamingToolCall{...}  // content_block_start
tool, ok := p.activeToolCalls[p.toolIndex]                // input_json_delta
delete(p.activeToolCalls, p.toolIndex)                    // content_block_stop
```

The counter and the block index are different quantities. They agree only under
an assumption the parser never states: that at most one tool block is open, and
that every event belonging to it arrives before any other block starts.
`content_block_stop` shows the mismatch most plainly — it deletes a tool entry
addressed by the counter regardless of which block actually stopped, including
when the event names a text or thinking block. The handler already unmarshals
the event and has `event.Index` in hand; it just was not read.

To be straight about the justification: I am **not** claiming a current model
emits an order under which the old key misroutes. This is conforming to the
documented contract rather than to an emission order the contract does not
promise — and the docs are explicit that a stream's shape is not frozen. They
already carve out a non-tool `fallback` block that arrives as a
`content_block_start`/`content_block_stop` pair with no deltas between them, and
they state that "new event types may be added, and your code should handle
unknown event types gracefully".

One defect here does not depend on ordering:

**A zero-argument tool streams invalid JSON.** Per the contract, `input: {}` on
`content_block_start` is a placeholder — "the empty object marks the slot in the
content array" — so the parser correctly skips it and emits `Arguments:""`. If
no `input_json_delta` then contributes any JSON text, accumulation ends at `""`.
On the Anthropic side a client is told to guard the parse; on the OpenAI side
`tool_calls[].function.arguments` is expected to be a JSON string the client
parses, so the gateway has to render "no arguments" as `{}` rather than passing
`""` through. This is reachable today from a tool called with no arguments.

**The change**

State is keyed by the Anthropic block index, so `input_json_delta` and
`content_block_stop` resolve the block the event names. The OpenAI `tool_calls`
index is dense across tool calls while the Anthropic block index counts every
block — including text and thinking — so the OpenAI index is recorded separately
on each call as `openaiIndex` and remains what reaches the wire. Block indexes
stay internal. A zero-argument tool emits a synthetic `{}` on its
`content_block_stop`.

**Impact:** the Anthropic-to-OpenAI streaming path, for any backend using this
parser (`GCPAnthropic`, `AWSAnthropic`). The zero-argument case produces a 200
with tool arguments that do not parse. The routing changes are contract
conformance, latent under today's emission order.

Rebased onto #2585, which forwards `eager_input_streaming` and cites the same
doc page. No overlap in the diff — that PR is request-side
(`translateOpenAItoAnthropicTools`), this one is the response parser — but it
makes conformance here more load-bearing, since with eager streaming the
fragments the gateway forwards are no longer server-validated.

`anthropic_helper_toolrouting_test.go` feeds events through the parser one at a
time, in wire order, and accumulates arguments per OpenAI `tool_calls` index the
way a client would. Three of the four fail before this change:

| test | what it pins | before |
|---|---|---|
| `TestStreamToolRoutingInterleavedDeltas` | step 2: deltas append to the block they name | FAIL — tool A's arguments arrive empty |
| `TestStreamToolRoutingStopEvictsOtherTool` | step 3: a stop closes only the block it names | FAIL — `received input_json_delta for unknown tool` |
| `TestStreamZeroArgumentToolEmitsEmptyObject` | accumulated arguments are valid JSON | FAIL — arguments accumulate to `""` |
| `TestStreamToolRoutingOpenAIIndexesStayDense` | block indexes do not leak to the wire | passes before and after |

The first two construct streams that are legal under the documented contract
rather than replaying an order observed from a model — they are conformance
tests, not regression tests for a reported failure, and I would rather say so
than dress them up. The third is a present-day defect. The fourth passes both
ways; it pins the dense OpenAI index mapping this PR introduces, which is the
part most likely to be got wrong later.

One judgement call worth flagging: the synthetic `{}` for zero-argument tools is
a behaviour change, not just a routing fix. The alternative is to emit `{}` from
`content_block_start` instead of `""`, which is a smaller diff but sends the
arguments before the block is known to be empty, and contradicts the placeholder
semantics the contract gives `input: {}`. Happy to switch if you prefer that
shape.

The four tests above and `go test ./internal/extproc/... -count=1` pass on the
rebased branch. Two sets of pre-existing failures reproduce unmodified on `main`
and are unrelated to this change: four `internal/controller` tests
(`TestBackendSecurityPolicyController_{ReconcileOIDC_Fail,RotateCredential,
RotateExpiredCredential,ExecutionRotation}`), and — only on Go 1.27 / arm64 —
`TestOpenAIToAWSAnthropicTranslatorV1ChatCompletion_ResponseBody/response_with_tool_use`,
which is a whitespace-only JSON difference from `sonic` falling back to
`encoding/json` outside its supported toolchain range.

1: https://platform.claude.com/docs/en/agents-and-tools/tool-use/fine-grained-tool-streaming#accumulating-tool-input-deltas
2: https://platform.claude.com/docs/en/build-with-claude/streaming

### DIFF
--- a/internal/translator/anthropic_helper.go
+++ b/internal/translator/anthropic_helper.go
@@ -836,6 +836,12 @@ type streamingToolCall struct {
 	id        string
 	name      string
 	inputJSON string
+	// openaiIndex is this call's index within the OpenAI tool_calls array,
+	// assigned in tool-block arrival order. It is not the Anthropic content
+	// block index: OpenAI indexes are dense across tool calls only, while
+	// Anthropic block indexes count every block, including text and
+	// thinking.
+	openaiIndex int64
 }
 
 // anthropicStreamParser manages the stateful translation of an Anthropic SSE stream
@@ -843,6 +849,10 @@ type streamingToolCall struct {
 type anthropicStreamParser struct {
 	buffer          bytes.Buffer
 	activeMessageID string
+	// activeToolCalls is keyed by the Anthropic content block index carried
+	// on every content_block_* event, so lookups are position-independent:
+	// text or thinking blocks interleaved between tool blocks cannot
+	// desynchronize the routing of input_json_delta or content_block_stop.
 	activeToolCalls map[int64]*streamingToolCall
 	toolIndex       int64
 	tokenUsage      metrics.TokenUsage
@@ -1003,7 +1013,7 @@ func (p *anthropicStreamParser) Process(body io.Reader, endOfStream bool, span t
 
 		// Add active tool calls to the final chunk.
 		var toolCalls []openai.ChatCompletionChunkChoiceDeltaToolCall
-		for toolIndex, tool := range p.activeToolCalls {
+		for _, tool := range p.activeToolCalls {
 			toolCalls = append(toolCalls, openai.ChatCompletionChunkChoiceDeltaToolCall{
 				ID:   &tool.id,
 				Type: openai.ChatCompletionMessageToolCallTypeFunction,
@@ -1011,7 +1021,7 @@ func (p *anthropicStreamParser) Process(body io.Reader, endOfStream bool, span t
 					Name:      tool.name,
 					Arguments: tool.inputJSON,
 				},
-				Index: toolIndex,
+				Index: tool.openaiIndex,
 			})
 		}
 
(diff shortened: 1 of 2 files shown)


kanurag94 (approved):
LGTM
Can you please rewrite the description. The justification should be about keying event.index is Anthropic's documented accumulation contract? The current description is not true completely.

khalidDaoud:
@kanurag94 updated as requested

PR #2580, opened 2026-08-24
ci: compute dev prerelease version for builds from main branch

herewasmike:
**Description**

Add automatically calculated and human readable version to builds from main branch in addition to 0.0.0-<commit sha> 

AI was used to produce this


GitVersion.yaml
```diff
@@ -0,0 +1,9 @@
+# Copyright Envoy AI Gateway Authors
+# SPDX-License-Identifier: Apache-2.0
+# The full text of the Apache license is available in the LICENSE file at
+# the root of the repo.
+
+# Main heads toward the next minor release per RELEASES.md, not the default "Patch" increment.
```
aabchoo, line 6: We do occasionally release patch increments for bug fixes. The gitVersion would ignore that in this case and start counting the gitVersions again.

From what I understand x.x.0 -> x.x+1.0 but also x.x.1 -> x.x+1.0. The gitVersion suffix incr will overlap with the versions. The image we push will have the date appended so nothing will change there but the incr will be a bit misleading.

.github/workflows/build_and_test.yaml
```diff
@@ -378,19 +378,51 @@ jobs:
...
+      - name: Determine next version
+        id: gitversion
+        continue-on-error: true
+        uses: gittools/actions/gitversion/execute@7417b1089e2c7de93510f1901d656ddf60bb024f # v4.7.0
+      - name: Compute dev tag
+        id: version
+        run: |
+          MMP="${{ steps.gitversion.outputs.majorMinorPatch }}"
+          if [ "${{ github.ref }}" = "refs/heads/main" ] && [ -n "$MMP" ]; then
+            echo "version=v${MMP}-dev.$(date -u +%Y%m%d%H%M)" >> "$GITHUB_OUTPUT"
```
aabchoo: do you mind updating this to include hyphens for ease of reading

GitVersion.yaml
herewasmike: That's the design -- if you opt in for `semver >= 0.0.0-0` you get w/e is fresh from the main as dev prerelease have higher minor version.
x.x+1.0-dev > x.x.1 > x.x.0

if you opt in for stable (no-prerelease)`semver >= 0.0.0` patch is still ordered correctly, and x.x+1.0-dev is ignored

.github/workflows/build_and_test.yaml
herewasmike: Do you mean `1.1.0-dev.2026-08-25-17-07`?
Or something else?

aabchoo: Yes, that or `1.1.0-dev.2026-08-25-17:07` whichever is fine

GitVersion.yaml
aabchoo: Let's go with the prior?

Would be nicer to loosely correspond images with release versions in my opinion. Open to your thoughts on the matter though.

.github/workflows/build_and_test.yaml
```diff
@@ -378,19 +378,51 @@ jobs:
...
+          fetch-depth: 0
+      - name: Install GitVersion
+        continue-on-error: true
+        uses: gittools/actions/gitversion/setup@7417b1089e2c7de93510f1901d656ddf60bb024f # v4.7.0
+        with:
+          versionSpec: "6.x"
+      - name: Determine next version
+        id: gitversion
+        continue-on-error: true
+        uses: gittools/actions/gitversion/execute@7417b1089e2c7de93510f1901d656ddf60bb024f # v4.7.0
```
mathetake, line 404: do we really need this dep?

GitVersion.yaml
herewasmike: I'm not sure I understand your concern fully.

In current setup dev build from main is always a minor version higher than the latest stable one.
I assume if patch's gonna happen it'd be tagged in branch from the previous tag?
Even if it wouldn't this still gives you exact correspondence to the release -- this commit when merged would get `1.2.0-dev.<bla-bla> `which put it in a correct semver order.

`1.2.0` (not created yet) > `1.2.0-dev.<bla-bla>` > `1.1.1` (patch version) > `1.1.0`

Am I missing something?

herewasmike: There's a possibility that `1.2.0-dev.<bla-bla>` and `1.1.1` would point to the same commit (if main is tagged with patch increment to release it) but I don't see any problem with that

aabchoo: I believe need to specify `label: dev` if you want the behavior of `1.2.0-dev.<bla-bla>`?

What I am pointing out is the scenario where:
1. we have 3 commits for v.1.1.0 then the expected version would be `1.2.0-dev.3`. 
2. we release a patch 1.1.1
3. a commit is merged so the version will start again at `1.2.0-dev.1`.

Please correct me if this is the wrong assumption. 

I wonder if having the semver provides any benefits ie what does `1.2.0-dev.31` correspond with? Curious to hear what additional benefits this gives over just having the date?

herewasmike: It won't because I used timestamp as a continuous increment

> echo "version=v${MMP}-dev.$(date -u +%Y%m%d%H%M)" >> "$GITHUB_OUTPUT"

so it'd be 1.2.0-dev.2026-08-26-00-00 then 1.1.1 and then 1.2.0-dev.2026-08-27-00-00

.github/workflows/build_and_test.yaml
herewasmike: That's the question I brought up on the last call, folks seemed to be fine with that. I don't see any harm in adding CI dependency, it's much cleaner than trying to implement this with the bash script

GitVersion.yaml
aabchoo: https://github.com/GitTools/GitVersion/blob/main/BREAKING_CHANGES.md

> In the GitHub and the Git Flow workflows the label property is by default set to an empty string on the main branch. This yields to a pre-release version on main with an empty tag. Instead of for instance 1.0.1+46 GitVersion generates the full semantic version 1.0.1-46 instead. This behavior can be changed to generate only stable versions (no pre-release version) with setting the label to null

Seems like we do need to set the label?

herewasmike: Do you mean to highlight that the build on main got breaking change?

I've added tagPreview so it's easier to reason about the tag -- check Summary on the run
https://github.com/envoyproxy/ai-gateway/actions/runs/33014399940?pr=2580

herewasmike: If we agree on the semantics for breaking changes -- can be added I guess.
I'd leave this as a followup, though since the idea was to add a simple human readable ordering to the builds from main

aabchoo: The docs implied that it would follow the format of `M.M.P-{commit #}-dev...` but the tagPreview says otherwise. Looks good to me then

.github/workflows/build_and_test.yaml
herewasmike: Done in the latest commit

aabchoo (approved):

aabchoo (approved):

PR #2583, opened 2026-08-25
fix(extensionserver): reject non-forwarding InferencePool routes

### LINKED ISSUE #2582: PostRouteModify should return an error, not pass through, when given an InferencePool with a non-forwarding route
## Description

After #1951, `PostRouteModify` (`internal/extensionserver/post_route_modify.go`) handles a route whose `GetRoute()` is nil by skipping the InferencePool configuration and returning the route unchanged. The regression test asserts `NoError` and an unmodified route.

That input is not a valid steady state, though. Envoy Gateway only hands the hook a `DirectResponse` for a rule that references an InferencePool when it could not resolve the pool backend, which in practice means the controller has not yet synced that InferencePool (see the companion readiness issue). Returning success publishes Envoy Gateway's `direct_response 500` for the model route as if it were the intended configuration.

## Why it matters

Envoy Gateway has an explicit policy for hook failures: `extensionManager.failOpen`. With `failOpen: false` (default) a hook error retains the last good snapshot; with `failOpen: true` it applies the unmodified xDS. Both are operator choices. Returning success from a state the extension knows is wrong removes that choice and makes the 500 unconditional.

## Proposal

When `PostRouteModify` receives InferencePool extension resources and `req.Route.GetRoute() == nil`, return an error (`codes.FailedPrecondition` or `Unavailable`) with a message naming the pool, instead of `nil`. The nil check from #1951 stays; only the return value changes. Envoy Gateway then applies `failOpen` as configured and retries on the next reconcile, which happens as soon as the InferencePool status transitions.

If there is a legitimate case where an InferencePool-referencing rule is meant to be a DirectResponse (a `RequestRedirect` filter on the same rule, say), gate the error on that not being present.

Observed on v0.5.0 (as a panic) and by inspection on v1.1.0; Envoy Gateway v1.6.2.

0YHR0:
**Description**

When Envoy Gateway cannot resolve an InferencePool backend, it invokes `PostRouteModify` with the pool extension resource but a non-forwarding route action. The hook currently returns that route successfully, which can publish Envoy Gateway's fallback direct-response 500 and bypass the configured `extensionManager.failOpen` policy.

This change returns a `FailedPrecondition` error naming the InferencePool and route whenever an InferencePool is paired with a non-forwarding route. Envoy Gateway can then retain the last good snapshot or fail open according to operator configuration. The regression test verifies both the status code and pool identity.

Fixes #2582

Validation:
- `go test ./internal/extensionserver -run '^TestPostRouteModify$' -count=1`
- `go test ./internal/extensionserver -count=1`
- `go vet ./internal/extensionserver`
- `git diff --check`


nacx (approved):
Thanks!

PR #2587, opened 2026-08-27
fix(mcp): accept JSON content type parameters

chengjingtao:
**Description**

The MCP proxy compared Content-Type using exact string equality with application/json. SonarQube returns application/json;charset=utf-8, so valid JSON-RPC responses were incorrectly sent through the SSE parser and dropped during multi-backend aggregation.

This change parses the media type with mime.ParseMediaType and applies the normalized JSON check to backend aggregation, single-backend response proxying, and response header/content-length handling. Tests cover charset parameters, case and whitespace, SSE, and invalid media types.


Hritik003 (commented):
LGTM, @chengjingtao can you paste some logs that caused this issue?

kanurag94 (approved):
LGTM

kanurag94:
cc @envoyproxy/ai-gateway-maintainers

nacx (approved):

Issue #2049, opened 2026-04-14
mcpoauth: support for `authorizationServerMetadata`


Hritik003:
_Description_:

When an MCP client (such as Cursor, Claude Desktop, or any RFC 9728-compliant client) connects to an OAuth-protected MCP endpoint, it needs to discover the authorization server's endpoints (authorization, token, JWKS, registration) to perform the browser-based OAuth login flow. Today, the ai-gateway controller attempts to auto-discover these endpoints by fetching the issuer's `/.well-known/oauth-authorization-server` or `/.well-known/openid-configuration` document. If auto-discovery fails, the controller falls back to **Keycloak-style hardcoded paths** (e.g., `/protocol/openid-connect/auth`, `/protocol/openid-connect/token`).

This approach breaks for authorization servers that:

1. **Do not serve well-known metadata at the issuer URL** — For example, the IDP we use have a legacy issuer path (`/api/idp/authn/.well-known/openid-configuration`) that returns `{}` (empty JSON), while the real metadata is only available at a versioned path (`/api/idp/v4/authn/.well-known/openid-configuration`).

2. **Use non-standard endpoint paths** — our IDP uses `/authn/v1/oidc/auth` and `/v4/authn/oidc/token`, not the Keycloak-style `/protocol/openid-connect/auth` and `/protocol/openid-connect/token`.

3. **Require authentication to access well-known endpoints** — Some deployments place the well-known endpoints behind an authentication gateway, returning 401 to unauthenticated requests from the controller.

4. **Have a mismatch between the issuer identity and the discovery path** — The `issuer` field in the MCPRoute must exactly match the `iss` claim in JWT tokens (Envoy does exact string comparison for token validation). But the well-known metadata might only be served at a different (e.g., versioned) path. These two concerns — token validation and metadata discovery — are coupled through the `issuer` field today, and they shouldn't be.

### Why the existing auto-discovery doesn't work

The controller derives well-known URLs from the `issuer` field. For an issuer like `https://example.com/api/iam/authn`, it tries:

| URL tried | Result |
|---|---|
| `https://example.com/.well-known/oauth-authorization-server/api/idp/authn` | 401 (gateway blocks) |
| `https://example.com/.well-known/openid-configuration/api/idp/authn` | 401 (gateway blocks) |
| `https://example.com/api/idp/authn/.well-known/oauth-authorization-server` | 401 or 404 |
[...]

PR #2592, opened 2026-08-28
fix: treat unusable OAuth metadata documents as discovery failures

zhaohuabing:
**Description**

An authorization server that answers `200` with `{}` at a well-known path it does not actually implement was treated as successful discovery. The empty struct unmarshaled cleanly, so the controller served MCP clients a metadata document with empty-string endpoints instead of falling through to the remaining well-known URL variants.

Related to #2049. This is the first of two changes discussed there; the `authorizationServerMetadataUrl` field is separate.


zhaohuabing:
cc @nacx

kanurag94 (approved):

nacx (approved):

PR #2596, opened 2026-08-29
fix(helm): support externally managed webhook secrets

### LINKED ISSUE #2591: helm: mutatingWebhook "custom secret" path is silently broken under GitOps rendering (lookup returns empty)
### Description

The non-cert-manager branch of `admission_webhook.yaml` relies on `lookup` to reuse an existing webhook cert Secret:

```gotemplate
{{- $existing := lookup "v1" "Secret" .Release.Namespace .Values.controller.mutatingWebhook.tlsCertSecretName }}
{{- if $existing }}
  {{/* reuse existing cert */}}
{{- else }}
  {{- $ca := genCA (printf "%s-ca" $serviceName) 3650 }}
  {{- $cert := genSignedCert (printf "%s.%s.svc" $serviceName .Release.Namespace) nil $dnsNames 365 $ca }}
{{- end }}
```

This works under `helm install/upgrade`, but Helm's `lookup` returns an empty value during `helm template` and any rendering without cluster access — which is how Argo CD, Flux, and similar GitOps tooling render charts. `$existing` is therefore always empty and `genCA`/`genSignedCert` mint a **new CA and server cert on every render**.

### Impact

1. The rendered desired state is non-deterministic, so the application is permanently OutOfSync by construction. With automated sync / self-heal enabled, the Secret is rewritten continuously with fresh certs, breaking webhook TLS. Since `failurePolicy` is hardcoded to `Fail`, pod CREATEs matching the webhook's `objectSelector` are then rejected — an outage, not a degradation.

2. `values.yaml` states:

   > The self-signed cert is embedded as part of the helm chart, so it is not recommended
   > for production use. You can use the configurations below to specify a custom secret
   > that contains the CA certificate and the server certificate.

   Under GitOps rendering this is actively misleading: pointing `tlsCertSecretName` at a Secret managed elsewhere (External Secrets Operator, a cert-manager Certificate in another chart, manual provisioning) does not make the chart *use* it — the chart renders over it at the same name, putting the two managers in a write-fight.

### Affected versions

v1.0.0, v1.1.0, and `main` — the template is unchanged across all three.

### Proposal

Add a value that skips rendering the Secret entirely, e.g. `controller.mutatingWebhook.createSecret: false` (or `existingSecret: true`), so a pre-provisioned Secret can be supplied without depending on a cluster-side `lookup`. That makes the documented "custom secret" behaviour true for GitOps users as well.

Today the only branch that emits no Secret is `certManager.enable: true`, which forces a cert-manager dependency on users who may not want or have one.

### Workarounds

- Enable cert-manager.
- In Argo CD, `ignoreDifferences` on the Secret's `/data` combined with the `RespectIgnoreDifferences=true` sync option. This freezes whatever cert is currently live, so it carries a 365-day expiry with nothing to rotate it, and breaks bootstrapping on a fresh cluster (the Secret gets created without `data`).

0YHR0:
**Description**

The Helm chart currently always renders and owns the mutating webhook TLS Secret when cert-manager is disabled. During GitOps rendering, Helm cannot reuse the live Secret through `lookup`, so it generates a new keypair on every render and can overwrite a Secret managed externally.

This change adds `controller.mutatingWebhook.createSecret`, defaulting to `true` for backward compatibility. Setting it to `false` keeps the webhook Deployment pointed at `tlsCertSecretName` but omits the Secret manifest, allowing External Secrets, another chart, or another GitOps-managed resource to own it deterministically.

The Helm test target now verifies that the default path still renders one Secret and that the external-secret path renders none.

Fixes #2591

The change is intentionally independent of #2584: this PR controls Secret ownership, while #2584 controls the webhook CA bundle. If #2584 merges first, I will rebase this condition onto its restructured template.

Local `git diff --check` passes. The repository-pinned Helm tool could not run locally because a transitive module download failed checksum verification; the added `helm-test` assertions will run in CI's clean environment.


nacx (approved):
LGTM!

PR #2599, opened 2026-08-30
fix(controller): gate readiness on cache sync

### LINKED ISSUE #2581: Controller readiness probe passes before InferencePool controller has synced, letting Envoy Gateway call the extension server with unresolved pools
## Description

The Helm chart's `readinessProbe` (and `livenessProbe`) for the controller is `grpc: port 1063`, the extension server port (`manifests/charts/ai-gateway-helm/templates/deployment.yaml`). In `cmd/controller/main.go` the extension server starts serving before `mgr.Start()` acquires the leader lease and starts the InferencePool controller. So the pod is Ready, and the Service routes Envoy Gateway's translation calls to it, while the controller has not yet reconciled its own InferencePools.

With `controller.leaderElection.enabled: true` (the default) on a single replica, that window is the lease acquisition time. Measured on v0.5.0 from a controlled restart:

```
18:00:52  extension server, webhook and metrics start
18:00:53  Attempting to acquire leader lease
18:01:10  Successfully acquired lease
18:01:10  Starting EventSource {"controller": "aigatewayroute" ...}
```

17 seconds during which `PostRouteModify` answers for an InferencePool it has not synced. Envoy Gateway cannot resolve the pool backend in that state and hands the hook a `DirectResponse` route for the rule.

## Impact

- v0.5.0: `PostRouteModify` dereferenced `GetRoute()` unconditionally and panicked (#1888, #1889, fixed by #1951). Because the pod is Ready again in ~5s and Envoy Gateway retries translation, the restart lands inside the same window and the loop self-sustains. We observed 27 restarts over ~45 minutes, during which Envoy Gateway retained its last xDS snapshot and no config change reached Envoy for any route on the shared gateway, not only the AI routes.
- v0.6.0+: #1951 skips the modification, so the rule is published as Envoy Gateway's `direct_response 500` until the InferencePool status update triggers a retranslation. Correct but a transient 500 on the model route after every controller restart.

## Proposal

1. Make readiness reflect readiness: register a `/readyz` (or gRPC health service state) that only reports healthy after leader election completes and `mgr.GetCache().WaitForCacheSync` returns, and point the chart's `readinessProbe` at it. Keep liveness on the gRPC port. This is the standard controller-runtime pattern and stops the Service from routing Envoy Gateway to a controller that cannot yet translate correctly.
2. Alternatively or additionally, have the extension server return `codes.Unavailable` from its hooks until the manager has synced, so Envoy Gateway's `extensionManager.failOpen` policy applies deliberately rather than publishing a 500.
3. Consider defaulting `controller.leaderElection.enabled` to `replicaCount > 1`. On one replica the lease protects nothing and is the bulk of the window.

Versions: ai-gateway v0.5.0 and code inspection of v0.6.0 / v1.1.0 `cmd/controller/main.go` (no readiness gating added); Envoy Gateway v1.6.2.

Aias00:
**Description**

Start the controller extension gRPC server only after the controller-runtime cache has synchronized, so Envoy Gateway cannot send extension calls before the controller has observed InferencePools and other watched resources.

The server is registered as a non-leader-election manager runnable. Controller-runtime starts this runnable on every replica after its local cache has synced, preserving horizontally scaled extension-server traffic on non-leader replicas.

The Helm deployment adds a gRPC startup probe with enough budget for the default cache sync timeout, preventing liveness from restarting the controller while the extension server is intentionally not serving.

Fixes #2581

Verification:

- `go test ./cmd/controller ./internal/controller ./internal/extensionserver -count=1`
- `go test -race ./cmd/controller -run TestExtensionServerRunnable -count=1`
- `go vet ./cmd/controller ./internal/extensionserver`
- `make helm-test`
- `make lint`

This change was developed with assistance from OpenAI Codex. I reviewed the implementation and take responsibility for it.


cmd/controller/main.go
```diff
@@ -467,6 +467,11 @@ func main() {
 		setupLog.Error(err, "failed to create extension server")
 		os.Exit(1)
 	}
+	go func() {
+		if mgr.GetCache().WaitForCacheSync(ctx) {
+			extSrv.MarkReady()
+		}
+	}()
```
mathetake: do we need to make this async? can we simply do `mgr.GetCache().WaitForCacheSync(ctx)` here without goroutine?

Aias00: Removed the explicit goroutine by using a non-leader-election manager runnable, which starts after cache sync on every replica.

mathetake: i meant not running extension server at all before cache sync - then we don't need to make any change to extension server, no?

Aias00: Updated — the extension server now starts only after cache sync, with a startup probe covering the delayed startup.

nacx (approved):
LGTM. @mathetake wanna ahve another look?

PR #2601, opened 2026-08-30
fix: skip non-HCM filter chains instead of failing translation

### LINKED ISSUE #2600: Extension server PostTranslateModify fails on listeners with non-HTTP (TCP/UDP) filter chains; default logRequestHeaderAttributes makes stock installs hit it
With the ai-gateway controller hooked as an Envoy Gateway extension service (`post: [Translation, Cluster, Route]`, listener IR included), `PostTranslateModify` fails whenever a listener contains a filter chain without an HTTP connection manager (TCPRoute/UDPRoute). The whole hook aborts, xDS is rejected, and all Gateways on that control plane stay `Programmed=False`.

Expected: skip such chains (they cannot host HTTP filters). Two `findHCM` callers already do; two don't.

# Repro steps

1. EG extensionManager → ai-gateway controller, `post: [Translation, Cluster, Route]`, `listener.includeAll: true`.
2. One Gateway with an AIGatewayRoute + one Gateway with TCP/UDP listeners (Syncthing :22000 TCP+UDP).
3. Controller with default flags (no `--logRequestHeaderAttributes=""`).

`findHCM` callers disagree on non-HCM chains:

```mermaid
flowchart TD
    A["insertRequestHeaderToMetadataFilters<br/>walks every listener unconditionally"] --> B{"findHCM(chain)"}
    A2["insertRouterLevelAIGatewayExtProc"] --> B
    B -- "no HCM" --> C{"caller behavior"}
    C -- "patchListenerWithInferencePoolFilters ✅<br/>findListenerRouteConfigs ✅" --> D["log + continue"]
    C -- "insertRequestHeaderToMetadataFilter ❌<br/>header_to_metadata.go:41<br/>insertRouterLevelAIGatewayExtProc ❌<br/>post_translate_modify.go:777" --> E["return error → hook aborts → xDS rejected → Gateways Programmed=False"]
    E --> F["Gateway Programmed=False (NoResources)<br/>dataplane crash-loops on rejected xDS"]
```

It hits stock installs: `logRequestHeaderAttributes` defaults to `agent-session-id:session.id` when unset (`internal/requestheaderattrs/resolve.go`), and `insertRequestHeaderToMetadataFilters` walks every listener unconditionally.

# Environment

- Envoy AI Gateway v1.1.0 (also v0.7.0, v1.0.0; unchanged on main)
- Envoy Gateway v1.9.0, Kubernetes 1.34

# Logs

```text
failed to insert request header metadata filter into FilterChain: tcproute/shikanime/syncthing: unable to find HTTPConnectionManager in FilterChain: tcproute/shikanime/syncthing
```

Workaround: `controller.logRequestHeaderAttributes: ""` — translation then succeeds. Proposed fix: skip non-HCM chains in the two failing callers, matching the existing pattern. Happy to send a PR.

shikanime:
**Description**

This commit skips filter chains without an HTTP connection manager in `insertRequestHeaderToMetadataFilter` and `insertRouterLevelAIGatewayExtProc` instead of returning an error that aborts the whole `PostTranslateModify` hook. Like the existing skip paths (`patchListenerWithInferencePoolFilters`, `findListenerRouteConfigs`, `insertMCPBackendFilters`, `insertQuotaRateLimitFilter`), chains without an HCM are skipped rather than failing the hook. Per review feedback, the skip keys on the `findHCM` sentinel (`hcmIndex == -1`) instead of the returned error, so a future genuine error is not silenced; the error is still logged.

Chains without an HCM (TCPRoute/UDPRoute listeners) cannot host HTTP filters, so finding no HCM in them is normal, not an error. A regression test is added: a listener mixing a TCP-proxy chain with an HCM chain must translate, with the filter injected only into the HTTP chain.

This hits stock installs: `logRequestHeaderAttributes` defaults to `agent-session-id:session.id` when the flag is unset (`internal/requestheaderattrs/resolve.go`), and `insertRequestHeaderToMetadataFilters` walks every listener unconditionally — so any Gateway carrying TCP/UDP listeners makes xDS translation fail for the entire control plane (all Gateways `Programmed=False`, dataplane crash-looping on rejected config).

Fixes #2600

`go test ./internal/extensionserver/`, `go vet`, and `gofmt` pass locally. Tested against a live fleet where this exact topology (HTTP inference Gateway + Syncthing TCP/UDP Gateway on one control plane) reproduced the failure.

### DIFF
--- a/internal/extensionserver/header_to_metadata.go
+++ b/internal/extensionserver/header_to_metadata.go
@@ -40,7 +40,9 @@ func (s *Server) insertRequestHeaderToMetadataFilter(listener *listenerv3.Listen
 	for _, currChain := range filterChains {
 		httpConManager, hcmIndex, err := findHCM(currChain)
 		if err != nil {
-			return err
+			// Chains without an HCM (e.g. TCP/UDP routes) cannot host HTTP filters.
+			s.log.Error(err, "skipping filter chain without an HTTP connection manager", "listener", listener.Name)
+			continue
 		}
 		if filterIndex, filter := findHeaderToMetadataFilter(httpConManager.HttpFilters); filter != nil {
 			typedConfig := filter.GetTypedConfig()
--- a/internal/extensionserver/header_to_metadata_test.go
+++ b/internal/extensionserver/header_to_metadata_test.go
@@ -146,4 +146,32 @@ func TestInsertRequestHeaderToMetadataFilter(t *testing.T) {
 		err = s.insertRequestHeaderToMetadataFilter(listener)
 		require.EqualError(t, err, "header_to_metadata filter missing typed_config")
 	})
+	t.Run("skip-non-hcm-chain", func(t *testing.T) {
+		s := &Server{logRequestHeaderAttributes: map[string]string{"x-tenant-id": "tenant.id"}}
+		hcm := &httpconnectionmanagerv3.HttpConnectionManager{
+			HttpFilters: []*httpconnectionmanagerv3.HttpFilter{{Name: wellknown.Router}},
+		}
+		hcmAny, err := toAny(hcm)
+		require.NoError(t, err)
+		listener := &listenerv3.Listener{
+			FilterChains: []*listenerv3.FilterChain{
+				// TCP proxy chain without an HCM must be skipped, not fail the walk.
+				{Filters: []*listenerv3.Filter{{Name: "envoy.filters.network.tcp_proxy"}}},
+				{
+					Filters: []*listenerv3.Filter{{
+						Name:       wellknown.HTTPConnectionManager,
+						ConfigType: &listenerv3.Filter_TypedConfig{TypedConfig: hcmAny},
+					}},
+				},
+			},
+		}
+		require.NoError(t, s.insertRequestHeaderToMetadataFilter(listener))
+		// The HTTP chain got the filter; the TCP chain is untouched.
+		hcmAfter := &httpconnectionmanagerv3.HttpConnectionManager{}
+		require.NoError(t, listener.FilterChains[1].Filters[0].GetTypedConfig().UnmarshalTo(hcmAfter))
+		require.Len(t, hcmAfter.HttpFilters, 2)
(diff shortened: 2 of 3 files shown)


internal/extensionserver/header_to_metadata.go
```diff
@@ -40,7 +40,9 @@ func (s *Server) insertRequestHeaderToMetadataFilter(listener *listenerv3.Listen
 	for _, currChain := range filterChains {
 		httpConManager, hcmIndex, err := findHCM(currChain)
 		if err != nil {
-			return err
+			// Chains without an HCM (e.g. TCP/UDP routes) cannot host HTTP filters.
```
PatilHrushikesh: Treating all findHCM errors as a missing HCM can hide actual unmarshaling or malformed config errors

internal/extensionserver/header_to_metadata.go
```diff
@@ -40,7 +40,9 @@ func (s *Server) insertRequestHeaderToMetadataFilter(listener *listenerv3.Listen
...
+			s.log.Error(err, "skipping filter chain without an HTTP connection manager", "listener", listener.Name)
```
nacx, line 44: Instead of relying on the error, can you change it to check for `hcmIndex == -1` instead?
If, in the future, we report any error in the method, we don't want to silence it by mistake. Checking the returned index unambiguously conveys that the filter was not found, regardless of any error.

shikanime: Done in bc2cfc88 — both skip sites now key on the sentinel:

```go
httpConManager, hcmIndex, err := findHCM(currChain)
if hcmIndex == -1 {
```

Applied in `header_to_metadata.go` and `post_translate_modify.go`; the error is still logged in the skip branch, so a future `findHCM` failure mode stays visible. One caveat: `findHCM` returns `-1` for a nil chain and an unmarshal failure too, so the sentinel alone does not distinguish not-found from those errors — the retained log line covers them.

`go test ./internal/extensionserver/`, `go vet`, and `gofmt` are green on the new head.

PR #2602, opened 2026-08-31
feat(mcp): add prefixMode=Never to expose bare tool/prompt names

### LINKED ISSUE #2316: mcp: allow disabling/customizing the `<backend>__` tool name prefix

deepika0326:
**Description**

**Summary**

- Adds a `prefixMode` field (`Always` = default, or `Never`) on MCPRoute and per-backend, so MCP tool/prompt names can be exposed **bare** (e.g. `search_dashboards`) instead of always prefixed with the backend name (`grafana__search_dashboards`).

---

MCPRoute currently always prefixes tool and prompt names with "<backendName>__" when aggregating multiple MCP backends behind one route. This is necessary to avoid collisions, but it's unnecessary friction for the common case of a single backend, or when backends already expose distinct, well-known tool names and clients expect the bare names (e.g. because they were written against the backend directly, or because an interactive MCP App has tool names hardcoded in server-authored HTML/JS that never carry the prefix).

This adds an optional PrefixMode field (Always|Never, default Always) at both MCPRoute and MCPRouteBackendRef. Route-level PrefixMode sets the default for all backends on the route; per-backend PrefixMode overrides the route-level default for that backend, so Never and Always backends can coexist on one route as a targeted escape hatch. A backend whose effective PrefixMode is Never must declare toolSelector.include (and promptSelector.include, if it wants its prompts exposed bare too); this is what lets tool/prompt-name uniqueness be proven at admission time, before the route is ever accepted, instead of only being discoverable as a runtime ambiguity.

Bare-name resolution for Never-mode backends (tools/call, prompts/get, completion/complete) is driven by static, route-level neverModeToolIndex/neverModePromptIndex maps built once in LoadConfig from each backend's admission-validated include list, so resolution is independent of any per-request or per-session state and survives separate POSTs and gateway replicas. mergeToolsList/mergePromptsList apply the same per-backend effective mode when building the aggregated list, and guard against a cross-mode collision: an Always-mode backend's prefixed name is dropped, deterministically and regardless of response order, if it collides with a bare name a Never-mode backend has declared ownership of. The route-reload comparison includes the new indexes, so a PrefixMode change is correctly detected as a catalog change and triggers the appropriate list_changed notification.

I validated this end-to-end against a real cluster (not just unit/e2e tests): a fresh MCP session calling tools/call with a bare name and no prior tools/list resolves and routes correctly, mixed-mode routing (bare + prefixed names on the same route) works as expected, and the admission-time collision rejection triggers with the expected error message.

I used generative AI (Cursor) to assist with this change; I reviewed and take full ownership of the code, tests, and this description.

Closes #2316

Two design points I'd like input on, both documented inline in the code:

- Per-backend PrefixMode override (mixing Always and Never on one route) is intentional, documented on the MCPRoutePrefixMode type. Happy to restrict this to route-level only if reviewers prefer one naming policy per route; the current design treats per-backend override as an explicit escape hatch, not the primary path.
- The exact-list requirement for Never mode (toolSelector.include/promptSelector.include, no unfiltered or regex selectors) is intentional and documented on validatePerBackendPrefixMode: it's what lets uniqueness be proven at admission time rather than only at runtime. Open to revisiting if there's a use case that needs Never mode without an exact list.


kanurag94:
I haven't reviewed the code changes, but what is the motivation for this? How are conflicts handled across providers? Say github and bitbucket both have `list_issues` -- what is the expectation?

deepika0326:
> I haven't reviewed the code changes, but what is the motivation for this? How are conflicts handled across providers? Say github and bitbucket both have `list_issues` -- what is the expectation?

@kanurag94  
Motivation: Clean tool names. Without prefixMode: Never, every tool is prefixed with the backend name (e.g. grafana__search_dashboards, github__list_issues). This leaks internal infrastructure naming into the surface that LLM agents and MCP clients consume. When you have a single-purpose route, the prefix adds noise with no value.

Conflict handling: Conflicts only arise when two backends are multiplexed on the same route. If github and bitbucket are on separate routes (e.g. /github and /bitbucket), they're completely isolated — list_issues on each route is unambiguous.

When they share a route with prefixMode: Never, the controller rejects the route at admission with NotAccepted if both backends declare list_issues in their toolSelector.include. The error is explicit: prefixMode Never requires tool names to be unique across backends, but "list_issues" is declared by both "github" and "bitbucket". The user either splits the route or uses different tool selectors to avoid overlap.

kanurag94:
@deepika0326 have you had the chance to review #2425? I think this is impactful and the approach needs more discussion.
On top of my head, in case of conflicting tool, they should be dropped. The mcp design currently depends on prefixing. I think we should contribute to @2425 first.

deepika0326:
Thanks @kanurag94! Yes, we reviewed #2425 — it's a proposal-only PR with no implementation. Our PR implements what #2425 proposes (plus the per-backend variant). We're happy to either:
1. Move the implementation into #2425 as the maintainers prefer, or
2. Keep this as a separate implementation PR that references #2425

On conflict handling: we do drop conflicting tools at runtime as a backstop (same as what #2425 proposes). The admission-time rejection is an additional safety layer to give operators a clear error at route creation rather than silently dropping tools at runtime.

Happy to follow whatever approach the maintainers prefer!

api/v1beta1/mcp_route.go
```diff
@@ -133,7 +163,20 @@ type MCPRouteBackendRef struct {
...
+	//
+	// This is also used to determine which prompt names can be safely exposed unprefixed when
+	// this backend's effective PrefixMode is Never: only prompts named in Include are validated
+	// for cross-backend uniqueness and exposed bare. If unset while PrefixMode is Never, this
+	// backend's prompts are not exposed bare (they keep the `<backendName>__` prefix) because
+	// their names cannot be proven unique at admission time.
+	//
+	// +kubebuilder:validation:Optional
+	// +optional
+	PromptSelector *MCPPromptFilter `json:"promptSelector,omitempty"`
```
nacx, line 207: Can we keep this PR focused just on the prefix and leave this to a follow-up PR? (Why prompt selector and not resourceSelector in this PR?) I'd rather focus on the prefix change here, and add additional selectors in follow-up PRs.

nacx (commented):
Thanks! The PR looks good to me overall, but I'd like to introduce the prompt in a different PR and keep this one just for the prefix change. Can you do that?

api/v1beta1/mcp_route.go
deepika0326: Thanks @nacx! Quick context on the prompt selector before I split anything out:

`promptSelector` isn't really a separate feature here — it's the same `prefixMode` change applied consistently. Tools and prompts are both **name-addressed** (`tools/call`, `prompts/get`, and `completion/complete`'s `ref/prompt` all key off a bare `name`), so they share the exact same `<backendName>__` prefixing and collision model. For `prefixMode: Never` to safely expose bare prompt names, it needs an enumerated, admission-validated unique set — which is just a 1:1 mirror of the existing `toolSelector.include`. Leaving prompts out would mean `Never` mode is inconsistent (bare tools, but still-prefixed prompts) on the same route.

I intentionally left **resources** out because they're **URI-addressed** rather than name-addressed — they're rewritten differently (both name and URI template), so "exposing bare" has different collision semantics, and there's no existing selector to mirror (I kept the `// TODO: resource selectors` in place for that reason). That felt like genuinely separate work; prompts did not.

Given prompts ride the same code path as tools rather than adding a new concept, do you still prefer I move them to a separate PR? Happy to split it out if so — just wanted to confirm since it'd leave `Never` mode only half-applied in the meantime.

nacx: Sounds good. Thanks!

nacx (approved):

PR #2604, opened 2026-08-31
fix: keep rotated credentials fresh in the filter config

### LINKED ISSUE #2603: stale token embedded in filter config due to informer cache lag after secret rotation
Description:

After upgrading to v1.1.0, the ai-gateway-controller can push a stale GCP Vertex access token to the ext-proc filter config, causing Vertex API requests to fail with HTTP 401.
Expected behavior: when gcp-token-rotator writes a fresh token to the K8s secret, reconcileFilterConfigSecret should read the newly written token value so ext-proc uses a valid bearer token on subsequent Vertex requests.

Repro steps:

1. Deploy ai-gateway v1.1.0 with a BackendSecurityPolicy referencing a GCP Vertex credential secret.
2. Wait for the gcp-token-rotator to rotate the token (fires every ~55 minutes).
3. Observe that within seconds of the rotation, reconcileFilterConfigSecret reconciles the filter config secret — but reads the pre-rotation token from the informer cache (cache lag window).
4. ext-proc picks up the stale token via the 5s file watcher and begins forwarding it as the bearer token on Vertex requests.
5. Vertex rejects requests with HTTP 401 until the next controller restart or reconcile that catches the cache update.

Root cause: PR #2494 ("feat: resolve backend credentials from the informer cache") changed reconcileFilterConfigSecret to read secrets from the informer cache rather than the API server directly. When a secret is updated, the reconcile can fire before the cache reflects the write, embedding the stale value.

Environment:

 - ai-gateway version: v1.1.0 (introduced by PR #2494)
 - Envoy Gateway version: v1.8.3
 - Envoy proxy: v1.38.3
 - Kubernetes: GKE
 - Credential type: GCP Vertex AI access token (1h TTL, rotated every 55 min)

Logs:

 Controller logs showing the race — token rotation fires, reconcile imle cache read:
 INFO controller.backend-security-policy.gcp-token-rotator  start rotating gcp access token   {"namespace": "abc-llm-gateway", "name": "llm-gateway-vertex"}
 INFO controller.backend-security-policy.gcp-token-rotator  updating gnamespace": "abc-llm-gateway", "name": "llm-gateway-vertex"}
 INFO controller.secret  Reconciling Secret  {"namespace": "abc-llm-gateway", "name": "ai-eg-bsp-llm-gateway-vertex"}
 INFO controller.secret  Reconciling Secret  {"namespace": "llm-internal-..."}

 Envoy access logs showing resulting 401s:
 {":authority":"us-xxxx-aiplatform.googleapis.com","response_code":40"via_upstream","duration":205, ...}

 401s begin ~10 minutes after the rotation fires and persist until a c fresh API server read.

kanurag94:
**Description**

After #2494 a rotated credential can sit one generation behind in the filter config, permanently.

What happens:
- `BackendSecurityPolicyController.reconcile` rotates a credential, then rebuilds the filter config in the same pass.
- The rebuild read the secret from the informer cache. The fan-out to it is in-process; the cache update needs a watch round trip.
- The fan-out wins, so the config gets the pre-rotation token. That token is valid for `preRotationWindow`, so requests fail about five minutes after a rotation, not at it.

Why nothing corrects it:
- The Secret watch fires. `secretController.syncSecret` looks policies up by `<secretName>.<namespace>`.
- `backendSecurityPolicyIndexFunc` never emitted the generated secret's name. AWS and Azure OIDC emitted `<policyName>.<namespace>`, which matches nothing; GCP WIF, Azure ClientSecretRef and GCP CredentialsFile emitted nothing for it.
- No policy is found, no rebuild runs, and the next rotation loses the same race. All five rotating configurations were affected, not just the reported GCP WIF one.

Fix:
- `backendSecurityPolicyIndexFunc` emits both keys, the referenced secret and the generated one, with the generated name from `getBSPGeneratedSecretName`.
- When the changed secret is a policy's generated one, `secretController` syncs that policy's targets directly. That reconcile was woken by the secret's own watch event, so its informer store already holds the new credential.
- A reconcile that rotated no longer fans out itself, so the config is never built from a stale read.

Credential reads are unchanged: `getSecretData` still resolves from the informer cache.

Also fixed:
- `getBSPGeneratedSecretName` returned `""` for Azure ClientSecretRef and GCP CredentialsFile, which do rotate. The owner-reference block is gated on that name, so their generated secrets were orphaned on policy delete.
- It also panicked on an unrecognised type, and now runs inside a field indexer, which is outside reconcile panic recovery. It returns `""`, and `rotateCredential` asserts the name exists.
- The owner reference is applied with a merge patch. `Update` from a cached read taken right after the rotator's write could conflict on `resourceVersion`.
- A `NotFound` on that read is no longer an error. The rotator's create is indexed to the policy, so its watch event brings the reconcile back.
- The index emitted a stray `""` key for policies that reference no secret.

Fixes #2603
Caused by #2494
Related: #2605

- Upgrade note: generated secrets for Azure ClientSecretRef and GCP CredentialsFile policies gain an owner reference on the first reconcile after upgrade, so they are garbage collected with the policy instead of being left behind.
- The targetRefs fan-out moved into `syncBackendSecurityPolicyTargets` so both controllers share it. Mechanical.
- The watch event is now the only route to a post-rotation rebuild, where before two paths reached it. The surviving one reads a cache that has observed the write.
- This covers most of #2605. What remains is `rotateCredential` returning early on rotation failure, and the missing AWS mutual exclusion between `credentialsFile` and `oidcExchangeToken`.

### DIFF
--- a/internal/controller/backend_security_policy.go
+++ b/internal/controller/backend_security_policy.go
@@ -118,9 +118,17 @@ func (c *BackendSecurityPolicyController) reconcile(ctx context.Context, bsp *ai
 	}
 
 	if requiresRotation {
-		res, err = c.rotateCredential(ctx, bsp)
+		var rotated bool
+		res, rotated, err = c.rotateCredential(ctx, bsp)
 		if err != nil {
 			return res, err
+		}
+		if rotated {
+			// The rotator wrote a new credential. Its watch event drives the rebuild from
+			// secretController, where the informer has already observed the write. Fanning out here
+			// as well would rebuild once from a cache that has not, publishing the previous
+			// credential until the second rebuild replaced it.
+			return res, nil
 		}
 	}
 	err = c.syncBackendSecurityPolicy(ctx, bsp)
@@ -128,7 +136,7 @@ func (c *BackendSecurityPolicyController) reconcile(ctx context.Context, bsp *ai
 }
 
 // rotateCredential rotates the credentials using the access token from OIDC provider and return the requeue time for next rotation.
-func (c *BackendSecurityPolicyController) rotateCredential(ctx context.Context, bsp *aigv1b1.BackendSecurityPolicy) (res ctrl.Result, err error) {
+func (c *BackendSecurityPolicyController) rotateCredential(ctx context.Context, bsp *aigv1b1.BackendSecurityPolicy) (res ctrl.Result, rotated bool, err error) {
 	var rotator rotators.Rotator
 
 	switch bsp.Spec.Type {
@@ -139,10 +147,10 @@ func (c *BackendSecurityPolicyController) rotateCredential(ctx context.Context, 
 			roleArn := bsp.Spec.AWSCredentials.OIDCExchangeToken.AwsRoleArn
 			rotator, err = rotators.NewAWSOIDCRotator(ctx, c.client, nil, c.kube, c.logger, bsp.Namespace, bsp.Name, preRotationWindow, oidc, roleArn, region)
 			if err != nil {
-				return ctrl.Result{}, err
+				return ctrl.Result{}, false, err
 			}
 		} else {
-			return ctrl.Result{}, nil
+			return ctrl.Result{}, false, nil
 		}
 	case aigv1b1.BackendSecurityPolicyTypeAzureCredentials:
 		clientID := bsp.Spec.AzureCredentials.ClientID
@@ -155,11 +163,11 @@ func (c *BackendSecurityPolicyController) rotateCredential(ctx context.Context, 
 			var oidcProvider tokenprovider.TokenProvider
 			oidcProvider, err = tokenprovider.NewOidcTokenProvider(ctx, c.client, oidc)
 			if err != nil {
-				return ctrl.Result{}, err
+				return ctrl.Result{}, false, err
 			}
 			provider, err = tokenprovider.NewAzureTokenProvider(ctx, tenantID, clientID, oidcProvider, options)
 			if err != nil {
-				return ctrl.Result{}, err
+				return ctrl.Result{}, false, err
 			}
 		} else if secretRef := bsp.Spec.AzureCredentials.ClientSecretRef; secretRef != nil {
 			secretNamespace := bsp.Namespace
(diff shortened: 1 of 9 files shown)


nacx (approved):

internal/controller/controller.go
```diff
@@ -469,40 +470,59 @@ func aiGatewayRouteIndexFunc(o client.Object) []string {
...
+			return secretRefKey(k.SecretRef, ns)
+		}
 	case aigv1b1.BackendSecurityPolicyTypeAnthropicAPIKey:
-		apiKey := backendSecurityPolicy.Spec.AnthropicAPIKey
-		key = getSecretNameAndNamespace(apiKey.SecretRef, backendSecurityPolicy.Namespace)
+		if k := bsp.Spec.AnthropicAPIKey; k != nil {
+			return secretRefKey(k.SecretRef, ns)
+		}
+	case aigv1b1.BackendSecurityPolicyTypeAWSCredentials:
+		if aws := bsp.Spec.AWSCredentials; aws != nil && aws.CredentialsFile != nil {
```
PatilHrushikesh, line 504: add check for OIDCExchangeToken as well

internal/controller/backend_security_policy.go
```diff
@@ -118,17 +118,25 @@ func (c *BackendSecurityPolicyController) reconcile(ctx context.Context, bsp *ai
...
+			return res, nil
+		}
 	}
 	err = c.syncBackendSecurityPolicy(ctx, bsp)
 	return res, err
 }
 
 // rotateCredential rotates the credentials using the access token from OIDC provider and return the requeue time for next rotation.
-func (c *BackendSecurityPolicyController) rotateCredential(ctx context.Context, bsp *aigv1b1.BackendSecurityPolicy) (res ctrl.Result, err error) {
+func (c *BackendSecurityPolicyController) rotateCredential(ctx context.Context, bsp *aigv1b1.BackendSecurityPolicy) (res ctrl.Result, rotated bool, err error) {
```
PatilHrushikesh, line 139: nit: can this function be split for each provider hanlding?

PR #2299, opened 2026-06-27
test: add OpenRouter real-provider data-plane coverage

amarrtech:
**Description**
Adds OpenRouter as an OpenAI-compatible real-provider data-plane test target. This wires TEST_OPENROUTER_API_KEY into the credential context, adds an OpenRouter backend using the OpenAI-compatible v1 schema, adds one gated chat-completions health-check case with openai/gpt-4o-mini, and exposes ENVOY_AI_GATEWAY_OPENROUTER_API_KEY to the data-plane CI job.

Closes #2272.

**Testing**
- go test ./internal/testing
- go test ./tests/data-plane -run '^$'
- TEST_OPENROUTER_API_KEY= go test ./tests/data-plane -run '^TestWithRealProviders/health-checking/chat/completions/openrouter$' -count=1 -v
- make precommit
- make test
- make test-data-plane

**Notes**
This PR is additive test infrastructure only; no production code paths are changed. The OpenRouter test skips unless TEST_OPENROUTER_API_KEY is set.

I used Codex/AI assistance to help prepare this change and reviewed the implementation and test results before submitting.

PR #2607, opened 2026-09-01
fix: write the base URL path to schema prefix, not version

### LINKED ISSUE #2606: aigw run zero-config discards the path in OPENAI_BASE_URL / ANTHROPIC_BASE_URL and always emits prefix: v1
_Description_:

The docs describe the CLI as experimental and under active development, so this is a report rather than a complaint — but it fails silently, which seemed worth writing down.

When `aigw run` generates its own configuration from environment variables, the path component of `OPENAI_BASE_URL` (and `ANTHROPIC_BASE_URL`) is discarded. The generated `AIServiceBackend` always carries `prefix: v1`, so any OpenAI-compatible provider that does not serve its API at the root is misrouted. Nothing is logged, and the request reaches the provider's website rather than its API. With OpenRouter (base path `https://openrouter.ai/api/v1`) the reply is that site's HTML 404 page wrapped in an `OpenAIBackendError`.

Reproduced on both the v1.1.0 `aigw-darwin-arm64` release binary and `envoyproxy/ai-gateway-cli:latest`, which reports `Envoy AI Gateway CLI: dev`.

```bash
OPENAI_API_KEY=dummy OPENAI_BASE_URL=https://openrouter.ai/api/v1 aigw run --run-id=probe
grep -A3 'schema:' ~/.local/state/aigw/runs/probe/envoy-ai-gateway-resources/config.yaml
```

Observed:

```yaml
        schema:
          name: OpenAI
          prefix: v1
```

Expected `prefix: api/v1`. Only the scheme, host and port reach the generated `Backend`.

Measured across several values, each with its own `--run-id`:

| variable | value | generated backend | generated prefix |
|---|---|---|---|
| `OPENAI_BASE_URL` | `https://openrouter.ai/api/v1` | `openrouter.ai:443` | `v1` |
| `OPENAI_BASE_URL` | `http://localhost:8099/deep/nested/v2` | `localhost:8099` | `v1` |
| `OPENAI_BASE_URL` | `https://api.openai.com/v1` | `api.openai.com:443` | `v1` |
| `ANTHROPIC_BASE_URL` | `https://example.com/api/v1` | `example.com:443` | `v1` |
| `ANTHROPIC_BASE_URL` | `https://api.anthropic.com/v1` | `api.anthropic.com:443` | `v1` |

Not provider-specific, and nested paths collapse to `v1` as well.

End to end, identical on the release binary and on the `dev` image:

```bash
OPENAI_API_KEY=$OPENROUTER_API_KEY OPENAI_BASE_URL=https://openrouter.ai/api/v1 aigw run
curl -H 'Content-Type: application/json' -XPOST http://localhost:1975/v1/chat/completions \
  -d '{"model":"google/gemini-3.5-flash-lite","messages":[{"role":"user","content":"say ok"}]}'
# 404, body is openrouter.ai's HTML page
```

Writing the configuration by hand and setting `prefix: api/v1` on the `AIServiceBackend` works correctly against the same provider, so the `prefix` plumbing itself is fine. It is the environment-variable autoconfiguration that does not populate it.

I have not read the source closely enough to point at a line, so I have not opened a PR. `internal/autoconfig` and its URL parsing helper look like the right area. I am happy to open one given a steer on which behaviour you want: carry the parsed path through to `prefix`, or reject a base URL whose path is not `/v1` with an explanatory error.

Two smaller things that would help under either choice:

- log the effective upstream URL, or the generated `prefix`, at startup, so this is visible without reading the run-state file
- the `aigw run` docs show only root-level examples such as `http://localhost:11434/v1`, so nothing currently signals that the path is dropped

_Relevant Links_:

- https://aigateway.envoyproxy.io/docs/cli/aigwrun/ documents `OPENAI_BASE_URL` and `ANTHROPIC_BASE_URL`
- https://aigateway.envoyproxy.io/docs/cli/ notes the CLI is experimental
- #2272 proposes OpenRouter as a data-plane test provider; that provider's `/api/v1` base path is exactly this case

thomasbergernz:
**Description**

`aigw run` autoconfiguration parses the path out of `OPENAI_BASE_URL` and `ANTHROPIC_BASE_URL` correctly, but renders it into the `AIServiceBackend` `schema.version` field. `VersionedAPISchema` documents that field as ignored for the OpenAI and Anthropic schemas, and points callers at `prefix` for custom request paths:

> This field is ignored for OpenAI, AWSBedrock, GCPVertexAI, and Anthropic.
> For OpenAI and Anthropic, use prefix to configure custom request paths.

So the parsed path had no effect and the upstream request used the default `/v1`. Any OpenAI-compatible provider that does not serve its API at the root was silently misrouted.

This renders `prefix` instead for OpenAI and Anthropic. `AzureOpenAI` keeps `version`, where it carries `OPENAI_API_VERSION` and is meaningful.

Before, with `OPENAI_BASE_URL=https://openrouter.ai/api/v1`:

```yaml
  schema:
    name: OpenAI
    version: "api/v1"     # ignored for OpenAI; upstream path stays /v1
```

After:

```yaml
  schema:
    name: OpenAI
    prefix: "api/v1"
```

Fixes #2606
Related PR: #2299, which adds OpenRouter data-plane coverage and already uses the correct `filterapi.VersionedAPISchema{Name: APISchemaOpenAI, Prefix: "api/v1"}`. That PR is unaffected by this bug, since it builds its backend config directly rather than through autoconfiguration.

Verified against the real provider, not only in tests. Building `cmd/aigw` from this branch and running:

```bash
OPENAI_API_KEY=<key> OPENAI_BASE_URL=https://openrouter.ai/api/v1 aigw run
curl -H 'Content-Type: application/json' -XPOST http://localhost:1975/v1/chat/completions \
  -d '{"model":"google/gemini-3.5-flash-lite","messages":[{"role":"user","content":"say ok"}]}'
```

On `main` this returns 404 carrying openrouter.ai's HTML page, wrapped in an `OpenAIBackendError`. With this change the request reaches the API and is answered there; with a deliberately invalid key it comes back as OpenRouter's JSON 401, which is the expected result.

The `openrouter` and `llamastack` golden files asserted the ineffective `version` field and are updated. I added an Anthropic case, since that schema had no golden coverage with a non-default base URL; reverting the Anthropic half of the template change fails it with `-prefix +version`.

`make check` and `make test` pass locally on go1.26.6, matching go.mod. On go1.27 five pre-existing tests in `internal/apischema/openai` and `internal/translator` fail, on `main` as well as here, because sonic falls back to `encoding/json` above go1.26.

One thing I left alone, in case you would rather it were done differently: the field carrying this value is still `OpenAIConfig.Version` / `AnthropicConfig.Version`, which now reads oddly given it renders as `prefix`. Renaming it to `Prefix` would touch the struct and a number of test fixtures, so I kept this change to the rendering bug. Happy to follow up if you want that.


nacx (approved):
Thanks! LGTM

Issue #2560, opened 2026-08-19
`ai-gateway-extproc` sidecar is injected and healthy but never attached to the Envoy filter chain — all requests bypass it (no credential injection, no model extraction)


ronyferdiansyah:
## Summary

On a clean install of Envoy AI Gateway v1.0.0, the `ai-gateway-extproc`
sidecar is injected into the Envoy data plane pod by the mutating webhook,
starts successfully, and registers its processors. However it is **never
added to Envoy's HTTP filter chain**, so no request ever reaches it.

Because that filter is what injects the upstream provider credential and
extracts the model name from the request body, the practical effects are:

- requests reach the provider **unauthenticated** → `401 x-api-key header is required`
- `x-ai-eg-model` is never set from the request body, so routing only works
  if the client sets that header manually

Every CR involved reports `Accepted` / `ResolvedRefs: True`, and the
controller logs no errors, so there is no signal that anything is wrong.

## Environment

| Component | Version |
|---|---|
| Envoy AI Gateway | v1.0.0 (`ai-gateway-helm`, `ai-gateway-crds-helm`) |
| ai-gateway-controller | `docker.io/envoyproxy/ai-gateway-controller:v1.0.0` |
| ai-gateway-extproc | `docker.io/envoyproxy/ai-gateway-extproc:v1.0.0` |
| Envoy Gateway | v1.8.1 (also reproduced on v1.8.3) |
| Kubernetes | v1.35.1 (minikube, hyperv driver) |

Envoy Gateway config has `extensionApis.enableBackend: true` and a Redis-backed
global rate limit service enabled.

## Reproduction

1. Install Envoy Gateway v1.8.1 with `extensionApis.enableBackend=true`.
2. Install `ai-gateway-crds-helm` and `ai-gateway-helm` v1.0.0.
3. Apply an `AIGatewayRoute` + `AIServiceBackend` (schema `Anthropic`) +
   `Backend` (fqdn `api.anthropic.com:443`, `tls.wellKnownCACertificates: System`,
   `tls.sni: api.anthropic.com`) + `BackendSecurityPolicy` (type `APIKey`).
4. Send a request to `/v1/messages` with a valid `x-ai-eg-model` header.

**Expected:** the extproc injects the API key; request succeeds.
**Actual:** `401` from the provider — the credential was never injected.

## Evidence

**The sidecar is present and healthy:**

```
[...]

PR #2614, opened 2026-09-02
controller: report when Envoy Gateway never calls the extension server

pjdurden:
**Description**

# Issue #2560 — extproc sidecar injected but never attached to the Envoy filter chain

## 1. Root cause

Provisioning of an AI Gateway data plane is split across two completely independent
mechanisms, and only one of them fails loudly:

1. **The pod mutating webhook** (`internal/controller/gateway_mutator.go`) injects the
   `ai-gateway-extproc` container and the UDS volume into the Envoy pod. It only needs the
   AIGatewayRoute/MCPRoute objects and the filter config secret, all of which the controller
   itself creates. This always succeeds in the reported setup.
2. **The Envoy Gateway extension server hook** `PostTranslateModify`
   (`internal/extensionserver/post_translate_modify.go`) is what actually puts traffic through
   that container. Everything that matters happens inside that one gRPC call: the
   `ai-gateway-extproc-uds` cluster is appended, `insertRouterLevelAIGatewayExtProc` inserts the
   `envoy.filters.http.ext_proc/aigateway` filter into every HCM filter chain, and
   `enableRouterLevelAIGatewayExtProcOnRoute` enables it per route.

Envoy Gateway only calls that hook if it was installed with the `extensionManager`
configuration AI Gateway ships in `manifests/envoy-gateway-values.yaml` (the
`hooks.xdsTranslator.post: [Translation, Cluster, Route]` list, the `translation.*.includeAll`
flags, and the `service.fqdn` pointing at the AI Gateway controller). The issue reporter
installed Envoy Gateway with only `extensionApis.enableBackend=true`, so the hook was never
invoked.

Nothing in AI Gateway observes that. The sidecar is healthy, every CR reports
`Accepted`/`ResolvedRefs: True`, the generated HTTPRoute and HTTPRouteFilters look correct, and
the controller logs no errors — while every request reaches the provider with no credential
injected and no model extracted from the body. This matches the reporter's evidence
(`config_dump | grep -c 'extproc-uds'` → `0`: the UDS cluster is added in the same hook, so its
absence proves the hook never ran) and the maintainer reproduction in the issue thread.

So the defect being fixed here is not a wrong xDS mutation; it is that a fatal misconfiguration
of the install is completely silent.

## 2. The fix and why

Make the missing wiring visible, which is what the maintainer asked for on the issue
("fail visible" instead of silently forwarding unmediated traffic).

* `internal/extensionserver`: the `Server` now records, in an `atomic.Bool`, that Envoy Gateway
  has invoked `PostTranslateModify` at least once, exposed as `Server.PostTranslateModifyInvoked()`.
  That single call is the right signal because every AI Gateway xDS mutation lives inside it.
* `internal/controller/extension_server_check.go`: a new `manager.Runnable` that ticks once a
  minute. Once there is at least one AIGatewayRoute or MCPRoute and the hook still has not been
  invoked, it logs an error naming the consequence (requests bypass credential injection and
  model extraction) and the remedy (install Envoy Gateway with
  `manifests/envoy-gateway-values.yaml`, and check its extension service hostname/port). It
  stops as soon as the hook is invoked, so a healthy install logs nothing at all.
* Wiring: `controller.Options.ExtensionServerInvoked` is an optional `func() bool`; the check is
  only registered when it is set. `cmd/controller/main.go` passes the extension server's method
  (the extension server and the controller manager already run in the same process). The
  standalone `aigw` path (`cmd/aigw/run.go`) leaves it nil and is unaffected.

Deliberately *not* done, and why:

* **No change to the xDS mutation logic.** The insertion path is correct; reproducing the bug
  needs a mis-installed Envoy Gateway, not a different filter-chain walk.
* **No status condition on AIGatewayRoute.** `newConditions` in `internal/controller/controller.go`
  models exactly one condition at a time (`Accepted` XOR `NotAccepted`) driven by reconciliation.
  Flipping routes to `NotAccepted` from a background timer would overload that meaning, would
  fight the reconciler, and would need a new re-trigger path to recover once the wiring appears.
  That is an API-level change, out of scope for this bug.
* **No reading of Envoy Gateway's `envoy-gateway-config` ConfigMap.** That would need new RBAC in
  the Helm chart, assumes a specific EG install layout, and would still miss the cases where the
  config is right but the extension server is unreachable (wrong `service.fqdn`, wrong port,
  NetworkPolicy). Observing the hook itself covers all of them.

## 3. Files changed

| File | Change |
| --- | --- |
| `internal/extensionserver/extensionserver.go` | `postTranslateModifyInvoked atomic.Bool` field + `PostTranslateModifyInvoked()` accessor |
| `internal/extensionserver/post_translate_modify.go` | Record the invocation at the top of `PostTranslateModify` |
| `internal/controller/extension_server_check.go` | **New.** `extensionServerCheck` runnable + `errExtensionServerNotInvoked` |
| `internal/controller/controller.go` | New `Options.ExtensionServerInvoked`; register the runnable via `mgr.Add` |
| `cmd/controller/main.go` | Pass `extSrv.PostTranslateModifyInvoked` into `controller.Options` |
| `internal/extensionserver/extensionserver_test.go` | Sub-test asserting the flag flips on the first hook call |
| `internal/controller/extension_server_check_test.go` | **New.** Tests for `check` and `Start` |

## 4. Risk / uncertainty

* **Multi-replica false positives.** Envoy Gateway keeps a single gRPC connection to the
  extension server Service, so with `controller.replicaCount > 1` a replica that Envoy Gateway is
  not connected to would consider itself unwired. The chart default is `replicaCount: 1`, and the
  runnable is registered with `mgr.Add` without implementing `LeaderElectionRunnable`, so
  controller-runtime runs it only on the leader — at most one replica can emit the message. The
  caveat is documented in the type comment. This is a log line only; it changes no behaviour and
  cannot break traffic.
* **Grace period.** The first check runs one tick (1 minute) after the manager starts, which
  doubles as the window for Envoy Gateway to connect and translate. On a very slow cluster a
  single spurious error line is possible before the hook lands; it stops on the next tick.
* **Scope of the diagnostic.** It detects "the hook never ran at all". It does not detect a hook
  that runs but produces no AI Gateway filter for one particular listener; that would need
  per-Gateway bookkeeping and is a larger change.
* I could not run the e2e suite (needs a cluster), so the end-to-end behaviour of the new log
  line under a deliberately mis-installed Envoy Gateway was not exercised.

## 5. Verification

Go toolchain at `/home/pjsump/go-sdk/go/bin`.

```
$ go build ./...
(no output)

$ go vet ./internal/... ./cmd/...
(no output)

$ go test ./internal/extensionserver/... ./internal/controller/...
ok  github.com/envoyproxy/ai-gateway/internal/extensionserver         0.291s
ok  github.com/envoyproxy/ai-gateway/internal/controller              4.019s
ok  github.com/envoyproxy/ai-gateway/internal/controller/rotators     0.171s
ok  github.com/envoyproxy/ai-gateway/internal/controller/tokenprovider 0.784s

$ go tool -modfile=tools/go.mod golangci-lint run \
    ./internal/controller/... ./internal/extensionserver/... ./cmd/controller/...
0 issues.

$ go tool -modfile=tools/go.mod golangci-lint fmt \
    ./internal/controller/... ./internal/extensionserver/... ./cmd/controller/...
(no files reformatted)
```

New tests, all passing:

* `TestServerPostTranslateModify/records_the_invocation` — `PostTranslateModifyInvoked()` is
  false on a fresh `Server` and true after one `PostTranslateModify` call.
* `TestExtensionServerCheck_check` — five sub-tests: already invoked (returns true, logs
  nothing); not invoked with no routes (silent); not invoked with an AIGatewayRoute (logs
  `errExtensionServerNotInvoked` and `aigatewayroute_count=1`); same for an MCPRoute; a List
  failure logs the list error and does *not* emit the false "not wired" error.
* `TestExtensionServerCheck_Start` — returns when the hook becomes invoked, and returns on
  context cancellation.

`make precommit` / `make test-e2e` were not run: they require Docker and a Kubernetes cluster,
which are not available in this environment.


PatilHrushikesh:
what would be behaviour in case of multiple controller pods?

PR #2636, opened 2026-09-06
fix(mcpproxy): match application/json ignoring charset parameter

### LINKED ISSUE #2568: MCP proxy silently drops backend JSON responses when Content-Type carries a charset parameter (exact string match in session.go)

chiliec:
**Description**

The MCP proxy's per-backend response collector (`session.go`) and the response proxy (`handlers.go`) decided whether to parse a backend response as JSON via an **exact string comparison** on the `Content-Type` header (`== "application/json"`).

A backend responding with `Content-Type: application/json;charset=UTF-8` — the servlet-container default emitted by the official MCP Java SDK's servlet transports (`HttpServletStatelessServerTransport`, which does `setContentType("application/json")` + `setCharacterEncoding("UTF-8")`) — failed that check. The body fell through to the SSE parser, which found no `data:` lines and yielded **zero events and zero errors**, so the backend's `tools/list` response was silently discarded (connected-but-toolless, nothing logged).

This parses the media type with `mime.ParseMediaType` and matches on the `application/json` media type, ignoring parameters like `charset` (per RFC 9110 §8.3; `charset` is meaningless for `application/json` per RFC 8259 §8.1). A small shared `isJSONContentType` helper replaces the three exact-match sites in the package (`sendRequestPerBackend`, `copyProxyHeaders`, `proxyResponseBody`).

Fixes #2568

Added a regression test `TestSendRequestPerBackend_JSONContentTypeWithCharset` mirroring the existing `TestSendRequestPerBackend_BOMPrefixedJSON`, using a backend that returns `application/json;charset=UTF-8`.

Verification (Go 1.26.6, `go test ./internal/mcpproxy/`):

- New test passes with the fix.
- RED→GREEN confirmed: reverting the `session.go` dispatch back to the exact string match makes the new test fail with `Diff: -message / +` (0 events collected); restoring the fix makes it pass.
- Full `internal/mcpproxy` package suite green (`ok ... 2.980s`), `go vet` and `gofmt -l` clean on the touched files.

AI-assisted; all results above are from real local runs. Happy to adjust.

### DIFF
--- a/internal/mcpproxy/handlers.go
+++ b/internal/mcpproxy/handlers.go
@@ -804,7 +804,7 @@ func (m *mcpRequestContext) handleToolCallRequest(ctx context.Context, s *sessio
 }
 
 func copyProxyHeaders(resp *http.Response, w http.ResponseWriter) {
-	isJSONResponse := resp.Header.Get("Content-Type") == "application/json"
+	isJSONResponse := isJSONContentType(resp.Header.Get("Content-Type"))
 	for k, v := range resp.Header {
 		// Skip content-length header for non JSON response since we might modify the response.
 		if !isJSONResponse && strings.EqualFold(k, "content-length") {
@@ -827,7 +827,7 @@ func (m *mcpRequestContext) proxyResponseBody(ctx context.Context, s *session, w
 	// Try to decode as a single JSON-RPC message first; if that fails, fall through to the
 	// SSE parser using the already-read bytes.
 	var sseReader io.Reader = resp.Body
-	if resp.Header.Get("Content-Type") == "application/json" {
+	if isJSONContentType(resp.Header.Get("Content-Type")) {
 		body, err := io.ReadAll(resp.Body)
 		if err != nil {
 			m.l.Error("failed to read response body", slog.String("error", err.Error()))
--- a/internal/mcpproxy/handlers_test.go
+++ b/internal/mcpproxy/handlers_test.go
@@ -1093,7 +1093,7 @@ func TestProxyResponseBody_JSONResponse(t *testing.T) {
 	require.NoError(t, err)
 
 	httpResp := &http.Response{
-		Header:     http.Header{"Content-Type": []string{"application/json"}},
+		Header:     http.Header{"Content-Type": []string{"application/json; charset=utf-8"}},
 		Body:       io.NopCloser(bytes.NewReader(body)),
 		StatusCode: http.StatusOK,
 	}
--- a/internal/mcpproxy/session.go
+++ b/internal/mcpproxy/session.go
@@ -14,7 +14,6 @@ import (
 	"fmt"
 	"io"
 	"log/slog"
-	"mime"
 	"net/http"
 	"os"
 	"strconv"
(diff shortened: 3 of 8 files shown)


nacx (commented):
This fix was already added in https://github.com/envoyproxy/ai-gateway/pull/2587, now merged. We can keep the PR for the additional test, but in htat case, mind deleting the now duplicate `isJSONContentType` method?

PR #2645, opened 2026-09-08
fix: propagate MCPRoute OAuth issuer to SecurityPolicy

### LINKED ISSUE #2641: MCPRoute securityPolicy.oauth.issuer is not propagated to the generated SecurityPolicy JWT provider
## Description

When an `MCPRoute` sets `spec.securityPolicy.oauth.issuer`, the value is accepted by the CRD and shown correctly on `kubectl get mcproute -o yaml`, but it is **not** copied onto `spec.jwt.providers[0].issuer` of the controller-generated `SecurityPolicy`.

As a result, Envoy's `jwt_authn` filter never validates the token's `iss` claim. Only signature and `exp`/`nbf` are enforced, plus `aud` if `audiences` was set (that field *does* propagate).

Envoy Gateway documents this explicitly on `JWTProvider.Issuer`: if the field is omitted, the JWT issuer is not checked. Envoy `jwt_authn` behaves the same way — an empty provider `issuer` skips `iss` matching.

## Impact

Any validly-signed, unexpired token from the configured JWKS source is accepted regardless of which authorization server issued it.

For routes where `audiences` is also unset, effective auth is signature + expiry only, with no issuer or audience binding — materially weaker than what the `MCPRoute` manifest declares.

The practical risk is highest when the JWKS can verify tokens from more than one issuer (shared / multi-tenant JWKS, e.g. Azure AD common, some Okta/Keycloak layouts, or a `remoteJWKS.uri` that serves keys for multiple issuers). If the JWKS is unique to one AS, a token from a different AS with different keys still fails signature check — but `iss` is still not enforced as declared.

`issuer` is still used for JWKS auto-discovery and for advertising `authorization_servers` in protected-resource metadata. Clients are told which AS to use; the gateway does not pin incoming tokens to that AS.

This also contradicts the intended contract described in the in-tree token-exchange proposal: `MCPRoute.spec.securityPolicy.oauth.issuer` is supposed to be the exact `iss` value the JWT filter matches.

## Root cause

In `ensureSecurityPolicy` (`internal/controller/mcp_route_security_policy.go`), the JWT provider is constructed as:

```go
jwtProvider := egv1a1.JWTProvider{
    Name:      name,
    Audiences: oauth.Audiences,
}
```

`Audiences`, `RemoteJWKS`/`LocalJWKS`, and `ClaimToHeaders` are assigned. `oauth.Issuer` is not.

This is isolated to the JWT provider path. It is not a broader compilation failure and does not depend on JWKS config shape (`remoteJWKS.uri` only, `remoteJWKS.backendRefs`, or `localJWKS`).

Controller tests set `Issuer` on the `MCPRoute` but never assert `securityPolicy.Spec.JWT.Providers[0].Issuer`, so the omission is untested.

## Steps to reproduce

1. Apply an `MCPRoute` with `securityPolicy.oauth.issuer` and `securityPolicy.oauth.jwks` set (with or without `audiences`).

2. Inspect the generated `SecurityPolicy` (same namespace, name `ai-eg-mcp-<mcproute-name>`):

```bash
kubectl get securitypolicy -n <ns> ai-eg-mcp-<mcproute-name> \
  -o jsonpath='{.spec.jwt.providers[0]}'
```

3. Observe `spec.jwt.providers[0]` has no `issuer` field, even though it was set on the source `MCPRoute`.

Reproduced across independently-authored `MCPRoute`s using both `remoteJWKS.uri`-only and `remoteJWKS.backendRefs`-based JWKS configs.

## Expected behavior

`spec.securityPolicy.oauth.issuer` on the `MCPRoute` should be copied to `spec.jwt.providers[0].issuer` on the generated `SecurityPolicy`, so `jwt_authn` enforces `iss` matching as declared.

## Actual behavior

`issuer` is omitted when constructing `egv1a1.JWTProvider` in `ensureSecurityPolicy`. `audiences`, `claimToHeaders`, and JWKS compile correctly.

## Suggested fix

Assign the field when building the provider:

```go
jwtProvider := egv1a1.JWTProvider{
    Name:      name,
    Issuer:    oauth.Issuer,
    Audiences: oauth.Audiences,
}
```

Add a unit assertion that `Providers[0].Issuer == oauth.Issuer`.

## Environment

- Envoy AI Gateway: observed in current `main` (`internal/controller/mcp_route_security_policy.go`)
- Envoy Gateway: v1.8.1 (dependency; `JWTProvider.Issuer` is optional and skips `iss` check when empty)

## Additional notes

Given the MCP spec's reliance on `iss`/`aud` for resource binding ([RFC 8707](https://datatracker.ietf.org/doc/html/rfc8707) / [RFC 9728](https://datatracker.ietf.org/doc/html/rfc9728)), this silently weakens the security posture of any `MCPRoute` that relies on the CRD's declared `issuer` for authorization-server pinning.

czczycz:
**Description**

Propagate `MCPRoute.spec.securityPolicy.oauth.issuer` to the generated `SecurityPolicy` JWT provider so Envoy Gateway validates the JWT `iss` claim.

Add regression assertions for explicit and auto-discovered JWKS configurations.

Fixes #2641

Tested with: `go test ./internal/controller -run ^TestMCPRouteController_syncMCPRouteSecurityPolicy$ -count=1`


Hritik003 (commented):
lgtm, can the CI be fixed?

nacx (approved):

aish1331 (approved):
LGTM, thanks for taking this up @czczycz!

PR #2660, opened 2026-09-09
feat: support AWS OpenAI-compatible APIs

xiaolin593:
**Description**

Adds `AWSOpenAI` as an OpenAI-compatible backend schema for the Chat Completions and Responses APIs. It defaults the AWS path prefix to `openai/v1`, supports custom prefixes and model overrides, and wires the schema through controller conversion and endpoint translator selection.

AWS SigV4 authentication signs `bodyMutation.GetBody()`, so the AWS-specific translators always emit the request body. This ensures Envoy forwards the exact payload bytes covered by the signature while leaving ordinary OpenAI passthrough behavior unchanged.

N/A

Validated with:

`GOTOOLCHAIN=auto GOSUMDB=sum.golang.org go test ./internal/controller ./internal/endpointspec ./internal/filterapi ./internal/translator`


aabchoo (approved):
lgtm

PR #2663, opened 2026-09-09
feat(mcp): introduce injectionPolicy for MCP API key handling in per-backend security policy

### LINKED ISSUE #2661: Allow MCP backend API key injection to preserve a forwarded credential header
_Description_:

When an MCP backend has both a security policy API key and `forwardHeaders` that land on the same header (typically `Authorization`), the gateway always overwrites the forwarded value with the configured credential.

That prevents a common setup: a default least-privilege service account on the backend, overridden by a caller-supplied personal access token when the client sends one. `forwardHeaders` is how clients inject that token today (usually via a dedicated header renamed with `backendHeader`), but the security policy wins and the backend only ever sees the service account.

**Desired behavior**

Add an `overwrite` field on `backendRef.securityPolicy.apiKey`

- `overwrite: true` (default) — always inject the configured API key. Current behavior; existing routes unchanged.
- `overwrite: false` — inject the API key only if the target header is absent. If `forwardHeaders` already populated that header, keep the client value.

Example:

```yaml
kind: MCPRoute
spec:
  backendRefs:
    - name: github
      kind: Backend
      group: gateway.envoyproxy.io
      securityPolicy:
        apiKey:
          secretRef:
            name: github-sa-token   # default least-privilege service account
          overwrite: false
      forwardHeaders:
        - name: X-GitHub-PAT
          backendHeader: Authorization
```

If the client sends `X-GitHub-PAT`, the backend receives that token as `Authorization`. If it does not, the gateway injects the service-account key.

**How it would be used**

- Default SA for `initialize` / `tools/list` and clients that have no PAT.
- User PAT for tool calls that need finer-grained backend access.
- Per backend: one backend on a route can set `overwrite: false`; others keep the default.

**Notes**

- Default must remain overwrite (`true`) so this is opt-in.
- `overwrite` is header injection only; it should not apply to `queryParam`.
- When the MCPRoute itself uses OAuth/API-key client auth, callers should **not** forward inbound `Authorization` (that is the gateway token). Use a dedicated client header and `backendHeader` to map it onto the backend credential header.
- Envoy Gateway `HTTPCredentialInjectionFilter` already has `overwrite`; secret-ref injection can wire this field through. Inline keys currently use `RequestHeaderModifier` `set`, which always replaces, so `overwrite: false` needs an explicit rule or a different injection path for inline.

[optional *Relevant Links*:]

- [MCP header forwarding](https://gateway.envoyproxy.io/docs/capabilities/mcp/#header-forwarding)
- Envoy Gateway `HTTPCredentialInjectionFilter.overwrite` (inject only if the header is missing when `false`)
- `MCPBackendAPIKey` in `api/v1alpha1/mcp_route.go` / `api/v1beta1/mcp_route.go`

aish1331:
**Description**

#### Problem

When an MCP backend has both a security policy API key and forwardHeaders that land on the same header (typically `Authorization` but can be any custom header key), the gateway always overwrites the forwarded value with the configured credential.

That prevents a common setup: a default least-privilege service account on the backend, overridden by a caller-supplied personal access token when the client sends one. forwardHeaders is how clients inject that token today (usually via a dedicated header renamed with backendHeader), but the security policy wins and the backend only ever sees the service account.

#### Fix

This PR adds an overwrite field on `backendRef.securityPolicy.apiKey` so a default backend credential can coexist with a caller-supplied token forwarded onto the same header.

This update adds an `InjectionPolicy` enum to the `MCPBackendAPIKey` struct (`backendRef.securityPolicy.apiKey`), controlling whether the injected API key replaces existing values in the target header. 
The default behavior is to overwrite the already present header (InjectionPolicy = Always), but when set to `IfNotPresent`, the API key is only injected if the header is absent, allowing for caller-supplied tokens to be preserved. 

This change includes validation rules to ensure that `InjectionPolicy` cannot be `IfNotPresent` when using query parameters, as query-parameter injection always rewrites the backend URL.

Additionally, tests have been added to verify the behavior of the new `InjectionPolicy` field in various scenarios, ensuring correct handling of inline and secret reference API keys.

Changes:
- Updated `MCPBackendAPIKey` struct to include `InjectionPolicy` field.
- Added validation rules for `InjectionPolicy` in the CRD.
- Enhanced controller logic to respect the `InjectionPolicy` setting.
- Expanded unit and integration tests to cover new functionality.

Fixes https://github.com/theagentrouter/agent-router/issues/2661

Related Issues:

https://github.com/theagentrouter/agent-router/issues/1966

- Default remains Always; this is opt-in per backend.
- Only the Authorization header gets a Bearer prefix; custom headers (X-Api-Key, X-GitHub-Token, …) are injected verbatim.
- If the MCPRoute itself uses OAuth or API-key client auth, callers must not forward inbound Authorization. Use a dedicated client header and backendHeader to map the backend token. This should be documented.


api/v1beta1/mcp_route.go
```diff
@@ -378,6 +379,19 @@ type MCPBackendAPIKey struct {
...
+	// header already set on the request, including values populated by forwardHeaders.
+	// When false, the API key is injected only if the target header is absent, so a
+	// caller-supplied token forwarded onto the same header is preserved.
+	//
+	// Overwrite applies only to header injection. It must not be set to false when queryParam
+	// is used, because query-parameter injection always rewrites the backend URL.
+	//
+	// +kubebuilder:default=true
+	// +optional
+	Overwrite *bool `json:"overwrite,omitempty"`
```
gavrissh: If the default is to have overwrite to be true. The field does not make sense as it was the behaviour before the client override option.

We should rather have a field which describes whether the client set keys are allowed to be passed?

aish1331: Yes, booleans with a default True are odd and confusing. 
I have the following reasons to add a field here:

1. There's only one BackendSecurity Policy header, forward headers is an array. The overwrite field on forwardHeaders would only be valid for 1 header at most.
2. The client headers are populated earlier in the filter chain. Adding a `overwrite` there doesn't make sense from a code perspective, as there's nothing to be overwritten at that time. But that's an implementation detail, and users need not know about the order.

Instead of `overwrite`, I can use an enum `injectionPolicy` with values `IfNotPresent` and `Always` as the default. WDYT?

nacx (commented):
Thanks! Ovewrall LGTM

nacx (approved):

PR #2664, opened 2026-09-09
fix: charge per-tenant token cost for Distinct bucket rules

### LINKED ISSUE #2551: bucketRules with type: Distinct never charge real per-tenant token cost, silently pooled into a shared bucket instead
## Description

The docs recommend `type: Distinct` for per-tenant budgets ("a per-tenant budget keyed on the tenant ID"). With token-based (quota_mode) buckets this doesn't work: real token cost is never charged to each tenant's own bucket, it's pooled into one shared bucket per rule.

**Expected**: each tenant's real token cost should be attributed to that tenant's own counter, matching the per-tenant check that already works correctly at pre-flight time.

**Root cause**: `buildStreamDoneHeaderMatchAction` (internal/extensionserver/quota_ratelimit.go#L873-L887) unconditionally returns a `GenericKey` action (fixed literal value) whenever `header.Type == Distinct`. The pre-flight action for the same header (`buildHeaderMatchAction`, #L911-L926) correctly uses `RequestHeaders` and reads the real value. Only the stream-done action, the one carrying `hits_addend` (the real token cost), has this fallback. Each tenant's counter only ever increments by the pre-flight default of 1 per request; the real cost from every tenant sharing a rule lands in one shared, genericly-keyed bucket.

The doc comment states this is intentional ("Distinct headers are treated as GenericKey because per-value bucketing is not applicable at stream-done time"). We think this may not fully hold: `buildDynamicMetadata` (internal/extproc/processor_impl.go#L819) runs at stream-done and already takes `requestHeaders` as a parameter, reading from it directly, so the real value is in scope exactly where the fallback happens. We also found `internal/extensionserver/header_to_metadata.go` already wires Envoy's native `header_to_metadata` filter to copy a header's real value into dynamic metadata before the rate limit filter runs, and it's already used today for an unrelated header, in the position a fix like this would need.

That said, PR #2347 (open) reframes QuotaPolicy as platform/backend-scoped, distinct from usage-based rate limiting, which is framed as the consumer/client-scoped mechanism. Per-tenant `Distinct` budgets may belong there instead. We confirmed Envoy Gateway's own rate limit translator (internal/xds/translator/ratelimit.go, envoyproxy/gateway) handles `Distinct` + stream-done cost correctly, it builds one set of descriptor actions and reuses them for both the pre-flight and response-cost rules, rather than splitting into two builders like ai-gateway does.

So we're unsure which fix is more appropriate:
1. Fix QuotaPolicy's stream-done handling for Distinct headers, or
2. Document Distinct + quota_mode as unsupported and point readers at usage-based rate limiting instead, consistent with PR #2347's scoping (which would also need updating there, since it still uses "a separate budget per tenant" as a QuotaPolicy example).

Happy to help test a PR either way.

## Repro steps

1. Configure a bucketRule with `type: Distinct` on a header (e.g. `x-tenant-id`), quota_mode enabled.
2. Send requests under two different header values with known, distinct token costs.
3. Inspect the Redis counters.

Observed: each tenant's own counter (keyed by the real value) increments by exactly 1 per request, regardless of token cost. A separate key, named after the descriptor key itself rather than a real value, accumulates the sum of the real costs from both tenants.

## Environment

ai-gateway v1.0.0, Envoy Gateway v1.8.3, Kubernetes.

## Logs

Ratelimit service debug log for the pre-flight check, showing the real header value correctly captured at that stage:

```
level=debug msg="ShouldRateLimit: domain:\"ai-gateway-quota\" descriptors:{entries:{key:\"rule-0-x-tenant-id-match-0\" value:\"tenant-a\"}} descriptors:{entries:{key:\"rule-1-match--1\" value:\"rule-1-match--1\"}}}"
```

The corresponding stream-done (hits_addend) call never appears under the real tenant's identity, since it uses the generic literal key instead. The definitive evidence is in Redis: the key for `tenant-a` holds `1` (the request count), while a separate key named after the literal descriptor key (`rule-0-x-tenant-id-match-0`) holds the sum of real token costs from every tenant sharing that rule.

vignesh-chaturvedi:
**Description**

`buildStreamDoneHeaderMatchAction` emits a `GenericKey` action for a `Distinct`
header, whose descriptor value is a constant. The request-time builder,
`buildHeaderMatchAction`, emits `RequestHeaders` for the same header, keyed on
the caller's actual value.

The two phases of the same rule therefore never resolve to the same bucket:

| Phase | Action | Descriptor value |
|---|---|---|
| Request-time (pre-flight) | `RequestHeaders` | the caller's `x-tenant-id` |
| Stream-done (real cost) | `GenericKey` | a constant |

Each tenant's counter takes only the pre-flight increment, while the real token
cost from every tenant sharing the rule settles into one pooled generic bucket.
That matches the Redis counters described in the issue.

This mirrors the request-time action so stream-done settles against the bucket
its pre-flight check reserved. One branch in one function.

Tests: `TestBuildStreamDoneHeaderMatchAction` covers a `Distinct` header
producing `RequestHeaders`, and an inverted `Exact` header still reporting with
`ExpectMatch=true`. `TestStreamDoneDescriptorsMatchRequestTime` asserts the
request-time and stream-done actions agree on both the descriptor kind and the
key. Comparing keys alone is not enough: `GenericKey` and `RequestHeaders` share
a key and differ only in the value, which is exactly how this hid.

Fixes #2551

The comment I removed claimed the fallback was deliberate: "Distinct headers
fall back to GenericKey because per-value bucketing is not applicable at
stream-done time." I think that rationale is wrong and would rather surface the
disagreement than quietly flip it. `ApplyOnStreamDone` entries are `RateLimit`
actions on the same route as the request-time ones, so the phase controls when
the report is sent rather than which actions can be evaluated, and a
`RequestHeaders` action resolves the same way it does pre-flight.

If the intent really was that per-value bucketing is unsupported here, the
descriptors still should not silently disagree, and the better change is the
issue's second option: reject or warn on `Distinct` combined with a token-based
quota mode rather than pooling the cost. Happy to take it that way instead.

The old behavior had no test coverage. The existing stream-done assertions only
exercise `Exact` headers and the default bucket, which is why this change breaks
nothing. Both new tests fail if the `GenericKey` branch is restored, and the
`Exact` subtest correctly keeps passing.

`make precommit` is clean and `go test ./internal/...` passes. I have not run
this against a live Redis-backed cluster, so the descriptor pairing is verified
at config-generation level only. If someone has a cluster handy, confirming the
per-tenant counters would be a good extra check.

I used an AI assistant while investigating and writing this change, noted per
the generative AI policy in the contributing guide. I have reviewed and
understand the diff and can revise it in review.

### DIFF
--- a/internal/extensionserver/quota_ratelimit.go
+++ b/internal/extensionserver/quota_ratelimit.go
@@ -826,8 +826,7 @@ func buildClientSelectorActions(
 }
 
 // buildClientSelectorStreamDoneActions is like buildClientSelectorActions but
-// always uses ExpectMatch=true on HeaderValueMatch actions. Distinct headers fall
-// back to GenericKey because per-value bucketing is not applicable at stream-done time.
+// always uses ExpectMatch=true on HeaderValueMatch actions.
 func buildClientSelectorStreamDoneActions(
 	ruleIndex int, selectors []egv1a1.RateLimitSelectCondition,
 ) []*routev3.RateLimit_Action {
@@ -869,18 +868,21 @@ func flattenAndSortClientSelectorHeaders(selectors []egv1a1.RateLimitSelectCondi
 }
 
 // buildStreamDoneHeaderMatchAction is like buildHeaderMatchAction but always uses
-// ExpectMatch=true. Distinct headers are treated as GenericKey.
+// ExpectMatch=true, so an inverted match still reports the cost it incurred.
 func buildStreamDoneHeaderMatchAction(
 	ruleIndex, matchIndex int, header egv1a1.HeaderMatch,
 ) *routev3.RateLimit_Action {
 	descriptorKey := translator.BucketRuleDescriptorKey(ruleIndex, matchIndex, header.Name, headerMatchKeyValue(header))
 
+	// Must mirror buildHeaderMatchAction: a descriptor that doesn't match the
+	// request-time one lands the real cost in a different bucket from the
+	// pre-flight reservation it is meant to settle.
 	if header.Type != nil && *header.Type == egv1a1.HeaderMatchDistinct {
 		return &routev3.RateLimit_Action{
-			ActionSpecifier: &routev3.RateLimit_Action_GenericKey_{
-				GenericKey: &routev3.RateLimit_Action_GenericKey{
-					DescriptorKey:   descriptorKey,
-					DescriptorValue: descriptorKey,
+			ActionSpecifier: &routev3.RateLimit_Action_RequestHeaders_{
+				RequestHeaders: &routev3.RateLimit_Action_RequestHeaders{
+					HeaderName:    header.Name,
+					DescriptorKey: descriptorKey,
 				},
 			},
 		}
--- a/internal/extensionserver/quota_ratelimit_test.go
+++ b/internal/extensionserver/quota_ratelimit_test.go
@@ -1611,6 +1611,75 @@ func TestBuildHeaderMatchAction(t *testing.T) {
 	})
 }
 
+func TestBuildStreamDoneHeaderMatchAction(t *testing.T) {
+	t.Run("distinct match reads the header value so each tenant keeps its own bucket", func(t *testing.T) {
+		header := egv1a1.HeaderMatch{Name: "x-tenant-id", Type: ptr.To(egv1a1.HeaderMatchDistinct)}
+
+		action := buildStreamDoneHeaderMatchAction(0, 0, header)
+
+		rh := action.GetRequestHeaders()
+		require.NotNil(t, rh, "a GenericKey here pools every tenant's cost into one bucket")
+		require.Equal(t, "x-tenant-id", rh.HeaderName)
+		require.Equal(t, translator.BucketRuleDescriptorKey(0, 0, "x-tenant-id", ""), rh.DescriptorKey)
+	})
+
+	t.Run("exact match keeps HeaderValueMatch with ExpectMatch true", func(t *testing.T) {
+		header := egv1a1.HeaderMatch{
+			Name:   "x-api-key",
+			Type:   ptr.To(egv1a1.HeaderMatchExact),
+			Value:  ptr.To("premium"),
+			Invert: ptr.To(true),
+		}
+
+		action := buildStreamDoneHeaderMatchAction(0, 0, header)
+
+		hvm := action.GetHeaderValueMatch()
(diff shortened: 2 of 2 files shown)


vignesh-chaturvedi:
Synced with main just now — merged rather than rebased, per the contributing
guide. The merge was conflict-free, nothing on main had touched either file, and
`make precommit` and `go test ./internal/...` are both green locally.

Two things that would help this along.

**Workflow approval.** Build and Test, Precommit Check and CodeQL are all sitting
at `action_required` and have never run on this branch, so the only green checks
here are Title, Description and DCO. Could a maintainer approve the workflow run?
As it stands there is no CI signal to review against, which is an awkward place
for a change whose argument is that the old behavior had no test coverage.

**Overlap with #2597.** That PR targets the same issue, was opened eleven days
before this one, and makes the same one-branch change. I am happy to close this
in its favour — I have no attachment to which one lands. The one thing I would
ask is that the tests here end up somewhere, either folded into #2597 or as a
follow-up:

- `TestStreamDoneDescriptorsMatchRequestTime` asserts that the request-time and
  stream-done actions agree on the descriptor *kind* as well as the key. That is
  the invariant that actually prevents this regression: `GenericKey` and
  `RequestHeaders` share a descriptor key and differ only in the value, so a test
  comparing keys alone still passes with the bug present. This is precisely how
  it stayed hidden.
- `TestBuildStreamDoneHeaderMatchAction` also covers an inverted `Exact` header
  still reporting with `ExpectMatch=true`, which is the other branch of the same
  function and was previously untested.

Both fail if the `GenericKey` branch is restored.

Whichever ordering the maintainers prefer is fine by me. Mainly flagging that
#2551 has been open since August with two unreviewed PRs attached to it, so a
steer on which one to take forward would unblock things.

Xunzhuo (approved):
L:GTM, thanks

vignesh-chaturvedi:
Thanks for the review!

The one red check is `E2E Test for Inference Extensions`, which is what fails
`ci-required`. It timed out in `InferencePoolResolvedRefsCondition` waiting on
pods and the InferencePool status. I believe it is a flake unrelated to this
change: that suite configures no QuotaPolicy, so the quota code returns before
reaching anything this PR touches, and the same job has failed in 8 of its last
23 completed runs, including on several unrelated PRs this week.

I cannot re-run jobs from a fork. Could someone re-run the failed job?

PR #2666, opened 2026-09-10
fix(mcpproxy): preserve valid SEP-2549 cache metadata in merged list responses

### LINKED ISSUE #2665: MCP proxy merge functions drop upstream cache metadata and emit invalid cacheScope: "" on list responses
## Summary

The MCP proxy (`internal/mcpproxy`) appears to unmarshal backend list responses that include valid cache metadata, then rebuild aggregated results in merge functions (e.g. `mergeToolsList`) without carrying that metadata forward. The response sent to clients can include:

```json
{
  "ttlMs": 0,
  "cacheScope": ""
}
```

Under MCP 2026-07-28 / SEP-2549, `cacheScope` is only valid as `"public"`, `"private"`, or **omitted**. An empty string is not valid. Strict MCP clients reject the otherwise-valid list payload and fail during tool discovery — even when OAuth and `initialize` succeed and the tool list itself is present on the wire.

This looks related to the class of proxy re-serialization issues discussed in [#2521](https://github.com/theagentrouter/agent-router/issues/2521) (dropped tool annotations after decode/re-encode), but manifests on cache metadata rather than tool hints.

## Environment

- **Gateway version observed:** `envoy-ai-gateway` v1.1.0 (from `serverInfo` in client captures)
- **Transport:** Streamable HTTP (`POST /mcp`)
- **Protocol:** MCP `2025-06-18` (default client path); related 2026-07-28 behavior tracked separately
- **Affected clients reported:** Codex CLI/Desktop 0.148–0.152 ([openai/codex#41437](https://github.com/openai/codex/issues/41437))

## Steps to reproduce

1. Deploy an `MCPRoute` with at least one authenticated backend MCP server that returns list responses with valid cache metadata (e.g. `cacheScope: "private"` on `tools/list`).
2. Complete OAuth and send an authenticated `tools/list` (and optionally `resources/list`, `resources/templates/list`) **through the gateway**.
3. Inspect the JSON-RPC result returned to the client (SSE `data:` frame or JSON body).

**Expected:** upstream cache metadata is preserved (`"public"` or `"private"`), or the field is omitted.

**Observed:** list payloads are present and structurally valid, but cache metadata is wrong:

```json
{
  "jsonrpc": "2.0",
  "id": 1,
  "result": {
    "tools": [ /* valid entries */ ],
    "ttlMs": 0,
    "cacheScope": ""
  }
}
```

The same empty `cacheScope` pattern has been observed on `resources/list` and `resources/templates/list` in the same session.

**Client symptom:** MCP startup reports success through auth/initialize, then fails before tool import — e.g. `Unexpected response type` with 0 discovered tools ([openai/codex#41437](https://github.com/openai/codex/issues/41437)).

## Analysis (suspected root cause)

In `internal/mcpproxy/legacy.go`, `sendToAllBackendsAndAggregateResponsesImpl` unmarshals per-backend responses into typed results, then calls a merge function and `json.Marshal`s the merged struct.

For `tools/list`, `mergeToolsList` (`handlers.go`) constructs a fresh `mcp.ListToolsResult` containing only the aggregated tools slice. Embedded `Cacheable` fields appear to remain at Go zero values, which serialize as `ttlMs: 0` and `cacheScope: ""`.

Similar merge helpers exist for other list methods (`mergeResourceList`, `mergeResourcesTemplateList`, `mergePromptsList`) and may share the same behavior.

The backend response **before** aggregation can contain valid cache metadata; the invalid empty value appears **after** gateway merge/re-encode.

## Impact

- OAuth and session setup can succeed while **tool/resource discovery fails** on strict clients.
- MCP integrations show as connected/authenticated but expose **zero tools**.
- Any deployment where the gateway aggregates list responses from authenticated, user-specific backends may be affected — not limited to a single vendor or backend implementation.

## What would "fixed" look like

- List responses through the gateway never include `cacheScope: ""`.
- Valid upstream values (`"public"` / `"private"`) are preserved where appropriate, or the field is omitted per spec.
- Behavior is covered by a regression test on the aggregated wire JSON (not only in-process struct inspection).

## Related issues

| Issue | Relationship |
|-------|----------------|
| [#2521](https://github.com/theagentrouter/agent-router/issues/2521) | Similar proxy re-encode metadata loss; `readOnlyHint` / `idempotentHint` |
| [#2323](https://github.com/theagentrouter/agent-router/issues/2323) | 2026-07-28 design; list aggregation / cache metadata |
| [#2218](https://github.com/theagentrouter/agent-router/issues/2218) | Different `tools/list` failure (Content-Type / SSE body parsing) |
| [openai/codex#41437](https://github.com/openai/codex/issues/41437) | Downstream client report with captured payloads |
| [rust-sdk#1242](https://github.com/modelcontextprotocol/rust-sdk/issues/1242) | Client-side discussion of empty `cacheScope` handling |

wangzhengzhuo05:
**Description**

Fixes #2665. The MCP proxy's aggregated list responses (`tools/list`, `resources/list`, `resources/templates/list`, `prompts/list`) now emit valid SEP-2549 cache metadata instead of the invalid `{"ttlMs":0,"cacheScope":""}` payload that strict MCP clients (e.g. Codex CLI 0.148-0.152) reject during tool discovery.

**What**

Added a small unexported helper `aggregateCacheable` in `internal/mcpproxy/handlers.go` that computes conservative cache metadata from the backend responses, and wired it into all four merge functions (`mergeToolsList`, `mergeResourceList`, `mergeResourcesTemplateList`, `mergePromptsList`):

- `ttlMs` - the minimum of all backends' TTL values (clamped to >= 0), so the aggregated response is never cached longer than any contributor's advertised freshness; `0` (immediately stale) when no backend advertises freshness.
- `cacheScope` - `"private"` if any backend's scope is private (most restrictive wins); otherwise the first non-empty backend scope; otherwise `"public"`, which is the SEP-2549 default when the field is absent and is always valid to emit explicitly.

The SDK's `Cacheable` struct (github.com/modelcontextprotocol/go-sdk v1.7.0) has no `omitempty` on `ttlMs`/`cacheScope`, so the previously-built fresh result structs marshaled the zero value as the invalid empty `cacheScope: ""` (verified: `{"ttlMs":0,"cacheScope":"","tools":[]}`). No other behavior changed: prefixing, selectors, authorization, and meta-URI rewriting are untouched.

**Why**

Under MCP 2026-07-28 / SEP-2549, `cacheScope` is only valid as `"public"`, `"private"`, or omitted - an empty string is not valid. Aggregated list responses are rebuilt from scratch in the merge functions, dropping upstream cache metadata and emitting an invalid document, which strict clients reject.

**Tests**

Added to `internal/mcpproxy/handlers_test.go`:

- `TestMergeToolsList_CacheableMetadata` - single-backend metadata preserved end-to-end (including `json.Marshal` shape), multi-backend conservative aggregation (min TTL + private-wins), no-upstream-metadata to valid public default, empty-responses slice validity.
- `TestMergeListResults_NeverEmitEmptyCacheScope` - `resources/list`, `resources/templates/list`, and `prompts/list` merges never marshal `"cacheScope":""` and preserve upstream values.

Verification run: `go test -count=1 ./internal/mcpproxy/...` returned ok (full package suite, ~3.0s). `gofmt -l` clean, `go vet` clean, go.mod/go.sum untouched.

**Mutation check**

Removing the four `Cacheable = aggregateCacheable(...)` wirings makes the new tests fail with exactly the reported symptom (`{"ttlMs":0,"cacheScope":"","tools":[]}` should not contain `"cacheScope":""`), then restoring the fix turns them green - the tests genuinely guard the fix.

**AI-generated code disclosure**

The change was implemented with AI assistance (OpenCode + muse-spark 1.3), then reviewed and verified for correctness, protocol conformance, and test coverage. The submitter understands the change and takes full ownership as required by CONTRIBUTING.md.

Fixes #2665

### DIFF
--- a/internal/mcpproxy/handlers.go
+++ b/internal/mcpproxy/handlers.go
@@ -386,6 +386,56 @@ type (
 	broadCastResponseMergeFn[T any] func(*session, []broadCastResponse[T]) T
 )
 
+// cacheableResult is the subset of mcp.Cacheable (mcp.CacheableResult) needed
+// to aggregate SEP-2549 cache metadata across backend responses.
+type cacheableResult interface {
+	GetTTLMs() int
+	GetCacheScope() string
+}
+
+// aggregateCacheable returns the conservative SEP-2549 cache metadata for an
+// aggregated list response. The proxy must never emit an empty cacheScope:
+// under SEP-2549 "cacheScope" is only valid as "public", "private", or
+// omitted, and the SDK marshals the zero value as an invalid "".
+//   - TTLMs: the minimum of all backends' TTL values, clamped to >= 0, so the
+//     aggregated response is never cached longer than any contributor's
+//     advertised freshness. 0 (immediately stale) is the safe default when no
+//     backend advertises freshness.
+//   - CacheScope: "private" if ANY backend scope is private (the most
+//     restrictive scope wins); otherwise the first non-empty backend scope;
+//     otherwise "public", which is the SEP-2549 default when the field is
+//     absent and therefore always safe to emit explicitly.
+func aggregateCacheable[T cacheableResult](responses []broadCastResponse[T]) mcp.Cacheable {
+	aggregated := mcp.Cacheable{CacheScope: "public"} // SEP-2549 default
+	if len(responses) == 0 {
+		return aggregated
+	}
+	minTTL := responses[0].res.GetTTLMs()
+	firstScope := responses[0].res.GetCacheScope()
+	anyPrivate := firstScope == "private"
+	for _, r := range responses[1:] {
+		if t := r.res.GetTTLMs(); t < minTTL {
+			minTTL = t
+		}
+		if s := r.res.GetCacheScope(); s == "private" {
+			anyPrivate = true
+		} else if firstScope == "" && s != "" {
+			firstScope = s
+		}
+	}
+	if minTTL < 0 {
+		minTTL = 0
+	}
+	aggregated.TTLMs = minTTL
+	switch {
+	case anyPrivate:
+		aggregated.CacheScope = "private"
+	case firstScope != "":
+		aggregated.CacheScope = firstScope
+	}
+	return aggregated
+}
+
 // mergeToolsList merges the list of tools from all backends and prepare the response message to be sent back to the client.
 func (m *mcpRequestContext) mergeToolsList(s *session, responses []broadCastResponse[mcp.ListToolsResult]) mcp.ListToolsResult {
 	// Use a non-nil empty slice so JSON encodes as [] not null; some clients reject tools:null.
@@ -439,6 +489,7 @@ func (m *mcpRequestContext) mergeToolsList(s *session, responses []broadCastResp
 			resp.Tools = append(resp.Tools, tool)
 		}
 	}
+	resp.Cacheable = aggregateCacheable(responses)
 
 	return resp
 }
@@ -456,6 +507,7 @@ func (m *mcpRequestContext) mergeResourceList(_ *session, responses []broadCastR
 			resp.Resources = append(resp.Resources, res)
 		}
 	}
+	resp.Cacheable = aggregateCacheable(responses)
 	return resp
 }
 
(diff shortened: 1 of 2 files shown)


Hritik003 (commented):
hi @wangzhengzhuo05 , 
the changes related to the new mcp spec 28-06-2026 are being taken up in #2545 along with the curated changes planned as per:
- #2431

PR #2675, opened 2026-09-10
translator: support field cache_write_tokens for openai

aabchoo:
**Description**

Adding an additional field `cache_write_tokens` for openai translation and marking `cache_creation_tokens` as deprecated in v1.3.0. This fixes cache writes not being accounted for for gpt-5.6+ models.

Openai translated fields returns `cache_write_tokens` as part of the usage response:

```
  "usage": {
    "prompt_tokens": 1116,
    "completion_tokens": 278,
    "total_tokens": 1394,
    "prompt_tokens_details": {
      "cached_tokens": 0,
      "cache_creation_tokens": 1113, # same
      "cache_write_tokens": 1113,
    },
}
```

https://developers.openai.com/api/docs/guides/prompt-caching/diagnostics#confirm-the-improvement

Updating the translation layer to parse the `cache_write_tokens` from/to openai as well. Keeping the legacy field in temporarily but will move to remove it in a later release. Cache write tokens created by openai are currently ignored.

This means the response of openai translation will be cache_write_tokens in addition to cache_creation_tokens.

n/a


internal/apischema/openai/openai.go
```diff
@@ -1577,7 +1577,44 @@ type PromptTokensDetails struct {
 	// Cached tokens present in the prompt.
 	CachedTokens int `json:"cached_tokens,omitzero"`
 	// Tokens written to the cache.
-	CacheCreationTokens int `json:"cache_creation_input_tokens,omitzero"`
+	CacheWriteTokens int `json:"cache_write_tokens,omitzero"`
```
sukumargaonkar, line 1580: i understand we are renaming because openai recently added the cache-write field
but since this will remove a field from the response users receives, i think we should do the field-deletion over atleast 2 minor releases
for this release we can have both `CacheCreationTokens` and `CacheWriteTokens` fields in response, note `CacheWriteTokens` as deprecated in upcoming release and remove it 2 versions down the line

yuzisun (approved):

PR #2676, opened 2026-09-11
feat(controller): allow protectedResourceMetadata.resource to be auto-derived from hostnames

### LINKED ISSUE #2406: Allow protectedResourceMetadata.resource to be optional, auto-derived from Gateway/Listener hostname
### Problem

The `protectedResourceMetadata.resource` field in `MCPRouteOAuth` is currently required and must be set to an explicit URI. This forces operators to hardcode the external gateway hostname inside the `MCPRoute` spec, coupling the route configuration to infrastructure-level concerns (e.g. which hostname is exposed on the `Gateway`/`Listener`).

### Current Behaviour

The resource field is mandatory:
```yaml
protectedResourceMetadata:
  resource: "https://api.example.com/mcp"  # must be hardcoded
```
It is used to:

-  Construct the `WWW-Authenticate` response header with `resource_metadata=`
-  Serve the OAuth protected resource metadata at `/.well-known/oauth-protected-resource{path}`

### Desired Behaviour

Make resource optional. When not set, the gateway should automatically derive the value from the hostname(s) configured on the parent `Gateway`/`Listener`, combined with the `MCPRoute`'s path.

### Why This Matters

-  Operators managing multiple environments (staging, production) would not need to maintain environment-specific `MCPRoute` manifests just to change the hostname.
-  It aligns with the principle that hostname exposure is a `Gateway`/`Listener` concern, not a route concern.

#### 1. Make resource optional in the API

Change the field from required to optional:
```golang
// +kubebuilder:validation:Optional
// +optional
Resource *string `json:"resource,omitempty"`
```

#### 2. Auto-derive the value at reconciliation time

In the controller, when resource is not set, derive it from the parent `Gateway`/`Listener` hostname and the `MCPRoute`'s path — following the same logic already used in `buildResourceMetadataURL`. [[controller/mcp_route_security_policy.go](https://github.com/envoyproxy/ai-gateway/blob/main/internal/controller/mcp_route_security_policy.go#L285-L339)]

The derived value would be: `https://{intersected_hostname}{mcpRoute.Spec.Path}`

This aligns with the Gateway API principle that the **intersected hostname** between the Gateway and the attached Route is the canonical representation of the public-facing URL. [[Hostnames in Gateway API](https://gateway-api.sigs.k8s.io/docs/concepts/hostnames/)]

### Caveats
- If the Gateway has **multiple hostnames**, it may be ambiguous which one to use — the implementation would need a defined selection strategy.
- The existing known limitation that only **one MCPRoute per listener** can use OAuth (due to the fixed well-known path) would still apply. [[controller/mcp_route.go](https://github.com/envoyproxy/ai-gateway/blob/main/internal/controller/mcp_route.go#L276-L363)]

HesandaLiyanage:
**Description**

Makes `protectedResourceMetadata.resource` optional in `MCPRouteOAuth` (both in `v1alpha1` and `v1beta1`).

When `resource` is omitted, the controller automatically derives the canonical public-facing resource identifier as `https://{hostname}{path}` (where `path` defaults to `spec.path` or `/mcp`) following the deterministic hostname resolution strategy discussed in Issue #2406:

1. **Route hostnames (`spec.hostnames`)**:
   - If exactly 1 non-wildcard hostname is specified, it is used.
   - If multiple hostnames are specified, derivation fails with a descriptive error because it is ambiguous.
   - If a wildcard hostname (`*`) is specified, derivation fails with an error because it cannot be resolved to a specific public URL.
2. **Parent Gateway listeners (`spec.parentRefs`)**:
   - When no hostnames are defined on the Route, inspects the parent Gateway:
     - If `sectionName` is specified on the parentRef, looks up that specific listener and verifies it has a non-wildcard configured hostname.
     - If `sectionName` is omitted, inspects all HTTP/HTTPS listeners on the parent Gateway and collects unique hostnames. If exactly 1 unique non-wildcard hostname exists (e.g. across HTTP:80 and HTTPS:443 listeners), it is used.
     - If multiple unique hostnames, no configured hostnames, or wildcard hostnames are present, derivation fails with a clear error.
3. If derivation fails, the route reconciliation marks the status condition as `ConditionTypeNotAccepted` with the descriptive error, informing the user that `resource` must be explicitly configured.
4. If an explicit `resource` URI is provided in the spec, it is used directly (with trailing slash trimmed).
5. Data plane configuration (`mcpConfig`) in `gateway.go` also resolves the resource URL when translating routes for `ResourceMetadataURL`.

Key changes:
- `api/v1alpha1/mcp_route.go`, `api/v1beta1/mcp_route.go`: changed `ProtectedResourceMetadata.Resource` from `string` to `*string` with `+kubebuilder:validation:Optional` and `+optional`.
- `internal/controller/mcp_route_security_policy.go`: added `resolveDeterministicHostname` and `resolveOAuthResourceURL`, updated `buildResourceMetadataURL`, `buildWWWAuthenticateHeaderValue`, and `ensureOAuthProtectedResourceMetadataHRF`/`BTP`.
- `internal/controller/gateway.go`: updated `mcpConfig` to resolve `ResourceMetadataURL` using `resolveOAuthResourceURL`.
- Regenerated manifests, deepcopy, CRDs, and API reference documentation (`make apigen codegen apidoc`).
- Added extensive unit tests covering all deterministic and ambiguous resolution scenarios in `internal/controller/mcp_route_security_policy_test.go` and `internal/controller/gateway_test.go`.
- Added CEL test case in `tests/crdcel/testdata/mcpgatewayroutes/resource_omitted.yaml` and `tests/crdcel/main_test.go`.

Validation performed:
- `make check` passed with 0 issues (linters, licenses, formatting, CRD generation).
- `make test` passed cleanly across all packages.

Fixes #2406

None.


internal/controller/mcp_route_security_policy.go
```diff
@@ -302,12 +302,140 @@ func (c *MCPRouteController) ensureOAuthProtectedResourceMetadataBTP(ctx context
...
+		path = "/mcp"
+	}
+	if !strings.HasPrefix(path, "/") {
+		path = "/" + path
+	}
+	path = strings.TrimSuffix(path, "/")
+
+	return fmt.Sprintf("https://%s%s", hostname, path), nil
+}
+
```
Hritik003, line 577: should the scheme be harcoded?

Hritik003: what if the listener to gateway is http://...80

internal/controller/gateway.go
```diff
@@ -654,7 +654,9 @@ func mcpConfig(mcpRoutes []aigv1b1.MCPRoute) (_ *filterapi.MCPConfig, hasEffecti
 			mcpRoute.Authorization = &filterapi.MCPRouteAuthorization{}
 
 			if route.Spec.SecurityPolicy.OAuth != nil {
-				mcpRoute.Authorization.ResourceMetadataURL = buildResourceMetadataURL(&route.Spec.SecurityPolicy.OAuth.ProtectedResourceMetadata)
+				if resourceURL, err := resolveOAuthResourceURL(ctx, k8sClient, route); err == nil {
+					mcpRoute.Authorization.ResourceMetadataURL = buildResourceMetadataURL(resourceURL)
+				}
```
Hritik003, line 659: `resolveOAuthResourceURL` is called across two places, one here and other in mcp_route_security_policy.go. In mcp_route_security_policy.go  we fail the mcp route to NotAccepted if there is an error. But here we skip err!=nil, shouldnt we make the route to not accepted also here?

internal/controller/mcp_route_security_policy.go
```diff
@@ -503,9 +631,6 @@ func buildOAuthProtectedResourceMetadataJSON(auth *aigv1b1.MCPRouteOAuth) string
 	if len(auth.ProtectedResourceMetadata.ScopesSupported) != 0 {
 		response["scopes_supported"] = auth.ProtectedResourceMetadata.ScopesSupported
 	}
-	if auth.ProtectedResourceMetadata.ResourceName != nil && *auth.ProtectedResourceMetadata.ResourceName != "" {
-		response["resource_name"] = auth.ProtectedResourceMetadata.ResourceName
-	}
 	if len(auth.ProtectedResourceMetadata.ResourceSigningAlgValuesSupported) > 0 {
```
Hritik003, line 779: reason for removing this?

internal/controller/mcp_route_security_policy.go
Hritik003: also we seem to be ignoring port here, that will be a problem when any https://host:8443 derives https://host/mcp silently implying 443

internal/controller/mcp_route_security_policy.go
```diff
@@ -302,12 +302,140 @@ func (c *MCPRouteController) ensureOAuthProtectedResourceMetadataBTP(ctx context
...
+	if err := k8sClient.Get(ctx, client.ObjectKey{Namespace: gwNamespace, Name: gwName}, &gw); err != nil {
+		return "", fmt.Errorf("failed to get parent Gateway %s/%s for OAuth resource derivation: %w", gwNamespace, gwName, err)
+	}
+
+	if parentRef.SectionName != nil && *parentRef.SectionName != "" {
+		sectionName := *parentRef.SectionName
+		var targetListener *gwapiv1.Listener
+		for i := range gw.Spec.Listeners {
+			if string(gw.Spec.Listeners[i].Name) == string(sectionName) {
+				targetListener = &gw.Spec.Listeners[i]
```
Hritik003: should skip tls,udp,tcp based protocols here

internal/controller/mcp_route_security_policy.go
```diff
@@ -302,12 +302,140 @@ func (c *MCPRouteController) ensureOAuthProtectedResourceMetadataBTP(ctx context
...
+	hostname, err := resolveDeterministicHostname(ctx, k8sClient, mcpRoute)
+	if err != nil {
+		return "", err
+	}
+
+	path := ptr.Deref(mcpRoute.Spec.Path, "/mcp")
+	if path == "" {
+		path = "/mcp"
+	}
+	if !strings.HasPrefix(path, "/") {
```
Hritik003, line 570: what if the resource path is https://host/v1/mcp and mcp route path is /mcp, in this case the suffixes dont match. And the well known uri in newHttpRoute is constructed on spec.path. We need to handle this case

internal/controller/mcp_route_security_policy.go
```diff
@@ -302,12 +302,140 @@ func (c *MCPRouteController) ensureOAuthProtectedResourceMetadataBTP(ctx context
...
+		}
+		if h == "" {
+			return "", errors.New("cannot derive OAuth protectedResourceMetadata.resource: route hostname is empty; resource must be explicitly configured")
+		}
+		return h, nil
+	}
+
+	// No hostnames on the route, resolve via parentRefs.
+	if len(mcpRoute.Spec.ParentRefs) == 0 {
+		return "", errors.New("cannot derive OAuth protectedResourceMetadata.resource: route has no hostnames or parentRefs configured; resource must be explicitly configured")
```
Hritik003: the crd already has a minimum=1 for parentrefs in mcp route spec

internal/controller/mcp_route_security_policy.go
```diff
@@ -302,12 +302,140 @@ func (c *MCPRouteController) ensureOAuthProtectedResourceMetadataBTP(ctx context
...
+		return "", errors.New("cannot derive OAuth protectedResourceMetadata.resource: route references multiple parent gateways; resource must be explicitly configured")
+	}
+	if k8sClient == nil {
+		return "", errors.New("cannot derive OAuth protectedResourceMetadata.resource: parent Gateway cannot be inspected without Kubernetes client; resource must be explicitly configured")
+	}
+
+	parentRef := mcpRoute.Spec.ParentRefs[0]
+	if parentRef.Kind != nil && *parentRef.Kind != "Gateway" {
+		return "", fmt.Errorf("cannot derive OAuth protectedResourceMetadata.resource: parentRef kind %q is not Gateway; resource must be explicitly configured", *parentRef.Kind)
+	}
```
Hritik003: the crd already has a validation for kind==Gateway in mcp route spec

Hritik003 (commented):
Thanks @HesandaLiyanage for taking this up and making the PR well structured. Added a few comments and would like to also address these scenarios:

1. Although we have the validation `format:uri` here, https://github.com/theagentrouter/agent-router/blob/472da7db7984f769c9a77671d12c0354e75fc3ba/api/v1beta1/mcp_route.go#L685
Any input like api.example.com will produce api.example.com/well-known/.... Can we add a CEL rule here that is should be either starting with HTTP or HTTPs?
2. What if the gateway listener hostname is changed? MCP Route reconcile doesnt happen again and leaves the BTP and Hrf stale. Good to add a watcher for mcp route to gateway in controller

internal/controller/mcp_route_security_policy.go
HesandaLiyanage: The reason https:// was originally used was to follow RFC 9728 §2.1 ("It should use the 'https' scheme") and the original issue proposal (https://{intersected_hostname}{path}). Also, when resolving from route spec.hostnames, Gateway API route hostnames don't carry a scheme or port.

That said, you're right that when resolving from parent Gateway listeners:

In dev/test or internal setups, a listener might only serve plain HTTP (e.g. port 80/8080).
If a listener uses a non-standard port (like 8443 for HTTPS or 8080 for HTTP), omitting the port gives an unreachable or mismatched resource URL.

We can handle this cleanly:

When resolving from Gateway listeners:
Scheme: use https:// if HTTPS listener(s) are present, http:// if only HTTP listener(s) exist. If both HTTP:80 and HTTPS:443 exist for the hostname, prefer HTTPS as usual.

Port: if the listener port is non-standard (not 443 for HTTPS, not 80 for HTTP), append :{port} (e.g. https://host:8443/mcp or http://host:8080/mcp).

When resolving from Route spec.hostnames: since these have no scheme or port, default to https://{hostname}{path} (or check the attached Gateway listener for matching port/protocol if parentRefs are present).
For more complex setups (e.g. cloud ALBs terminating TLS on 443 but forwarding to 8443), operators can still explicitly set resource.

does this approach work for you?

internal/controller/gateway.go
HesandaLiyanage: This is actually intentional , it comes down to separation of concerns between the two controllers.

gateway.go is part of GatewayController, which reconciles Gateways to build the extproc config secret. MCPRoute.Status is owned exclusively by MCPRouteController. If GatewayController also wrote to MCPRoute.Status, we'd get concurrent writes and resourceVersion conflicts between the two controllers.

Also, when resolveOAuthResourceURL fails, MCPRouteController already handles it . It sets ConditionTypeNotAccepted with a proper error message during its own reconciliation. So the route does get marked not accepted, just not from here.

Lastly, one route failing to resolve its metadata shouldn't take down the whole Gateway reconciliation and affect other healthy routes on it.

So leaving ResourceMetadataURL empty here and letting MCPRouteController own the acceptance status is the intended design, not an oversight.

internal/controller/mcp_route_security_policy.go
HesandaLiyanage: This block was accidentally duplicated in buildOAuthProtectedResourceMetadataJSON (once before scopes_supported, once after). I just cleaned up the duplicate , resource_name is still there at lines 628–630.

internal/controller/mcp_route_security_policy.go
HesandaLiyanage: I'll add that check.

Separately , in the sectionName lookup, we match the listener by name and verify its hostname, but we never check targetListener.Protocol. If someone's sectionName points to a TCP, UDP, or raw TLS listener, it'd currently be accepted, which is invalid since MCPRoute and OAuth metadata are strictly HTTP/HTTPS.

Should we validate that targetListener.Protocol is HTTP or HTTPS here too, and reject TCP/UDP/TLS with a descriptive error? I can fold it into this fix if so.

internal/controller/mcp_route_security_policy.go
HesandaLiyanage: In the auto derivation path (420–425), resource comes directly from mcpRoute.Spec.Path, so those paths always match by definition, so no mismatch possible there.

But if an operator explicitly sets resource with a different path (e.g. resource: "https://host/v1/mcp", spec.path: "/mcp"), buildResourceMetadataURL advertises /.well-known/oauth-protected-resource/v1/mcp, while newHTTPRoute still matches on /.well-known/oauth-protected-resource + spec.path. That's a mismatch.

Docs say resource "should match the MCPRoute's URL." . So two options-

Validate -reject with ConditionTypeNotAccepted if explicit resource's path doesn't match spec.path.
Align -derive newHTTPRoute's match path from the resolved resource URL instead.

Which do you prefer?

internal/controller/mcp_route_security_policy.go
HesandaLiyanage: Thank you! I'll update this.

internal/controller/mcp_route_security_policy.go
HesandaLiyanage: Thank you!! , I'll fix it.

HesandaLiyanage: Resolved

internal/controller/mcp_route_security_policy.go
HesandaLiyanage: Resolved

api/v1alpha1/mcp_route.go
```diff
@@ -588,9 +588,14 @@ type ProtectedResourceMetadata struct {
...
-	// +kubebuilder:validation:Required
+	// When omitted, the controller automatically derives the resource identifier from the
+	// hostname configured on the Route or parent Gateway/Listener, combined with the MCPRoute's path.
+	//
+	// +kubebuilder:validation:Optional
 	// +kubebuilder:validation:Format=uri
-	Resource string `json:"resource"`
+	// +kubebuilder:validation:XValidation:rule="self.startsWith('https://') || self.startsWith('http://')",message="resource must start with http:// or https://"
+	// +optional
+	Resource *string `json:"resource,omitempty"`
```
aish1331: so far, we have been taking newer changes in beta, instead of both. Is there any reason why you want to add it to alpha API?

api/v1beta1/mcp_route.go
```diff
@@ -681,9 +681,14 @@ type ProtectedResourceMetadata struct {
...
 	// This should match the MCPRoute's URL. For example, if the MCPRoute's URL is
 	// "https://api.example.com/mcp", the Resource should be "https://api.example.com/mcp".
-	// +kubebuilder:validation:Required
+	// When omitted, the controller automatically derives the resource identifier from the
+	// hostname configured on the Route or parent Gateway/Listener, combined with the MCPRoute's path.
+	//
+	// +kubebuilder:validation:Optional
 	// +kubebuilder:validation:Format=uri
-	Resource string `json:"resource"`
+	// +kubebuilder:validation:XValidation:rule="self.startsWith('https://') || self.startsWith('http://')",message="resource must start with http:// or https://"
```
aish1331, line 718: why is this `self.startsWith('https://') || self.startsWith('http://')` validation added?

api/v1beta1/mcp_route.go
```diff
@@ -681,9 +681,14 @@ type ProtectedResourceMetadata struct {
...
+	// +optional
+	Resource *string `json:"resource,omitempty"`
```
aish1331: Converting the datatype of Resource from `string` to `*string` is a breaking change; let's avoid that. Can we continue to keep `Resource` as a string and make it optional?

api/v1alpha1/mcp_route.go
HesandaLiyanage: It was added to v1alpha1 purely for schema parity while both versions are served, but keeping new changes strictly in v1beta1 makes total sense if v1alpha1 is frozen. I'll make that change

api/v1beta1/mcp_route.go
HesandaLiyanage: This was added based on review feedback from @Hritik003 . While +kubebuilder:validation:Format=uri validates RFC 3986 generic URI syntax, it permits non HTTP schemes (or inputs like api.example.com), which leads to malformed discovery URLs (e.g., api.example.com/.well-known/oauth-protected-resource missing a scheme) when constructing RFC 9728 metadata endpoints.

The CEL rule fails fast at CRD admission time to ensure the resource is a valid HTTP/HTTPS URL.

api/v1beta1/mcp_route.go
HesandaLiyanage: On itt !!!!

HesandaLiyanage: Resolved

api/v1alpha1/mcp_route.go
HesandaLiyanage: Resolved

internal/controller/mcp_route_security_policy.go
Hritik003: So from what i have observed, clients construct well known urls on top of the mcp route URL path. 

Yes,
If resource path exists and does not match with spec path, reject the route with not accepted. This seems to be right option. 

thoughts @nacx ?

internal/controller/mcp_route_security_policy.go
Hritik003: yes

internal/controller/mcp_route_security_policy.go
Hritik003: yes the approach sounds good

Hritik003 (commented):
Hi @HesandaLiyanage, have replied to the approaches/suggestions on the comments.

internal/controller/mcp_route_security_policy.go
HesandaLiyanage: Hi @Hritik003  , What about this approach?
When protectedResourceMetadata.resource is explicitly configured, parse the URL path and compare it against mcpRoute.Spec.Path (normalizing trailing slashes and defaulting omitted spec.path to /mcp). If the paths don't match, reject the route with ConditionTypeNotAccepted.

Hritik003 (commented):
@HesandaLiyanage can you fix the tests?

PR #2681, opened 2026-09-13
docs: link to Agent Router as a service

zinuga:
**Description**

Adds a link for users who want to try Agent Router as a hosted service instead of running it themselves. The link appears in the README (nav row and Quick start), in the Getting Started docs for both the current and 1.1 versions, and as CTAs on the homepage hero and quickstart panel.

The hero and quickstart CTAs are data-driven from src/data/home/index.ts. The quickstart panel is dark in both themes, so its CTA carries its own on-dark colours rather than the light-ground .button--secondary.


kanurag94 (approved):
lgtm, thank you!

missBerg (changes requested):
See PR https://github.com/theagentrouter/agent-router/pull/2688 

Could we change this PR to only cover the docs site, and not the homepage? Above PR addresses homepage

PR #2691, opened 2026-09-14
feat: protectedResourceMetadata.resource from the request instead of requiring it to be configured

### LINKED ISSUE #2690: Derive protectedResourceMetadata.resource from the request instead of requiring it to be configured
_Description_:

`MCPRoute.spec.securityPolicy.oauth.protectedResourceMetadata.resource` is currently
required. It is the RFC 9728 resource identifier — the canonical URL of the protected MCP
endpoint — and the gateway advertises it both in the Protected Resource Metadata document
and in the `resource_metadata` parameter of the `WWW-Authenticate` challenges it returns.
Clients compare that value against the URL they actually used, so it has to match exactly,
including scheme and port.

Requiring it to be configured means operators hand-write their own external URL into the
MCPRoute and keep it in sync by hand. That is duplicated knowledge the gateway already has
at request time, and it is impossible to get right in several ordinary cases:

- The same MCPRoute reachable on more than one hostname. A single static value can only be
  correct for one of them.
- A gateway listening on a non-standard port. The port is part of the identifier, and it is
  not recoverable from any Gateway listener field the control plane can read.
- A plain-HTTP gateway, or one where TLS terminates upstream. The scheme is only knowable
  from the forwarded request.
- Local development and CI. `examples/mcp/mcp_oauth_keycloak.yaml` currently has to hardcode
  `resource: "http://127.0.0.1:1975/mcp"`, and the e2e suite reaches the gateway through a
  port-forward on a port chosen at run time, which no static value can match.

Desired behavior: make `resource` optional and derive it per request from the scheme
(forwarded protocol), authority (host and port) and path the client actually used. An
explicitly configured value keeps working and takes precedence, for deployments behind
something that rewrites the external URL without setting forwarded headers.

Used as:

```yaml
securityPolicy:
  oauth:
    issuer: "https://keycloak.example.com/realms/master"
    protectedResourceMetadata:
      # resource omitted - derived per request
      scopesSupported: ["profile", "email"]
```

The identifier surfaces in three places, and they have different constraints:

1. **The metadata document.** Served today by an `HTTPRouteFilter` direct response, whose
   body is static — Envoy Gateway's `CustomResponseBody` is `Inline` or a ConfigMap
   `ValueRef`, with no templating — so the value must be known when the HTTPRoute is
   generated. Deriving it requires serving the document from the MCP proxy instead, which
   already sees the request and already receives the `x-ai-eg-mcp-route` header identifying
   the route.
2. **The 403 `insufficient_scope` challenge**, built in the MCP proxy. It already has the
   request in hand.
3. **The 401 challenge**, produced by Envoy's JWT filter before the request reaches the
   proxy. This one cannot be computed in Go and needs an Envoy substitution format string.

Deriving the identifier from `Host` / `X-Forwarded-Proto` means trusting those headers.
Envoy sanitizes them by default, but this is worth stating explicitly in the docs so that a
listener configured to trust downstream forwarded headers is a deliberate choice.

*Relevant Links*

- RFC 9728, Protected Resource Metadata: https://datatracker.ietf.org/doc/html/rfc9728
- MCP authorization spec: https://modelcontextprotocol.io/specification/2025-06-18/basic/authorization
- Related: https://github.com/theagentrouter/agent-router/pull/2676 takes a different
  approach, deriving the identifier at reconcile time from `MCPRoute.spec.hostnames` or the
  parent Gateway listener hostname. That keeps the static direct response but fails when no hostname is configured anywhere — which is the shape
  both `examples/mcp/mcp_oauth_example.yaml` and `tests/e2e/testdata/mcp_route_oauth.yaml`
  use today.

aish1331:
**Description**

The OAuth Protected Resource Metadata document (RFC 9728) was served as a static HTTPRouteFilter direct response, so the `resource` identifier it advertises had to be known when the HTTPRoute was generated. That forced `protectedResourceMetadata.resource` to be a required field, with operators hand-writing their own external URL into the MCPRoute and keeping it in sync by hand. It also made the field impossible to get right in ordinary cases: the control plane cannot see the scheme, authority or port a client will actually use, so one configured value is wrong as soon as the route is reachable on a second hostname, sits behind a non-standard port, or is served over plain HTTP. The Keycloak example had to hardcode `resource: "http://127.0.0.1:1975/mcp"` for exactly this reason.

This commit serves the metadata document from the MCP proxy instead. The well-known route rule now forwards to the shared MCP proxy Backend with the same `x-ai-eg-mcp-route` header the main rule already sets, so the proxy resolves the route's OAuth configuration and computes the identifier from the request it is handling: `X-Forwarded-Proto` for the scheme, the preserved downstream authority for host and port, and the request path. The 403 `insufficient_scope` challenge is built the same way. One configuration therefore advertises the correct identifier on every address the route is reachable on, with nothing to keep in sync.

`protectedResourceMetadata.resource` becomes optional. An explicitly configured value still takes precedence everywhere, so deployments fronted by something that rewrites the externally visible URL without setting forwarded headers keep working unchanged, and existing manifests are unaffected. The field stays a plain string rather than becoming a pointer, so this is not a breaking change for Go consumers of the `api` packages; the only CRD change is that `resource` is dropped from the schema's `required` list.

The 401 challenge is the one value the proxy cannot produce, because Envoy's JWT filter rejects the request before it is ever proxied. It uses an Envoy substitution format string, which Envoy Gateway passes through to the local response policy's `response_headers_to_add`.

Fixes https://github.com/theagentrouter/agent-router/issues/2690
Related PR: https://github.com/theagentrouter/agent-router/pull/2676

The 401 `WWW-Authenticate` header is the only part of this change that cannot be verified by a unit test. It relies on Envoy expanding `%REQ(X-FORWARDED-PROTO)%` and `%REQ(:AUTHORITY)%` in a header value that Envoy Gateway funnels into the local response policy's `response_headers_to_add`. A unit test pins the exact string the controller emits, but only e2e proves the expansion happens, so please give that job's result particular weight. If the expansion does not work, the header would contain the literal format string rather than a URL, which fails loudly rather than silently.

`tests/e2e/testdata/mcp_route_oauth.yaml` now deliberately omits `resource`, so the e2e suite exercises the derived path end to end: it reaches the gateway through a port-forward on a port chosen at run time, which no statically configured identifier could ever have matched. `tests/e2e/testdata/mcp_route_authorization.yaml` keeps `resource` pinned, so the override path stays covered.

Upgrade path: a cluster reconciled by an older controller has an HTTPRouteFilter serving the metadata as a direct response, and the route rule no longer references it. `ensureOAuthResources` deletes that superseded filter rather than leaving it dangling. There is a test for this, and the delete costs one cached Get per reconcile once nothing is left to remove.

Deriving the identifier from `Host` and `X-Forwarded-Proto` means the advertised value depends on request headers. Envoy sanitizes both by default, and the MCP capability docs now say so explicitly, but it is worth a reviewer confirming that framing is the one the project wants to commit to.

`externalPath` deliberately does not consult `x-ai-eg-original-path`. Nothing strips `x-ai-eg-*` off inbound requests, so honoring it would let a client choose which document the proxy serves. The generated route rules carry no URLRewrite filter, so the routed path is already the path the client sent. There is a test asserting client-supplied values are ignored.


missBerg:
Direction is right, and the 401 substitution works end to end per the e2e log. Three things before this can land:

1. **The metadata endpoint times out in e2e on all three EG versions.** `modifyMCPGatewayGeneratedCluster` in `internal/extensionserver/mcproute.go` only rewrites the `rule/0` cluster to the in-process proxy, so the new well-known rule's cluster still points at the dummy IP `192.0.2.42`. `maybeUpdateMCPRoutes` has the same `rule/0` check. Match on the dummy-IP endpoint instead of the rule index. Until this is fixed the proxy-served document and the derived 403 path are unverified.

2. **Derived `resource` must agree with `audiences`.** Clients send the advertised resource as the RFC 8707 `resource` parameter and the AS binds `aud` to it, but `audiences` is still static. A token minted for `http://127.0.0.1:1975/mcp` fails validation against `https://api.example.com/mcp`. The docs need to say this, and "nothing to keep in sync" is overclaiming when audiences are set.

3. **Docs say Envoy sanitizes `Host` by default. It does not.** Only X-Forwarded-Proto is sanitized. Low impact, but the sentence should be corrected.

examples/mcp/mcp_oauth_keycloak.yaml
```diff
@@ -21,7 +21,8 @@ spec:
         remoteJWKS:
           uri: http://localhost:8080/realms/master/protocol/openid-connect/certs
       protectedResourceMetadata:
-        resource: "http://127.0.0.1:1975/mcp"
```
Hritik003, line 24: nit: do not remove here, and add a new example?

internal/controller/mcp_route_security_policy.go
```diff
@@ -597,9 +538,11 @@ func (c *MCPRouteController) ensureOAuthResources(ctx context.Context, mcpRoute
...
 	}
 
-	// Create HTTPRouteFilter for OAuth protected resource metadata endpoint.
-	if hrfErr := c.ensureOAuthProtectedResourceMetadataHRF(ctx, mcpRoute); hrfErr != nil {
-		return fmt.Errorf("failed to ensure HTTPRouteFilter: %w", hrfErr)
+	// The OAuth protected resource metadata document used to be served by an HTTPRouteFilter
+	// direct response. It is now served by the MCP proxy, which can derive the resource
+	// identifier from the request, so delete the filter left behind by an older version.
+	if delErr := c.deleteOAuthProtectedResourceMetadataHRF(ctx, mcpRoute); delErr != nil {
+		return fmt.Errorf("failed to delete legacy HTTPRouteFilter: %w", delErr)
```
Hritik003, line 545: nit: can add a todo to remove this block after 2.0?

internal/filterapi/mcpconfig.go
```diff
@@ -45,6 +45,11 @@ type MCPRoute struct {
 	// Authorization is the authorization configuration for this route.
 	Authorization *MCPRouteAuthorization `json:"authorization,omitempty"`
 
+	// OAuth is the OAuth protected resource metadata (RFC 9728) served for this route.
+	// When set, the MCP proxy serves the protected resource metadata document and includes
+	// a resource_metadata challenge in WWW-Authenticate headers it emits.
+	OAuth *MCPRouteOAuth `json:"oauth,omitempty"`
```
Hritik003, line 51: if this represents only PRM, can we rename it to ProtectedResourceMetadata, since the present name sounds confusing

internal/controller/mcp_route_security_policy.go
```diff
@@ -343,8 +359,11 @@ func buildResourceMetadataURL(metadata *aigv1b1.ProtectedResourceMetadata) strin
...
 // * https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization#protected-resource-metadata-discovery-requirements
 // * https://datatracker.ietf.org/doc/html/rfc9728#name-www-authenticate-response
-func buildWWWAuthenticateHeaderValue(metadata *aigv1b1.ProtectedResourceMetadata) string {
-	resourceMetadataURL := buildResourceMetadataURL(metadata)
+func buildWWWAuthenticateHeaderValue(metadata *aigv1b1.ProtectedResourceMetadata, servingPath string) string {
+	resourceMetadataURL := envoyDerivedResourceMetadataURL(servingPath)
+	if metadata.Resource != "" {
+		resourceMetadataURL = buildResourceMetadataURL(metadata.Resource)
+	}
 	headerValue := `Bearer error="invalid_token", error_description="The access token is missing or invalid"`
```
Hritik003, line 367: we already have the PRM metadata, so we should prefer metadata.Resource and only fall back to the derived URL when it's empty

internal/controller/mcp_route_security_policy.go
```diff
@@ -303,37 +305,51 @@ func (c *MCPRouteController) ensureOAuthProtectedResourceMetadataBTP(ctx context
...
+// The 401 challenge is produced by Envoy's JWT filter before the request ever reaches the MCP
+// proxy, so unlike every other place the resource identifier appears it cannot be computed in
+// Go. Envoy Gateway funnels a ResponseOverride's header values into the local response policy's
+// response_headers_to_add, where command operators are expanded.
+//
+// Reference: https://www.envoyproxy.io/docs/envoy/latest/configuration/observability/access_log/usage#command-operators
+func envoyDerivedResourceMetadataURL(servingPath string) string {
+	return fmt.Sprintf("%%REQ(X-FORWARDED-PROTO)%%://%%REQ(:AUTHORITY)%%%s%s",
+		oauthWellKnownProtectedResourceMetadataPath, strings.TrimSuffix(servingPath, "/"))
+}
```
Hritik003, line 320: qq: if we already have the url info here, why not hardcode the same in the HRF without the mcpproxy impl this PR suggests?

is it because clients would then get the literal string:

%REQ(X-FORWARDED-PROTO)%://%REQ(:AUTHORITY)%/.well-known/oauth-protected-resource/mcp
inside the JSON of PRM response and envoy does not expand command operators in that body?

internal/mcpproxy/oauth.go
```diff
@@ -0,0 +1,197 @@
...
+			"route", routeName, "path", externalPath(r))
+		http.NotFound(w, r)
+		return
+	}
+
+	if r.Method == http.MethodOptions {
+		writeProtectedResourceMetadataCORSHeaders(w.Header())
+		w.WriteHeader(http.StatusNoContent)
+		return
+	}
```
Hritik003, line 144: qq: this is purely for browser clients like mcp inspector?

tests/e2e/testdata/mcp_route_oauth.yaml
```diff
@@ -56,7 +56,10 @@ spec:
         - claim: name
           header: x-jwt-claim-name
       protectedResourceMetadata:
-        resource: "https://foo.bar.com/mcp"
+        # resource is deliberately omitted: the gateway derives the identifier from the
+        # scheme, authority and path of each request. This test reaches the gateway through a
+        # port-forward on a port picked at run time, so no statically configured value could
+        # ever match. mcp_route_authorization.yaml pins resource to cover the override path.
         resourceName: "example-resource"
         scopesSupported:
```
Hritik003, line 64: nit: can let this be, and create another tc without resource url?

PR #2693, opened 2026-09-15
fix(aigw): support custom auth header for MCP servers

kanurag94:
**Description**

`aigw run --mcp-config` only recognized a credential when it was supplied as `Authorization: Bearer <token>`. MCP servers that require a different header for auth, like Composio's `x-consumer-api-key`, never got their credential sent, because any non-`Authorization` header was silently dropped by the config generation template.

This extends `AddMCPServers` to also recognize a single non-`Authorization` header as the server's credential, and threads the header name through to the `MCPRoute`'s `securityPolicy.apiKey.header` field, which the CRD and its controller already fully support, reusing that existing mechanism rather than adding anything new at the CRD/controller layer.

- `Authorization: Bearer <token>` keeps working exactly as before.
- A single custom header (e.g. `x-consumer-api-key: ${COMPOSIO_API_KEY}`) is now extracted and re-injected into that same header name on the upstream request, unprefixed.
- Multiple simultaneous non-auth headers remain unsupported (pre-existing, separate limitation, out of scope here).

Test plan:
- `go test ./internal/autoconfig/...` - new unit test for the custom-header extraction, plus a new golden-file test (`testdata/composio.yaml`) exercising the full template render end-to-end
- `go test ./cmd/aigw/...`
- `go build ./...`
- `gofmt -l` / `go vet` clean on changed files

None.

None.


Hritik003 (commented):
Hi @kanurag94, can you please help me understand the difference between forward headers and this change we are making?

Issue #2128, opened 2026-05-11
Governance policy hook in LLM ExtProc pipeline


imran-siddique:
## Problem

AI Gateway already has strong routing, translation, rate limiting, and upstream auth, plus general Envoy security integrations (JWT, ext_authz) on the data plane. The MCP side goes further with scoped authorization, CEL expressions, and tool selectors.

What's missing is a first-class, body-aware governance hook in the LLM ExtProc pipeline where translated requests and responses can be evaluated by an external policy engine before dispatch and after completion.

A few existing issues point to this recurring need:
- #1919 raises external security provider integration with request/response scanning
- #1409 covers prompt guard/moderation
- #2073 discusses human-in-the-loop for MCP tool calls

These aren't identical problems, but they all need a governance extension point that doesn't exist in the LLM path today.

## Proposal

Add an optional, phased governance hook to the LLM ExtProc pipeline. Starting narrow:

### Phase 1: Pre-request governance + decision metadata

A route-level config field pointing to an external governance service (gRPC). The hook runs in `upstreamProcessor.ProcessRequestHeaders()` (in `internal/extproc/processor_impl.go`) **after translation and header/body mutation are finalized, before upstream auth and dispatch**.

The governance service receives:
- Translated request headers and body (or a configurable subset/redaction)
- Backend and route name
- Caller identity from already-authenticated upstream filters, validated JWT claims, or trusted forwarded headers

It returns allow/deny with a reason and optional metadata (matched policy rule, trust level, request classification).

On deny, the ExtProc returns an `ImmediateResponse` with the policy engine's reason. On allow, governance decision metadata is written to Envoy dynamic metadata so it flows into access logs, OTel traces, and rate limit descriptors.

### Phase 2: Async response audit

After the response completes, send response metadata to the governance service asynchronously for audit logging and compliance tracking. Non-blocking; the response is not held.

### Phase 3: Buffered response enforcement (later)

For non-streaming responses only, optionally gate the response through the governance service before returning to the client. Streaming responses stay audit-only since tokens already left the proxy.

## What this enables

- **Model access control**: "team-X can only use gpt-4o-mini" enforced at the gateway, not in application code
- **Content governance**: request scanning delegated to an external engine (prompt injection detection, PII checks)
- **Audit trail**: every LLM call gets a structured governance decision alongside existing token/latency metrics
- **Compliance**: operators in regulated industries can prove every AI call went through policy evaluation

## Implementation

[...]

PR #2696, opened 2026-09-16
feat: add configurable request and response guardrails

### LINKED ISSUE #1409: Prompt Gaurd/Moderation
_Description_: Users often would like to enforce some prompt guard or moderation policies to protect applications which uses the LLMs. The feature would fit for some enterprise users who build internal LLM/GenAI gateway by Envoy AI Gateway.

> Describe the desired behavior, what scenario it enables and how it
> would be used.

Image

[optional *Relevant Links*:]
- Concept:
  - [Navigating the LLM Landscape: Uber’s Innovation with GenAI Gateway](https://www.uber.com/blog/genai-gateway/)
- Gateway Implementations
  - [kgateway - Prompt Guards](kgateway.dev/docs/latest/ai/prompt-guards)
- Guard/Moderation API
  - [Open AI - Moderation](https://platform.openai.com/docs/guides/moderation)
  - [Anthoropic - Content moderation](https://docs.claude.com/en/docs/about-claude/use-case-guides/content-moderation)
  - [Gemini for safety filtering and content moderation](https://cloud.google.com/vertex-ai/generative-ai/docs/multimodal/gemini-for-filtering-and-moderation)

> Any extra documentation required to understand the issue.

From the API level perspective, I just imagined that the feature might be integrated into `AIGatewayRoute` or would introduce new API(CRD) to configure the guard/moderation policy because such policies are complex.

Perhaps, `BackendTrafficPolicy` or `SecurityPolicy` could be a candidate for integration target. But, these APIs are not Envoy AI Gateway's one. So, I think we should design more _generic_ apis so that it can call external apis in this case.

Note: although Envoy AI Gateway does not support this feature natively, users could achieve this by implementing [External Processing](https://gateway.envoyproxy.io/v1.4/tasks/extensibility/ext-proc/) or [External Authrization](gateway.envoyproxy.io/docs/tasks/security/ext-auth)

### LINKED ISSUE #1415: Native Integraion with vLLM Semantic Router
_Description_:

vLLM tracked in https://github.com/vllm-project/semantic-router/issues/486

Add Native Integraions with vLLM SR, providing:
+ intelligent routing
+ semantic caching
+ prompt guard, pii detection

### LINKED ISSUE #1919: Adding support for external security provider integration
_Description_:

We are a security vendor looking to start a discussion on the best way to integrate our platform with the Envoy AI Gateway to provide guardrails for LLM traffic.
Our goal is to enable a standardized way for users to trigger request and response scanning as part of their gateway flow.

We would appreciate guidance from the maintainers on the best way to kick off this project.

Thank you,
Matan

fernandoescolar:
**Description**

This change adds native content guardrails to Envoy AI Gateway through a new `GuardrailPolicy` resource and request/response enforcement in the external processor.

The design proposal is included at `docs/proposals/013-guardrails/proposal.md`. It compares the ext-proc architecture with the Built on Envoy Bedrock Guardrails and Azure Content Safety dynamic-module extensions and records the adopted decisions, current scope, follow-up work, and open questions.

The API is available in both `v1alpha1` and `v1beta1` and targets `AIServiceBackend` resources. Rules support request or response evaluation, deterministic multi-policy composition, configurable payload limits and timeouts, and explicit fail-open or fail-closed behavior. Actions can block, monitor, or mask matching content. CRD admission validates provider configuration, actions, endpoints, targets, and unique rule names.

The runtime supports:

- deterministic in-process regular expression rules;
- Presidio analyzer requests;
- AWS Bedrock ApplyGuardrail requests signed with SigV4; and
- Azure AI Content Safety text analysis.

Provider credentials are resolved from Kubernetes Secrets or, for Bedrock, from the standard AWS credential chain. Secret changes and policy deletion trigger reconciliation of affected routes. Missing or revoked provider configuration follows the rule's failure mode and does not leave stale gateway configuration active.

External providers receive schema-extracted text instead of serialized request envelopes. Mask actions write transformed text back to its original JSON field. Response guardrails keep streaming responses buffered until evaluation completes, preventing unsafe content from being partially delivered. Violations return HTTP 403 with the `GuardrailViolation` error type.

The implementation adds structured logs without payload content, an OpenTelemetry evaluation counter, and guardrail span events. It also includes example manifests and task-oriented documentation.

Validation includes unit tests, CRD admission tests, controller lifecycle and Secret rotation tests, HTTP-stub integration tests, Envoy dataplane tests for request and response blocking, and a Testcontainers test against the pinned official Presidio analyzer image. Live Azure AI Content Safety and AWS Bedrock Guardrails tests passed against real provider resources. Credential-gated live tests remain skipped by default.

GitHub Copilot was used to assist with implementation, tests, documentation, and review. The contributor has reviewed and takes responsibility for the resulting changes.

fixes: #1409 (Prompt guard moderation)
fixes: #1415 (pii detection)
fixes: #1919 (external security provider)
related: #2128 (content governance)

The main review areas are the dual-version CRD contract, Secret lifecycle behavior, backend scoping, fail-open/fail-closed semantics, response buffering, and external provider request formats.

Live provider validation requires local credentials and is intentionally not part of default CI. Presidio requires Docker but no credentials.

```shell
# Presidio
export TEST_PRESIDIO_ENDPOINT="https://..."
export TEST_PRESIDIO_API_KEY="..." # optional
export TEST_PRESIDIO_BLOCKED_TEXT="text that your policy blocks"

# Azure Content Safety
export TEST_AZURE_CONTENT_SAFETY_ENDPOINT="https://..."
export TEST_AZURE_CONTENT_SAFETY_API_KEY="..."
export TEST_AZURE_CONTENT_SAFETY_BLOCKED_TEXT="text that should be blocked"

# Bedrock Guardrails
export TEST_AWS_BEDROCK_GUARDRAIL_REGION="us-east-1"
export TEST_AWS_BEDROCK_GUARDRAIL_ID="..."
export TEST_AWS_BEDROCK_GUARDRAIL_VERSION="1"
export TEST_AWS_BEDROCK_GUARDRAIL_BLOCKED_TEXT="text that should be blocked"
# AWS credentials for Bedrock Guardrails
export AWS_ACCESS_KEY_ID="..."
export AWS_SECRET_ACCESS_KEY="..."
```


missBerg:
I haven't checked, as there are many PRs open at the moment, but have you submitted a design doc PR for this? Would be helpful!
Also, have you checked these extensions:

https://builtonenvoy.io/extensions/bedrock-guardrails/
https://builtonenvoy.io/extensions/azure-content-safety/

fernandoescolar:
I’ve added a proposal/technical design document (`docs/proposals/013-guardrails/proposal.md`) to help clarify the motivation, architecture, implementation details, and current scope of this PR.

I’m sorry I did not follow the project’s contribution and review protocol correctly from the beginning. I’ve reviewed the guidelines and updated the PR to align with them.

jrderuiter:
Would love to see native guardrail support! Any thoughts on extensibility? E.g. if users want to add another (not yet supported) guardrail service with a HTTP interface, how could they do that?

(I see that you mention calling external executables as being out of scope, but also assume extensibility should have some role in this proprosal).

fernandoescolar:
Yes, I think extensibility should definitely be part of the proposal.

One possible approach would be to define a small HTTP interface that custom guardrail services can implement, similar in spirit to Presidio’s API. For example, the runtime could send something like:

```json
{
  "text": "some user input",
  "context": {
    "stage": "input"
  }
}
```

to a configured endpoint such as:

```http 
POST /analyze
```

and expect a normalized response:

```json
{
  "action": "allow",
  "findings": [
    {
      "type": "PII",
      "start": 10,
      "end": 20,
      "score": 0.92
    }
  ]
}
```

action could be something like allow, block, or modify, with optional findings / replacement content depending on the guardrail.

Presidio is a useful precedent here because the analyzer is exposed behind a simple HTTP contract, while the implementation of the detection remains completely independent.

That would let us provide native integrations for common guardrail providers, while also having a generic HTTP guardrail adapter for services we don’t support yet. It also avoids the security / portability concerns of invoking arbitrary local executables.

PR #2698, opened 2026-09-16
fix: preserve Anthropic streaming overload errors in-band

### LINKED ISSUE #2697: Preserve Anthropic streaming overload errors for Chat Completions clients
## Description

Anthropic reports overloads differently for non-streaming and streaming requests. Its [HTTP error documentation](https://platform.claude.com/docs/en/api/errors#http-errors) assigns HTTP 529 to a non-streaming `overloaded_error`. Its [streaming documentation](https://platform.claude.com/docs/en/build-with-claude/streaming#error-events) explains that a streaming request can return HTTP 200 and report the same condition later through the following Server-Sent Events (SSE) event:

```http
HTTP/1.1 200 OK
Content-Type: text/event-stream

event: error
data: {"type":"error","error":{"type":"overloaded_error","message":"Overloaded"}}
```

The current [Chat Completions translator](https://github.com/envoyproxy/ai-gateway/blob/1fd110eb6f45fe7413709a23f6c7c3eba5918a6e/internal/translator/anthropic_helper.go#L1243-L1249) parses the Anthropic event but returns a Go error instead of a translated stream item:

```go
case string(constant.ValueOf[constant.Error]()):
	var errEvent anthropic.ErrorResponse
	if err := json.Unmarshal(data, &errEvent); err != nil {
		return nil, fmt.Errorf("unparsable error event: %s", string(data))
	}
	return nil, fmt.Errorf("anthropic stream error: %s - %s", errEvent.Error.Type, errEvent.Error.Message)
```

The [response processor](https://github.com/envoyproxy/ai-gateway/blob/1fd110eb6f45fe7413709a23f6c7c3eba5918a6e/internal/extproc/processor_impl.go#L596-L599) propagates that translator error instead of returning a response-body mutation or an immediate response:

```go
newHeaders, newBody, tokenUsage, responseModel, err := u.translator.ResponseBody(
	u.responseHeaders, decodingResult.reader, body.EndOfStream, u.parent.span,
)
if err != nil {
	return nil, fmt.Errorf("failed to transform response: %w", err)
}
```

The propagated Go error closes the external-processing remote procedure call (RPC). Under [Envoy's external processor failure behavior](https://github.com/envoyproxy/envoy/blob/v1.37.0/api/envoy/service/ext_proc/v3/external_processor.proto#L44-L48), what the client receives depends on whether Envoy has committed the downstream response:

| Downstream state | Envoy behavior | Client-visible result |
| --- | --- | --- |
| HTTP response not committed | Envoy replaces the upstream HTTP 200 response with a local HTTP 500 response. | The request fails with a generic HTTP 500 before stream iteration begins. |
| HTTP 200 response committed | Envoy cannot change the status, so it terminates the existing stream. | The client receives HTTP 200, but stream iteration later fails because the stream is truncated. |

Neither result exposes the provider's `overloaded_error` classification, so users cannot distinguish an overload from a generic gateway failure. The gateway needs a downstream representation that preserves this distinction for Chat Completions clients.

## Option 1: synthesize HTTP 529

If Envoy has not committed the downstream response, the gateway can replace the upstream HTTP 200 stream with a non-streaming error response:

```http
HTTP/1.1 529
Content-Type: application/json

{"error":{"type":"overloaded_error","message":"Overloaded"}}
```

This puts the overload on the HTTP error path. Clients receive the failure before stream iteration begins, status-based monitoring records it, and HTTP 5xx retry policies may treat it as retryable. However, a 529-only implementation still has two possible results:

| Downstream state when overload arrives | 529 result |
| --- | --- |
| HTTP response not committed | Return HTTP 529 with a JSON error. |
| HTTP 200 response committed | Terminate the stream because the status can no longer be changed. |

The limitation is inconsistency: the same provider error has two client-visible outcomes because an HTTP status cannot be changed after the response is committed. Returning 529 also differs from Anthropic's documented streaming behavior, which reports the error in-band under HTTP 200.

## Option 2: preserve HTTP 200 and emit an in-band error

The gateway can preserve the upstream streaming status and translate the Anthropic event into a data-only top-level error envelope:

```http
HTTP/1.1 200 OK
Content-Type: text/event-stream

data: {"error":{"type":"overloaded_error","message":"Overloaded"}}
```

This option behaves consistently for every streaming overload: it preserves HTTP 200 and sends the same terminal error event downstream. It follows Anthropic's documented streaming model—HTTP 200 followed by an in-band error—while translating the envelope for Chat Completions clients.

The compatibility limitation is that the [Chat Completions streaming reference](https://developers.openai.com/api/reference/resources/chat/subresources/completions/streaming-events/) documents `chat.completion.chunk` events but does not define a terminal top-level `error` event. The proposed event is handled by the SDKs' streaming layers rather than parsed as a `ChatCompletionChunk`.

The official OpenAI [Python v2.13.0](https://github.com/openai/openai-python/blob/v2.13.0/src/openai/_streaming.py#L83-L98), [Node v7.10.0](https://github.com/openai/openai-node/blob/v7.10.0/src/core/streaming.ts#L105-L130), and [Go v3.41.0](https://github.com/openai/openai-go/blob/v3.41.0/packages/ssestream/ssestream.go#L176-L221) stream parsers check for a top-level `error` before parsing a normal stream item. The Python SDK raises `APIError` during iteration and exposes the nested object as `error.body`.

A structural review of gateway response logs also found that direct OpenAI Chat Completions streams used a data-only SSE item containing a top-level `error` object and no named `event:` line:

```text
data: {"error":{"code":"internal_server_error","message":"...","param":null,"type":"server_error"}}
```

The cited SDK versions therefore handle the same top-level envelope observed in direct OpenAI Chat Completions traffic. Python clients can classify the proposed overload event by checking `error.body["type"] == "overloaded_error"` instead of parsing exception text. However, the event is not a documented `ChatCompletionChunk` variant, and support cannot be assumed for every third-party Chat Completions client.

## Proposed resolution

Use HTTP 200 with a terminal in-band error for every streaming overload. Of the two options, this is the only one that preserves `overloaded_error` after an HTTP 200 stream has started. It also avoids changing client behavior based on response timing. The event should contain only the overload type and message; provider request IDs and diagnostic details must not be copied.

The gateway should:

1. Return a typed translator error for the exact Anthropic `overloaded_error` type while retaining the serialized top-level error event.
2. Return that event as a normal response-body mutation under HTTP 200.
3. Mark the stream terminal and suppress every later provider body chunk.
4. Record the request as failed in metrics and tracing, and omit `[DONE]`.
5. Continue propagating every other Anthropic streaming error through the existing failure path.

## Acceptance criteria

- An overload as the first provider SSE event returns HTTP 200 and `data: {"error":{"type":"overloaded_error","message":"Overloaded"}}`.
- An overload after one or more translated chunks returns the same error event after those chunks.
- Content followed by overload in the same response-body callback preserves the content and ends with the error event.
- No later provider event and no `[DONE]` appears after the overload.
- The request is recorded as failed in metrics and tracing.
- Non-overload Anthropic SSE errors retain their current behavior.
- Tests exercise the error as the first event, after content, in the same callback as content, and with later provider chunks.
- Compatibility is verified with supported OpenAI client versions by checking the structured error body during iteration.

hustxiayang:
**Description**
Anthropic can return HTTP 200 for a streaming request and later report an `overloaded_error` in the Server-Sent Events (SSE) body. The current translator propagates that event as a Go error, so the client receives either a generic HTTP 500 before the response is committed or a terminated HTTP 200 stream after commitment. Neither result exposes the overload classification.

This change preserves HTTP 200 and translates an Anthropic streaming overload into a terminal Chat Completions SSE error:

```text
data: {"error":{"type":"overloaded_error","message":"Overloaded"}}
```

The translator returns a typed `AnthropicStreamError`, and the response processor returns the serialized event as a normal body mutation only when its type is `overloaded_error`. The event is appended to any output already translated in the same callback. The processor then suppresses later provider chunks, omits `[DONE]`, and records the request as failed. Other streaming error types and non-streaming errors retain their existing behavior.

## Related issue

Fixes https://github.com/theagentrouter/agent-router/issues/2697


hustxiayang:
@yuzisun Could you please help to review it? Thanks a lot!

PR #2705, opened 2026-09-17
fix: preserve JSON schema property order for Claude structured outputs

hustxiayang:
**Description**
Claude structured outputs currently return fields in alphabetical order instead of the order declared in the client's JSON schema. This change preserves the client-defined property order when translating OpenAI `response_format` requests to Anthropic `output_config.format`.

### Root cause

The OpenAI request schema initially arrives as raw JSON, which retains the order of its object properties. The Anthropic translator unmarshaled that schema into a Go `map[string]any` and assigned the map to the Anthropic SDK's `JSONOutputFormatParam.Schema` field.

Go maps do not retain insertion order. When the Anthropic SDK serializes the map, its JSON encoder intentionally sorts map keys alphabetically to produce deterministic JSON. Claude therefore receives an alphabetically ordered schema and generates its structured output in that order.

The response translator does not cause the reordering: Claude's generated JSON is carried back as text without being parsed into another map.

### Fix

- Continue unmarshaling the schema into `map[string]any` to preserve the existing validation and typed representation.
- Override only the serialized `schema` field with the original raw JSON bytes.
- Preserve the shared implementation for AWS Anthropic and GCP Anthropic backends.


internal/translator/anthropic_helper.go
```diff
@@ -755,16 +755,23 @@ func buildAnthropicParams(openAIReq *openai.ChatCompletionRequest, apiSchema fil
...
+		if err = json.Unmarshal(rawSchema, &schemaMap); err != nil {
 			return nil, fmt.Errorf("failed to parse JSON schema: %w", err)
 		}
+		format := anthropic.JSONOutputFormatParam{
+			Type:   constant.JSONSchema("json_schema"),
+			Schema: schemaMap,
+		}
+		// Override only the serialized schema with the validated raw JSON. Keeping
+		// Schema populated above preserves the typed representation for callers.
+		format.SetExtraFields(map[string]any{"schema": rawSchema})
```
xiaolin593, line 772: format will contain both `json_schema` and `schema` fields?

internal/translator/anthropic_helper_test.go
```diff
@@ -1500,6 +1500,36 @@ func TestBuildAnthropicParamsWithStructuredOutput(t *testing.T) {
...
+			},
+		},
+	}
+
+	params, err := buildAnthropicParams(request, filterapi.APISchemaAWSAnthropic, "")
+	require.NoError(t, err)
+
+	body, err := json.Marshal(params)
+	require.NoError(t, err)
+	require.Contains(t, string(body), `"schema":`+string(rawSchema))
```
xiaolin593, line 1530: can you add expected request payload in plain json and compare it with output after `buildAnthropicParams`? it would be much easier to view the difference

internal/translator/anthropic_helper.go
hustxiayang: `SetExtraFields` is just to override the `schema` field

internal/translator/anthropic_helper_test.go
hustxiayang: sorry, what difference you want to view?  `Conains` should already be byte-to-byte identical?

yuzisun (approved):

Issue #2715, opened 2026-09-20
Route backend MCP ping responses through Envoy AI Gateway


waw2637:
_Description_:

When an MCP backend sends a `ping` request with a numeric JSON-RPC ID, Envoy AI Gateway forwards the request to the client without applying its backend-aware request-ID encoding. The client's response therefore returns the original numeric ID and is rejected with HTTP 400:

Playwright is an example that expects this behavior and is used to cleanup sessions

```
invalid response ID type: 1
```

`mcpRequestContext.maybeServerToClientRequestModify` recognizes other server-to-client request methods before applying the common ID rewrite, but `ping` currently falls through the default case and returns early.

Expected behavior:

- A backend `ping` with numeric ID `1` is exposed to the client as a backend-qualified string ID such as `1__i__mcp-playwright`.
- The client response is routed only to the backend that originated the ping.
- The backend receives its original numeric response ID `1`.
- Different backends can reuse the same original ping ID without collision.
- `ping` is recorded as a recognized server-to-client method rather than unsupported.

This is protocol routing only. It does not enable, schedule, or otherwise change heartbeat behavior.

_Protocol version scope_:

Envoy AI Gateway v1.1.0 negotiates MCP protocol version `2025-06-18`. In that version, the Ping specification states that either the client or server can initiate a `ping` request and requires the receiver to respond with the same request ID.[1]

MCP protocol version `2026-07-28` removes `ping` from the protocol.[2] This issue is therefore limited to the legacy `2025-06-18` MCP path and does not propose adding ping handling to the modern `2026-07-28` path.

_Repro steps_:

1. Check out `v1.1.0` (`c217da8a539570ba3f6014a60c2687c29a945f1f`).
2. Pass a JSON-RPC request with method `ping` and numeric ID `1` through `maybeServerToClientRequestModify` for backend `mcp-playwright`.
3. Observe that the ID remains numeric `1` instead of becoming `1__i__mcp-playwright`.
4. Return a client response with numeric ID `1` and observe HTTP 400 from `handleClientToServerResponse`.

A focused regression test on the unpatched source reports:

```
expected: string("1__i__backend")
actual  : int64(1)
```

_Environment_:

- Envoy AI Gateway v1.1.0
- Linux amd64
- Go 1.26.4 (selected by the repository toolchain directive)

_Logs_:

```
invalid response ID type: 1
```

[1]: https://modelcontextprotocol.io/specification/2025-06-18/basic/utilities/ping
[...]

PR #2716, opened 2026-09-20
fix(mcp): route backend ping responses through the gateway

waw2637:
**Description**

This change recognizes backend-originated `ping` requests in the legacy MCP proxy's server-to-client request path. Pings do not need a payload rewrite, but now continue into the existing common request-ID encoding so a numeric ID such as `1` becomes a backend-qualified ID such as `1__i__mcp-playwright`.

Envoy AI Gateway v1.1.0 negotiates MCP protocol version `2025-06-18`. That version allows either the client or server to initiate a ping and requires the receiver to respond with the same request ID.[1] MCP protocol version `2026-07-28` removes `ping` from the protocol.[2] This change is therefore limited to the legacy `2025-06-18` path and does not add ping handling to the modern protocol path.

When the client responds, the existing response path uses that qualified ID to select only the originating backend and restores the backend's original numeric ID. Qualifying IDs with the backend name also allows multiple backends to reuse the same original ping ID without collision.

The response metrics path now treats `ping` as a recognized server-to-client method instead of recording and logging it as unsupported. This does not enable or schedule heartbeats and does not change heartbeat policy, backend-specific behavior, or session lifecycle behavior.

Tests replace the previous ping pass-through case with numeric-ID routing coverage, add ping to the recognized response-method cases, and retain the existing verification that `1__i__backend1` is routed to `backend1` and decoded to numeric ID `1`.

This change was prepared with assistance from OpenAI Codex. I reviewed the generated code and tests and take responsibility for the contribution.

**Related Issues

Fixes #2715

**Special notes for reviewers

This patch is based on v1.1.0 and targets `release/v1.1`. Current `main` moved the legacy MCP handler into `internal/mcpproxy/legacy.go`; the equivalent ping fallthrough is still present there and may require a separate forward fix according to the maintainers' preferred backport workflow.

Validation completed locally:

- `make build.extproc`
- `make precommit test`

[1]: https://modelcontextprotocol.io/specification/2025-06-18/basic/utilities/ping
[2]: https://modelcontextprotocol.io/specification/2026-07-28/changelog


Hritik003 (commented):
LGTM

Issue #2522, opened 2026-08-13
Validate Valkey with the existing global rate-limit and quota backend


atao2004:
_Description_:

Envoy AI Gateway’s documented token rate-limit and QuotaPolicy paths use Envoy Gateway’s external `envoyproxy/ratelimit` service and a Redis-labelled backend URL. AI Gateway does not own a Redis client or counter store.

Could maintainers confirm whether an additive compatibility-validation change for Valkey is welcome, and whether this repository is the correct ownership location?

## Motivation

Organizations that standardize on Valkey for Redis-compatible operational state need a supported, tested way to use the existing AI Gateway rate-limit and quota path without maintaining a separate Redis deployment solely for these counters. Valkey is available in self-managed deployments and managed offerings such as Google Cloud Memorystore for Valkey. This proposal is intentionally limited to validating the existing integration boundary, not adding a new gateway datastore feature.

## Proposed scope, pending confirmation

- Run the existing token-rate-limit and quota e2e scenarios against a pinned Valkey version.
- Preserve the current `rateLimit.backend.type: Redis` enum and configuration shape; use a Valkey service endpoint rather than introduce a new API.
- Add a minimal Valkey deployment/example and documentation stating exactly which version and scenarios were validated.
- Preserve existing Redis examples and behavior.

Validation would cover the existing counter, expiry, and enforcement paths through the external rate-limit service. No new AI Gateway runtime datastore client or semantic-cache feature is proposed.

Relevant Links:
- https://github.com/envoyproxy/ai-gateway/blob/main/examples/token_ratelimit/envoy-gateway-values-addon.yaml
- https://github.com/envoyproxy/ai-gateway/blob/main/tests/e2e/backend_quota_ratelimit_test.go
- https://github.com/envoyproxy/ai-gateway/blob/main/site/docs/capabilities/traffic/usage-based-ratelimiting.md
- https://cloud.google.com/memorystore/docs/valkey

siddharth1036:
I agree this makes sense for the AI Gateway repo since the rate-limit integration is configuration-only via Envoy Gateway's external service. Valkey's Redis compatibility suggests it should work as a drop-in replacement. Happy to leave the final decision on implementation scope and prioritization to the maintainers.

MennoLaanAtos:
ALso very intrested in this feature being enabled as we use valkey as well.

PR #2719, opened 2026-09-21
test: validate Valkey as a rate-limit backend

daric93:
**Description**

Envoy Gateway's global rate limit backend type is `Redis`, and the external `envoyproxy/ratelimit` service reaches that backend over the Redis protocol. Valkey speaks the same protocol, so it should drop into that position unchanged, but nothing in the repo demonstrated it.

This adds Valkey as a selectable backend for the rate-limit examples and end-to-end tests. Redis stays the default and its behavior is unchanged:

- `examples/token_ratelimit/` gains `valkey.yaml`, which pins `valkey/valkey:8.1.8-alpine`, and `envoy-gateway-values-valkey-addon.yaml`. The addon is the Redis addon with a different URL; `rateLimit.backend.type` stays `Redis`.
- `E2E_RATELIMIT_STORAGE` selects the backend for the rate-limit e2e tests. Unset or `redis` is the existing behavior; `valkey` deploys the Valkey example and asserts counters through `valkey-cli`.
- `make test-e2e-ratelimit-valkey` runs `Test_Examples_TokenRateLimit` and `Test_Examples_BackendQuotaRateLimit` against Valkey, and a new `test_e2e_ratelimit_valkey` CI job runs that target.
- The rate-limit docs state that both backends are supported and name the Valkey version the tests exercise.

There are no API, controller, or data plane code changes. This establishes that the existing Redis-protocol path works with Valkey; it does not add a new backend type, and it is neither semantic caching nor an AI Gateway datastore. What is validated is `valkey/valkey:8.1.8-alpine` against the two documented flows: token counters charged from LLM responses, and `429` enforcement once a bucket is over its limit. The docs say exactly that rather than claiming general Valkey support, since that is the scope the tests cover. Semantic caching and using a Redis-protocol store as an AI Gateway datastore are out of scope, and the docs say so.

Related issue: #2522
Earlier attempt at the same change: #2577

The original implementation is by @atao2004. This branch carries that work onto current `main`, resolves the conflicts from the docs rename, and addresses review feedback. Anna is preserved as commit author with `Co-authored-by`.

- The Valkey addon duplicates the Redis addon apart from the URL. I kept them as two files to match how the existing addons in `examples/` are structured, and added a cross-reference comment in each so the `envoyproxy/ratelimit` image pin does not drift. Happy to collapse them if you would rather.
- `SelectedRateLimitStorage()` returns an error rather than panicking. It is called from `initEnvoyGateway`, which is shared cluster setup for every e2e test, so a typo in the env var would otherwise take down the whole suite with an opaque panic instead of a message naming the bad value.
- `test_e2e_ratelimit_valkey` is not in `ci-required`, matching how `test_e2e_namespaced` is treated. It still runs and reports. Say the word if it should block.
- The quota manifest is rendered through stdin only when running against Valkey. The checked-in `testdata/backend_quota_ratelimit.yaml` hardcodes the Redis URL, so the Redis path still applies the file directly and the default case is byte-for-byte what it was. The `REDIS_*` env names and Redis-derived deployment names stay unchanged in the rendered Valkey variant, since those are the rate limit service's own names.
- The new job is pinned to Envoy Gateway v1.9.1. v1.8.1 rejects `examples/token_ratelimit/token_ratelimit.yaml`'s global rate limit rules with a int32 boundary validation error, which is unrelated to the backend swap and never surfaced in the default `test_e2e` matrix.

Testing: `make lint`, including `actionlint` on the new CI job, plus `make format`, `make spellcheck`, `make editorconfig` and `make tidy` are all clean, and `go test ./tests/internal/e2elib/...` passes. The new `test_e2e_ratelimit_valkey` job passed on my fork. The two failures in that fork run are unrelated to this change: a tokenless codecov upload, which a fork cannot do, and a port-collision flake in the macOS data-plane VCR tests.

AI usage disclosure, per the repository's GenAI policy: Claude was used to carry this work onto current `main`, resolve the doc conflicts, apply the review fixes listed above, write the CI job, and draft this description. All of it was reviewed by me, and the results above were run locally or in CI.


mathetake:
why do we need to test the redis-compatibility of Valkey?

jbrinkman:
The PR includes documentation updates as well as tests to confirm the compatibility. While Redis and Valkey were fully compatible for Redis 7.2, they have since continued development on independent paths. There are subtle differences between the two platforms which can lead to some different behaviors depending on the specific commands being used. Even if you don't think the tests are warranted, please consider the documentation changes as they help clarify to users that they can confidently use Valkey with Agent Router.

mathetake:
I am saying why do WE need to test that here. It's completely wrong abstraction layer to test that. I would suggest send a PR to https://github.com/envoyproxy/ratelimit - this is 100% wrong place to send this PR

daric93:
@mathetake Fair point on the layer — the counter and expiry behavior belongs to the rate limit service, not here. I'll take the e2e coverage to https://github.com/envoyproxy/ratelimit.

That leaves the docs. The part that is specific to this repo is that the `rateLimit.backend.type: Redis` setting in the token rate limit and quota examples accepts a Valkey endpoint, which is the question users land on here rather than in the ratelimit repo. Do you want to keep that? I'm happy to either trim this PR down to the doc note plus the Valkey example manifest, or drop it entirely and put everything in envoyproxy/ratelimit. Whichever you prefer.

